Issue #9, opened 2019-06-03
Rewrite resolver format


wtrocki:
Rewrite resolver format to work with typescript source code format.

ankitjena:
## Format
### CRUD operations
```
const resolvers = {
    Query: {
        ....
    }
    Mutations: {
        ....
    }
}
```
### Relations resolvers
TODO

wtrocki:
I will suggest to:
- Rewrite format without relations now 
- Drop some simple apollo template into templates folder
- Test them end to end with Altair or Playground.

This will be a really nice and beneficial work to understand the current schema.
If you have questions about the current schema let me know.

wtrocki:
Awesome!

PR #16, opened 2019-06-04
change resolver format

ankitjena:
Changed resolver format highlighted in #9


ankitjena:
A good approach would be to define implementation as a custom type. Something like
```
type customType = (...args: any[]) => any;
```

src/resolvers/ResolverManager.ts
```diff
@@ -1,8 +1,8 @@
 import { Type } from 'graphql-codegen-core';
 import { DatabaseContextProvider } from '../datasource/DatabaseContextProvider';
 import { logger } from '../logger';
-import { ResolverBuilder } from './ResolverBuilder';
-import { ResolverInstance } from './ResolverInstance'
+import { ResolverBuilder } from './NewResolverBuilder';
```
wtrocki: Naming is hard, but I think we can do better with `NewResolverBuilder`

wtrocki:
I do not fully understand intentions here so sorry.
New file we creating in example still using the old json format.
We save file to json where issue mentions that we going to do typescript file.

`JSON.stringify(generated.resolvers, undefined, 2)` is not going to be used to save file because the content will not be json so original problem should not exist.

Example resolvers look like this:
https://github.com/aerogear/ionic-showcase/blob/master/server/src/tasks/schema.js#L52-L102

wtrocki:
It simply cannot be string. It needs to be code we can run. 
If we start with the template first (basic apollo one) and add resolves there. 
Then make the generator to replace the resolvers file - it should still compile.

wtrocki:
I put some thoughts into this and I think that we might avoid using large handlebars template with for loop etc. and stay with each resolver type from CRUD having it is own template. This can be initially small but they might grow large over the time so having them as  sepatate template file makes sense

example/index.ts
```diff
@@ -1,7 +1,10 @@
-import { readFileSync } from 'fs';
+import { readFileSync, writeFile } from 'fs';
+import { promisify } from 'util'
 import { join } from 'path'
 import { enableDebug, GraphQLBackendCreator, IGraphQLBackend, KnexResolverManager, PostgresSchemaManager } from '../src/index';
 
+const writeFileAsync = promisify(writeFile)
```
wtrocki: There is async counterpart of this method so no need for this

src/GraphQLBackend.ts
```diff
@@ -85,27 +86,39 @@ export class GraphQLBackendCreator {
...
+        const outputSchema = generator.generateNewSchema(context, this.resolvers, this.config);
         backend.schema = outputSchema;
       } else {
         logger.info("Schema generation skipped.")
       }
 
+      backend.resolvers = {}
+      backend.resolvers["Query"] = {}
+      backend.resolvers["Mutation"] = {}
+      for (const value of this.resolvers) {
```
wtrocki: Clean Code. Responsibility. This should be in resolvers module. Really we should avoid format.of data being wrong in the first place

src/GraphQLBackend.ts
```diff
@@ -128,5 +141,5 @@ export interface IGraphQLBackend {
   // Human readable schema that should be replaced with current one
   schema?: string,
   // Resolvers that should be mounted to schema`
-  resolvers?: ResolverInstance[]
+  resolvers?: object
```
wtrocki: They should be still this ResolverInstance. It just need to change its format

ankitjena:
> It simply cannot be string. It needs to be code we can run.
> If we start with the template first (basic apollo one) and add resolves there.
> Then make the generator to replace the resolvers file - it should still compile.



> I put some thoughts into this and I think that we might avoid using large handlebars template with for loop etc. and stay with each resolver type from CRUD having it is own template. Each can be initially small but they might grow large over the time so having them as separate template file makes sense

But that would mean a lot of handlebars files.

wtrocki:
To not overcomplicate - we already using javascript templates in generator so we can have those or use handlebars - your choice. 
Use and build whatever you feel will be gold for the job. Separate files were to not overload single file with so many concerns but ignore that for now

src/GraphQLBackend.ts
```diff
@@ -107,18 +108,12 @@ export class GraphQLBackendCreator {
...
-        }
-        if(value.resolverType === 'Mutation') {
-          backend.resolvers["Mutation"][value.fieldName] = value.implementation
-        }
-      }
-
+      /**
+       * Generate TS resolvers from the ResolverInstance[] generated
+       */
+      const generator = new GraphQLResolverGenerator()
```
wtrocki: Generator should be injected. Hardcoding genrator is not best option. 
What will happen if we will want to add another generator in future

src/GraphQLBackend.ts
```diff
@@ -141,5 +136,5 @@ export interface IGraphQLBackend {
   // Human readable schema that should be replaced with current one
   schema?: string,
   // Resolvers that should be mounted to schema`
-  resolvers?: object
+  resolvers?: string
```
wtrocki: It is good to have this here as string for the moment but we might need to introduce another abstraction (outputformater)

src/resolvers/ResolverBuilder.ts
```diff
@@ -19,14 +18,14 @@ export class ResolverBuilder {
    * @param argumentContext context for generated arguments
    * @param knexContext name of knex object that was exposed
    */
-  constructor(argumentContext: string = 'resolve.args', knexContext: string = 'db') {
+  constructor(argumentContext: string = 'args', knexContext: string = 'db') {
```
wtrocki: Since we control the entire implementation now we do not need these hacks. 
We can declare that args directly in the template.

src/resolvers/ResolverBuilder.ts
```diff
@@ -38,7 +37,9 @@ export class ResolverBuilder {
       schemaDefinition: `${fieldName}(input: ${gqlType.name}Input!): ${gqlType.name}!`,
       action: ResolverType.CREATE,
       resolverType: "Mutation",
-      implementation: `return ${this.knexContext}('${tableName}').insert(${this.argumentContext}.input).returning('*')`
+      implementation: `(root, args, context, info) => {
```
wtrocki: Implementation is js or typescript? This code is looks like javascript.

src/resolvers/ResolverBuilder.ts
```diff
@@ -66,11 +68,12 @@ export class ResolverBuilder {
       fieldName: fieldName,
       typeName: gqlType.name,
       schemaDefinition: `${fieldName}(id: ID!): ID!`,
-      action: ResolverType.DELETE,
+      action: ResolverType.UPDATE,
       resolverType: "Mutation",
```
wtrocki: Let's remove this field and aggregate results into format of resolvers: 
```javascript
mutations: MutationResolver[]
queries: QueriesResolvers[]
subscriptions: SubscriptionsResolvers[]
```

src/resolvers/ResolverBuilder.ts
```diff
@@ -49,12 +50,13 @@ export class ResolverBuilder {
     return {
       fieldName: fieldName,
```
wtrocki: Do we need all those fields. If we define implementation to contain name as well we might not need it as separate element

src/resolvers/ResolverBuilder.ts
```diff
@@ -49,12 +50,13 @@ export class ResolverBuilder {
...
       typeName: gqlType.name,
-      schemaDefinition: `${fieldName}(id: ID!, input: ${gqlType.name}Input!): ${gqlType.name}!`,
+      schemaDefinition: `${fieldName}(id: ID!, input: ${gqlType.name}Input!): ${gqlType.name}!`,  
       action: ResolverType.UPDATE,
```
wtrocki: So this was for historical reasons as we needed type to be displayed in the UI that is not longer relevant. We do not need all of those times. I think we might just need implementation. 
If you find any place here that requires action. Having typeName seems ok but rest of the fields are just legacy/not neeed.

image

src/resolvers/ResolverBuilder.ts
```diff
@@ -100,7 +105,9 @@ export class ResolverBuilder {
       schemaDefinition: `${fieldName}(fields: QueryFilter): [${gqlType.name}]!`,
       action: ResolverType.FIND,
       resolverType: "Query",
-      implementation: `return ${this.knexContext}.select().from('${tableName}').where(${this.argumentContext}.field.name, '=', ${this.argumentContext}.field.value)`
+      implementation: `(root, args, context, info) => {
+        return ${this.knexContext}.select().from('${tableName}').where(${this.argumentContext}.field.name, '=', ${this.argumentContext}.field.value)
```
wtrocki: This is not properly implemented. Let's create separate issues for pagination and filtering as they are simply not there.

src/resolvers/ResolverGenerator.ts
```diff
@@ -0,0 +1,68 @@
...
+    input.resolvers = this.splitByTypes(resolvers)
+
+    return template(input)
+  }
+
+  /**
+   * Split resolvers into Query and Mutations
+   * @param resolvers Array of ResolverInstance
+   */
+  private splitByTypes(resolvers: ResolverInstance[]) {
```
wtrocki: As mentioned above - this code makes no sense. 
Why not actually generate those objects so we do not need to split data by types - we will have everything ready.

src/resolvers/ResolverGenerator.ts
```diff
@@ -0,0 +1,68 @@
...
+    const computedResolvers = { query: [], mutation: [] };
+    computedResolvers.mutation = resolvers.filter((value: ResolverInstance) => {
+      return value.resolverType === "Mutation"
+    })
+    computedResolvers.query = resolvers.filter((value: ResolverInstance) => {
```
wtrocki: See my proposition above. Please do not feel attached to implementation that was made in 2 days under time pressure to demo this project to stackholders. Code here is really rough and it requires refactoring. Let's not base on this and adapt it to our needs.

src/resolvers/partials/mutation.handlebars
```diff
@@ -0,0 +1,5 @@
+Mutation: {
```
wtrocki: This split do not make much sense. 
Would we ever need to modify this? Probably not. 
We can have implementation doing all and not using handlebars at all.

src/resolvers/ResolverGenerator.ts
```diff
@@ -0,0 +1,68 @@
...
+      return value.resolverType === "Query"
+    })
+
+    return computedResolvers;
+  }
+
+  /**
+   * Read partials for the template
+   */
+  private registerPartials() {
```
wtrocki: DRY - this is helper method that has nothing to do with generators. It is very generic and probably should not be here.

src/resolvers/ResolverBuilder.ts
ankitjena: Just fieldname, schemaDefinition and implementation is required as of now. Can merge fieldName and implementation into one.

src/resolvers/ResolverBuilder.ts
ankitjena: Yes, the way they are defined now, they are unable to run as some of the stuff are undefined.
 `db` for instance in the implementations.

src/resolvers/ResolverGenerator.ts
ankitjena: Hmm, seems better than this.

src/resolvers/ResolverGenerator.ts
ankitjena: How do you propose we go with implementing the resolvers now?

src/resolvers/ResolverGenerator.ts
ankitjena: Doesn't this read the handlebars templates from the directory?

src/resolvers/ResolverBuilder.ts
wtrocki: hmm.. me thinking - we can still keep field name
but implementation will contain full resolver including field name

src/resolvers/ResolverBuilder.ts
wtrocki: Yes.. That is why I recomend you to take look into showcase and remove implementation specific stuff from it. 
It has db mounted. Take look ionic-showcase/server has everything and we should probably push that first. If you need help on this I can also come over and do it

wtrocki: Let's have quick chat about this.

src/resolvers/ResolverGenerator.ts
wtrocki: We do not need handlebars for the moment.

src/resolvers/ResolverBuilder.ts
wtrocki: If we follow template approach we can have the flag for different languages

docs/usage.md
```diff
@@ -68,3 +68,4 @@ type Note {
 }
 ```
 See ./testss for use cases and tests on relations.
```
wtrocki: that is probably my error

src/resolvers/MetadataInstance.ts
```diff
@@ -5,7 +5,7 @@ import { ResolverType } from './ResolverType'
  */
 // FIXME maybe diferent name - this is not resolver specific
```
wtrocki: Remove invalid comment later.

wtrocki:
GREAT JOB ANKIT! 💯

PR #36, opened 2019-06-11
feat: init command

ankitjena:
## Approach
- choose template from existing templates
- copy template to the new folder(user's input)
- install dependencies

## Usage
```cmd
graphback init <project-name>
```

Screenshot from 2019-06-11 19-29-43


wtrocki:
@ankitjena I would love to see another steps:

- Drop sample file to template (we can always start with some sample model)
- Generate schema for this sample file
- Build project
- Print instruction what to do next to try out project:
you can run it by executing `npm run start`

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const isDirectory = (dir: string) => lstatSync(dir).isDirectory()
+
+function getDirectories(templateDir: string): string[] {  
+  return readdirSync(templateDir).filter((name: string) => isDirectory(join(templateDir, name)))
+}
+
+async function chooseTemplate(): Promise<string> {
+  const { templateName } = await ask([
+    {
```
wtrocki: My recommendation is to keep template files as separate file.

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,57 @@
...
+import { promisify } from 'util'
+import * as yargs from 'yargs'
+import { logInfo } from '../utils'
+
+const copyAsync = promisify(ncp)
+
+type Params = { name?: string }
+
+function getTemplateDir(): string {
+  return `${__dirname.split('packages')[0]}templates`
```
wtrocki: This is not going to work! This command line will not have access to the templates. 
We need to download them from github!

ankitjena:
> @ankitjena I would love to see another steps:
> 
>     * Drop sample file to template (we can always start with some sample model)
>       This can be part of the command line or generator (not sure now)
> 
>     * Generate schema for this sample file
>       Just trigger generator for this
> 
>     * Build project (npm run build)
> 
>     * Print instruction what to do next to try out the project:
>       you can run it by executing `npm run start`

Yes, making a separate `build` command and will do build within the init.

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,62 @@
...
+import ora from 'ora'
+import * as yargs from 'yargs'
+import { extractTemplate } from '../helpers/extractTemplate'
+import { allTemplates, Template } from '../helpers/template'
+import { logError, logInfo } from '../utils'
+
+type Params = { name?: string }
+
+
+async function installDependencies(name: string): Promise<void> {
```
wtrocki: Move those methods to separate file like utils or something. 
Let's keep the right architecture and actual command file clear so anyone can come and see straightaway what are the inputs and execution logic.

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    accessSync(path)
+    logError(`A folder with name ${name} exists. Remove it or try another name.`)
+    process.exit(0)
+  } catch (error) {
+    mkdirSync(path)
+  }
+}
+
+async function chooseTemplate(): Promise<Template> {
```
wtrocki: I would put that into separate file as well.
```
./commands/init.ts
./templates/templates.ts
./templates/templateMetadata.ts
./util
```

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,62 @@
...
+      name: 'templateName',
+      message: 'Choose a template to bootstrap',
+      choices: allTemplates.map((t: Template) => t.name)
+    }
+  ])
+
+  return allTemplates.find((t: Template) => t.name === templateName)
+}
+
+export const command = 'init <name>'
```
wtrocki: Missing option to provide template name directly.

packages/graphback-cli/src/commands/init.ts
```diff
@@ -0,0 +1,62 @@
+import * as execa from 'execa'
+import { accessSync, mkdirSync } from 'fs'
+import { prompt as ask } from 'inquirer'
+import ora from 'ora'
+import * as yargs from 'yargs'
+import { extractTemplate } from '../helpers/extractTemplate'
+import { allTemplates, Template } from '../helpers/template'
+import { logError, logInfo } from '../utils'
+
+type Params = { name?: string }
```
wtrocki: interface?

templates/apollo-starter-ts/model/Types.graphql
```diff
@@ -0,0 +1,14 @@
+type Note {
```
wtrocki: I would prefer for templates to not have this embedded. 
We going to introduce the concept of example schemas. 
For the moment it is best to just have those inside cmd and copy that after.
We can also provide a parameter for people to init command with their own schema later.

tslint.json
```diff
@@ -18,6 +18,7 @@
     "prefer-object-spread": false,
     "prefer-readonly": false,
     "no-unsafe-any": false,
+    "no-shadowed-variable": false,
```
wtrocki: -1000 this is usually bug

wtrocki:
Let's apply this in separate PR. Merging now to add more stuff to it.
Great job!

Issue #14, opened 2019-06-03
Create graphback init cli command


wtrocki:
Init should go thru templates and pick one. Then generate initial project structure. In this stucture, we need to have an example type. 
We can reuse example type we have created for example app.

PR #44, opened 2019-06-13
fix: restructure graphback init

ankitjena:
#14 Restructure the init command
- Separate concerns into methods
- Lint fixes
- Remove .graphql file from model in templates
- add post command message


packages/graphback-cli/src/utils/initHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return template
+}
+
+function postSetupMessage(name: string): string {
+  return `
+  GraphQL server successfully bootstrapped :rocket:
+  Next Steps:
+  1. change directory into project folder - ${chalk.cyan(`cd ${name}`)}
```
wtrocki: ❤️ 
```suggestion
  1. Change directory into project folder - ${chalk.cyan(`cd ${name}`)}
```

packages/graphback-cli/src/utils/initHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+  2. Create a graphql file inside ${chalk.cyan(`model`)} containing your Types
```
wtrocki:
```suggestion
  2. Create a files with graphql extension inside ${chalk.cyan(`model`)} containing only GraphQL Types
```

packages/graphback-cli/src/utils/initHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+  3. run ${chalk.cyan(`graphback build`)} to generate schema and resolvers
```
wtrocki:
```suggestion
  3. Run ${chalk.cyan(`graphback build`)} to generate schema and resolvers
```

packages/graphback-cli/src/utils/initHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+  4. run ${chalk.cyan(`graphback db`)} to create database resources in postgres
```
wtrocki: This will be multi-database project. Not progress only (although this will be the first).  

I think we missing command to run this thing after. 
`npm run start`?

templates/apollo-starter-ts/model/Types.graphql
```diff
@@ -1,14 +0,0 @@
-type Note {
```
wtrocki: So we might need separate issue where we going to introduce another question in init:

- Do you want to initialize project with sample schema?
yes
- What example project you want:
   x Note Example
   x Shop example (etc.)

We might have those examples as another folder in the repository and allow people to contribute their own.

wtrocki:
Lovely, nice and really really appreciated change.
Good work!

PR #49, opened 2019-06-13
feat: ask for model in init workflow

ankitjena:
## Major
During `graphback init` ask if user want's to include model or not. If yes, include one of the example models, if not create a file with commented out text

### Minor
Fix indentation in template literals(new line strings)

### DIFF
--- a/packages/graphback-cli/src/templates/modelTemplates.ts
+++ b/packages/graphback-cli/src/templates/modelTemplates.ts
@@ -1,0 +1,97 @@
+import { writeFileSync } from 'fs';
+import { prompt as ask } from 'inquirer'
+import ora from 'ora'
+import { Model } from './templateMetadata'
+
+export const defaultModel: string = `# type Note {
+#   id: ID!
+#   title: String!
+#   description: String!
+#   ## Relationship
+#   comment: [Comment]!
+
+# }
+
+# type Comment {
+#   id: ID!
+#   title: String!
+#   description: String!
+# }
+`
+
+export const allModels: Model[] = [
+  {
+    name: 'Note',
+    content: `type Note {
+  id: ID!
+  title: String!
+  description: String!
+  ## Relationship
+  comment: [Comment]!
+
+}
+
+type Comment {
+  id: ID!
+  title: String!
+  description: String!
+}
+`
+  },
+  {
+    name: 'Shop',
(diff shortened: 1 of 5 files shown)


packages/graphback-cli/src/templates/modelTemplates.ts
```diff
@@ -0,0 +1,97 @@
+import { writeFileSync } from 'fs';
+import { prompt as ask } from 'inquirer'
+import ora from 'ora'
+import { Model } from './templateMetadata'
+
+export const defaultModel: string = `# type Note {
```
wtrocki: Would it be possible to have them as separate files. each as separate graphql file. We can read them directly using fs

packages/graphback-cli/src/templates/modelTemplates.ts
```diff
@@ -0,0 +1,97 @@
...
+    content: `type Note {
+  id: ID!
+  title: String!
+  description: String!
+  ## Relationship
+  comment: [Comment]!
+
+}
+
+type Comment {
```
wtrocki: love templates great ideas

packages/graphback-cli/src/templates/modelTemplates.ts
```diff
@@ -0,0 +1,97 @@
...
+        type: 'list',
+        name: 'modelName',
+        message: 'Choose one of the example models',
+        choices: allModels.map((m: Model) => m.name)
+      }
+    ])
+
+    const content = allModels.find((m: Model) => m.name===modelName).content
+    
+    return [modelName, content]
```
wtrocki: How about default

packages/graphback-cli/src/templates/templateMetadata.ts
```diff
@@ -8,4 +8,9 @@ export interface TemplateRepository {
   uri: string
   branch: string
   path: string
+}
+
+export interface Model {
```
wtrocki: lack of comments. Rename to make it more clear what this is. ExampleTypes ?

packages/graphback-cli/src/templates/modelTemplates.ts
ankitjena: We would have to copy them in build then, Examples aren't that large anyways

PR #53, opened 2019-06-14
feat: generate command

ankitjena:
- Generate schema and resolvers using graphback-core
- Write them into generated folder


packages/graphback-cli/src/commands/generate.ts
```diff
@@ -0,0 +1,11 @@
+import { generate } from '../utils/generateHelper'
```
wtrocki: utils? I would say have separate implementation for each command somewhere but utils are not the best place. GenerateCommand maybe as name?

packages/graphback-cli/src/utils/generateHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Generate schema and resolvers using graphback-core and
+ * write them into generated folder
+ */
+async function generateBackend(): Promise<void> {
+  try{
+    accessSync(`${process.cwd()}/model`)
+    const model = new GlobSync('model/*.graphql', { cwd: process.cwd()})
+    
+    if(model.found.length === 0) {
+      logError(`No graphql file found inside model.`)
```
wtrocki:
```suggestion
      logError(`No graphql file found inside ./model folder .`)
```

packages/graphback-cli/src/utils/generateHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+      process.exit(0)
+    }
+
+    const path = `${process.cwd()}/${model.found[0]}`
```
wtrocki: Why only single file. Why we cannot load more than one and just concatenate they are types so this should be nice. We should promote more than one file. Our examples could be one file but devs may have 50 to 150 types inside their schema. It is not feasible to have it in single file

packages/graphback-cli/src/utils/generateHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+    const schemaText = readFileSync(path, 'utf8')
+
+    const outputSchemaPath = `${process.cwd()}/generated/schema.graphql`
+    const outputResolverPath = `${process.cwd()}/generated/resolvers.ts`
+    
+    const backend = new GraphQLBackendCreator(schemaText)
+    
+    const connectionConfig = {
```
wtrocki: WOW - what is that? and Why we have it here? 
Why we need to have hardcoded data for postgress connecting to localhost for generate comment (btw.. we will support more than postgress later.)

Are we no longer doing this split between generate and db commands?

wtrocki: If that is just temporary then ok, but I would love to see it gone and refactor underlying generator to not require it.

packages/graphback-cli/src/utils/generateHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+      'user': 'postgresql',
+      'password': 'postgres',
+      'database': 'users',
+      'host': '127.0.0.1',
+      'port': '5432'
+    }
+    
+    const manager = new PostgresSchemaManager(connectionConfig);
```
wtrocki: So we do db in generate command?
Would it be possible to introduce questions for this command:

- What database you want to choose
 x Postgress (just for now but there willk be more options in future)

I would love also to not generate db here (separate db command), but if you like idea if the generate doing all then let's remove db command. 

We will need to remember user choices inside the template. 
IMHO template can have something like graphback.json that will contain information about the database connection and any other choices

wtrocki:
This looks really nice but I would love to clarify our strategy before we merge that command. It will be nice to also address hardcoded config, however, you may add fixme or todo comment and we can have written an agreement that the following PR will address that. WDYT?

packages/graphback-cli/src/commands/generate.ts
ankitjena: How does helpers sound?

packages/graphback-cli/src/utils/generateHelper.ts
ankitjena: Hadn't thought about this, but this is a good idea. We should do it.

packages/graphback-cli/src/utils/generateHelper.ts
ankitjena: We have separated generate and db. You might not have rebuild the graphback package why it's creating db too.

ankitjena: I think separation is better since db command will be pretty hefty with implementation.

packages/graphback-cli/src/utils/generateHelper.ts
ankitjena: Actuallly db config is not needed here. I put them cuz I thought it was needed. My bad

packages/graphback/package.json
```diff
@@ -1,13 +1,16 @@
 {
-  "name": "graphback",
+  "name": "graphback-core",
```
wtrocki: we alrrady published package so rename not possible

PR #55, opened 2019-06-17
feat: add db config in init

ankitjena:
## Approach
create a db.json file in user's home directory which the cli can use

We would have to create env and edit the docker-compose being generated in init with database name, which is currently hardcoded to `users`


packages/graphback-cli/src/helpers/initHelper.ts
```diff
@@ -79,6 +80,18 @@ async function assignTemplate(templateName: string): Promise<Template> {
   return template
 }
 
+function createDBConfig(name: string) {
+  const configPath = `${homedir}/.db.json`
```
wtrocki: -10000 on db specific config. We will end up with 30 different config files in user root directory?

-10000 on actually having this globally. This could lead to really nasty bugs. 
It should be always project based not global

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+]
+
+
+/**
+ * Choose from the available db choices
+ */
+export async function chooseDB(projectName: string): Promise<object> {
+  const { dbName } = await ask([
+    {
```
wtrocki: kinda pointless really. We can pick the name but we cannot really setup different port or something. How db will be running on localhost? Do we require people use docker compose before running CLI or CLI will deal with starting the database. How about error handling when database is not created?

ankitjena: @wtrocki Should the user be able to customize about name, port, password etc ?

One way would be to custom generate a docker-compose file with user's inputs which can be later used by the db command before creating database.
Error handling would be handled in the db command.

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Different database choices
+ */
+export const dbChoices: Config[] = [
+  {
+    name: 'postgres',
+    config: (projectName: string) => {
+      return {
+        'user': 'postgresql',
+        'password': 'postgres',
+        'database': `${projectName}`,
```
wtrocki: IMHO let's hardcode name to graphback for the moment and print it. This stuff will need to change in the near future anyway so there is no point of investing time into this. 

On different operating systems, docker containers will be running under a different IP so localhost may be good for most of the cases but not always. We really need to document full flow in documentation first to make sure that we are on the same page here, before we merge that. 

PR do not need to resolve problems but I would want to avoid pushing things without having end to end vision and how we get there.

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: > Should the user be able to customize about name, port, password etc ? 

Yes. But we keep this file editable right in the project so we have this covered. 

Hmm.. we already have docker compose but this has some things hardcoded.
I think we can:

- have db abstraction later (mysql, postgress, mongo) and for the moment keep postgres, but to cather that IMHO let's move docker copose from template and inject it. So then we will have folder with databases and their starting config which is really nice way. 

Like we need this not only for database. The same requirement will go for templates. They are currently in the codebase of cmd but that is IMHO temporary. We need diagram of all moving pieces:

- Database of your choice
- Starting template of your choice
- Default client config
- Template user picks in that is missing all those and it is just runtime. 

This is just an idea and does not need to be executed now. 
The obvious drawback here is that we going to have this complexity of dropping different drivers into templates (currently it is all postgress) so another - much simpler option is to have also db dedicated templates, but this means a lots of replication of the code. 

So then we can have another idea:
Our templates execution layer can be abstracted to separate package and we can have templates that just share db specific code + typescript/Javascript etc.

All those probably should be spiked and we can play with this and see the best option.

ankitjena: > > Should the user be able to customize about name, port, password etc ?
> 
> Yes. But we keep this file editable right in the project so we have this covered.
Wouldn't  the docker-compose files will be different for templates depending on their simplicity.

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+      'user': 'postgresql',
+      'password': 'postgres',
+      'database': `${projectName}`,
+      'host': '127.0.0.1',
+      'port': '5432'
+    }
+  }
+}
+
+const pgDockerCompose = (projectName: string) => {
```
wtrocki: I really do not like this. I think we should have them as files. No strong opinion, but it will be much cleaner approach and we can actually test them .current approach do not allow us to test it

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -0,0 +1,52 @@
...
+
+services:
+  postgres:
+    image: postgres:9.6
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_USER: postgresql
+      POSTGRES_DB: ${projectName}
```
ankitjena: @wtrocki Well thought behind putting them in code was using the project name

PR #62, opened 2019-06-24
feat: add db command

ankitjena:
Create database using config created during init.


ankitjena:
@wtrocki I'm getting error `backend.createDatabase()` is not a function. But I has separated it and it works in the example in graphback package. Can you check?

packages/graphback-cli/src/commands/db.ts
```diff
@@ -0,0 +1,11 @@
+import { createDB } from '../helpers/dbHelper'
+
+export const command = 'db'
+
+export const desc = 'Create project'
```
wtrocki: ? :)

packages/graphback-cli/src/commands/db.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const builder = {}
```
wtrocki: No arguments?

packages/graphback-cli/src/helpers/dbHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+  try{
+    accessSync(`${process.cwd()}/model`)
+    const models = new GlobSync('model/*.graphql', { cwd: process.cwd()})
+    
+    if(models.found.length === 0) {
+      logError(`No graphql file found inside ${process.cwd()}/model folder.`)
+      process.exit(0)
+    }
+
+    const path: string = `${process.cwd()}`
```
wtrocki: Strange to use a template for this

packages/graphback-cli/src/helpers/dbHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+    const backend: GraphQLBackendCreator = new GraphQLBackendCreator(schemaText)
+    
+    const manager = new PostgresSchemaManager(dbConfig);
+    backend.registerDataResourcesManager(manager);
+
+    await backend.createDatabase()
+
+  } catch(err) {
+    // tslint:disable-next-line: no-console
+    console.log(err)
```
wtrocki: This will be really nasty and antipatern. We should be providing more human redable output for cmd. Everytime you need to disable ts-lint think if this can be avoided. Ideally we looking for some output class that can format it properly.

wtrocki: LogInfo used bellow

PR #67, opened 2019-06-24
Create watch command

ankitjena:
## Implementation
Use `chokidar` to watch for changes in `model` folder and run generate and db.

## Todo
Drop database implementation


packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -47,14 +47,16 @@ async function generateBackend(): Promise<void> {
...
 }
 
 /**
  * exported generate handler
  */
 export async function generate(): Promise<void> {
+  checkDirectory()
   await generateBackend()
   postCommandMessage()
+  process.exit(0)
```
wtrocki: reason why this is here?

packages/graphback-cli/src/helpers/watch.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as execa from 'execa'
+import { Stats } from 'fs';
+import { checkDirectory } from './common';
+import { createResources } from './db';
+import { generateBackend } from './generate';
+
+const modelsPath = `${process.cwd()}/model`
+
+const watcher = chokidar.watch(modelsPath, {
+  ignored: /(^|[\/\\])\../,
```
wtrocki: comment what we ignoring

packages/graphback-cli/src/helpers/init.ts
```diff
@@ -114,4 +114,5 @@ Bootstraping graphql server :dizzy: :sparkles:`)
   await installDependencies(name)
   createDBConfig(name)
   logInfo(postSetupMessage(name))
+  process.exit(0)
```
wtrocki: Why this? This seems to be bad pattern for reusing this code

wtrocki:
Executed watch but after chage it crashed with no error. Is chokidar an reputable package?

packages/graphback-cli/src/helpers/init.ts
ankitjena: After I added the watch command, all other commands got stuck after execution so added this as temporary fix for the workflow

ankitjena:
> Executed watch but after chage it crashed with no error. Is chokidar an reputable package?

It's quite reputable actually

packages/graphback-cli/src/helpers/common.ts
```diff
@@ -0,0 +1,12 @@
+import { accessSync } from 'fs'
+import { logError } from '../utils'
+
+
+export const checkDirectory = (): void => {
+  try{
+    accessSync(`${process.cwd()}/model`)
+  } catch(err) {
+    logError(`model directory not found. Make you sure you are in the root of your project.`)
+    process.exit(0)
```
wtrocki: Please remove process exit.

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -39,6 +30,8 @@ export const createResources = async(): Promise<void> => {
 }
 
 export const createDB = async(): Promise<void> => {
+  checkDirectory()
   await createResources()
   logInfo("Database Created")
+  process.exit(0)
```
wtrocki: Remove this

packages/graphback-cli/src/helpers/watch.ts
```diff
@@ -0,0 +1,22 @@
...
+  persistent: true
+});
+
+export const watch = async(): Promise<void> => {
+  checkDirectory()
+  execa('npm', ['run', 'start']).stdout.pipe(process.stdout);
+  watcher.on('change', async(path: string, stats: Stats) => {
+    await Promise.all([generateBackend(),createResources()])
```
wtrocki: Create resources will override existing files? what this is doing?

packages/graphback-cli/src/helpers/watch.ts
```diff
@@ -0,0 +1,22 @@
...
+const modelsPath = `${process.cwd()}/model`
+
+const watcher = chokidar.watch(modelsPath, {
+  ignored: /(^|[\/\\])\../,
+  persistent: true
+});
+
+export const watch = async(): Promise<void> => {
+  checkDirectory()
+  execa('npm', ['run', 'start']).stdout.pipe(process.stdout);
```
wtrocki: I think that we need to cancel return object.everytime this is the reason you got stuck initially. Lets do not make changes just to het things running but resolve problems in source

packages/graphback-cli/src/helpers/common.ts
```diff
@@ -0,0 +1,12 @@
+import { accessSync } from 'fs'
+import { logError } from '../utils'
+
+
+export const checkDirectory = (): void => {
+  try{
+    accessSync(`${process.cwd()}/model`)
+  } catch(err) {
+    logError(`model directory not found. Make you sure you are in the root of your project.`)
```
wtrocki: mkdir instead?

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -4,13 +4,26 @@ import { GraphQLBackendCreator, PostgresSchemaManager } from 'graphback';
...
-  const manager = new PostgresSchemaManager(dbConfig);
-  // tslint:disable-next-line: await-promise
-  await manager.getConnection().raw('DROP SCHEMA public CASCADE;')
-  // tslint:disable-next-line: await-promise
-  await manager.getConnection().raw('CREATE SCHEMA public;')
+  try {
+    const { dbConfig } = JSON.parse(readFileSync(`${process.cwd()}/config.json`, 'utf8'))
+    const manager = new PostgresSchemaManager(dbConfig);
+    // tslint:disable-next-line: await-promise
+    await manager.getConnection().raw('DROP SCHEMA public CASCADE;')
```
wtrocki: 👍 This is what we need for the moment.
Problem with this is that this is postgres specific and public schema may not exist but let's not bother with that for the moment. This code will be replaced soon.
Some warning message here will be cool - your data is lost. 
Do not push that to production etc.

PR #84, opened 2019-06-28
generate resolver types from schema

ankitjena:
## Implementation
Generate Resolver typings using `@graphql-codegen/core` and `typescript-resolvers` plugin.

### DIFF
--- a/packages/graphback/example/index.ts
+++ b/packages/graphback/example/index.ts
@@ -28,7 +28,7 @@ async function main() {
   // tslint:disable-next-line: no-console
   console.log(generated.resolvers)
 
-  await backend.createDatabase()
+  // await backend.createDatabase()
 }
 
 // tslint:disable-next-line: no-floating-promises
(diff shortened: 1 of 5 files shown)


packages/graphback/example/index.ts
```diff
@@ -28,7 +28,7 @@ async function main() {
   // tslint:disable-next-line: no-console
   console.log(generated.resolvers)
 
-  await backend.createDatabase()
+  // await backend.createDatabase()
```
wtrocki: Remove this?

packages/graphback/src/resolvers/ResolversTypeGenerator.ts
```diff
@@ -0,0 +1,50 @@
...
+   */
+  constructor(schemaText: string) {
+    this.schema = buildSchema(schemaText)
+    this.parsedSchema = parse(printSchema(this.schema))
+    this.config = {
+      filename: '',
+      schema: this.parsedSchema,
+      plugins: [
+        {typescript: {}},
+        {typescriptResolvers: {}} // Here you can pass configuration to the plugin
```
wtrocki: Are defaults ok for us?

wtrocki:
More info? Verification steps. Documentation change?
We should start adding more than just code as it will be hard to cover that after!

PR #94, opened 2019-07-02
Plugin using graphql-codegen for schema gen

ankitjena:
## Implementation
Custom plugin using `graphql-codegen`. The implementation follows this approach
- Use `visitor` pattern to return type names along with the respective fields. The fields are returned as `InputContext`.
- Use the `InputContext` to create `TargetContext` which is fed to string templates.

## Example
Input
```
type Note {
  id: ID!
  title: String!
  description: String!
  ## Relationship
  comment: [Comment]!
}

type Comment {
  id: ID!
  title: String!
  description: String!
}

```
Output Schema
```
interface Node {
  id: ID!
}

input NoteInput {
  title: String
  description: String
}

input CommentInput {
  title: String
  description: String!
}

type Note implements Node {
  id: ID
  title: String
  description: String
}

type Comment implements Node {
  id: ID
  title: String
  description: String!
}

type NoteFilter {
  id: ID
  title: String
  description: String
}

type CommentFilter {
  id: ID
  title: String
  description: String!
}

type PaginationInfo {
  totalPages: Int!
  totalItems: Int!
  page: Int!
  perPage: Int!
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
}

type NotePagination {
  items: [Note]!
  pageInfo: PaginationInfo!
}

type CommentPagination {
  items: [Comment]!
  pageInfo: PaginationInfo!
}

type Query {
  findNote(fields: NoteFilter!): Note!
  findComment(fields: CommentFilter!): Comment!
  findAllNote: NotePagination
  findAllComment: CommentPagination
}

type Mutation {
  createNote(input: NoteInput!): Note!
  createComment(input: CommentInput!): Comment!
  updateNote(id: ID!, input: NoteInput!): Note!
  updateComment(id: ID!, input: CommentInput!): Comment!
  deleteNote(id: ID!): ID!
  deleteComment(id: ID!): ID!
}

type Subscription {
  newNote: Note!
  newComment: Comment!
  updatedNote: Note!
  updatedComment: Comment!
  deletedNote: ID!
  deletedComment: ID!
}


```


wtrocki:
Great work! Is it possible to try adding missing elements in this PR?

ankitjena:
> Great work! Is it possible to try adding missing elements in this PR?

Working on it

packages/graphback-schema/src/plugin.ts
```diff
@@ -0,0 +1,15 @@
...
+import { visitor } from './visitor'
+
+
+
+export const plugin = (schema: GraphQLSchema) => {
+  const printedSchema = printSchema(schema)
+  const astNode = parse(printedSchema)
+  
+  const result = visit(astNode, { leave: visitor })
+  result.definitions.reverse()
```
wtrocki: Reason for this?

packages/graphback-schema/src/schema.ts
```diff
@@ -0,0 +1,98 @@
+const node = `interface Node {
+  id: ID!
+}
+`
+
+const pagination = (types: string[]): string => {
```
wtrocki: Love it!

packages/graphback-schema/src/visitor.ts
```diff
@@ -0,0 +1,7 @@
+import { ObjectTypeDefinitionNode } from 'graphql';
+
+export const visitor = {
+  ObjectTypeDefinition: (node: ObjectTypeDefinitionNode) => {
```
wtrocki: I think we miss field definition here. 
Without that we do not have access to fields

packages/graphback-schema/tests/empty-test.ts
```diff
@@ -0,0 +1,5 @@
+import ava, { ExecutionContext } from 'ava';
+
+ava('TODO', async (t: ExecutionContext) => {
+  t.assert(true)
```
wtrocki: lol

packages/graphback-schema/src/visitor.ts
wtrocki: generally add more definitions here and see how we can process things

packages/graphback-schema/src/visitor.ts
```diff
@@ -1,7 +1,31 @@
-import { ObjectTypeDefinitionNode } from 'graphql';
+import { FieldDefinitionNode, ListTypeNode, NamedTypeNode, NameNode, NonNullTypeNode, ObjectTypeDefinitionNode } from 'graphql';
 
 export const visitor = {
```
wtrocki: So visitor is going to collect whatever data that is needed and map that to digestible object. This is IMHO much cleaner approach than before where you were getting all.
Drawback here is that for each plugin we will need to visit entire graphql node couple times so it will affect overal performance. It is up to discussion/spike if we need couple dedicated plugins or large one. WDYT about this? Let's have bigger conversation on what we have in this pR already as we slowly heading towards something that will work for our use cases.

packages/graphback-schema/src/schema.ts
```diff
@@ -21,6 +40,60 @@ ${types.map((name: string) => `type ${name}Pagination {
...
+
+    return false
+  })
+}
+
+const inputFields = (fields: string[]): string[] => {
+  return fields.filter((f: string) => !f.startsWith('id'))
+}
+
+const inputs = (defs: Definition[]): string => {
```
wtrocki: I think I get what this code does but overall readability is quite bad here. 
We do a lot of type juggling and it can be very difficult to get that. 

We do lots of map objects here and mixing it with actual schema definitions. 
Do you thing we can detach that into 2 areas:

- code here can build context object we need 
- We will have a strongly typed context object that we expect in schema definitions
- schema definitions could be still in template language
- we validate output schema before sending success. 

We need more than just get things working. 
We need to figure out architecture and tools on top of graphql code generator that will work for our use cases.

If you think that this approach is clean and easy we can stay with that, just not really convinced that this is a good option in terms of unit testing this.

ankitjena: Yes it would be better

wtrocki: Can we explore that approach then?

wtrocki:
Is this finished. I see context being build but we still not used to build functions

wtrocki:
From my point of view I'm happy with everything you think it is best but lets have discussion on how we see this working in this PR

ankitjena:
> From my point of view I'm happy with everything you think it is best but lets have discussion on how we see this working in this PR

Updated the PR description

wtrocki:
> Use visitor pattern to return type names along with the respective fields. The fields are returned as string.

We will need to know the types of fields. Can we do that?

wtrocki:
> Build a context object which has the following structure.

This structure looks very strange. Can you explain the reason why we have it. 

Our case will be that we always supply only the busines model type and generate everything else. Really confused.

Issue #83, opened 2019-06-28
Code generation engine V2.0


ankitjena:
## Background
We are using `graphql-code-generator` version `0.18.2` in our code base to help generate the schema. Particularly we are using the method `schemaToTemplateContext` to parse datamodel file and use handlebars to generate schema.

## Version 1.0
`graphql-code-generator` has released version `1.0.0` which removes the above mentioned method as stated in docs
>Remove the old, deprecated code from this repostory (everything related to flattenDocuments and buildSchemaContext).

Current version's method generate string in output file format. Check PR #84 

## Requirement
- We need to use version `1.0.0` to generate Resolver Typings to generate resolvers(these are generated to an output file)
- Figuring out Schema generation in current version.

wtrocki:
We decided we need to rearchitecture currently graphback package to rely on the plugins and utilize latest version of the tool. 

![Screenshot 2019-06-28 at 14 58 13](https://user-images.githubusercontent.com/981838/60349061-c3d41880-99b8-11e9-9446-5957a16fc00d.png)

ankitjena:
Upon further inspection we might not need separate packages after all. Using the visitor pattern we can replace the deprecated context generation part. And redesign the core along the way. Graphql-code-generator doesn't provide much value at out current scope. But we might use it to provide out of the box solutions in future.

ankitjena:
Closing this

dotansimha:
@ankitjena the changes in https://github.com/aerogear/graphback/pull/112 looks great :)
What do you think about wrapping these visitors as `graphql-codegen` plugin? this way it will be available as ready-to-use and independent generator.
It might also make it easier for you to load the schema (because the codegen does that with graphql-toolkit, and it can easily load schema from any file - code/graphql ast/exports/url).

wtrocki:
+100 for wrapping our code and using extended features of codegen

PR #102, opened 2019-07-08
Reformat Input Context structure

ankitjena:
## Changes
- Reformat the context to be reused by graphback packages #83 
### Structure of the Context
```
export interface FieldContext {
  name: string
  type: string
  isArray: boolean
  isType: boolean
  isNull: boolean
  directives: object
  hasDirectives: boolean
}

export interface InputContext {
  name: string
  fields: FieldContext[]
  config: object
}
```
`config` is an object which has the following structure
```
{
  paginate: boolean
  create: boolean,
  update: boolean,
  delete: boolean,
  find: boolean,
  findAll: boolean
}
```
By default the config is 
```
{
  paginate: false,
  create: true,
  update: true,
  delete: true,
  find: true,
  findAll: true
}
```


packages/graphback-schema/src/plugin.ts
```diff
@@ -1,15 +1,13 @@
...
+import { buildTargetContext, TargetContext } from './targetType';
 
 export const plugin = (schema: GraphQLSchema) => {
-  const printedSchema: string = printSchema(schema)
-  const astNode: DocumentNode = parse(printedSchema)
-  
-  const inputTypes = visit(astNode, { leave: inputTypeVisitor })
 
-  const context: TargetContext = buildTargetContext(inputTypes.definitions.reverse())
+  const inputContext = new GraphbackContextCreator(printSchema(schema)).createInputContext()
```
ankitjena: We are initializing the Input Context here. Current way using visitor pattern doesn't require using codegen plugin as we can pass this Input Context to different use cases. 

Codegen further needs to accept schema instance everytime which is really not required.

@wtrocki  WDYT?

packages/graphback/src/InputTypeVisitor.ts
```diff
@@ -1,7 +1,16 @@
-import { DirectiveDefinitionNode, FieldDefinitionNode, ListTypeNode, NamedTypeNode, NameNode, NonNullTypeNode, ObjectTypeDefinitionNode } from 'graphql';
+import { ArgumentNode, BooleanValueNode, DirectiveDefinitionNode, DirectiveNode, FieldDefinitionNode, ListTypeNode, NamedTypeNode, NameNode, NonNullTypeNode, ObjectTypeDefinitionNode, StringValueNode } from 'graphql';
 
 const scalars = ['ID', 'Int', 'Float', 'String', 'Boolean']
 
+const defaultConfig = {
```
wtrocki: Quick question. Do we still use directives here and override this values?

packages/graphback/src/directives.ts
```diff
@@ -1,7 +1,12 @@
-const directives = `
-  directive @OneToOne(field: String) on FIELD
-  directive @OneToMany(field: String) on FIELD
-  directive @ManyToMany(tablename: String) on FIELD
+const directives = `directive @OnetoMany(field: String) on FIELD_DEFINITION
+directive @OneToOne(field: String) on FIELD_DEFINITION
+directive @ManyToMany(tablename: String) on FIELD_DEFINITION
+directive @config(paginate: Boolean,
```
wtrocki: I will suggest using `Generator` or `Model`. Let's use capitals maybe.

wtrocki:
@ankitjena What is the plan here? Are we planning to conform to this config in this PR and do resolves or there will be another PR for that? I would love to have pagination enabled by default (sorry :)).

Can I verify that or it is still in the works?

packages/graphback/src/InputTypeVisitor.ts
ankitjena: Yes we do that while visiting the nodes.

ankitjena:
> @ankitjena What is the plan here? Are we planning to conform to this config in this PR and do resolves or there will be another PR for that? I would love to have pagination enabled by default (sorry :)).
> 

Yes, will do resolvers in separate PR.

wtrocki:
Let's merge that and work on new aproach

PR #112, opened 2019-07-12
Reformat(MAJOR) - Code Generation engine 2.0

ankitjena:
Solves issue #83 
## Changes
- Remove dependency from `graphql-code-generator` version `0.18.2`
- Use Visitor pattern to create custom Context object
- Remove dependency from `handlebars`, Use string templates instead
- Rewrite schema generation and resolver generation
- Change schema file output format


packages/graphback/src/schema/targetSchemaContext.ts
```diff
@@ -0,0 +1,119 @@
...
+    return `${f.name}: ${f.type}${f.isNull ? '!': ''}`
+  }
+}
+
+export const buildTargetContext = (inputContext: Type[]) => {
+  const context: TargetContext = {
+    nodes: [],
+    inputFields: [],
+    filterFields: [],
+    pagination: [],
```
wtrocki: I thought we agreed to ignore pagination for the moment?

packages/graphback/src/schema/targetSchemaContext.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  })
+  context.filterFields = inputContext.map((t: Type) => {
+    return {
+      "name": t.name,
+      "fields": t.fields.filter((f: Field) => !f.isType)
+                  .map(arrayField)
+    }
+  })
+  context.pagination = inputContext.filter((t: Type) => t.config.paginate)
```
wtrocki: Actually let's leave it hardcoded to false but still leave the code.

wtrocki:
Good job!

PR #113, opened 2019-07-15
fix: change resolver implementation, edit template mapping

ankitjena:
Checked working of graphback by going through the flow of commands
Fixed:
- resolver implementation of create, update
- schema def of find method
- template mapping

### DIFF
--- a/packages/graphback/src/resolvers/knex/resolverImplementation.ts
+++ b/packages/graphback/src/resolvers/knex/resolverImplementation.ts
@@ -2,8 +2,9 @@ const knexContext = `context.db`
 const argumentContext = `args`
 
 export const createTemplate = (fieldName: string, tableName: string): string => {
-    return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
+    return `${fieldName}: async (root, args, context, info) => {
+      const result = await ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
+      return result[0]
     }`
   }
 
@@ -22,13 +23,14 @@ export const findAllTemplate = (fieldName: string, tableName: string): string =>
 
 export const findTemplate = (fieldName: string, tableName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}.select().from('${tableName}').where(...${argumentContext}.fields)
+      return ${knexContext}.select().from('${tableName}').where(${argumentContext}.fields)
     }`
 }
 
 export const updateTemplate = (fieldName: string, tableName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}('${tableName}').where('id', '=' , ${argumentContext}.id).update(${argumentContext}.input).then( () => {
-        return ${knexContext}.select().from('${tableName}').where('id', '=' , ${argumentContext}.id);
+      return ${knexContext}('${tableName}').where('id', '=' , ${argumentContext}.id).update(${argumentContext}.input).then( async () => {
+        const result = await ${knexContext}.select().from('${tableName}').where('id', '=' , ${argumentContext}.id);
+        return result[0]
     })}`
 }
--- a/templates/apollo-starter-ts/generated/resolvers.ts
+++ b/templates/apollo-starter-ts/generated/resolvers.ts
@@ -1,0 +1,11 @@
+export const resolvers = {
+  Query: {
+      hello: (obj, args, context, info) => {
+          // we can access the request object provided by the Voyager framework
+
+          // we can access the context added below also
+          // console.log(context.serverName)
+          return `Hello world from ${context.serverName}`;
+      }
+  }
+}
--- a/templates/apollo-starter-ts/src/config/config.ts
+++ b/templates/apollo-starter-ts/src/config/config.ts
@@ -12,8 +12,8 @@ class Config {
     this.port = process.env.PORT || 4000
     this.db = {
       database: process.env.DB_NAME || 'users',
-      user: process.env.DB_USERNAME || 'postgres',
-      password: process.env.DB_PASSWORD || 'password',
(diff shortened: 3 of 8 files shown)


packages/graphback/src/resolvers/knex/resolverImplementation.ts
```diff
@@ -2,8 +2,9 @@ const knexContext = `context.db`
 const argumentContext = `args`
 
 export const createTemplate = (fieldName: string, tableName: string): string => {
-    return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
+    return `${fieldName}: async (root, args, context, info) => {
+      const result = await ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
```
wtrocki: Formatting of those can be off. 
We can do:

A) Ignore formatting completely and use external formatter after (like vscode headless) based on the editorconfig in the template (we need to add it actually)
B) Try to format code ourselves here to look nicely without formatter with some opiniated format (spaces vs tabs ;P)

Ideally, we need both :)

packages/graphback/src/resolvers/knex/resolverImplementation.ts
```diff
@@ -22,13 +23,14 @@ export const findAllTemplate = (fieldName: string, tableName: string): string =>
...
-      return ${knexContext}.select().from('${tableName}').where(...${argumentContext}.fields)
+      return ${knexContext}.select().from('${tableName}').where(${argumentContext}.fields)
     }`
 }
 
 export const updateTemplate = (fieldName: string, tableName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}('${tableName}').where('id', '=' , ${argumentContext}.id).update(${argumentContext}.input).then( () => {
-        return ${knexContext}.select().from('${tableName}').where('id', '=' , ${argumentContext}.id);
+      return ${knexContext}('${tableName}').where('id', '=' , ${argumentContext}.id).update(${argumentContext}.input).then( async () => {
```
wtrocki: ${argumentContext} is being used here but we actually have `args` above. This is looks really strange. We can hardcode args as we define it.

packages/graphback/src/resolvers/knex/resolverImplementation.ts
```diff
@@ -22,13 +23,14 @@ export const findAllTemplate = (fieldName: string, tableName: string): string =>
 
 export const findTemplate = (fieldName: string, tableName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
-      return ${knexContext}.select().from('${tableName}').where(...${argumentContext}.fields)
+      return ${knexContext}.select().from('${tableName}').where(${argumentContext}.fields)
     }`
 }
 
 export const updateTemplate = (fieldName: string, tableName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
```
wtrocki: What happened to our type safety claim? Does this compile with TypeScript engine (it should not)

templates/apollo-starter-ts/generated/resolvers.ts
```diff
@@ -0,0 +1,11 @@
+export const resolvers = {
```
wtrocki: Add git ignore for generated folder outside template - in root of the repo!

wtrocki: Or do we want this to be there so it will work without generation?

wtrocki: IF yes then we need to documenent that.

templates/apollo-starter-ts/src/config/config.ts
```diff
@@ -12,8 +12,8 @@ class Config {
     this.port = process.env.PORT || 4000
     this.db = {
       database: process.env.DB_NAME || 'users',
-      user: process.env.DB_USERNAME || 'postgres',
-      password: process.env.DB_PASSWORD || 'password',
+      user: process.env.DB_USERNAME || 'postgresql',
```
wtrocki: It looks strange and it is very postgres specific. Do we still do pluggable db idea?

wtrocki:
Pointed couple problems that can possibly land as issues.
Please do:
A) Log things as issue if you think it is issue and resolve comment with issure reference.
B) Address problem directly.

For me this can be merged as long as my concerns will be addressed in near future.

packages/graphback/src/resolvers/knex/resolverImplementation.ts
ankitjena: Haven't made them type safe. Should we use graphql-codegen and generate types?

templates/apollo-starter-ts/generated/resolvers.ts
ankitjena: This was for mapping. Once we do `generate` it will be replaced.

wtrocki:
I actualy realized that I need that change to play with stuff on master. Merged but feel free to create new PR.

templates/apollo-starter-ts/src/config/config.ts
ankitjena: Because the files for db connection are specific, maybe db specific templates is a way to go
like this can be `apollo-postgres-typescript`

wtrocki: hmm.. We agreed that we will have separate abstraction for databases added to generator. We have this in form of docker-compose file when we choose this in CLI. This is tech debt - let's create separate issue to address it.

packages/graphback/src/resolvers/knex/resolverImplementation.ts
wtrocki: Yes.. Why not? Any challenge?

templates/apollo-starter-ts/src/config/config.ts
wtrocki: TL:DR let's not do it now today - but this is code smell that needs to be addressed before 1.0 :)

PR #115, opened 2019-07-15
Implement Subscriptions resolvers

ankitjena:
## Implementation
For each type generate
- `new`
- `updated`
- `deleted`

Use redis pubsub for subscription

## Example
For input
```
type Note {
  id: ID!
  title: String!
  description: String!
}
```

the output resolver file
```
enum Subscriptions {
  NEW_NOTE = 'newnote',
  UPDATED_NOTE = 'updatednote',
  DELETED_NOTE = 'deletednote'
}

export const resolvers = {
  Query: {
    findNotes: (_, args, context) => {
      return context.db.select().from('note').where(args.fields)
    },
    findAllNotes: (_, __, context) => {
      return context.db.select().from('note')
    }
  },

  Mutation: {
    createNote: async (_, args, context) => {
      const result = await context.db('note').insert(args.input).returning('*')
      context.pubsub.publish(Subscriptions.NEW_NOTE, {
        newNote: result[0]
      })
      return result[0]
    },
    updateNote: (_, args, context) => {
      return context.db('note').where('id', '=' , args.id).update(args.input).then( async () => {
        const result = await context.db.select().from('note').where('id', '=' , args.id);
        context.pubsub.publish(Subscriptions.UPDATED_NOTE, {
          updatedNote: result[0]
        })
        return result[0]
    })},
    deleteNote: (_, args, context) => {
      return context.db('note').where('id', '=' , args.id).del().then( () => {
        context.pubsub.publish(Subscriptions.DELETED_NOTE, {
          deletedNote: args.id
        })
        return args.id;
    })}
  },

  Subscription: {
    newNote: {
      subscribe: (_, __, context) => context.pubsub.asyncIterator(Subscriptions.NEW_NOTE)
    },
    updatedNote: {
      subscribe: (_, __, context) => context.pubsub.asyncIterator(Subscriptions.UPDATED_NOTE)
    },
    deletedNote: {
      subscribe: (_, __, context) => context.pubsub.asyncIterator(Subscriptions.DELETED_NOTE)
    }
  }
}

```

### DIFF
--- a/packages/graphback-cli/src/templates/resources/docker/pg-docker-compose.yml
+++ b/packages/graphback-cli/src/templates/resources/docker/pg-docker-compose.yml
@@ -10,10 +10,7 @@ services:
       POSTGRES_USER: postgresql
       POSTGRES_DB: users
 
-  # Mosca is a simple MQTT Broker
-  # In OpenShift/Production we would use the Red Hat AMQ broker
-  mosca:
-    image: matteocollina/mosca
+  redis:
+    image: redis:3.2
     ports:
-      - "1883:1883" # MQTT
-      - "80:80" # web interface
+      - "6379:6379"
--- a/packages/graphback/src/resolvers/knex/resolverImplementation.ts
+++ b/packages/graphback/src/resolvers/knex/resolverImplementation.ts
@@ -1,17 +1,22 @@
 const knexContext = `context.db`
-const argumentContext = `args`
 
-export const createTemplate = (fieldName: string, tableName: string): string => {
+export const createTemplate = (fieldName: string, tableName: string, typeName: string): string => {
     return `${fieldName}: async (root, args, context, info) => {
-      const result = await ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
+      const result = await ${knexContext}('${tableName}').insert(args.input).returning('*')
+      pubsub.publish(Subscriptions.NEW_${typeName.toUpperCase()}, {
+        new${typeName}: result[0]
+      })
       return result[0]
     }`
   }
 
-export const deleteTemplate = (fieldName: string, tableName: string): string => {
+export const deleteTemplate = (fieldName: string, tableName: string, typeName: string): string => {
     return `${fieldName}: (root, args, context, info) => {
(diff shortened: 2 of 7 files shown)


packages/graphback-cli/src/templates/resources/docker/pg-docker-compose.yml
```diff
@@ -10,10 +10,7 @@ services:
       POSTGRES_USER: postgresql
       POSTGRES_DB: users
 
-  # Mosca is a simple MQTT Broker
-  # In OpenShift/Production we would use the Red Hat AMQ broker
-  mosca:
-    image: matteocollina/mosca
+  redis:
```
wtrocki: pg docker compose has redis? We are ok with that but we might create issue to abstract it

packages/graphback/src/resolvers/knex/resolverImplementation.ts
```diff
@@ -1,17 +1,22 @@
 const knexContext = `context.db`
-const argumentContext = `args`
 
-export const createTemplate = (fieldName: string, tableName: string): string => {
+export const createTemplate = (fieldName: string, tableName: string, typeName: string): string => {
     return `${fieldName}: async (root, args, context, info) => {
-      const result = await ${knexContext}('${tableName}').insert(${argumentContext}.input).returning('*')
+      const result = await ${knexContext}('${tableName}').insert(args.input).returning('*')
+      pubsub.publish(Subscriptions.NEW_${typeName.toUpperCase()}, {
```
wtrocki: pubsub must be on context. Not sure where this comes from

ankitjena: Done

packages/graphback-cli/src/templates/resources/docker/pg-docker-compose.yml
ankitjena: Redis will be common, as it is for subscriptions. pg docker compose specifies database as postgres

wtrocki: ok

wtrocki:
If we do tests using snapshots that will be cool

ankitjena:
> If we do tests using snapshots that will be cool

Merging this and doing test in another PR

Issue #122, opened 2019-07-17
Compile template in Typescript


ankitjena:
Current template is missing tsconfig.json and hence has issues with typescript compilation. Plus resolvers are missing strong typings.

PR #123, opened 2019-07-17
fix: make apollo-template compile in typescript

ankitjena:
Fix for #122
- add tsconfig.json, build script and configuration
- Fix typescript compile errors
- Create Context interface for resolver context
```
export interface Context {
  pubsub: RedisPubSub
  req: express.Request
  db: knex<any, unknown[]>
}
```
- Typings in resolvers
```
key: (_: any, args: any, context: Context) => { }
```

### DIFF
--- a/packages/graphback/src/resolvers/knex/resolverImplementation.ts
+++ b/packages/graphback/src/resolvers/knex/resolverImplementation.ts
@@ -2,7 +2,7 @@ const knexContext = `context.db`
 const pubsub = `context.pubsub`
 
 export const createTemplate = (fieldName: string, tableName: string, typeName: string): string => {
-    return `${fieldName}: async (_: any, args: any, context: Context) => {
+    return `${fieldName}: async (_: any, args: any, context: GraphQLContext) => {
       const result = await ${knexContext}('${tableName}').insert(args.input).returning('*')
       ${pubsub}.publish(Subscriptions.NEW_${typeName.toUpperCase()}, {
         new${typeName}: result[0]
@@ -12,7 +12,7 @@ export const createTemplate = (fieldName: string, tableName: string, typeName: s
   }
 
 export const deleteTemplate = (fieldName: string, tableName: string, typeName: string): string => {
-    return `${fieldName}: (_: any, args: any, context: Context) => {
+    return `${fieldName}: (_: any, args: any, context: GraphQLContext) => {
       return ${knexContext}('${tableName}').where('id', '=' , args.id).del().then( () => {
         ${pubsub}.publish(Subscriptions.DELETED_${typeName.toUpperCase()}, {
           deleted${typeName}: args.id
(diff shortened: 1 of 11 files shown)


wtrocki:
So this is the place where I would love to see cli being used to generate code and at least check if it will compile

templates/apollo-starter-ts/tslint.json
```diff
@@ -14,7 +14,8 @@
       "object-literal-sort-keys": false,
       "trailing-comma": false,
       "object-literal-shorthand": false,
-      "no-console": false
+      "no-console": false,
+      "interface-name": false
```
wtrocki: Reason?

templates/apollo-starter-ts/package.json
```diff
@@ -3,9 +3,11 @@
   "version": "0.3.0",
   "description": "",
   "private": true,
-  "main": "dist/index.js",
+  "main": "dist/src/index.js",
   "scripts": {
-    "start": "ts-node src/index.ts"
+    "start": "node dist/src/index.js",
+    "build": "tsc",
+    "dev": "ts-node src/index.ts"
```
wtrocki: I will add the watch option.

packages/graphback/src/resolvers/knex/resolverImplementation.ts
```diff
@@ -2,7 +2,7 @@ const knexContext = `context.db`
 const pubsub = `context.pubsub`
 
 export const createTemplate = (fieldName: string, tableName: string, typeName: string): string => {
-    return `${fieldName}: async (_, args, context) => {
+    return `${fieldName}: async (_: any, args: any, context: Context) => {
```
wtrocki: Where is context going to be defined? Use something like `GraphBackContext | any`

wtrocki:
We need additional script - can be npm or bash that will execute command line tool on build and generate project. Then we should check if the generated project compiles (follow up will be if it actually works)

packages/graphback/src/resolvers/knex/resolverImplementation.ts
ankitjena: It will be defined in the context.ts in the template

ankitjena:
> We need additional script - can be npm or bash that will execute command line tool on build and generate project. Then we should check if the generated project compiles (follow up will be if it actually works)

When we run npm run build at the root of the project, it also executes the build commands in templates. If there is any error, the build will fail. Thus we can see if the template actually compiles. As we have made changes for the type error in resolvers

packages/graphback/src/resolvers/resolverTemplate.ts
```diff
@@ -1,7 +1,11 @@
 import { TargetResolverContext } from './knex/targetResolverContext';
 
+const imports = `import { Context } from '../src/context'`
```
ankitjena: Doing the import here

templates/apollo-starter-ts/tslint.json
ankitjena: Demanding resolver name to start with "I"

packages/graphback/src/resolvers/resolverTemplate.ts
wtrocki: Rename it to something like GeneratedContext or something similar.

wtrocki:
> When we run npm run build at the root of the project, it also executes the build commands in templates. If there is any error, the build will fail. Thus we can see if the template actually compiles. As we have made changes for the type error in resolvers

Yes, but do we generated resolvers on it? This is very simple task we just need to generate resolvers on top of the sample template to see if this compiles as whole (t did not before so let's do it)

ankitjena:
> > When we run npm run build at the root of the project, it also executes the build commands in templates. If there is any error, the build will fail. Thus we can see if the template actually compiles. As we have made changes for the type error in resolvers
> 
> Yes, but do we generated resolvers on it? This is very simple task we just need to generate resolvers on top of the sample template to see if this compiles as whole (t did not before so let's do it)

Then we would need a example model file to generate resolvers. Not sure if I am getting this right but we can have script like
```
graphback db && tcs
```
So we generate resolvers and then compile. Just the template will need to have a model by default.

wtrocki:
I think we can just copy model from CLI right?

ankitjena:
> I think we can just copy model from CLI right?
> 
> WDYT:
> 
> ```
> cp {model} ./model
> graphback generate
> graphback db 
> ```

I think we should do this in integration test in cli. And if we do strict typing on generated resolvers it will compile as well. Plus it will make the build command the user can reuse, which won't alter the generated files.

PR #127, opened 2019-07-18
fix: change scripts

ankitjena:
Currently `npm start` runs the built js files. Since `graphback watch` runs `npm start`(currently not working as rebuild is not done),
- changed `start` script to `ts-node`
- add `serve` script for built files


wtrocki:
Documentation will be cool. Merging to get things fixed

PR #145, opened 2019-07-22
Relation support in graphback

### LINKED ISSUE #130: Relationship support in schema and resolvers
## Background
`graphback` version `0.3.1` creates database structure for relationships.

## TODO
Add relationships in `schema` and `resolvers`
For a model
```
type Note {
    id: ID!
    title: String!
    comments: [Comment!]!
}

type Comment {
    id: ID!
    content: String!
}
```
we can create the schema with types like
```
type Note {
    id: ID!
    title: String!
    comments: [Comment]!
}

type Comment {
    id: ID!
    content: String!
    noteId: ID!
}
```
Not sure how to work with n:m relations for now
Resolvers can be implemented using either
- Dataloader
- Graphql-query-mapper

@wtrocki What should be the methods we should provide for relationships? like `findXsOfY` for example.

ankitjena:
## Work done
Added relationships info to schema generation

### Example
```
type Note {
  id: ID!
  title: String!
  description: String!
  ## Relationship
  comment: [Comment]!
}

type Comment {
  id: ID!
  title: String!
  description: String!
}
```
output schema looks like
```
type Note {
  id: ID!
  title: String!
  description: String!
  comment: [Comment]!
}

type Comment {
  id: ID!
  title: String!
  description: String!
  noteId: ID!
}

input NoteInput {
  title: String!
  description: String!
}

input CommentInput {
  title: String!
  description: String!
  noteId: ID!
}

input NoteFilter {
  id: ID
  title: String
  description: String
}

input CommentFilter {
  id: ID
  title: String
  description: String
  noteId: ID
}
......
```
Changes in output resolvers
```
Note: {
  comments: (parent: any, _: any, context: GraphQLContext) => {
    return context.db.select().from('comment').where('noteId', '=', parent.id)
  }
},
......
```
Resolves #130

### DIFF
--- a/packages/graphback/src/resolvers/resolverTemplate.ts
+++ b/packages/graphback/src/resolvers/resolverTemplate.ts
@@ -3,8 +3,32 @@ import { TargetResolverContext } from './knex/targetResolverContext';
 const imports = `import { GraphQLContext } from '../src/context'`
 
 export const generateResolvers = (context: TargetResolverContext): string => {
-  return `${imports}
+  if(context.relations.length) {
+    return `${imports}
+  
+enum Subscriptions {
+  ${context.subscriptionTypes.join(',\n  ')}
+}
 
+export const resolvers = {
+  ${context.relations.join(',\n\n  ')},
+
+  Query: {
+    ${context.queries.join(',\n    ')}
+  },
+
+  Mutation: {
+    ${context.mutations.join(',\n    ')}
+  },
+
+  Subscription: {
+    ${context.subscriptions.join(',\n    ')}
+  }
+}
+`
+  } else {
+    return `${imports}
+  
 enum Subscriptions {
   ${context.subscriptionTypes.join(',\n  ')}
 }
@@ -23,4 +47,5 @@ export const resolvers = {
   }
 }
 `
+}
 }
--- a/packages/graphback-cli/package.json
+++ b/packages/graphback-cli/package.json
@@ -59,7 +59,7 @@
     "execa": "2.0.3",
     "figlet": "1.2.3",
     "glob": "7.1.4",
-    "graphback": "^0.3.1",
+    "graphback": "0.3.1",
     "inquirer": "6.5.0",
     "node-emoji": "1.10.0",
     "ora": "3.4.0",
--- a/packages/graphback/src/resolvers/knex/resolverImplementation.ts
+++ b/packages/graphback/src/resolvers/knex/resolverImplementation.ts
@@ -68,15 +68,15 @@ export const deletedSub = (typeName: string): string => {
     }`
 }
 
-export const typeRelation = (relation: string, typeName: string, fieldName: string, tableName: string): string => {
+export const typeRelation = (relation: string, columnName: string, fieldName: string, tableName: string): string => {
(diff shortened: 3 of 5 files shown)


wtrocki:
Is this PR going to have implementation for resolvers

ankitjena:
> Is this PR going to have implementation for resolvers

Yes

packages/graphback/src/resolvers/resolverTemplate.ts
```diff
@@ -3,8 +3,32 @@ import { TargetResolverContext } from './knex/targetResolverContext';
...
-  return `${imports}
+  if(context.relations.length) {
+    return `${imports}
+  
+enum Subscriptions {
+  ${context.subscriptionTypes.join(',\n  ')}
+}
+
+export const resolvers = {
+  ${context.relations.join(',\n\n  ')},
```
wtrocki: I would see them on bottom actually

wtrocki:
>    return context.db.select().from('comment').where('noteId', '=', parent.id)

Newbie question
How we know that `parent.id` is the way to filter comments. 
How this will work with a relationship table - it might be tricky.

wtrocki:
Need some time to play with this to check how it works with all the cases. 
Generally, the code looks sane. I will suggest to not block you and go 

We might badly need integration tests as things can break quite fast with the amount of features we will add so verification could take very long time. 
WDYT?

ankitjena:
> > return context.db.select().from('comment').where('noteId', '=', parent.id)
> 
> Newbie question
> How we know that `parent.id` is the way to filter comments.
> How this will work with a relationship table - it might be tricky.

It's the way we deal with the database. For 1:1 and 1:m we create column. And for how we create these resolvers the approach is known

ankitjena:
> Need some time to play with this to check how it works with all the cases.
> Generally, the code looks sane. I will suggest to not block you and go
> 
> We might badly need integration tests as things can break quite fast with the amount of features we will add so verification could take very long time.
> WDYT?

Yes we should add integration tests to cli

wtrocki:
> How we know that parent.id is the way to filter comments.

I mean here that parent type very often not have `id` as field at all.
We need to get it from relationship field (in fact we have this in directive already)

ankitjena:
> > How we know that parent.id is the way to filter comments.
> 
> I mean here that parent type very often not have `id` as field at all.
> We need to get it from relationship field (in fact we have this in directive already)

Well `id` will be present in database as a column and for the cases we are fetching all the columns from database now. But I see what you mean - that when we use selective querying. 

Didn't quite get the directive reference.

wtrocki:
type may not have id. It is just that our sample model have it :)

wtrocki:
Let's update docs as well

ankitjena:
> type may not have id. It is just that our sample model have it :)

But the postgres table would have

wtrocki:
ah.. ok so this is a bug on postgress. If we specify ID on field it should be added. Lets go with this for now

ankitjena:
> ah.. ok so this is a bug on postgress. If we specify ID on field it should be added. Lets go with this for now

Not a bug exactly. We are creating that in `table.primary`

PR #149, opened 2019-07-23
fix: add missing documentation

wtrocki:
Feedback provided shown us that we are missing critical parts of the docs and it is not clear how graphback works. This should do the job.

### DIFF
--- a/docs/working.md
+++ b/docs/working.md
@@ -2,7 +2,29 @@
 id: working
 title: How Graphback works
 ---
+## GraphQL Schema Input
+
+Graphback is processing GraphQL Schema DSL language to generate server and client side artifacts to speed up development.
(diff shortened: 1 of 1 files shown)


docs/working.md
```diff
@@ -11,10 +33,23 @@ Currently we provide `apollo-typescript-starter` only, and we'll be supporting m
...
+
+`./model` folder will contain input schema
+`./resolvers` folder will contain generated and custom GraphQL resolvers for the project.
+
 ## Databases
 Graphback allows you to configure your database according to your needs. Graphback provides support to multiple databases(MySQL, MongoDB etc).
 Currently, we are supporting `PostgreSQL`.
 
+
+## GraphQL Modules
```
ankitjena: This is to be done

wtrocki: Yes.. we can mention that in documentation to give overview.

PR #150, opened 2019-07-23
fix: add generation configuration to config

wtrocki:
Let's chat about this? WDYT?
Addressing community feedback here.


packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -16,6 +16,17 @@ const pgDockerCompose = () => {
   return readFileSync(`${dockerFilesPath}/pg-docker-compose.yml`, 'utf8')
 }
 
+const generationConfig = `{
```
ankitjena: Are we providing user the option to pass config through the `config.json` file and not use directives? Because we already have default config in the core.

wtrocki: We still need to use directives, but you can imagine schema looking very very verbose if we stat to disable every type. We need a global config for generation and then individual types

ankitjena: So let's have directives for applying stuff and users can disable stuff in the config

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -16,6 +16,17 @@ const pgDockerCompose = () => {
...
+  "create": true,
+  "update": true,
```
ankitjena: We should keep `delete` as `true` and maybe `update` as `false`

packages/graphback-cli/src/templates/configTemplates.ts
ankitjena: This seems good as config, and we can include pagination in this too if we want later. But what should be the directives we provide now `@sub(new, update) @paginate` etc

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: hmm.. Delete is very rarely used. Almost never - usually people have something like archive data. Data is very rarely deleted in production systems. 

Update is the main method being used.

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: Yes.. in fact if you look into TODO config already have it so I will add it here as well

wtrocki:
@ankitjena Leaving that for you

PR #155, opened 2019-07-24
Sofa integration

wtrocki:
## Motivation

Not everyone is fully into GraphQL. While using DSL as a way to model data seems very cool and people like it sometimes there is need to use rest for some clients where GraphQL cannot be used (embedded etc.) Adding new template that will offer both REST and GraphQL options. index.ts has main changes.

We are using here Sofa from @urigo that makes this task very simple. 
Going to follow up with testing this template with production ready schemas we have to see how this plays up and if we need to add much more later.

## Verification
```
npm run start
curl http://localhost:4000/rest/me
```

### DIFF
--- a/docs/working.md
+++ b/docs/working.md
@@ -2,6 +2,14 @@
 id: working
 title: How Graphback works
 ---
+
+## Introduction
+
+The target of GraphBack is to bind industry templates and sample applications with business logic.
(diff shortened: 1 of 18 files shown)


Urigo:
that is AMAZING!
the vision is happening!

Issue #157, opened 2019-07-24
Sample model (Note) Missing relationships


wtrocki:
Our sample Notes model - generated on init do not have any relationships we can showcase. It is important to have a relationship by default for people to try things

ankitjena:
It has relationship, just missing directives - we are doing relationships without directives too

Issue #154, opened 2019-07-24
Templates in CLI are hardcoded to master branch


wtrocki:

wtrocki:
@ankitjena We can simply extract templates config to json and patch it on release - see release scripts.

wtrocki:
Lets do release and leave this as it is. In future templates should not change that much

PR #163, opened 2019-07-25
fix: change cli templates source to graphback-cli branch

ankitjena:
#154 #157
Change template source to `graphback-cli` branch


wtrocki:
How this will work now? Are we pushing to this branch on release? I think that:
- we should keep master for dev
- we should swap this on release to tagname - I have done that manually but we can have some script

Issue #159, opened 2019-07-24
Templates should use graphql playground


wtrocki:
Playground got better. Additionally we should generate example queries and mutations or playground for people to explore

PR #166, opened 2019-07-26
fix: change playground from altair to graphql-playground

ankitjena:
#159

### DIFF
--- a/templates/apollo-rest-starter/package.json
+++ b/templates/apollo-rest-starter/package.json
@@ -12,7 +12,6 @@
   },
   "license": "Apache 2.0",
   "dependencies": {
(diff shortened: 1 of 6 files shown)


imolorhe:
Hey, just out of curiosity (and not to start any conversations around this 😊), may I ask why the move from Altair to Playground?

wtrocki:
@imolorhe Things were repeatedly broken and fixed with releases. 
For example with the current release, when having more than one query in tab - selection option in execution do not appear. I have seen issue on the repository, so we are really happy to restore it after that is fixed

imolorhe:
Okay. Thanks for the explanation!

PR #170, opened 2019-07-29
Change output resolvers format

ankitjena:
## Implementation
- Divide resolvers into types based on user's input
- change template structure

## Structure
```
|----src
      |----resolvers
            |----generated
            |----custom
            index.ts
      |----schema
           generated.ts
```

### Schema output
```
type Query {
    findNotes(fields: NoteFilter!): [Note!]!
    findComments(fields: CommentFilter!): [Comment!]!
    findSubscriptions(fields: SubscriptionFilter!): [Subscription!]!
    findAllNotes: [Note!]!
    findAllComments: [Comment!]!
    findAllSubscriptions: [Subscription!]!
    ## Custom queries
    getLikedNotes: [Note]!
    getPublishedNotes: [Note]
  }
```
### Resolver output
```
export const customResolvers = {
  Query: {
    getLikedNotes: (_: any, args: any, context: GraphQLContext) => {
      // Implementation here
    },
    getPublishedNotes: (_: any, args: any, context: GraphQLContext) => {
      // Implementation here
    }
  },
......
```


ankitjena:
This PR changes the structure of templates. We need to deal with #154 first I think.

packages/graphback/src/resolvers/outputResolvers/apollo.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const generateTypeResolvers = (context: TargetResolverContext, name: string): string => {
+  if(context.relations.length && context.subscriptions.length) {
+    return `${imports}
+  
+enum Subscriptions {
+  ${context.subscriptionTypes}
+}
+
+export const ${name}Resolvers = {
```
wtrocki: This should be rather part of the template I think. The template should shape resolvers to the format that is required.

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -42,17 +42,26 @@ export async function generateBackend(): Promise<void> {
...
 
     writeFileSync(outputSchemaPath, generated.schema)
-    writeFileSync(outputResolverPath, generated.resolvers)
+
+    writeFileSync(`${outputResolverPath}/index.ts`, generated.resolvers.index)
+    
+    if(!existsSync(`${outputResolverPath}/custom`)) {
+      mkdirSync(`${outputResolverPath}/custom`)
+    }
+    //tslint:disable-next-line
```
wtrocki: why?

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -42,17 +42,26 @@ export async function generateBackend(): Promise<void> {
...
+    generated.resolvers.resolvers.forEach((output: any) => writeFileSync(`${outputResolverPath}/generated/${output.name}.ts`, output.output))
+    
   } catch (err) {
     logError(err)
+    process.exit(0)
```
wtrocki: ?? :>

packages/graphback/src/resolvers/outputResolvers/apollo.ts
ankitjena: Yes this would be for apollo-server

packages/graphback-cli/src/helpers/generate.ts
ankitjena: because in case of error, the successive message were displayed so

packages/graphback/src/resolvers/outputResolvers/apollo.ts
ankitjena: When there will be multiple templates, I was thinking of having a key in the template and use respective template like apollo for now

ankitjena: > This should be rather part of the template I think. The template should shape resolvers to the format that is required.

How would that work, in the cli ?

packages/graphback-cli/src/helpers/generate.ts
ankitjena: fixed this

packages/graphback/src/resolvers/outputResolvers/apollo.ts
wtrocki: Let's say we generate resolvers using Apollo model and this will not change. From Apollo Model we can move to graphql.js model easily but not vice versa. Then Graphq.js template will just transform it in code.

packages/graphback/src/resolvers/knex/targetResolverContext.ts
```diff
@@ -78,51 +105,63 @@ const createRelationResolvers = (types: string[], relations: RelationImplementat
...
+          implementation: knex.typeRelation('OneToMany', columnName, f.name, f.type.toLowerCase())
+        })
       }
-    })
-  });
+    }
+  })
 
-  const context = {
+  const typeContext = {
```
ankitjena: This has the details for each type

packages/graphback/src/resolvers/knex/targetResolverContext.ts
```diff
@@ -78,51 +105,63 @@ const createRelationResolvers = (types: string[], relations: RelationImplementat
...
+  typeContext.mutations = [createResolver(context), updateResolver(context), deleteResolver(context)].filter((s: string) => s!==undefined)
+  typeContext.subscriptions = [newSub(context), updatedSub(context), deletedSub(context)].filter((s: string) => s!==undefined)
+  typeContext.subscriptionTypes = createSubscriptionTypes(context)
+
+  return typeContext
+}
+
+export const buildResolverTargetContext = (inputContext: Type[]) => {
+  const output: TypeContext[] = []
+  inputContext.forEach((t: Type) => {
```
ankitjena: For graphql-js we can simply spread this like you said

templates/apollo-starter-ts/src/schema/generated.ts
```diff
@@ -1,7 +1,12 @@
 import gql from 'graphql-tag'
 
 export const typeDefs = gql`
+type User {
```
wtrocki: Why this is published!

packages/graphback/src/schema/schemaTemplate.ts
```diff
@@ -30,53 +30,107 @@ const filters = (defs: TargetType[]): string => {
 
 const imports = `import gql from 'graphql-tag'`
 
+/**
+ * Patch together generated and custom queries
+ * @param queries queries generated from the types (CRUD)
+ * @param custom custom queries inputted by the user
+ */
+const generateQueries = (queries: string[], custom: Field[]) => {
+  if(custom.length) {
```
wtrocki: This code can be much simpler. We can calculate this map beforehand and have:
```
${queries.join('\n')}
${customQueries.join('\n')}
```

packages/graphback/src/schema/schemaTemplate.ts
```diff
@@ -30,53 +30,107 @@ const filters = (defs: TargetType[]): string => {
...
+ * Patch together generated and custom mutations
+ * @param mutations mutations generated from the types (CRUD)
+ * @param custom custom mutations inputted by the user
+ */
+const generateMutations = (mutations: string[], custom: Field[]) => {
+  if(custom.length) {
+    return `type Mutation {
+    ${mutations.join('\n    ')}
+    ## Custom mutations
+    ${custom.map((f: Field) => {
```
wtrocki: I think this will be hard to maintain. Let's do all the required preprocessing before.

packages/graphback/src/schema/targetSchemaContext.ts
```diff
@@ -86,6 +85,16 @@ const maybeNullField = (f: Field) => {
   }
 }
 
+export const maybeNullFieldArgs = (f: Field) => {
```
wtrocki: Missing comment what this does.

wtrocki: Worried that all this code comes with no unit tests. It is hard for me to test it now with all possible cases. We should have unit tests for such examples.

wtrocki:
I will wait for responses and then verify.

packages/graphback/src/schema/schemaTemplate.ts
ankitjena: Yeah but since those are string templates, there will be unrequired whitespaces in the output in case of no custom input

packages/graphback/src/schema/schemaTemplate.ts
ankitjena: I think if else is required, but doing the preprocessing

wtrocki:
@ankitjena We are Missing custom method in sample model so we can demo it instantly.

ankitjena:
> @ankitjena We are Missing custom method in sample model so we can demo it instantly.

Yes, it's also missing directives at the moment(for customization)

wtrocki:
What kind of directives we talking about?

wtrocki:
This is one of the best PR's I have seen in my life. Really good work. You have applied all.my feedback and we got results. I'm happy after very quick testing so we can integrate.it now and I will play on top of the master

PR #171, opened 2019-07-29
fix: removed package bug in templates

ankitjena:


wtrocki:
UUUUU.... We really need integration tests on build!

PR #175, opened 2019-08-01
fix: missing generated folder

ankitjena:
Hadn't added `.gitkeep`, generated folder was not present in template


wtrocki:
This is good quick fix, but CLI should handle that itself.

PR #181, opened 2019-08-03
update docs

ankitjena:
Add documentation about generation config, directives and custom generation with examples.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,0 +1,43 @@
+---
+id: config
+title: CRUD Config
+---
+
+GraphBack takes input model and generates CRUD operations as queries and mutations.
+ These include `create`, `update`, `delete`, `find` and `findAll`. These are generated for each `type` in your model.
+ Further Graphback also generates three predefined subscriptions, namely `new`, `updated` and `deleted`.
+  
+ Graphback allows user to customize the generation process by using `configuration` and `directives`.
+
+## Configuration
+Graphback uses flags to allow user to choose between the CRUD operations and subscriptions. These are present in `config.json`, created during `init`,
+in the root of your project folder under the `generation` key. The default config is
+```json
+{
+  ...,
+  "generation": {
+    "create": true,
+    "update": true,
+    "findAll": true,
+    "find": true,
+    "delete": false,
+    "subCreate": false,
+    "subUpdate": false,
+    "subDelete": false
+  }
+}
+```
+Changing these flags and performing `generate`, regenerates your `schema` and `resolvers` with provided config.
+>Note: For subscriptions, the user needs to change value of respective operation to `true`. For example, changing
+`subDelete` to `true` won't work unless, `delete` is `true`.
+
+## Directives
+Changing the generator config applies the config to all the types in your schema. Graphback allows you change these for any single type using directives. All the config flags are available as directives - `@create`, `@update`, `@delete`, `@find`, `@findAll`, `@subCreate`, `@subUpdate` and `@subDelete`.
+
+User can use these directives to have more control on individual elements. For example, 
+```
+type Note @delete {
+  ...
+}
+```
+will create the `delete` mutation for `Note` type only.
--- a/docs/config.md
+++ b/docs/config.md
@@ -4,8 +4,9 @@ title: CRUD Config
 ---
 
 GraphBack takes input model and generates CRUD operations as queries and mutations.
- These include `create`, `update`, `delete`, `find` and `findAll`. These are generated for each `type` in your model.
+ These include `create`, `update`, `delete`, `find` and `findAll`. These operations can be generated for each `type` in your model.
  Further Graphback also generates three predefined subscriptions, namely `new`, `updated` and `deleted`.
+We can use them in clients to receive live updates for every change that is happening on the server.
   
  Graphback allows user to customize the generation process by using `configuration` and `directives`.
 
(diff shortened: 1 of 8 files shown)


docs/config.md
```diff
@@ -0,0 +1,43 @@
+---
+id: config
+title: CRUD Config
+---
+
+GraphBack takes input model and generates CRUD operations as queries and mutations.
+ These include `create`, `update`, `delete`, `find` and `findAll`. These are generated for each `type` in your model.
```
wtrocki:
```suggestion
 These include `create`, `update`, `delete`, `find` and `findAll`. These operations can be generated for each `type` in your model.
```

docs/config.md
```diff
@@ -0,0 +1,43 @@
...
+ Further Graphback also generates three predefined subscriptions, namely `new`, `updated` and `deleted`.
```
wtrocki:
```suggestion
 Further Graphback also generates three predefined subscriptions, namely `new`, `updated` and `deleted`.
 We can use them in clients to receive live updates for every change that is happening on the server.
```

docs/config.md
```diff
@@ -0,0 +1,43 @@
...
+    "find": true,
+    "delete": false,
+    "subCreate": false,
+    "subUpdate": false,
+    "subDelete": false
+  }
+}
+```
+Changing these flags and performing `generate`, regenerates your `schema` and `resolvers` with provided config.
+>Note: For subscriptions, the user needs to change value of respective operation to `true`. For example, changing
```
wtrocki:
```suggestion
> Note: For subscriptions, the user needs to change the value of the respective operations to `true`. For example, changing 
```

docs/config.md
```diff
@@ -0,0 +1,43 @@
...
+`subDelete` to `true` won't work unless, `delete` is `true`.
+
+## Directives
+Changing the generator config applies the config to all the types in your schema. Graphback allows you change these for any single type using directives. All the config flags are available as directives - `@create`, `@update`, `@delete`, `@find`, `@findAll`, `@subCreate`, `@subUpdate` and `@subDelete`.
```
wtrocki:
```suggestion
Changing the generator config applies the config to all the types in your schema. Graphback allows you to change these for any single type using directives. All the config flags are available as directives 

- `@create`
- `@update` 
- `@delete` 
- `@find`
- `@findAll`
- `@subCreate`
- `@subUpdate` 
- `@subDelete`
```

docs/config.md
```diff
@@ -0,0 +1,43 @@
...
+
+User can use these directives to have more control on individual elements. For example, 
+```
+type Note @delete {
+  ...
+}
+```
+will create the `delete` mutation for `Note` type only.
```
wtrocki:
```suggestion
will create only the `delete` mutation for the `Note` type.
```

docs/config.md
```diff
@@ -0,0 +1,43 @@
...
+Changing these flags and performing `generate`, regenerates your `schema` and `resolvers` with provided config.
+>Note: For subscriptions, the user needs to change value of respective operation to `true`. For example, changing
+`subDelete` to `true` won't work unless, `delete` is `true`.
+
+## Directives
+Changing the generator config applies the config to all the types in your schema. Graphback allows you change these for any single type using directives. All the config flags are available as directives - `@create`, `@update`, `@delete`, `@find`, `@findAll`, `@subCreate`, `@subUpdate` and `@subDelete`.
+
+User can use these directives to have more control on individual elements. For example, 
+```
+type Note @delete {
```
wtrocki: Missing: Disabling defaults. I think we should document that directives override config so if you pass delete if will override delete from config, but how we can disable them if the global config is true.

docs/config.md
wtrocki: Is that true? I was thinking that we going to use directives to change global flag. 
This is very counter intuitive to what normally people will expect.

docs/custom.md
```diff
@@ -0,0 +1,60 @@
...
+  id: ID!
+  name: String!
+}
+```
+and we want to implement a mutation to like a Post, then we can simply
+```graphql
+...
+...
+
+#Custom methods
```
wtrocki:
```suggestion
#Custom mutations
```

docs/custom.md
```diff
@@ -0,0 +1,60 @@
...
+export const likePost = {
+  Mutation: {
+    likePost: (_: any, args: any, context: GraphQLContext) => {
+        // Implementation here
+    }
+  }
+}
+```
+
+>This works similarily for `Query` and `Subscription`.
```
wtrocki:
```suggestion
> **Note:** This works similarily for `Query` and `Subscription`.
```

wtrocki:
I have found some issues, but they might be related to implementation. Please check comments and suggestions.

wtrocki:
If this really work as documented we can merge that and then work on improving it in code and changing docs.

docs/config.md
ankitjena: generator config applies to all types and directives add certain operation to a single type.

docs/config.md
ankitjena: Yes adding that

wtrocki:
Yes. I think we have bug. I have true in config but want to disable this using directive. WDYT?

ankitjena:
> Yes. I think we have new feature here. I have true in config but want to disable this using directive. WDYT?

Use directives to invert the config value?

wtrocki:
Yup.. We can always enable it by adding directive but disabling is not there

wtrocki:
Minor priority though for the moment

PR #184, opened 2019-08-05
Add sqlite3 database

ankitjena:
## Work done
- add another option in `init` command to choose database
Screenshot from 2019-08-05 16-22-29
- modify resolver output according to database selection, currently only the create method for `mysql` and `sqlite`


packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -1,22 +1,30 @@
...
-  name: 'postgres',
-  config: readFileSync(`${configFilesPath}/pgconfig.json`, 'utf8')
-}
-
-const dockerFilesPath = `${__dirname}/resources/docker`
 
-const pgDockerCompose = () => {
-  return readFileSync(`${dockerFilesPath}/pg-docker-compose.yml`, 'utf8')
+const getConfig = (database: string) => {
+  if(database === 'pg') {
```
wtrocki: I would preffer postgress

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -16,12 +19,18 @@ const handleError = (err: { code: string; message: string; }): void => {
...
   try {
-    const { dbConfig } = JSON.parse(readFileSync(`${process.cwd()}/config.json`, 'utf8'))
-    const manager = new PostgresSchemaManager(dbConfig);
-    // tslint:disable-next-line: await-promise
-    await manager.getConnection().raw('DROP SCHEMA public CASCADE;')
-    // tslint:disable-next-line: await-promise
-    await manager.getConnection().raw('CREATE SCHEMA public;')
+    const { database, dbConfig } = JSON.parse(readFileSync(configPath, 'utf8'))
+    if(database === 'sqlite3') {
+      if(existsSync(`${process.cwd()}/db.sqlite`)) {
```
wtrocki: I think we need configuration where this DB will be placed relative to project.
Not in this PR but great to have separate issue

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -36,21 +45,18 @@ export const createDBResources = async(): Promise<void> => {
...
-    const configPath = `${process.cwd()}/config.json`
-
-    const file = readFileSync(configPath, "utf8");
-    let genConfig = {};
-    if (file) {
-      genConfig = JSON.parse(file).generation
+    const { database, dbConfig, generation } = JSON.parse(readFileSync(configPath, "utf8"))
+    
+    if(database === 'sqlite3') {
+      await execa('touch', ['db.sqlite'])
```
wtrocki: Creating empty file seems strange. Maybe driver has option to created it on conenction

packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * Base resolver class for knex implementation with implementations that
+ * doesn't differ across various relation database like sqlite3 or postgresql
+ */
+
+export class BaseKnexResolver {
```
wtrocki: 😍

packages/graphback/src/resolvers/knex/KnexResolver.ts
```diff
@@ -0,0 +1,47 @@
+import { BaseKnexResolver } from './BaseKnexResolver'
+
+/**
+ * Resolver implementation for some operations which vary according to database,
+ * extends all the methods of the Base Resolver class
+ * For now only includes the `create` method
+ */
+export class KnexResolver extends BaseKnexResolver {
+  public createTemplate = (database: string, subscription: boolean, fieldName: string, tableName: string, typeName: string): string => {
+    if(database === 'sqlite3' || database === 'mysql') {
```
wtrocki: Instead of if's maybe we can have separate classes. 
How much duplication it will be?

wtrocki: Just in case we will want to test it later etc. :D

templates/apollo-starter-ts/package.json
```diff
@@ -20,7 +20,8 @@
     "graphql-tag": "2.10.1",
     "ioredis": "4.14.0",
     "knex": "0.19.1",
-    "pg": "7.12.0"
+    "pg": "7.12.0",
```
wtrocki: I think that is the smell that we want to address. We do not want every template 
to have access to library that it might not use. Additionaly - template will have postgress related code. I remember we talked about this before so we might drop some new requirements or create new issue to cover this problem

wtrocki:
WOW! Thinks are improving so fast! Really great job! 
I'm astonished by quality of your PR. This one is true masterpiece. 
Really like the new architecture config approach. Logged couple questions and suggestions but nothign major. I will wait with verification after you response.

Graphback for the win!

templates/apollo-starter-ts/package.json
ankitjena: I think we may have to inject package.json too, like other files. And the only config it depends upon can be imported from the config.

wtrocki: Or have list of depenencies that will be extecuted on top of npm instal. Lets create issue to chat about it. I think having extra dependencies could be cool

ankitjena: Oh yes I didn't think of that

packages/graphback/src/resolvers/knex/KnexResolver.ts
ankitjena: Methods or classes ?

packages/graphback-cli/src/templates/configTemplates.ts
ankitjena: Kept it as `pg` since knex client going to use it, in display it's `PostgreSQL`

templates/apollo-starter-ts/package.json
ankitjena: The only files that were postgres specific were `config.ts` and `db.ts`. Used `resolveJsonModule` in the recent commit to read config from the `config.json` file

packages/graphback-cli/src/helpers/db.ts
ankitjena: Didn't find any option in `knex`

packages/graphback/src/resolvers/knex/KnexResolver.ts
wtrocki: Classes - we can use strategy pattern and inject different provider based on config. That was initial idea

PR #188, opened 2019-08-05
feat: Add new Tasks schema

wtrocki:
# Motivation

I wanted to see if we can use graphback with showcase schema along with some extra features like enums. Purpose is to create this template that will help us to see how showcase schema can evolve around graphback.

CC: @Lavanyagaur22 

## Verification
No linking is needed - using a globally released version is desired.

```
cd graphback-cli 
npm run cmd -- init test   
```

Follow classical generation flow


wtrocki:
@Lavanya Feel free to PR against this to extend schema for your own needs

packages/graphback-cli/src/templates/resources/models/Tasks.graphql
```diff
@@ -0,0 +1,32 @@
...
+
+type User {
+  id: ID!
+  firstName: String!
+  lastName: String!
+  title: String!
+  email: String!
+}
+
+enum TaskStatus {
```
ankitjena: We are not supporting enums now as visitor is currently not parsing enum. This will be good feature to add

Lavanyagaur22:
@wtrocki @ankitjena For now probably we can use strings instead of enum and it should work. I'll make a pr for it if it sounds good to you.

ankitjena:
@Lavanyagaur22 Don't understand what you mean by using strings, an example maybe ?

Lavanyagaur22:
I mean for the task schema @wtrocki added, as it doesn't support enum at the time being, we can just simply keep `status: String!`.

wtrocki:
@Lavanyagaur22 Yeah. I'm going to integrate this soon so feel free to PR against master. Any changes are welcomed.

wtrocki:
@ankitjena This is failing now: 
```
╰─$ graphback generate
TypeError: Cannot read property 'forEach' of undefined

```

wtrocki:
Checking what is the problem. Leave that with me

packages/graphback-cli/src/templates/resources/models/Tasks.graphql
```diff
@@ -0,0 +1,31 @@
...
+  id: ID!
+  firstName: String!
+  lastName: String!
+  title: String!
+  email: String!
+}
+
+## Custom queries and mutations
+type Query {
+  allTasks(first: Int, after: String): [Task],
```
ankitjena: `allTasks` in pretty similar to the `findAll` method that's gonna be generated. Maybe switch to `allCompletedTasks`?

Issue #187, opened 2019-08-05
@disablegen and @rest directive


wtrocki:
`@rest` directive should disable db generation and leave empty implementation.

wtrocki:
Leaving as a good first issue for the community. 
CC @MStokluska

wtrocki:
`@rest` and `@disablegen` directives should have the same behaviour. 
disablegen is needed because sometimes we want to have type 
that is being used only for relational purposes and this will mean that 
we will need to put all possible directives on it to disable it. 
For rest we will keep this behavior temporarily - we can add some extra parameter later and improve generation.

richcorbs:
I don't know how to link this to the PR for this issue but here is the (merged) PR:

https://github.com/aerogear/graphback/pull/214

wtrocki:
We might need discussion how restful ones can work. Making empty resolvers would not justify existence of REST. Lets leave it for the moment and see what community will expect from rest first

PR #193, opened 2019-08-06
Modified Tasks.graphql

Lavanyagaur22:
Added `createdWhen: TimeDetails` and replaced enum with String.

### DIFF
--- a/packages/graphback-cli/src/templates/resources/models/Tasks.graphql
+++ b/packages/graphback-cli/src/templates/resources/models/Tasks.graphql
@@ -4,10 +4,14 @@ type Task {
   title: String!
   description: String!
   status: String!
-  createdBy: User
+  creationMetadata: CreationMetadata
   assignedTo: User
-  createdWhen: TimeDetails
 }
+
+type CreationMetadata {
+   createdDate: String
+   createdBy: User
+ }
 
 type User {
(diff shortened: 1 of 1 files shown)


packages/graphback-cli/src/templates/resources/models/Tasks.graphql
```diff
@@ -3,9 +3,10 @@ type Task {
   version: Int
   title: String!
   description: String!
-  status: TaskStatus
+  status: String!
   createdBy: User
   assignedTo: User
+  createdWhen: TimeDetails
```
wtrocki: WDYT about `created`

packages/graphback-cli/src/templates/resources/models/Tasks.graphql
```diff
@@ -3,11 +3,16 @@ type Task {
...
   title: String!
   description: String!
-  status: TaskStatus
-  createdBy: User
+  status: String!
+  creationMetadata: CreationMetadata
   assignedTo: User
 }
 
+type CreationMetadata {
```
wtrocki: @ankitjena - good example of the model that needs to be `@genignore`'d. 
We use it for relationships only and do not need full crud on this. 
See #187

PR #203, opened 2019-08-07
fix: disable option in directives

ankitjena:
## Background
```
type Test @update {
.....
}
```
will enable the operation while
```
type Test @update(enable: false) {
.....
}
```
will disable the operation.


wtrocki:
Can we use this directive in the Task template? I will test it now

ankitjena:
> Can we use this directive in the Task template? I will test it now

Yes we can

wtrocki:
@ankitjena I meant can you add disable into tasks or Notes schemas so we will have usage of this pattern somewhere.

PR #210, opened 2019-08-09
feat: add graphql-js template

ankitjena:
## Implementation
Modify core to output according to template type. Use a key `templateType` in `.template.json` in the project root.


wtrocki:
Let's chat about this. It works nicely.

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -43,7 +45,7 @@ export async function generateBackend(): Promise<void> {
 
     const backend: GraphQLBackendCreator = new GraphQLBackendCreator(schemaText, generation)
 
-    const generated: IGraphQLBackend = await backend.createBackend(database)
+    const generated: IGraphQLBackend = await backend.createBackend(database, templateType)
```
wtrocki: -1 for template engine to be aware of the templates. This introduces complexity and kinda breaks the onion architecture of the solution.

If we generate resolvers in some specific way that have all the info and then just map output format directl in template we going to prove extensivility.

Developers should be able to create tons of templates without changing CLI or core of graphback.

ankitjena: But then it will output according to what the template needs right ?

packages/graphback/src/resolvers/index.ts
```diff
@@ -19,20 +20,20 @@ export class ResolverGenerator {
...
     }
 
-    return {
-      types: generateResolvers(this.context),
-      index: generateIndexFile(this.context, hasCustomElements),
-      custom: generateCustomResolvers(customContext)
+    if(templateType === 'graphql-js') {
+      return generateGraphQLjsResolvers(this.context, customContext, hasCustomElements)
+    } else {
+      return generateApolloResolvers(this.context, customContext, hasCustomElements)
```
ankitjena: According to the template

packages/graphback-cli/src/helpers/generate.ts
wtrocki: Yeah, but think about complexity that this is putting. 
There could be many templates that we will support with different variations. This is also making things quite hard to test - every change in generation needs to be tested in 2 variations. 


I think we can simply generate resolvers in format that provides a lots of metadata then we can adjust that to new format directly in template. 

For example:
```
{ Queries: { find() }, Mutations: {}
```

can be transformed to
```
{ ...Queries, ...Mutations} 
```

ankitjena: Hmm we are doing that only, we generate metadata using `targetResolverContext` and inject them to `apollo-template` and `graphqljs-template`, respectively.

ankitjena: I removed the bit where we were using template info inside metadata generation in newest commit.

packages/graphback/src/resolvers/index.ts
wtrocki: Yes. This is quite tricky to test and makes the generator much more complex. 
Let's imagine that we want to generate a different format than knex and in this generator, we will need to also have template type covered. This change makes things much more complex and less contributor friendly. Our success depends on extensibility - if it is hard to extend

ankitjena:
> Approved by mistake. Do you think that my arguments are sane?

Yes, what do you suggest then?

ankitjena:
> Yes. This is quite tricky to test and makes the generator much more complex.
Let's imagine that we want to generate a different format than knex and in this generator, we will need to also have template type covered. This change makes things much more complex and less contributor friendly. Our success depends on extensibility - if it is hard to extend

If we would want a different format then knex, similarily we would be generating metadata we are doing here and using it in template. How do we make it extensible. Maybe we can separate the generating metadata and putting it in a template?

templates/graphql-js-starter/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "main": "dist/src/index.js",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "build": "tsc",
+    "serve": "node dist/src/index.js",
+    "watch": "graphback watch"
+  },
+  "license": "Apache 2.0",
+  "dependencies": {
+    "apollo-server-express": "2.8.1",
```
wtrocki: It still has apollo?

templates/graphql-js-starter/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+- Start the server
+```
+npm start
+```
+
+### Directory structure
+
+```
+|-----generated
```
wtrocki: I think we might move this from templates to docs.and remove it here. it is hard to maintain it

packages/graphback/src/resolvers/index.ts
```diff
@@ -29,10 +29,6 @@ export class ResolverGenerator {
       hasCustomElements = true
     }
 
-    return {
-      types: generateResolvers(this.context),
-      index: generateIndexFile(this.context, hasCustomElements),
-      custom: generateCustomResolvers(customContext)
-    }
+    return generateApolloResolvers(this.context, customContext, hasCustomElements)
```
wtrocki: How about not call it apollo but create.interface and document that this format is aligned.with apollo

templates/graphql-js-starter/package.json
ankitjena: This is by mistake.

templates/graphql-js-starter/README.md
ankitjena: The folder structure ?

PR #214, opened 2019-08-10
Adds and uses disableGen Type directive

richcorbs:
This is literally my very first attempt at contributing to open source so please be gentle. :-)

Sorry for the changes reflected in removing trailing spaces. Must be an editor configuration. I didn't do that intentionally.

I did my best to follow the contribution guidelines and everything seemed to run just fine.

I spent an hour or more reasoning through the ticket and walking through code. I **think** I did the right thing.

I tried to break tests by setting `"disableGen": true` in `empty-test.js` but couldn't get the tests to fail so I probably need a nudge in the right direction to update tests.

If `@disableGen` is looking right then I'll try and replicate it for `@rest`.


wtrocki:
Love it! Extra whitespaces are not problem. Those were actually off our standard formatter. From my side this looks like perfect contribution. I really cannot point anything to correct. 

Thank you so much for contribution.

wtrocki:
@ankitjena - WDYT?

PR #224, opened 2019-08-12
fix: reformat string templates, disableGen usage

ankitjena:
## Work done
- restructure schema and resolver templates to not leave empty spaces
- enforce disableGen  for all operations
- add info about disableGen in docs


wtrocki:
Verified using Tasks schema and PostgressDB. 
Verified on an existing project by adding @disablegen again. 

We need release notes mention on this feature.

wtrocki:
Verified using Tasks schema and PostgressDB. 
Verified on an existing project by adding @disablegen again. 

We need release notes mention on this feature.

Issue #209, opened 2019-08-09
Ability to query relationship is for One type only


ankitjena:
## Background
Currently given a relation
```
type User {
    comments: [Comment]
}
```
We are only generating comments field for the `User` type. We should also generate resolver for the 2nd type such that a query like this would be possible
```
query {
    findAllComments {
        user {
            name
        }
    }
}
```

ankitjena:
So if we do
```
type User {
    ....
    comments: [Comment]
}
```
we are generating these in the schema
```
type User {
    ....
    comments: [Comment]
}
```
and 
```
type Comment {
    ....
    userId: ID!
}
```
If we do 
```
type Comment {
    ....
    user: User!
}
```
instead, this will allow to do the query and implement the resolver accordingly.

wtrocki:
Good catch! We do not need any extra generation flag or directive for this as it will be expected to have this capability all the time.

wtrocki:
@richcorbs thank you for logging this issue

richcorbs:
If someone is not already working on this and if I had time to start looking at it where would you suggest I start looking?

richcorbs:
![Screen Shot 2019-08-10 at 1 56 40 PM](https://user-images.githubusercontent.com/437167/62826317-b775da80-bb76-11e9-9e44-21e9b1fd25d8.png)

Is this the right change in `targetSchemaContext.ts`?

richcorbs:
Is this close to the right place and the right thing?

![Screen Shot 2019-08-10 at 2 05 23 PM](https://user-images.githubusercontent.com/437167/62826401-eb053480-bb77-11e9-99d1-d155e5d161b0.png)

ankitjena:
> If someone is not already working on this and if I had time to start looking at it where would you suggest I start looking?

Yes sure, go ahead.

> Is this close to the right place and the right thing?

Yes this is it. 

> If I made changes how would I run the updated graphback to see if my changes worked?

There's a command in project root `bootstrap`, do `npm run bootstrap` after running `npm run build`(after making the changes)

wtrocki:
If you hit any problems with contribution let us know. We can add more info to Contributing.md.
If code changes were made feel free to create PR, our team will help with verification and merge.

richcorbs:
PR #225 submitted.

richcorbs:
[...]

PR #225, opened 2019-08-12
Tweak one-to-one relationships

richcorbs:
This PR tweaks one-to-one relationships to work as described in issue #209. It touches database, schema, and resolvers.

### DIFF
--- a/packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
+++ b/packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
@@ -87,16 +87,23 @@ public deletedSub = (typeName: string): string => {
 public typeRelation = (relation: string, columnName: string, fieldName: string, tableName: string): string => {
   if(relation === 'OneToOne') {
     return `${fieldName}: (parent: any, _: any, context: GraphQLContext) => {
-      return context.db.select().from('${tableName}').where('id', '=', parent.${fieldName}Id)
+      return ${this.knexContext}.select().from('${tableName}').where('${columnName}', '=', parent.id)
                                 .then((result) => result[0])
     }`
   } else if(relation === 'OneToMany') {
     return `${fieldName}: (parent: any, _: any, context: GraphQLContext) => {
-      return context.db.select().from('${tableName}').where('${columnName}', '=', parent.id)
+      return ${this.knexContext}.select().from('${tableName}').where('${columnName}', '=', parent.id)
     }`
   } else {
     return undefined
   }
+}
+
+public invertTypeRelation = (columnName: string, fieldName: string, tableName: string): string => {
+  return `${fieldName}: async (parent: any, _: any, context: GraphQLContext) => {
+      const result = await ${this.knexContext}.select().from('${tableName}').where('id', '=', parent.${columnName})
+      return result[0]
+    }`
 }
 
 public blankResolver = (name: string) => {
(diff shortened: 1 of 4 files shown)


packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
```diff
@@ -87,7 +87,7 @@ public deletedSub = (typeName: string): string => {
 public typeRelation = (relation: string, columnName: string, fieldName: string, tableName: string): string => {
   if(relation === 'OneToOne') {
     return `${fieldName}: (parent: any, _: any, context: GraphQLContext) => {
-      return context.db.select().from('${tableName}').where('${columnName}', '=', parent.id)
+      return context.db.select().from('${tableName}').where('id', '=', parent.${fieldName}Id)
```
ankitjena: parent might not have a `${fieldName}Id` always

packages/graphback/src/schema/targetSchemaContext.ts
```diff
@@ -132,37 +132,43 @@ export const buildTargetContext = (input: Type[]) => {
...
-          })
-          relations.push({
+          }
+          if (!relations.includes(relation)) relations.push(relation)
+          relation = {
             "name": f.type,
             "type": 'ID',
-            "relation": `${t.name.toLowerCase()}Id: ID!`
-          })
+            "relation": `${f.name}Id: ID!`
```
ankitjena: I think this should be `${f.type.toLowerCase()}: ${f.type}`, this will give you the `user: User` field

ankitjena:
Thanks for the PR! @richcorbs . I will take a look how it works.

wtrocki:
@richcorbs Thank you so much for contribution.
We going to get that sorted, but some additional work will be required to merge that to master. We are working on set of automated tests and this change will be great oportunity to test it.

packages/graphback/src/schema/targetSchemaContext.ts
richcorbs: Ah, I think you are right. It's because `f.type !== f.name` if someone used a different name for the field. Good catch. I was lucky because in my example schema they are the same. I'll make that change.

packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
richcorbs: If it has a OneToOne relationship it should always have `${fieldName}Id`.

That's what defines the OneToOne relationship. What other circumstance could there be to have a OneToOne relationship and NOT have `${fieldName}Id`?

Maybe I'm missing something? All of the scenarios worked `notes.user`

packages/graphback/src/schema/targetSchemaContext.ts
richcorbs: I pushed that change to my branch but am not seeing how to merge it in to this PR. Is that possible?

ankitjena: If you pushed it to your branch, it should automatically reflect in the PR.

packages/graphback/src/resolvers/knex/BaseKnexResolver.ts
richcorbs: If relationships are defined in the `.graphql` file using `user: User!` and `comments: [Comment!]!` constructs then are OneToOne and OneToMany directives still needed?

wtrocki:
Wow. Thank you guys for this awesome collaboration. Played with this and it simply works. This is cool that we might do 0.7.1 release to include it.

Issue #240, opened 2019-08-19
Avoid using ts-node in npm run start


wtrocki:
Use compiled code instead

Issue #216, opened 2019-08-11
Getting error while running server. 


Lavanyagaur22:
### Description

I tried to run the project (server) using the following steps.
- graphback generate
- graphback db
And then `npm start` command to be followed after this.

While running `graphback db` command I got this error: `password authentication failed for user "name"`. How to resolve this?

wtrocki:
problem is our config.json require config. 
We require file that is outside typescript source path.
Compiling source code to dist will still work but running npm run start without
compilation is causing config to be empty.

I could not replicate that on my machine and debug it but I think we might need to improve the way we use config in templates or force compile when template is created. @Lavanyagaur22 Thank you for logging issue

wtrocki:
Workaround now is to use `graphback watch` or simply `npm run build` before `npm run start`

ankitjena:
> Workaround now is to use graphback watch or simply npm run build before npm run start

@Lavanyagaur22 did this work for you?

Lavanyagaur22:
> > Workaround now is to use graphback watch or simply npm run build before npm run start
> 
> @Lavanyagaur22 did this work for you?

@ankitjena Yep 👍

ankitjena:
addressed it in #249

PR #249, opened 2019-08-21
change start script and update docs

ankitjena:
- Change `start` to script use compiled code.
- remove `watch` command from package.json
- add `develop` script for ts-node

#240 #216

### DIFF
--- a/templates/graphql-js-starter/package.json
+++ b/templates/graphql-js-starter/package.json
@@ -5,10 +5,10 @@
   "private": true,
   "main": "dist/src/index.js",
   "scripts": {
-    "start": "ts-node src/index.ts",
     "build": "tsc",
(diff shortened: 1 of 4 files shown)


templates/graphql-js-starter/package.json
```diff
@@ -5,10 +5,9 @@
   "private": true,
   "main": "dist/src/index.js",
   "scripts": {
-    "start": "ts-node src/index.ts",
+    "develop": "ts-node src/index.ts",
     "build": "tsc",
-    "serve": "node dist/src/index.js",
-    "watch": "graphback watch"
```
wtrocki: watch can be still watching typescript code.` ts-node -w`. Then if you do graphback generate it will reload. This means we do not need to have graphback watch at all

PR #256, opened 2019-08-22
In memory subscriptions for sqlite3

ankitjena:
Remove dependency of docker-compose in case of `sqlite3` database.

### DIFF
--- a/packages/graphback-cli/src/helpers/init.ts
+++ b/packages/graphback-cli/src/helpers/init.ts
@@ -21,7 +21,7 @@ async function installDependencies(name: string, database: string): Promise<void
   if (database === 'pg') {
     await execa('npm', ['i', '-S', 'pg'])
   } else if (database === 'sqlite3') {
-    await execa('npm', ['i', '-S', 'sqlite3'])
+    await execa('npm', ['i', '-S', 'sqlite3', 'graphql-subscriptions'])
   }
   spinner.succeed()
 }
--- a/packages/graphback-cli/src/helpers/init.ts
+++ b/packages/graphback-cli/src/helpers/init.ts
@@ -22,6 +22,7 @@ async function installDependencies(name: string, database: string): Promise<void
     await execa('npm', ['i', '-S', 'pg'])
   } else if (database === 'sqlite3') {
     await execa('npm', ['i', '-S', 'sqlite3', 'graphql-subscriptions'])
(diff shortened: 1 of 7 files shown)


packages/graphback-cli/src/helpers/init.ts
```diff
@@ -21,7 +21,7 @@ async function installDependencies(name: string, database: string): Promise<void
   if (database === 'pg') {
     await execa('npm', ['i', '-S', 'pg'])
   } else if (database === 'sqlite3') {
-    await execa('npm', ['i', '-S', 'sqlite3'])
+    await execa('npm', ['i', '-S', 'sqlite3', 'graphql-subscriptions'])
```
wtrocki: I think for apollo we might need `apollo-server` as it has in mem implementation. Not 100% sure but I have used it there.

packages/graphback-cli/src/helpers/init.ts
```diff
@@ -21,7 +21,7 @@ async function installDependencies(name: string, database: string): Promise<void
   if (database === 'pg') {
     await execa('npm', ['i', '-S', 'pg'])
```
wtrocki: LEt's drop redis subscriptions there - it is now added into package.json for each template

templates/graphql-js-starter/src/context.ts
```diff
@@ -1,9 +1,10 @@
 import express from 'express'
 import { RedisPubSub } from 'graphql-redis-subscriptions';
+import { PubSub } from 'graphql-subscriptions';
 import knex from 'knex'
 
 export interface GraphQLContext {
-  pubsub: RedisPubSub
+  pubsub: RedisPubSub | PubSub
```
wtrocki: I would sacrifice strong typing to remove RedisSubscriptions from it. 
We can also use conditionals from TypeSctipt that will check if package exist

packages/graphback-cli/src/helpers/init.ts
ankitjena: We would have to check for template then, because graphql-js would need this package.

templates/graphql-js-starter/src/context.ts
ankitjena: Remove pubsub from here?

wtrocki: RedisPubSUb is specific. We can simply use interface ?

Issue #72, opened 2019-06-25
Generate sample client side queries based on schemas.


wtrocki:
We can use the graphQL code generator to generate client-side typings and queries. Additionally, we can investigate generating react based forms based on schema. For that we can use https://uniforms.tools/

wtrocki:
I have done some playing with this and the way I see it is that.
On Init, we will ask if we want to generate client-side code.
We will have a template for the client-side. 
```
graphback init

Do you want to generate client? Yes/No
```

This puts flag in config and client folder in server.
Client should contain:
- Queries and mutations (typesafe)
- Very very simple test that will allow people to use client and leave query/mutate unimplemented. 

Bonus - we can link queries into the playground (server)

Follow up. Blog post - `end to end development experience on graphback`

wtrocki:
Update. Uniforms on it's own provides really good solution, however we need to generate sample queries and mutations for GraphQL first including input types. Then we will be able to feed then into uniforms or other elements. 

I have done POC as part of the https://github.com/aerogear/modern-appdev-workshop/

wtrocki:
Generally generating sample queries and mutations that accept parameters like here:
https://github.com/aerogear/modern-appdev-workshop/tree/master/template/client/src/graphql

will be the most critical step. Then we can get fragments and other things

ankitjena:
Do we generate a fixed query and mutation like `findAllTasks` and `createTask` for example?

wtrocki:
Take look into the example linked above. Variables are passed from client

ankitjena:
Going to generate CRUD queries for each type with arguments as variables into specific folders like in the example.

wtrocki:
generate sample queries. We will follow up with uniforms tool to pick that up

PR #265, opened 2019-08-27
Generate client queries based on datamodel

ankitjena:
## How it works
Generate queries and mutation with fragments according to config(which are enabled). #72 

### Mutation
```
import gql from "graphql-tag"
import { NoteFragment } from "../fragments/Note"

export const updateNote = gql`
  mutation updateNote($id: ID!, title: String!, description: String!, published: Boolean!) {
    updateNote(id: $id, input: {title: $title, description: $description, published: $published}) {4
      ...NoteFields
    }
  }

  ${NoteFragment}
`
```

### Query
```
import gql from "graphql-tag"
import { NoteFragment } from "../fragments/Note"

export const findAllNotes = gql`
  query findAllNotes {
    findAllNotes {
      ...NoteFields
    }
  }

  ${NoteFragment}
`
```
### Subscription
```
import gql from "graphql-tag"
import { CommentFragment } from "../fragments/Comment"

export const newComment = gql`
  subscription newComment {
    newComment {
      ...CommentFields
    }
  }

  ${CommentFragment}
`
```

### Fragment
```
import gql from "graphql-tag"

export const NoteFragment = gql`
  fragment NoteFields on Note {
    id
    title
    description
    published
  }
`
```

### DIFF
--- a/docs/clientsidequeries.md
+++ b/docs/clientsidequeries.md
@@ -1,0 +1,90 @@
+---
+id: clientqueries
+title: Client-side Queries
+---
+
+Graphback allows users to generate client-side queries based on the data model. All the queries/mutations/subscriptions
+which are generated through `graphback generate` for the user can also be generated as client-side queries. Users may follow up 
+with using `create-react-app` or `apollo-client` to kickstart client experience.
+
+## Autogenerating forms based on server-side schema
+These can be used by client side frameworks like [uniform.tools](https://uniforms.tools) to create
+GraphQL client applications.
+
+## How it works
+Graphback asks whether you want to generate client-side queries in the `graphback init` command. Answering it yes,
+will set `client` in `config.json` to `true`.
+
+For a data model having
+```
+type Note {
+  id: ID!
+  title: String!
+  description: String!
+}
+```
+running `graphback generate` will generate all operations which are enabled in `config.json` or using `directives`.
+
+## Examples
+### Fragment
+```
+import gql from "graphql-tag"
+
+export const NoteFragment = gql`
+  fragment NoteFields on Note {
+    id
+    title
+    description
+    published
+  }
+`
+```
+
+### Query
+```
+import gql from "graphql-tag"
+import { NoteFragment } from "../fragments/Note"
+
+export const findAllNotes = gql`
+  query findAllNotes {
+    findAllNotes {
+      ...NoteFields
+    }
+  }
+
+  ${NoteFragment}
+`
+```
+
+### Mutation
+```
+import gql from "graphql-tag"
+import { NoteFragment } from "../fragments/Note"
+
+export const updateNote = gql`
+  mutation updateNote($id: ID!, title: String!, description: String!, published: Boolean!) {
+    updateNote(id: $id, input: {title: $title, description: $description, published: $published}) {4
+      ...NoteFields
+    }
+  }
+
+  ${NoteFragment}
+`
+```
+
+### Subscription
+```
+import gql from "graphql-tag"
+import { CommentFragment } from "../fragments/Comment"
+
+export const newComment = gql`
+  subscription newComment {
+    newComment {
+      ...CommentFields
+    }
+  }
+
+  ${CommentFragment}
+`
+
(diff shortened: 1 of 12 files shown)


packages/graphback/src/resolvers/outputResolvers/resolverTemplate.ts
```diff
@@ -40,7 +40,7 @@ const generateTypeResolvers = (context: TargetResolverContext, name: string): st
   let output = `${imports}`
 
   if(context.subscriptions.length) {
-    output += `\n\n enum Subscriptions {
+    output += `\n\nenum Subscriptions {
```
wtrocki: ??

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -56,10 +56,20 @@ export const chooseDatabase = async(): Promise<string> => {
   return database
 }
 
+export const askForClient = async(): Promise<boolean> => {
+  const { client } = await ask({
+    type: 'confirm',
+    name: 'client',
+    message: 'Do you want to generate client?'
```
wtrocki:
```suggestion
    message: 'Do you want to generate client-side GraphQL queries?'
```

wtrocki:
We need documentation and release notes. I will need to play with that much longer over the comming days.

packages/graphback/src/resolvers/outputResolvers/resolverTemplate.ts
ankitjena: No its `\n` it left empty space so had indentation error.

ankitjena:
> Good work. I think we might need to have a blog post about it.
> Please add docs and we are sorted

But we haven't integrated uniforms, shouldn't we download client template in init

wtrocki:
@ankitjena No. Client side will be never focus of graphback, however we can deliver client side artifacts to work on clients. 

WE need documentation that our cli will generate queries in some folder. 
Then people can simple use `create-react-app`, but I'm working with uniforms team to pick that up and have `create-uniforms-app` that will be giving us end to end client support
thru thrid party.

I think it is important to be able to change folder of the client queries generation.

docs/clientsidequeries.md
```diff
@@ -0,0 +1,85 @@
+---
+id: clientqueries
+title: Client-side Queries
+---
+
+Additionaly graphback also allows users to generate client-side queries based on the data model.
```
wtrocki:
```suggestion
Graphback allows users to generate client-side queries based on the data model.
```

docs/clientsidequeries.md
```diff
@@ -0,0 +1,85 @@
...
+These can be used by client side frameworks like [uniform.tools](https://uniforms.tools) to create
```
wtrocki: I would move this to separate . chapter here. 
Something like

## Autogenerating forms based on server-side schema

wtrocki: Then we will collaborate with @radekme to get proper integration

wtrocki: Then we will collaborate with @radekmie to get proper integration

docs/clientsidequeries.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+running `graphback generate` will generate all operations which are enabled in `config.json` or using `directives`.
+
+## Examples
+### Fragment
+```
+import gql from "graphql-tag"
+
+export const NoteFragment = gql`
```
wtrocki: It is important that we done it this way.

Uniforms can base on NoteFragments to build forms input. Separate tooling I have build will apply common patterns - for example hide id field and generate labels based on the metadata provided in directives

docs/clientsidequeries.md
wtrocki: Missing info that users still need to follow up with any available clients like Apollo and use something like create-react-app to kickstart experience.

docs/clientsidequeries.md
ankitjena: Sure

PR #269, opened 2019-08-28
Website improvements

wtrocki:
I'm not skilled in layout. Target is to put demo video on the website plus diagram.
While this layout is still quite bad it will give additional value to our users.

Screenshot 2019-08-28 at 13 32 49


wtrocki:
Actually this was published by mistake: https://graphback.dev/

ankitjena:
> I'm not skilled in layout. Target is to put demo video on the website plus diagram.
> While this layout is still quite bad it will give additional value to our users.

This looks awesome.

wtrocki:
I have noticed that layout on mobile is kinda broken. I will check why

darahayes:
@wtrocki sorry I only got to look at this now. I'm mostly happy with it except the text underneath the Graphback Workflow heading is center aligned which looks pretty bad. Instead it should be left aligned.

PR #275, opened 2019-08-28
fix: issue with missing files in templates cli

wtrocki:
Reverting breaking changes. We need to think how to do this properly.


wtrocki:
We need to wipe dist for every build

Issue #191, opened 2019-08-06
Graphback support for interfaces


ankitjena:
Graphback currently doesn't support enums and interfaces as the visitor is not parsing it. It would be a nice feature to add.

wtrocki:
Hmm.. another inconvinience is that for custom mutations we will need to know input types that can be used (they will be generated) we need to document how input types are named because we never know until they are generated.

wtrocki:
To do this first we will need to:

1) Extend model in template to have interface
```
interface A { ...}
type B implements interface A
```

2) Investigate generated AST using ast explorer. Check blog post on graphback page

3) Investigate visitor we have build:
https://github.com/aerogear/graphback/blob/master/packages/graphback/src/InputTypeVisitor.ts#L11

4) Add handler for an interface for visitor and stage information about the interface to separate object

https://github.com/aerogear/graphback/blob/master/packages/graphback/src/InputTypeVisitor.ts#L11

5) Extend schema generation to include interface.

PR #297, opened 2019-09-02
Add GraphQL interface support

craicoverflow:
[AEROGEAR-9814](https://issues.jboss.org/browse/AEROGEAR-9814)

## Description.

Add support for interfaces. See #191 for more information.

## Tasks

- [x] Collect interface nodes
- [x] Add interface types to generated schema
- [x] Remove query, subscription, mutation generation for interface types
- [x] Remove resolver generation for interfaces
- [x] Remove missing resolver warning in templates
- [ ] Add automated tests

## Verification Steps

Add the following to your template's schema file:

```graphql
interface Searchable {
  searchPreviewText: String!
}

interface Item {
  price: Int!
}

type Movie implements Searchable & Item {
  directory: String!
  searchPreviewText: String!
  price: Int!
}
```

Run `graphback generate && graphback db`.

Next, inspect the file `src/schema/generated.ts`. It should include the `Searchable`, `Item` interfaces. The `Movie` type should implement `Searchable` and `Item`.


packages/graphback/src/ContextCreator.ts
```diff
@@ -1,26 +1,55 @@
...
 export const createInputContext = (schemaText: string, defaultConfig: Config): Type[] => {
   const schema = applyGeneratorDirectives(schemaText)
   try {
     const astNode = parse(schema)
 
-    return visit(astNode, { leave: inputTypeVisitor }).definitions.map((t: Type) => {
+    const schemaDef = visit(astNode, { leave: inputTypeVisitor });
+
+    return schemaDef.definitions.map((t: Type) => {
+
```
wtrocki: I would extract that to separate function for redability.

It is that we have 2 steps for building this definitions: 

- Collect actual interfaces in visitors
- Map interface usage in types

It will be cool to have it in single place (visitor itself), but this code is ok as well. We just need to make it clear for future contributors that this is connected to usage of interface

packages/graphback/src/GraphQLBackend.ts
```diff
@@ -89,10 +94,24 @@ export class GraphQLBackendCreator {
...
 
+  /**
+   * Filters duplicate fields (by name) from a list of Fields
+   *
+   * @private
+   * @param {Field[]} fields The original list of fields from the type
+   * @returns {Field[]} Return the filtered list of fields
+   * @memberof GraphQLBackendCreator
+   */
+  private getUniqueFields(fields: Field[]): Field[] {
```
darahayes: minor nit: This is O(n^2) when it could be linear and it's really hard to understand. This would suffer performance degradation as the schema grows. There's nothing wrong with that but we might as well not let these perf things creep up on us.

There's a really nice stackoverflow answer that uses a set https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array

The code would look something like:

```
private getUniqueFields(fields: Field[]): Field[] {
    const seen = {};
    return fields.filter(function(field) {
		if (seen.hasOwnProperty(field.name)) {
			return false
		}
        seen[field.name] = true
		return true
    })
}
```

In this example we iterate over the array once using `filter`. For each item we check to see if `seen.hasOwnProperty(field.name)` is true. If it's not true, then then add it and return true. If `seen.hasOwnProperty(field.name)` is true that means we've hit a duplicate, so we return false, this ensures the duplicate is not in the newly returned array.

This code has the benefit of being faster (linear time) and in my opinion, a lot easier to understand.

wtrocki: Question - why inputContext will have duplicates at this point?

Do we need to handle cases when someone shadows the same field with different type? Which one wins now interface or direct definition? This code looks like direct definition will be omited.

packages/graphback/src/InputTypeVisitor.ts
```diff
@@ -50,24 +50,42 @@ export const inputTypeVisitor = {
     }
   },
 
+  InterfaceTypeDefinition: (node: InterfaceTypeDefinitionNode) => {
+    let config = { create: false, update: false, subCreate: false, subUpdate: false, subDelete: false };
+
+    node.directives.map((directive: object) => {
```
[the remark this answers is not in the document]
craicoverflow: Yeah, for this to be honest I just cloned the `ObjectTypeDefinition` function which does this:

https://github.com/aerogear/graphback/blob/35e6b09271fe3923ace40e76954bc16611b23b75/packages/graphback/src/InputTypeVisitor.ts#L72

Both should be changed.

packages/graphback/src/InputTypeVisitor.ts
```diff
@@ -50,24 +50,42 @@ export const inputTypeVisitor = {
     }
   },
 
+  InterfaceTypeDefinition: (node: InterfaceTypeDefinitionNode) => {
+    let config = { create: false, update: false, subCreate: false, subUpdate: false, subDelete: false };
```
wtrocki: [Documentation] So this implies that our logic is going to ignore anything that is applied to interface will be ignored and users should be applying those to types. 

This can be ensured by directive validation so it will not be possible to do it or we can simply mention that in docs.

packages/graphback/src/schema/schemaTemplate.ts
```diff
@@ -10,21 +10,21 @@ import { TargetContext, TargetType } from './targetSchemaContext';
 // }
 
 const inputs = (defs: TargetType[]): string => {
-  return `${defs.map((d: TargetType) => `input ${d.name}Input {
-    ${d.fields.join('\n    ')}
+  return `${defs.map((d: TargetType) => `${d.type} ${d.name}Input {
+    ${[...new Set(d.fields)].join('\n    ')}
```
wtrocki: That is very creative way to remove duplicates. However we do not control what field takes precedence. Set will literally remove duplicate randomly which may be a problem if duplicate removal from the db will remove different one. We can end up with incompatible types. 

I'm not 100% sure about this as I could not replicate it, but basing on my understanding of the code this could happen.

wtrocki:
I'm having problems to verify this change as I have right code in graphback. I have it compiled. 
Cli is linked globally and uses proper graphback package, however I still see interfaces with generated resolvers and database tables.

packages/graphback/src/GraphQLBackend.ts
craicoverflow: At this point, inputContext should not have duplicates. Looking to change this now.

We need to see which field takes precedence when both are the same.

craicoverflow:
> Nice PR, I left some non critical suggestions in there. Any plan to add some tests? It would really nice to see a test that tests the verification steps you put in the PR description

I agree that tests should be included as part of this. I will start adding them to this PR once the code stabilises (things are changing a lot right now).

wtrocki:
Checking now

templates/apollo-starter-ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "apollo-rest-template",
-  "version": "0.8.3",
+  "version": "0.8.4",
```
craicoverflow: @wtrocki I wasn't sure if I am to add this manually or is this step done (semi) automatically?

wtrocki:
I will create separate issue for automated tests. 
I have added release notes in #308

Issue #289, opened 2019-08-30
Create plugin for graphql cli


wtrocki:
A prototype of the plugin was created here: 
https://github.com/ankitjena/graphql-cli-graphback-plugin

Project:
https://github.com/graphql-cli/graphql-cli

PR #301, opened 2019-09-03
Plugin for graph ql cli

MStokluska:
## Motivation
#289 

## Tasks

- [x] Code
- [x] Fix build/tests
- [ ] Verification

## Verification

```sh
npm install
```
```sh
npm run bootstrap
```
```sh
npm run build
```
```sh
graphql
```
Should produce GraphQL-CLI options along with Graphback options.


package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
```
wtrocki: Remove this file.

packages/graphback-cli/src/index.ts
```diff
@@ -2,10 +2,14 @@
...
+  // tslint:disable-next-line: no-unused-expression
+  yargs
+    .commandDir('commands')
+    .demandCommand(1)
+    .help()
+    .version()
+    .argv;
+}
+
+export * from './helpers'
```
wtrocki: I would put it on top.

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
...
+  <br/>
+  Auto generate database structure, <br/>
+  GraphQL Resolvers and Queries from GraphQL types 🚀
+</p>
+
+## Motivation 
+
+GraphQL can be hard and intimidating initially.
+Graphback helps you to kickstart your experience with any existing GraphQL implementation
+by generating backend layer using your data model.
```
wtrocki: This description was changed in readme. We changed to to Node.js backend server instead of layer with database changes.

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
...
+```sh
+graphql backend-generate
+```
+To create database resources:
+
+```sh
+graphql backend-db
+```
+
+**Documentation**: https://graphback.dev
```
wtrocki: I would put this on the top.

packages/graphql-cli-graphback-plugin/.gitignore
```diff
@@ -0,0 +1,4 @@
+dist
```
darahayes: This file is unnecessary

packages/graphql-cli-graphback-plugin/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "graphql-cli-graphback-plugin",
+  "version": "0.2.1",
+  "description": "Auto generate resolvers on top of the graphql and database of your choice",
```
wtrocki:
```suggestion
  "description": "GraphQL CLI plugin for Graphback - Node.js backend generator",
```

packages/graphql-cli-graphback-plugin/.npmignore
```diff
@@ -0,0 +1 @@
+src
```
darahayes: This file is also unecessary

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
...
+GraphQL can be hard and intimidating initially.
+Graphback helps you to kickstart your experience with any existing GraphQL implementation
+by generating backend layer using your data model.
+
+## Commands
+
+To generate backend using Graphback:
+
+```sh
+graphql backend init <name>
```
darahayes: should this be `graphql backend-init <name>` ?

packages/graphql-cli-graphback-plugin/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "ts-node": "8.3.0",
+    "tslint": "5.19.0",
+    "tslint-config-prettier": "1.18.0",
+    "tslint-microsoft-contrib": "6.2.0",
+    "tsutils": "3.17.1",
+    "typescript": "3.5.3",
+    "typescript-eslint-parser": "21.0.2"
+  },
+  "dependencies": {
+    "chalk": "2.4.2",
```
wtrocki: None of those are actually required. Most of the dependencies will come from CLI package.

packages/graphql-cli-graphback-plugin/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "graphql-cli-graphback-plugin",
+  "version": "0.2.1",
+  "description": "Auto generate resolvers on top of the graphql and database of your choice",
+  "homepage": "https://github.com/ankitjena/graphql-cli-graphback-plugin",
```
darahayes: incorrect homepage

packages/graphql-cli-graphback-plugin/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(tsx?|jsx?)$",
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
+      "js",
+      "json",
+      "jsx"
+    ]
+  },
+  "license": "ISC"
```
wtrocki: We need to relicense everything to MIT

packages/graphql-cli-graphback-plugin/package.json
```diff
@@ -0,0 +1,72 @@
...
+  "author": "wtrocki@redhat.com",
+  "scripts": {
+    "clean": "rimraf coverage dist types",
+    "build": "npm run clean && tsc",
+    "cmd": "node ./dist/index.js",
+    "watch": "tsc -w",
+    "lint": "tslint -t stylish --project \"tsconfig.json\""
+  },
+  "engines": {
+    "node": "12.9.1"
```
darahayes: minor nit: this is inaccurate. I can't remember the exact syntax but it should probably be something like `> 8.x.x`. Refer to npm's `package.json` reference documentation to find out the correct syntax

packages/graphql-cli-graphback-plugin/package.json
darahayes: License should be Apache-2.0

packages/graphql-cli-graphback-plugin/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{ 
+    "include": [
+      "./src/**/*.ts"
+    ],
+    "compilerOptions": {
```
wtrocki: See other packages that do `  "extends": "../../tsconfig.json",` those were copied as this was separate package.

packages/graphql-cli-graphback-plugin/README.md
MStokluska: following Graphql-Cli format init is without `-` and other commands are with '-' but I can change it if you think it's unnecessary.

packages/graphql-cli-graphback-plugin/src/commands/backend-init.ts
```diff
@@ -0,0 +1,28 @@
+import yargs from 'yargs';
+import { init } from 'graphback-cli'
+
+type Params = { name?: string, templateName?: string, templateUrl: string }
+
+export const command = 'backend init <name>'
```
darahayes: cc @wtrocki @ankitjena we prefixed the commands with `backend` so we wouldn't clash with the other commands provided by `graphql-cli`. What do you think of these ones? I would prefer if we could figure out how to shorten those names but still keep them meaningful. Any ideas?

packages/graphql-cli-graphback-plugin/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{
```
darahayes: This `tsconfig.json` is not extending the top level one. It should probably be replaced with the exact same `tsconfig.json` you would find in the `graphback` and `graphback-cli` packages

packages/graphql-cli-graphback-plugin/package.json
darahayes: The repo has an Apache-2.0 License

darahayes: This raises an important question - Should the license file always be distributed with the source? Because in our monorepo we have the license at a top level but if you actually installed one of these packages from npm, the license would not be in there. WDYT?

packages/graphql-cli-graphback-plugin/README.md
wtrocki: I think we need to have subcomand as `backend` adding many commands for single extension is antipattern and it will not be merged to cli.

packages/graphql-cli-graphback-plugin/package.json
wtrocki: Graphql-cli is MIT and to integrate it we need to be MIT. Rest of the packages can be in any license.

packages/graphql-cli-graphback-plugin/src/commands/backend-init.ts
wtrocki: hmm.. requirements are - single plugin single command - so whatever we will use - backend or anything else as prefix will work.

packages/graphql-cli-graphback-plugin/package.json
darahayes: @wtrocki I think we should probably think about how/where the license files are placed then. Most companies/end users simply cannot touch code if there is any ambiguity around licensing. If we're planning to have different packages using different licenses all within the one repo, I think we should probably be very explicit and include the appropriate license file directly in each package folder. Otherwise the released distribution they download from npm will not actually contain a license file. What do you think?

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
...
+<p align="center">
+  <img width="400" src="https://github.com/aerogear/graphback/raw/master/website/static/img/graphback.png">
+  <br/>
+  Auto generate database structure, <br/>
+  GraphQL Resolvers and Queries from GraphQL types 🚀
+</p>
+
+## Motivation 
+
+GraphQL can be hard and intimidating initially.
```
wtrocki: I would remove this sentence

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
+**Documentation**: https://graphback.dev
```
wtrocki: Move this to line 11

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
...
+```sh
+graphql backend-generate
+```
+To create database resources:
+
+```sh
+graphql backend-db
+```
+
+**Repository**: https://github.com/aerogear/graphback/
```
wtrocki: Move this to line 12

packages/graphql-cli-graphback-plugin/README.md
```diff
@@ -0,0 +1,36 @@
+**Documentation**: https://graphback.dev
+
+## GraphBack
```
craicoverflow:
```suggestion
## Graphback
```

PR #308, opened 2019-09-04
fix: Add documentation for interface support

wtrocki:
Every change we do - skipping some trivial ones will need to land into release notes.
There is no need for documentation for this feature.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -14,7 +14,7 @@ Please follow indidividual releases for more information.
 #### Features
 
 - Support for Interfaces in GraphQL schema
-GraphQL Schema supports now interfaces that can be used to ensure consistency of the data.
+GraphQL Schema now supports interfaces that can be used to ensure consistency of the data.
 Example:
 
 ```graphql
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,6 +8,33 @@ title: Releases
 This file contains changes and migration steps for Graphback project. 
 Please follow indidividual releases for more information.
 
+# Next
+
+### Graphback
+#### Features
+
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -8,6 +8,33 @@ title: Releases
 This file contains changes and migration steps for Graphback project. 
 Please follow indidividual releases for more information.
 
+# Next
+
+### Graphback
+#### Features
```
MStokluska: Correct if I'm wrong but don't we need to keep the same format as others? ie 
```
## 0.8.3 (29th August, 2019)
### Templates
#### Fixes
```

docs/releases.md
```diff
@@ -8,6 +8,33 @@ title: Releases
...
+
+- Support for Interfaces in GraphQL schema
+GraphQL Schema supports now interfaces that can be used to ensure consistency of the data.
```
craicoverflow:
```suggestion
GraphQL Schema now supports interfaces that can be used to ensure consistency of the data.
```

docs/releases.md
craicoverflow: Yes, I am also unsure about this. This new section doesn't really seem to "fit" in with the rest.

wtrocki: It is not clear what will be the next version so I'm using Next to stage it and going to replace it on release.

Format is right see bellow:

```
## 0.8.0 (28th August, 2019)
### Graphback
#### Features	
```

wtrocki: > ## 0.9.0 (5th September, 2019)

So I do not know the version number and date- this is going to be done on release.

Issue #307, opened 2019-09-04
CLI instructions


MStokluska:
## Summary

I think we should rethink how Graphback instructions are displayed as currently, we are repeating steps unnecessarily. 

## Issue

The first set of instructions are displayed after `init`
```sh
graphback init <name>
```
Which produces following output:
![image](https://user-images.githubusercontent.com/43473912/64266771-17b92b80-cf2d-11e9-8e0d-fe274d9a2385.png)

Another set of instructions are displayed `generate`
```sh
graphback generate
```
![image](https://user-images.githubusercontent.com/43473912/64266899-50f19b80-cf2d-11e9-84cd-0e07afa379f6.png)

I don't think there's any need for the second set of instructions and that step 4. could have been separated into 2 individual steps.

ping @wtrocki @darahayes @craicoverflow @ankitjena

wtrocki:
For init we should only print step:

1,2,3

For generate we should only print 

1.

For db we should print 

2.

ankitjena:
@wtrocki agree with this

PR #313, opened 2019-09-05
Reshaping cli

MStokluska:
## Motivation
#307 

## Summary
I've updated CLI with few changes, it looks slightly different now, matching preferred format from issue #307 and updates instruction depending on whether graphback was accessed from graphql-CLI or graphback-CLI.

### Graphback:

1. `graphback init <name>` produces :

image

2. `graphback generate` produces: 

image

3. `graphback db` produces:

image

### GraphQL-CLI

1. `graphql backend-init <name>` produces:

image

2. `graphql backend-generate` produces:

image

3. `graphql backend-db` produces:

image

## Verification

For Graphback run:

1. `graphback init <name>
2. `graphback generate`
3. `docker-compose up -d`
4. `graphback db`

For GraphQL-CLI

1. `graphql backend-init <name>`
2. `graphql backend-generate`
3. `docker-compose up -d` if no docker container is running
4. `graphql backend-db`

### DIFF
--- a/packages/graphback-cli/src/cliName.ts
+++ b/packages/graphback-cli/src/cliName.ts
@@ -1,1 +1,0 @@
-export const CLI_NAME = "graphback "
--- a/packages/graphback-cli/src/cliName.ts
+++ b/packages/graphback-cli/src/cliName.ts
@@ -1,0 +1,1 @@
+export const CLI_NAME = "graphback "
--- a/packages/graphback-cli/src/commandRoot.ts
+++ b/packages/graphback-cli/src/commandRoot.ts
@@ -1,0 +1,1 @@
+export const commandRoot = "graphback "
--- a/packages/graphback-cli/src/commands/generate.ts
+++ b/packages/graphback-cli/src/commands/generate.ts
@@ -1,3 +1,4 @@
+import { CLI_NAME } from '../cliName'
 import { generate } from '../helpers'
 
 export const command = 'generate'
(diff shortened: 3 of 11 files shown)


packages/graphback-cli/src/cliName.ts
```diff
@@ -0,0 +1 @@
+export const CLI_NAME = "graphback "
```
wtrocki: WDYT? `commandRoot`

wtrocki: I would avoid space here and use it in actual templates

wtrocki:
This works for me! Good job!

wtrocki:
Can we add release notes that we have CLI as bonus - see docs/release.md

wtrocki:
Rebase needed!

packages/graphback-cli/src/cliName.ts
MStokluska: @wtrocki there's an issue with space as if I delete it and add space to print out in [init.ts]( 
https://github.com/aerogear/graphback/pull/313/files#diff-7b17caceb59bc1f2a0a7bbf6a6a954afR96) graphback print outs will look fine but graphql generate command will look like this:
```
graphql backend- generate
```
while it should be :
```
graphql backend-generate
```
Not sure if there's any workaround that if this is the approach I took.

wtrocki: Yes. I think we might need to do some backend subcomand at some poit. Merge it as it is

PR #319, opened 2019-09-06
Add paths config

### LINKED ISSUE #294: All input output folders should be configurable
Currently graphback expects model to be in some hardcoded locations and outputs to some specific folders as well. This can lead to many problems and wrong assumptions. 
All folders should be configurable in separate part of the configuration file. 

## Spec
Currently available folders:

Input folders: 
- model

Output folders:
- resolvers
- schema

Additional we should provide documentation for all input and outputs of the generation process - including configuration so that it will be clear how things work and how to build an template.

MStokluska:
## Motivation

fix #294

## Summary

Enable fully configurable paths to:
- `model` input
- `generatedResolvers` output
- `customResolvers` outup
- `schema` outup
- `client` output
All paths can be specified in `.graphback` configuration file. Which serves as previous `config.json` file.

### How it works
1. By default, we can still use graphback as we had before as default routes are pointing to the same dir as previous config.json

image

2. Let's move `model` folder a level up and edit path to `model`

image

3. `graphback generate` and `graphback db` will work as previously.

image

4. Now let's select a different folder for generated resolvers:

image

Generated resolvers are generated in specified file

5. Next, a different folder for custom resolvers:

image

6. Custom location for `schema` :

image

7. Custom location for `client`:

image


packages/graphback-cli/src/helpers/db.ts
```diff
@@ -6,7 +6,7 @@ import { DatabaseSchemaManager, GraphQLBackendCreator } from 'graphback';
 import { logError, logInfo } from '../utils'
 import { checkDirectory } from './common'
 
-const configPath = `${process.cwd()}/config.json`
+const configPath = `${process.cwd()}/.graphback`
```
ankitjena: Just `.graphback` or `.graphback.json`

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -37,7 +37,15 @@ const generationConfig = {
   "disableGen": false
 }
 
-export const chooseDatabase = async(): Promise<string> => {
+const paths = {
+  "model": "model",
```
ankitjena:
```suggestion
  "model": "/model",
```

ankitjena:
How does it work with relative paths like `../model` for example?

packages/graphback-cli/src/helpers/db.ts
MStokluska: @ankitjena hey, 
We have just realized that for now it needs to stay as .json, changed it graphback.json

packages/graphback-cli/src/templates/configTemplates.ts
MStokluska: thank Ankit. I'm going to close this PR as there's a slightly different approach to this that I want to try out following another format of providing paths with : `./something`
Thanks for the valuable feedback!

MStokluska:
Closing this draft as I want to try a little bit different approach to support `./xx` paths

wtrocki:
> Concatenating that with process.cwd() should not work I think?

`path.join will help`

PR #330, opened 2019-09-10
Add path config

MStokluska:
## Motivation
#294 

## Solution

Paths for `model` `schema` `generatedResolvers` and `customResolvers` can be configured in graphback.json with default values being the same as they were. 

## Verification

1. `graphback init <name>` to create graphback project
2. navigate to project and run `graphback generate`
3. graphback should work as usual
4. create new folder in you graphback project called `foo` and move models to `foo` folder
5. in `graphback.json` change model path to `/foo/model/` and `graphback generate` and `graphback db` should work as usual
6. in `graphback.json` change the output location for resolvers and run `graphback generate` resolvers should be generated into a new path
7. in `graphback.json` change the output of `client` and run `graphback generate`, `client` queries should be generated in a specified path
8. in `graphback.json` change output path of `schema`, `graphback generate` should create schema in a new folder
9. if any of new folders don't exist a folder should be created
10. if folder already exists it should not be overwritten but instead, files generated should be added to existing foler.


packages/graphback-cli/src/helpers/db.ts
```diff
@@ -6,23 +6,23 @@ import { DatabaseSchemaManager, GraphQLBackendCreator } from 'graphback';
...
   }
   process.exit(0)
 }
 
-export const dropDBResources = async(): Promise<void> => {
+export const dropDBResources = async (): Promise<void> => {
   try {
     const { database, dbConfig } = JSON.parse(readFileSync(configPath, 'utf8'))
-    if(database === 'sqlite3') {
+    if (database === 'sqlite3') {
```
wtrocki: Wrong formatting. It is strange to see this as we all using visual studio code.

packages/graphback-cli/src/helpers/common.ts
```diff
@@ -1,12 +1,27 @@
...
+
+    const { paths } = JSON.parse(readFileSync(configPath, "utf8"))
+  
+    pathForModel = join(process.cwd(), paths.model)
+
+    accessSync(`${pathForModel}`)
+
+  } catch (err) {
+    logError(`Model not found, make sure you are in root directory of your project and that you have
+    specified correct path to your .graphql file`)
```
craicoverflow:
```suggestion
    specified the correct path to your .graphql file`)
```

packages/graphback-cli/src/helpers/common.ts
```diff
@@ -1,12 +1,27 @@
...
+    const configPath = `${process.cwd()}/graphback.json`
+
+    const { paths } = JSON.parse(readFileSync(configPath, "utf8"))
+  
+    pathForModel = join(process.cwd(), paths.model)
+
+    accessSync(`${pathForModel}`)
+
+  } catch (err) {
+    logError(`Model not found, make sure you are in root directory of your project and that you have
```
wtrocki: Not issue with your code itself but it can be improved. 
Code above can fail in 2 places:

- on reading config - then we need to say we have a config issue
- on reading graphql files in the model

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -37,23 +37,23 @@ export const dropDBResources = async(): Promise<void> => {
   }
 }
 
-export const createDBResources = async(): Promise<void> => {
-  try{
-    const models = new GlobSync('model/*.graphql', { cwd: process.cwd()})
+export const createDBResources = async (pathForModel: string): Promise<void> => {
+  try {
+
```
wtrocki: Let's avoid extra linebreaks etc.

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -75,10 +75,10 @@ Run ${chalk.cyan(`npm run develop`)} to start the server.
   `)
 }
 
-export const createDB = async(): Promise<void> => {
-  checkDirectory()
+export const createDB = async (): Promise<void> => {
+  const pathForModel = checkDirectory()
```
wtrocki: It will be cool if we calculate all paths in single place and pass entire object instead of individual paths.
This can be separate PR though.

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -6,7 +6,7 @@ import { DatabaseSchemaManager, GraphQLBackendCreator } from 'graphback';
 import { logError, logInfo } from '../utils'
 import { checkDirectory } from './common'
 
-const configPath = `${process.cwd()}/config.json`
+const configPath = `${process.cwd()}/graphback.json`
```
wtrocki: [Idea] Can we have function that will append all `process.cwd()` to all paths in config. Then we can have interface for it and pass it to every module to use it. This will be the only one `extra field` that we will need to require from outside as it is path to config itself.

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -37,6 +37,14 @@ const generationConfig = {
   "disableGen": false
 }
 
+const paths = {
```
wtrocki:
```suggestion
const defaultPaths = {
```

packages/graphback-cli/src/helpers/common.ts
```diff
@@ -1,12 +1,27 @@
...
-    accessSync(`${process.cwd()}/model`)
-  } catch(err) {
-    logError(`model directory not found. Make you sure you are in the root of your project.`)
+export const checkDirectory = (): string => {
+
+  let pathForModel
+
+  try {
+
+    const configPath = `${process.cwd()}/graphback.json`
```
wtrocki: `${process.cwd()}/graphback.json` is used in couple places maybe we can setup it as constant in separate folder and simply require it?

templates/apollo-rest-starter/graphback.json
```diff
@@ -17,5 +17,12 @@
     "subDelete": false,
     "disableGen": false
   },
+  "paths": {
+    "model": "./model",
+    "generatedResolvers": "./src/resolvers",
+    "customResolvers": "./src/resolvers",
+    "schema": "./src/schema",
+    "client": "./client"
```
wtrocki:
```suggestion
    "client": "./client/src/graphql"
```

templates/apollo-rest-starter/src/config/config.ts
```diff
@@ -1,5 +1,5 @@
 import knex from 'knex'
-import * as config from '../../config.json'
+import * as config from '../../graphback.json'
```
wtrocki: We can create issue about this problem to see if we can decompose config from the database.

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -21,61 +22,65 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
  * Generate schema and resolvers using graphback-core and
  * write them into generated folder
  */
-export async function generateBackend(): Promise<void> {
+export async function generateBackend(pathForModel: string): Promise<void> {
```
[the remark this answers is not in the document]
wtrocki: See comment above

wtrocki:
> It is possible to set the paths outside of the project. For example:

We are making sure that we are always using CWD. People can use `../` which do not make sense as your source code or model will not be included into the project, but if they want it they could do it. There is no sense to invest time preventing some 0.000001% use case that someone might actually found useful :D

wtrocki:
> It might be nice to be able to define what paths you want on graphback init. So something like:

Nice idea! @MStokluska this could be separate PR/Issue depending on your judgment.

packages/graphback-cli/src/helpers/generate.ts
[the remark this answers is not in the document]
MStokluska: 100% going to investigate

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -37,6 +37,14 @@ const generationConfig = {
...
+  "model": "./model",
+  "generatedResolvers": "./src/resolvers",
+  "customResolvers": "./src/resolvers",
+  "schema": "./src/schema",
+  "client": "./client/src/graphql"
```
ankitjena: This will throw error since `mkdir` creates folder if parent folder exists. To be precise the error
```
Error: ENOENT: no such file or directory, mkdir '/home/ankitj/apollo/client/src/graphql'
```
Are we assuming the user already has the folder structure created?

wtrocki: I think we can create a structure if it is not there.

MStokluska: hi there,
Ankit did you check this? We are first mkdir with recursive { true } which solves the issue.

ankitjena: @MStokluska I checked a bit and seems to be an issue the node versioning. I had downgraded my node version to v9.1.1 for some reason. The recursive true option works for anything above v10

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -21,61 +22,65 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
...
       }
     }
 
-    writeFileSync(outputSchemaPath, generated.schema)
-
-    writeFileSync(`${outputResolverPath}/index.ts`, generated.resolvers.index)
-    
-    if(!existsSync(`${outputResolverPath}/custom`)) {
-      mkdirSync(`${outputResolverPath}/custom`)
+    if(!existsSync(`${pathForSchema}`)) {
```
ankitjena:
```suggestion
    if(!existsSync(pathForSchema)) {
```

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -21,61 +22,65 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
...
+      mkdirSync(`${pathForSchema}`, { recursive: true })
```
ankitjena:
```suggestion
      mkdirSync(pathForSchema, { recursive: true })
```

ankitjena:
@MStokluska Works perfect. Good work on this

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -21,61 +22,65 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
...
     }
-
-    if(!existsSync(`${outputResolverPath}/generated`)) {
-      mkdirSync(`${outputResolverPath}/generated`)
+    if(!existsSync(`${customResolvers}/custom`)) {
```
ankitjena: Why not use `./src/resolvers/custom` directly in paths instead of appending it here. Same for generated resolvers.

templates/graphql-js-starter/graphback.json
```diff
@@ -17,5 +17,12 @@
     "subDelete": false,
     "disableGen": false
   },
+  "defaultPaths": {
```
darahayes: Minor question/suggestion. Is `defaultPaths` the name that will be present in the `graphback.json` file that users will modify? As an end user, the name `defaultPaths` is kinda weird because they're not default paths right? They're paths that I can change. Maybe something like `paths` or `files` would make a little more sense.

wtrocki:
@ankitjena @MStokluska do you mind takling a look. I'm not sure about config format. 
Still need to update some docs.

MStokluska:
> @darahayes @ankitjena @MStokluska do you mind takling a look. I'm not sure about config format.
> Still need to update some docs.

I like the new config a lot. I can see I've been on a good track with interfaces myself, just noticed thanks to your PR what I have been doing wrong! Looks like a really great job! Thanks @wtrocki

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -23,73 +25,68 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
...
         generatedClient[folder].forEach((c: ClientImplementation) => writeFileSync(`${currentFolder}/${c.name}.ts`, c.implementation))
       })
     }
-
   } catch (err) {
     logError(err)
     process.exit(0)
   }
+
+  function checkAndCreateFolders(pathForSchema: string, customResolvers: string, generatedResolvers: string) {
```
craicoverflow: This could be made more generic:

```js
function checkAndCreateFolders(...paths: string[]) {
	for const path in paths {
       if (!existsSync(path)) {
          mkdirSync(path, { recursive: true });
       }
	}
}
```

wtrocki: Good idea!

darahayes: Could throw an error mid loop if there were permissions problems

wtrocki: So IMHO let's not change that but.. also try catch it. 
@darahayes is an walking node.js reference :)

wtrocki:
#worksforme

packages/graphback-cli/src/helpers/generate.ts
```diff
@@ -23,73 +25,68 @@ followed by ${chalk.cyan(`${cliName}db`)} to create database.
...
     generated.resolvers.custom.forEach((output: OutputResolver) => {
-      if(!existsSync(`${outputResolverPath}/custom/${output.name}.ts`) || output.name === 'index') {
-        writeFileSync(`${outputResolverPath}/custom/${output.name}.ts`, output.output)
+      if (!existsSync(`${customResolvers}/${output.name}.ts`) || output.name === 'index') {
+        writeFileSync(`${customResolvers}/${output.name}.ts`, output.output)
       }
     })
 
-    const resolverFiles = readdirSync(`${outputResolverPath}/generated`)
-    resolverFiles.forEach((file: string) => unlinkSync(`${outputResolverPath}/generated/${file}`))
```
ankitjena: This was for the case that the user changes his types. So there would be no need for a resolver file from the older model.

packages/graphback-cli/src/config/ProjectConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Configuration for input/output folders
+     */
+    folders: FileConfig
+
+    /**
+     * Empty object config that can be created to generate client
+     * Reserved for future usage
+     */
+    client: object
```
ankitjena: Shouldn't this be `boolean` instead of `object`

ankitjena:
```suggestion
    client: boolean
```

packages/graphback-cli/src/helpers/generate.ts
wtrocki: Removing people's code - that they could have made modifications on is risky business. Especially that changes might not be committed. 
So people will be able to manually remove them later and there is no need for them to exist due to the fact that they will not be required in index. It is tradeoff for now. 

More work will be done in #340

packages/graphback-cli/src/config/ProjectConfig.ts
wtrocki: One rule in building software that people depend on (libraries and frameworks) is to not introduce breaking changes at the beginning. I know that we assign boolean but if we will have any other field for the client later it will require people to change all their configs. 

So I have decided to use object as form of enablement and did not change underlying implementation from boolean.

wtrocki:
So the only outstanding element here is to add extra try catch for fs/path methods.

craicoverflow:
I cannot complete step 1. `graphback init` fails.

Screenshot from 2019-09-17 12-39-10

MStokluska:
> I cannot complete step 1. `graphback init` fails.
> 
> Screenshot from 2019-09-17 12-39-10

Hey, I'm having the same issue.

packages/graphback-cli/src/config/ConfigBuilder.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Abstraction for Graphback configuration
+ */
+export class ConfigBuilder {
+
+    public config: ProjectConfig
+
+    constructor(creator: () => ProjectConfig) {
+        this.config = creator()
+        this.applyAbsolutePaths();
```
ankitjena: We should not be calling this here in `init`, this can be done later in `generate`. Moreover `process.cwd()` is different during `init`. This causes init to fail

wtrocki:
Have all fixed. Just need more time for tests

Issue #331, opened 2019-09-10
Defining a type extension in input schema generates duplicate queries, models and mutations.


craicoverflow:
## Steps to reproduce

Define a schema:

```graphql
type Note {
    title: String
}

extend type Note {
    hello: String
}
```

Run `graphback generate`.

The generated output schema will look like this:

```graphql
type Note {
  title: String
}

type Note {
  hello: String
}

input NoteInput {
  title: String
}

input NoteInput {
  hello: String
}

input NoteFilter {
  title: String
}

input NoteFilter {
  hello: String
}

type Query {
  findNotes(fields: NoteFilter!): [Note!]!
  findNotes(fields: NoteFilter!): [Note!]!
  findAllNotes: [Note!]!
  findAllNotes: [Note!]!
}

type Mutation {
  createNote(input: NoteInput!): Note!
  createNote(input: NoteInput!): Note!
  updateNote(id: ID!, input: NoteInput!): Note!
  updateNote(id: ID!, input: NoteInput!): Note!
}
```

Note the duplicate inputs, queries and mutations.

wtrocki:
Related issue #298

PR #332, opened 2019-09-11
fix: duplicate schema in type extension

ankitjena:
## Issue
#331 

## Summary
- Modified visited output to combine `ObjectTypeDefinition` and `ObjectTypeExtension` nodes.

## Verification
- Use type extension in model
```
type Note {
  title: String
}

extend type Note {
  description: String
}
```
- Run `graphback generate` and check for duplication in schema as in #331

### DIFF
--- a/packages/graphback/src/ContextCreator.ts
+++ b/packages/graphback/src/ContextCreator.ts
@@ -2,6 +2,7 @@ import { parse, visit } from 'graphql';
 import { Config, INTERFACE_TYPE_DEFINITION, InterfaceType, OBJECT_TYPE_DEFINITION, OBJECT_TYPE_EXTENSION, Type } from './ContextTypes'
 import { applyGeneratorDirectives } from './directives';
 import { inputTypeVisitor } from './InputTypeVisitor';
+import { filterInterfaceTypes, filterObjectExtensions, filterObjectTypes } from './utils';
 
 /**
  * create input context to be reused for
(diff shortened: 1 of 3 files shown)


wtrocki:
@ankitjena Great work on this. However, extend keyword has an additional issue I want to check. IF we move to graphql modules I will need to check if dependent modules will be able to see it thru extend (because we will get limited schema. This works now because we are using global contatenated schema). Leave that with me.

ankitjena:
@wtrocki Sure

packages/graphback/src/ContextCreator.ts
```diff
@@ -15,12 +15,35 @@ export const createInputContext = (schemaText: string, defaultConfig: Config): T
...
+    let context =  schemaDef.definitions.map((t: Type) => {
       return {
         ...t,
         config: { ...defaultConfig, ...t.config }
       }
     })
+
+
+
+    const extendNodes = context.filter((t: Type) => t.kind === OBJECT_TYPE_EXTENSION)
```
craicoverflow: You could use the available util function to do this filter.

https://github.com/aerogear/graphback/blob/2aeaeac9d57c2fdcc6aaf5cc8baeb698d548d766/packages/graphback/src/utils/index.ts#L21

packages/graphback/src/ContextCreator.ts
```diff
@@ -15,12 +15,35 @@ export const createInputContext = (schemaText: string, defaultConfig: Config): T
...
+
+    const interfaces = context.filter((t: Type) => t.kind === INTERFACE_TYPE_DEFINITION)
```
craicoverflow: You could use the available util function for this filter.

https://github.com/aerogear/graphback/blob/2aeaeac9d57c2fdcc6aaf5cc8baeb698d548d766/packages/graphback/src/utils/index.ts#L23

wtrocki: @ankitjena This is required for merge

wtrocki:
Tried and tested

PR #335, opened 2019-09-12
refactor: use graphql modules

craicoverflow:
https://github.com/aerogear/graphback/issues/337

The purpose of this pull request is to investigate the feasibility and benefits of refactoring Graphback to use [graphql-modules](https://graphql-modules.com/).

## Tasks

- [x] Separate type definitions into individual module folder.
- [x] Separate resolvers into their own module folder.
- [x] Update database generation process.
- [x] Figure out how to dynamically import modules into each other, based on required relationship between types defined in each schema.
- [ ] Move database access logic from resolvers into providers.
- [ ] Clean up.

## Verification

1. Define 3 .graphql files in the model directory.

**Interfaces.gql**

```gql
interface Person {
    name: String
}

```

**Comment.graphql**

```gql
type Comment {
    text: String
}
```
**Note.graphql**

```gql
type Note {
    title: String
    text: String
    comments: [Comment!]!
    author: Author
}

type Author implements Person {
    id: ID
    name: String
}

type Reader implements Person {
  name: String
}
```

2. Run `graphback generate`.
3. Run `graphback db`.
4. Run `npm run develop`.
5. Open your GraphQL playground and create/fetch some data.

### DIFF
--- a/packages/graphback-cli/src/helpers/db.ts
+++ b/packages/graphback-cli/src/helpers/db.ts
@@ -54,7 +54,7 @@ export const createDBResources = async(): Promise<void> => {
     }
 
      const modelDefinitions: IGraphbackModel[] = models.found.map((m: string) => {
-      const name = path.posix.basename(`${process.cwd()}/${m}`, '.graphql');
+      const name = path.basename(`${process.cwd()}/${m}`, '.graphql');
       const schemaText = readFileSync(`${process.cwd()}/${m}`, 'utf8');
 
       return {
--- a/packages/graphback-cli/src/helpers/db.ts
+++ b/packages/graphback-cli/src/helpers/db.ts
@@ -2,9 +2,10 @@ import chalk from 'chalk';
 import * as execa from 'execa'
 import { readFileSync, unlinkSync } from 'fs'
 import { GlobSync } from 'glob'
-import { DatabaseSchemaManager, GraphQLBackendCreator } from 'graphback';
+import { DatabaseSchemaManager, GraphQLBackendCreator, IGraphbackModel } from 'graphback';
 import { logError, logInfo } from '../utils'
 import { checkDirectory } from './common'
+import * as path from 'path';
 
 const configPath = `${process.cwd()}/config.json`
 
@@ -52,10 +53,17 @@ export const createDBResources = async(): Promise<void> => {
       await execa('touch', ['db.sqlite'])
     }
 
-    const path: string = process.cwd()
-    const schemaText: string = models.found.map((m: string) => readFileSync(`${path}/${m}`, 'utf8')).join('\n')
+     const modelDefinitions: IGraphbackModel[] = models.found.map((m: string) => {
+      const name = path.posix.basename(`${process.cwd()}/${m}`, '.graphql');
+      const schemaText = readFileSync(`${process.cwd()}/${m}`, 'utf8');
 
-    const backend: GraphQLBackendCreator = new GraphQLBackendCreator(schemaText, generation)
+      return {
+        name,
+        schema: schemaText
+      }
+    });
+
+    const backend: GraphQLBackendCreator = new GraphQLBackendCreator(modelDefinitions, generation)
 
     const manager = new DatabaseSchemaManager(database, dbConfig);
     backend.registerDataResourcesManager(manager);
--- a/packages/graphback-cli/src/helpers/generate.ts
+++ b/packages/graphback-cli/src/helpers/generate.ts
@@ -1,5 +1,5 @@
 import chalk from 'chalk';
-import { existsSync, mkdirSync, readdirSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
+import { existsSync, mkdirSync, readdirSync, readFileSync, unlinkSync, writeFileSync, exists } from 'fs';
 import { GlobSync } from 'glob'
 import { ClientImplementation, GraphQLBackendCreator, IGraphQLBackend, OutputResolver } from 'graphback'
 import { logError, logInfo } from '../utils';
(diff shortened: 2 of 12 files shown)


packages/graphback-cli/src/helpers/db.ts
```diff
@@ -52,10 +53,17 @@ export const createDBResources = async(): Promise<void> => {
       await execa('touch', ['db.sqlite'])
     }
 
-    const path: string = process.cwd()
-    const schemaText: string = models.found.map((m: string) => readFileSync(`${path}/${m}`, 'utf8')).join('\n')
+     const modelDefinitions: IGraphbackModel[] = models.found.map((m: string) => {
+      const name = path.posix.basename(`${process.cwd()}/${m}`, '.graphql');
```
darahayes: Is there any reason why you specifically needed to use `path.posix`? From the [Node Docs](https://nodejs.org/api/path.html#path_path_posix) I see that `The path.posix property provides access to POSIX specific implementations of the path methods.`

I may be missing something but Ideally we shouldn't need it and it could introduce problems later on.

craicoverflow: There was a reason I chose it, but that reason was wrong. I thought POSIX was for making `path` work on both Windows and Linux/Unix systems. Looking at the docs again I see that I just misunderstood the documentation. Thanks :+1:

craicoverflow:
@wtrocki I've gotten module imports working okay now. Tested with interfaces and types.

The developer can run into circular dependency issues, but in my view this should not be for us to resolve and it should be down to the developer to design and separate their modules to prevent this from happening.

wtrocki:
My thoughts after checking this:

- Can we still have a generic generator engine and let apply GraphQL modules on top of it. Especially for service layer that will need to be shared with rest (Do we need DI at all?)?

- Is GraphQL modules separation per file really working for us. 
Can entire generator (CRUD engine with types) be single module that people can just integrate with other modules they have to achieve customization? What value we will get from the modules in this case

- Can we use modules to bring extra common elements like:

- Scalar support (executable schema is missing those now)
- Directives we currently add in really hacky way

- If we provide extra package with some helper methods (graphback-runtime), can we automatically add some extras to the schemas 

- Would loaders (fb dataloader) (usually attached to context) be shared across different modules? How extensions will work in this case.

craicoverflow:
> My thoughts after checking this:
> 
> Can we still have a generic generator engine and let apply GraphQL modules on top of it. Especially for service layer that will need to be shared with rest (Do we need DI at all?)?

I'm not sure I understand. Let's talk about this.
 
>  Is GraphQL modules separation per file really working for us.
>  Can entire generator (CRUD engine with types) be single module that 

There are two ways we could do this:

1. If the developer wants to have a single module, they could just have a single input schema file.
2. We could merge all input schema files to generate a single schema file, like how it works in the current release, and then output a single module.

> 
>  Can we use modules to bring extra common elements like:
>     - Scalar support (executable schema is missing those now)
>     - Directives we currently add in really hacky way

I don't see how GraphQL Modules can bring scalar support.

> Would loaders (fb dataloader) (usually attached to context) be shared across different modules? How extensions will work in this case.

https://graphql-modules.com/docs/recipes/data-loader

My thoughts:

- GraphQL Modules are great for structuring an app in a modular, scalable way. However, this is very opinionated and does not let the developer choose how they wish to structure their application. If we want to offer full flexibility this might not be the way to go.
- We _could_ have a flag (disabled by default), to use GraphQL Modules. Something like `graphback init graphback-app --graphql-modules=true`. However this would require a lot of work for something that is not of core importance and **might not be used**. I don't think we should do this.
- I'm not sure if this is something we should implement. Let's talk about this.

craicoverflow:
@wtrocki do you think https://github.com/aerogear/graphback/issues/337#issuecomment-532258426 is a good approach? Should I do it?

craicoverflow:
> @wtrocki
> 
> > My thoughts after checking this:
> > Can we still have a generic generator engine and let apply GraphQL modules on top of it. Especially for service layer that will need to be shared with rest (Do we need DI at all?)?
> 
> I'm not sure I understand. Let's talk about this.
> 
> > Is GraphQL modules separation per file really working for us.
> > Can entire generator (CRUD engine with types) be single module that
> 
> There are two ways we could do this:
> 
>     1. If the developer wants to have a single module, they could just have a single input schema file.
> 
>     2. We could merge all input schema files to generate a single schema file, like how it works in the current release, and then output a single module.
> 
> 
> > Can we use modules to bring extra common elements like:
> > - Scalar support (executable schema is missing those now)
> > - Directives we currently add in really hacky way
> 
> I don't see how GraphQL Modules can bring scalar support.
> 
> > Would loaders (fb dataloader) (usually attached to context) be shared across different modules? How extensions will work in this case.
> 
> [graphql-modules.com/docs/recipes/data-loader](https://graphql-modules.com/docs/recipes/data-loader)
> 
> My thoughts:
> 
>     * GraphQL Modules are great for structuring an app in a modular, scalable way. However, this is very opinionated and does not let the developer choose how they wish to structure their application. If we want to offer full flexibility this might not be the way to go.
> 
>     * We _could_ have a flag (disabled by default), to use GraphQL Modules. Something like `graphback init graphback-app --graphql-modules=true`. However this would require a lot of work for something that is not of core importance and **might not be used**. I don't think we should do this.
> 
>     * I'm not sure if this is something we should implement. Let's talk about this.
> 
>     * What is clear is that we should have better separation between schema, resolver, provider and database. We may be able to use [`@graphql-modules/di`](https://www.npmjs.com/package/@graphql-modules/di) for DI which is independent from `@graphql-modules/core`.

wtrocki:
## Update

the way I want to merge this PR is to

- Introduce APILayerLayout abstraction in separate PR and merge it first
- Rebase your PR and put changes as separate abstractions rather than changing current code. 

For the time being, we will have 2 abstractions to deal with 
but IMHO I do not want to make some strong architectural decisions in very short time.

 I think that if we go with layout you can:

- use graphql code modules per file
- Use graphql modules outside and simply wrap generated schema and resolvers into module in 

If we merge it now we going to loose initial implementation thus 
it will be hard to have case study for extensibility reasons.
my recommendation will be to move out from the graphql modules and explore capabilities of external helpers like:

- Module for logging
- Module for authentication etc.

That was initial idea of the spike - I have done some research that I need to document
but you can do this as well

wtrocki:
Closing as integrating this work will be quite hard now but we will take bits of this work in new architecture

PR #347, opened 2019-09-20
fix: remove chokidar

ankitjena:


wtrocki:
@ankitjena Any motivation for this change? Why we removing this package. 
There are couple unused packages so if you want to clean this up let's use some tool to check them

ankitjena:
We removed `watch` so no need of this, and used to `depcheck`

Issue #360, opened 2019-09-24
Resolvers generation


MStokluska:
### Issue summary

Currently, if our model contains an entity that has 2 or more attributes of the same type it will generate resolvers as many times as the type has been used. 

### To recreate

- use `graphback init` with shop example model
- navigate to `model/Shop.graphql` and add `owner` field of type `Person!` to the model
- run graphback generate
- navigate to `src/resolvers/generated/person.ts` 

![image](https://user-images.githubusercontent.com/43473912/65525318-38025780-dee7-11e9-97c8-098b474bf108.png)
- run `graphback db` and `npm run develop`, the following error should be produced:

![image](https://user-images.githubusercontent.com/43473912/65525431-67b15f80-dee7-11e9-9631-40e2031cb26d.png)

wtrocki:
I think this feature should not be there as it is strange to have this type of connection as person do not have capabilities to fetch shop.

ankitjena:
Since this is created as part of nested relation, it's just repeated. Keeping one will work for both. So we can check if it's already been added and skip append.

wtrocki:
Which leads to the point that we need object representation of the resolvers first and then map it to the template. This will mean that we can test things nicely without inspecting strings. 

Another problem I see is that I'm not sure if the relation should be there in the first place.
There is no shop relation in any of the types so this cannot be queried.

https://github.com/aerogear/graphback/blob/master/packages/graphback-cli/src/templates/resources/models/Shop.graphql#L1-L17

We are missing product relations for example etc.

PR #363, opened 2019-09-27
fix: creating unnecesarry types

MStokluska:
#360 

Leaving it as a draft for now as I'm still investigating whether that's a solution for the issue

Issue before:
image

When for example Shop had an attribute of type of another entity it would generate resolver for that entity also. Creating ` *:* ` relationship out of ` 1:* ` causing it to fail.

Fix:

image

Given above schema our person resolvers look like this now:
image

and shop resolvers :

image

### DIFF
--- a/packages/graphback-cli/src/templates/resources/models/Shop.graphql
+++ b/packages/graphback-cli/src/templates/resources/models/Shop.graphql
@@ -4,6 +4,7 @@ type Shop {
   ## Relationship
   products: [Product]!
   customers: [Person]!
+  owner: Person!
 }
 
 type Person {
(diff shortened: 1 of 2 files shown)


wtrocki:
@MStokluska Great job. As bonus we can add unit test to confirm there is no problem - it will require exposing additional data in generator:
https://github.com/aerogear/graphback/blob/master/packages/graphback/src/GraphQLBackend.ts#L99-L104

wtrocki:
@MStokluska Completely unrelated to issue but, I think there is also separate issue with resolver
I tried this PR and I think we cannot fetch customers. We are missing customer resolver for Shop. 
Is that right or I'm missing something?

MStokluska:
> @MStokluska Completely unrelated to issue but, I think there is also separate issue with resolver
> I tried this PR and I think we cannot fetch customers. We are missing customer resolver for Shop.
> Is that right or I'm missing something?

Hi,

Currently, as it stands:
image

I am able to fetch customers from the shop as you can see above, need a bit of clarification on the resolvers part.

PR #378, opened 2019-10-07
[Feat] Service architecture

wtrocki:
## Motivation

Reorganize graphback structure into individual subfolders that can be later stripped into individual packages. Introduce abstraction of service layer. 

This is SUPER BREAKING PR that is going to be long-running to cover most of the critical changes we need to in graphback. Creating early for visibility and collaboration.
I would try to maintain backward compatibility where possible

## New folder structure

- Inputs (abstraction for input types
- Generators (generators that can be future - GraphQL-Code-Generator packages)
- Components (like logging, subscription etc.)
Additional capabilities added to generation
- Layers (configurable and extendable data and service layers for generated applications)
- Migrations (database migrations - for db layers that need it)

Screenshot 2019-10-07 at 16 03 28

Each of these folders will be later split into individual packages or pushed to upstream.

## New architecture

Screenshot 2019-10-07 at 13 48 45

TODO explain service
TODO explain components

## Progress

- [x] Implement database layer 
- [x] Implement service layer
- [x] Add abstraction for ts/js generation in schema
- [x] Implement resolvers generator for service layer (file and inmem)
- [x] Adjust template for the changes
- [x] Adjust the cli for the changes
- [x] SQL Lite support
- [x] Remove legacy not needed generators
- [x] Integration tests for db layer
- [x] Integration tests for service layer
- [x] Reorganize package structure

## Breaking changes notes

- Changed the way templates connect to the database. 
The template itself needs to define a connection to the database
- Generated resolvers will no longer contain SQL queries 
- Schema templates do not compile automatically - developers need to use gql themselves

### Runtime layer

Temporary changes were added to the sample app.
Please review `runtime.ts` file.

The sample application will work now without generation when `config.runtume === true`
Steps to verify
1. Execute graphback migrate (as migration at runtime is not supported yet)
2. Start server :)
3. Enjoy runtime API :)

Useful queries:
```
# Write your query or mutation here
query find {
  findAllUsers{
    name
  }
}

mutation create {
  
  createUser(input:{name: "test"}){
    id
  }
}
```


packages/graphback/src/generators/resolvers/legacy-knex/targetResolverContext.ts
```diff
@@ -157,19 +133,19 @@ export const buildResolverTargetContext = (input: Type[], database: string) => {
...
-          if(f.directives.OneToOne) {
+          if (f.directives.OneToOne) {
             columnName = f.directives.OneToOne.field
           }
           relations.push({
             typeName: t.name,
             implementation: knex.typeRelation('OneToOne', columnName, f.name, f.type.toLowerCase())
           })
-        } else if(f.directives.OneToMany || f.isArray) {
+        } else if (f.directives.OneToMany || f.isArray) {
```
wtrocki: Default formatter is worng. Going to deal with this.

packages/graphback/src/generators/resolvers/services/ServiceRuntimeResolverGen.ts
```diff
@@ -0,0 +1,72 @@
+import { getFieldName, getTableName, ResolverType } from '../../..'
+import { Type } from '../../../input/ContextTypes'
+import { GraphbackCRUDService } from '../../../layers/service/GraphbackCRUDService'
+
+/**
+ * Generate runtime resolver layer using Apollo GraphQL format 
+ * and injected service layer. Service layer offers various capabilities like monitoring, cache etc. 
+ * so resolver logic can be kept simple and interchangable.
+ */
+export class ServicesRuntimeResolverGenerator {
```
wtrocki: @b1zzu You would love this. 
This will create a runtime resolver layer for testing. 
I created that first for building a suite of integration tests.
Going to push more examples

packages/graphback/src/layers/data/InMemoryDataProvider.1.ts
```diff
@@ -0,0 +1,23 @@
+import { GraphbackDataProvider } from './GraphbackDataProvider';
+/**
+ * In Mem data provider used for testing capabilities
+ */
+// TODO
+// tslint:disable-next-line: no-any
+export class InMemoryDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext> {
```
wtrocki: @b1zzu This is pushed as an example of InMemoryData provider. 
I would need to pick something. From what I see you have been researching SqlLite InMem that might make this implementation obsolete (as knex will work with that as well)

wtrocki:
@craicoverflow 942ce7d contains most of the API changes. I will rebase and append changes to that to make it complete.

packages/graphback/src/layers/data/PgKnexDBDataProvider.ts
```diff
@@ -0,0 +1,68 @@
+// tslint:disable-next-line: no-require-imports
+import * as Knex from 'knex';
+import { AdvancedFilter, GraphbackDataProvider } from './GraphbackDataProvider';
+import { NoDataError } from './NoDataError';
+ 
+/**
+ * Knex.js database data provider exposing basic CRUD operations. 
+ */
+// tslint:disable-next-line: no-any
+export class PgKnexDBDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext>{
```
b1zzu: @wtrocki why don't have just a Knex resolver which will be compatible with SQLite and Postgres?

packages/graphback/src/layers/data/GraphbackDataProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+    readObject(name: string, id: string, context?: GraphbackContext): Promise<Type>;
+    updateObject(name: string, id: string, data: Type, context?: GraphbackContext): Promise<Type>;
+    deleteObject(name: string, id: string, context?: GraphbackContext): Promise<string>;
+
+    // Find
+    findAll(name: string, context?: GraphbackContext): Promise<Type[]>;
+    findBy(name: string, filter: Type | AdvancedFilter, context?: GraphbackContext): Promise<Type[]>;
+
+    // Dataloader support
+    // Caching support for nested queries
```
b1zzu: @wtrocki I think the caching should be handled by the CRUDService, and for batching we should only expose one additional method `batchRead(name: string, ids: [string]): Promise<Type[]>`

packages/graphback/src/layers/data/GraphbackDataProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+    // readObjectWithCache(id: String, data: Type);
+}
```
b1zzu: The DataProvider should handle the relationships in order to take advantage of joins

packages/graphback/src/layers/data/InMemoryDataProvider.ts
```diff
@@ -0,0 +1,35 @@
+import { GraphbackDataProvider } from './GraphbackDataProvider';
+
+
+/**
+ * In Mem data provider used for testing capabilities
+ */
+// TODO
+// tslint:disable-next-line: no-any
+export class InMemoryDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext>{
```
b1zzu: Ad in-memory database we can use SQLite, but there are also alternatives like https://github.com/techfort/LokiJS that we could take in consideration, or also just write a javascript hash

packages/graphback/src/layers/data/InMemoryDataProvider.1.ts
b1zzu: If we create a generic Knex we can use SQLite as in-memory database so we don't have to create a separate Provider, but there are also alternative like https://github.com/techfort/LokiJS that we could explor

packages/graphback/src/generators/resolvers/services/ServiceRuntimeResolverGen.ts
b1zzu: I think the ResolverGenerator should receive a more advanced `inputContext` that contains also the fields names and the table names because both TS and Runtime resolvers will need them

b1zzu: Same for all of the `if config.create`, `if config.update`, `...` the config should be applied outside the Generator and the generator should only loop over the resolvers.

packages/graphback/src/layers/service/GraphbackCRUDService.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * TODO
+ */
+// tslint:disable-next-line: no-any
+export interface GraphbackCRUDService<Type = any, GraphbackContext = any> {
+    // FIXME Pagination support
+    // CRUD
+    createObject(name: string, data: Type, context?: GraphbackContext): Promise<Type>;
+    updateObject(name: string, id: string, data: Type, context?: GraphbackContext): Promise<Type>;
+    deleteObject(name: string, id: string, context?: GraphbackContext): Promise<string>;
```
b1zzu: If I would like to extend the GraphbackCRUDService to an Optimistic Locking Service that checks the model version before performing any mutation it would be impossible because with the current interface I couldn't read also the version args. I know it's probably a corner case but maybe should take it into consideration.

b1zzu: I've created a dummy example: https://github.com/aerogear/graphback/blob/alternative-runtime-resolver/packages/graphback/src/generators/resolvers/layered/LayeredRuntimeResolverGen.ts

packages/graphback/src/layers/data/PgKnexDBDataProvider.ts
wtrocki: Because there compatibility is theoretical. There will be difference between create method. Most DB do not have ability to return all data on create but postgress has. I  will simply extend this class and change this small element.

packages/graphback/src/layers/data/GraphbackDataProvider.ts
wtrocki: Very interesting. I like the batchRead idea in db level, however I was thinking about read with checking if id is already cached. Do you think we should be using some different approach?

As for data loading.. I think we can have composite (more than one service layers so loader will be there. I would want to add loader support only for nested data - where it makes sense.

packages/graphback/src/layers/data/GraphbackDataProvider.ts
wtrocki: Generally there are two ideas that we can use:
- use execution plan of graphql for relationships and make individual queries to db with cache for relationships. 

context.cache['user'][5] can contain cached data

- use root resolvers only and something like our own graphql-query-mapper.
This approach would work nicely with joins, but it will require devs to build their own equivalent of execution plan. 

For the moment we going to take approach nr1. I think you have really interesting ideas. If you have some article or blog that gives.more.info let me know

-

packages/graphback/src/layers/data/PgKnexDBDataProvider.ts
wtrocki: Another reason is the actual dependency to specific db. If we split to packages we do not want them to bring extra dependency thst will not be used

b1zzu: I see than it's better to have a dedicated one for PG, but with can still have a generic one from SQLite (MariaDB could also enter in the generic one)

packages/graphback/src/generators/resolvers/services/ServiceRuntimeResolverGen.ts
b1zzu: I've created a dummy example: [LayeredRuntimeResolverGen.ts](https://github.com/aerogear/graphback/blob/alternative-runtime-resolver/packages/graphback/src/generators/resolvers/layered/LayeredRuntimeResolverGen.ts)

packages/graphback/src/generators/resolvers/layered/LayeredRuntimeResolverGen.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+      if (resolverElement.config.delete) {
+        const deleteField = getFieldName(resolverElement.name, ResolverType.DELETE);
+        // tslint:disable-next-line: no-any
+        resolvers.Mutation[deleteField] = (_: any, args: any, context: any) => {
+          return this.service.deleteObject(tableName, args.id, context)
+        }
+      }
+      if (resolverElement.config.findAll) {
+        const findAllField = getFieldName(resolverElement.name, ResolverType.FIND_ALL, 's');
```
craicoverflow:
```suggestion
        const findAllField = getFieldName(pluralize(resolverElement.name), ResolverType.FIND_ALL);
```

This is pretty minor in the grand scheme of things, but we should use a library like [https://github.com/blakeembrey/pluralize](pluralize) to ensure the correct plural form of a word is used when creating resolvers.

In the current implementation, there is a possibility of incorrect words, such as "sheeps", "fishs".

I'm not expecting this to be changed here, in fact - I could create a Hacktoberfest issue?

templates/apollo-starter-ts/src/runtime.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Runtime test
+import Knex = require('knex');
+import * as jsonConfig from '../graphback.json'
+
+/**
+ * Method used to create runtime schema
+ * It will be part of of the integration tests
+ */
+export const createRuntime = async (client: Knex) => {
```
craicoverflow: Will runtime eventually be created from the Graphback package?

packages/graphback/src/generators/resolvers/layered/LayeredRuntimeResolverGen.ts
wtrocki: +1000

packages/graphback/src/layers/service/GraphbackCRUDService.ts
wtrocki: Target is to pass more to delete method than just id in case we need to make sure that delete is possible

packages/graphback/src/layers/data/GraphbackDataProvider.ts
```diff
@@ -4,32 +4,83 @@
...
 
-    // Find
+    /**
+     * Implementation for reading object 
+     * 
+     * @param name name of the object to create
+     * @param id id of the object
+     * @param context context object passed from graphql or rest layer
+     */
+    readObject(name: string, id: string, context?: GraphbackContext):  Promise<Type>;
```
b1zzu: @wtrocki Isn't simplest to just name it `read(` because otherwise to uniform the other methods we should rename the other methods (`findAllObjects`, `findObjectBy` and `batchReadObjects`)

wtrocki: I will also create separate PR to fix this interface both read and delete will after accepting full object instead of id only to provide extra eventual consistency. There is just simply too much in this PR to properly change that

packages/graphback/src/components/Logger.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Offers basic logging component providing ability to print some extra messages
+ */
+export interface GraphbackMessageLogger {
+    /**
+     * @param message message to log
+     * @param level additional context information to log
+     */
+    // tslint:disable-next-line: no-any
+    log(message: string, context?: any): void
```
b1zzu: @wtrocki did you ever tried to use opentracing + jaeger? or elasticsearch? I see both tools as an implementation of the GraphbackMessageLogger but for both of them to work well in addition to the log message is best practice to provide also some key=value tags to filter the logs more easily. Another requirement for jaeger in order to trace requests across a microservice is to put a session id in the request header (or somewhere else) and to report the session id with the log, but it should be possible to retrieve the session id from the context.
Maybe I'm out of topic here but I wanted to say that maybe and additional arguments to log interface could be useful to provide these key=value

wtrocki: Love this. Yes.. This is just the beggining to setup some places in the code for architecture reasons. This logging interface is wrong - it do not support levels etc. 
I just wanted to provide the whole architecture ion the code and start documenting it properly. Logging was the easiest one to have as it will give some message in the server that can be useful for offline use cases etc.

craicoverflow:
There seems to be an issue with `findBy`.

Executing a basic query using the generic filter query fails. See example:

```gql
query fetchTasks {
  findTasks(fields: {id:"1"}) {
    id
    text
  }
}
```

Console output:

```sh
querying object task with filter undefined
```

Prints the following stacktrace:

```sh
{
  "errors": [
    {
      "message": "The operator \"undefined\" is not permitted",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "findTasks"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: The operator \"undefined\" is not permitted",
            "    at Formatter.operator (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/formatter.js:203:13)",
            "    at QueryCompiler_PG.whereBasic (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/compiler.js:657:22)",
            "    at QueryCompiler_PG.where (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/compiler.js:379:34)",
            "    at components.map (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/compiler.js:116:69)",
            "    at Array.map (<anonymous>)",
            "    at QueryCompiler_PG.select (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/compiler.js:116:35)",
            "    at QueryCompiler_PG.toSQL (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/compiler.js:64:29)",
            "    at Builder.toSQL (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/query/builder.js:76:44)",
            "    at /home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/knex/lib/runner.js:30:36",
            "    at tryCatcher (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/util.js:16:23)",
            "    at /home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/using.js:185:26",
            "    at tryCatcher (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/util.js:16:23)",
            "    at Promise._settlePromiseFromHandler (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:547:31)",
            "    at Promise._settlePromise (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:604:18)",
            "    at Promise._settlePromise0 (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:649:10)",
            "    at Promise._settlePromises (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:729:18)",
            "    at Promise._fulfill (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:673:18)",
            "    at PromiseArray._resolve (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise_array.js:127:19)",
            "    at PromiseArray._promiseFulfilled (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise_array.js:145:14)",
            "    at Promise._settlePromise (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:609:26)",
            "    at Promise._settlePromise0 (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:649:10)",
            "    at Promise._settlePromises (/home/endaphelan/code/src/github.com/aerogear/graphback/templates/apollo-starter-ts/node_modules/bluebird/js/release/promise.js:729:18)"
          ]
        }
      }
    }
  ],
  "data": null
}
```

wtrocki:
@craicoverflow fixed

packages/graphback/src/layers/data/GraphbackDataProvider.ts
```diff
@@ -0,0 +1,86 @@
...
+    findBy(name: string, filter: Type | AdvancedFilter, context?: GraphbackContext): Promise<Type[]>;
+
+
+    /**
+     * Read multiple items by their id's (used for lazy data loading purposes)
+     * 
+     * @param name 
+     * @param ids array of identifiers that needs to be fetched
+     */
+    batchRead(name: string, ids: [string]): Promise<Type[]>
```
b1zzu:
```suggestion
    batchRead(name: string, ids: string[]): Promise<Type[]>
```

wtrocki:
I'm working on integration tests for service layer. Target is to merge this PR with still some things unfinished - like relationships and from now on create smaller PR's for codegeneration.

As PR is a huge preferred way to verify it is to review docs in the new sample application and try all methods that are currently implemented. I'm not going to provide review steps as it is best to test it using some empirical way an see if this breaks.

Huge thanks for @b1zzu and @craicoverflow for initial ideas. 
@craicoverflow once this is merged we can create a suggested issue for hacktoberfest.

wtrocki:
I will work on draft for extra tasks in the gist: https://gist.github.com/wtrocki/a0811bf32245f43a77927d281c02d0a3

wtrocki:
@craicoverflow I think we can merge migrations work into this branch and make it pass

packages/graphback/src/generators/schema/SchemaGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+        this.context = buildTargetContext(this.inputContext)
+        const customContext = createCustomSchemaContext(this.inputContext)
+        const schemaString = generateSchemaString(this.context, customContext)
+        if (this.formatter) {
+            return this.formatter.format(schemaString);
+        }
+
+        return schemaString
+    }
+
```
wtrocki: Remove those

templates/apollo-starter-ts/package.json
```diff
@@ -15,12 +15,13 @@
...
     "cors": "2.8.5",
     "express": "4.17.1",
+    "graphback": "0.9.2",
     "graphql": "14.5.8",
     "graphql-subscriptions": "1.1.0",
     "graphql-tag": "2.10.1",
     "knex": "0.19.4",
     "pg": "7.12.1",
-    "sqlite3": "4.1.0"
+    "sqlite3": "^4.1.0"
```
craicoverflow: This should be pinned.

templates/apollo-starter-ts/src/runtime.ts
```diff
@@ -0,0 +1,28 @@
...
+import { makeExecutableSchema } from 'graphql-tools';
+import Knex = require('knex');
+import { resolve } from "path";
+import * as jsonConfig from '../graphback.json'
+
+/**
+ * Method used to create runtime schema
+ * It will be part of of the integration tests
+ */
+export const createRuntime = async (client: Knex, pubSub: PubSub) => {
```
craicoverflow: Are we going to remove runtime support from `apollo-starter-ts`?

templates/apollo-starter-ts/package.json
craicoverflow: This should be pinned.

templates/apollo-starter-ts/src/runtime.ts
wtrocki: Yep

packages/graphback/src/generators/resolvers/layered/LayeredRuntimeResolverGen.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+
+      this.createSubscriptions(resolverElement, resolvers)
+    }
+    // TODO relationships
+
+    return resolvers;
+  }
+
+  // tslint:disable-next-line: no-any
```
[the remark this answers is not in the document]
wtrocki: hmm.. it is easier to disable this rule IMHO rather than letting all things to become any. 
Our usage of any is around generics. Usually people define single type like "Unknown" and then map that everywhere but I did not wanted to do this till we split to packages. 

No strong opinions

packages/graphback/src/generators/resolvers/legacy-knex/BaseKnexResolver.ts
```diff
@@ -70,7 +70,7 @@ public newSub = (typeName: string): string => {
 
 public updatedSub = (typeName: string): string => {
     return `updated${typeName}: {
-      subscribe: (_: any, __: any, context: GraphQLContext) => {
+      m  : (_: any, __: any, context: GraphQLContext) => {
```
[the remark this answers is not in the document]
wtrocki: WOW. I think I have broken that one by accident :)

packages/graphback/src/generators/schema/SchemaGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+ * 
+ * Schema is generated by passing type `inputContext` array containing information about all 
+ * provided types that needs to be included.
+ */
+export class SchemaGenerator {
+    private context: TargetContext
+    private inputContext: InputModelTypeContext[]
+    private formatter: SchemaFormatter;
+
+    constructor(inputContext: InputModelTypeContext[], formatter?: SchemaFormatter) {
```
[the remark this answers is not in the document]
wtrocki: Runtime. We getting plain schema string in runtime

packages/graphback/src/input/directives.ts
```diff
@@ -1,3 +1,5 @@
+
+// FIXME split directives into individual plugins
```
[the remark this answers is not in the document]
wtrocki: Like all migration, services, generators etc will need to be able to insert and consume directives.

templates/apollo-runtime-ts/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+ - KnexDBDataProvider
+ - PgKnexDBDataProvider
+ - RestDataProvider (Looking for contributions)
+
+Then developers can create runtime instance:
+
+```ts
+    const backend = new GraphQLBackendCreator(`your model string here`, {});
+    const dbClientProvider = new PgKnexDBDataProvider(client);
```
craicoverflow: It's not clear here what `client` is, or how to define it. An example should be include above this maybe?

templates/apollo-runtime-ts/README.md
```diff
@@ -0,0 +1,44 @@
...
+    const runtime = await backend.createRuntime(dbClientProvider, pubSub);
```
craicoverflow: It might not be clear what `pubSub` is. Does it have to be a specific library (graphql-subscriptions)?

templates/apollo-runtime-ts/README.md
```diff
@@ -0,0 +1,44 @@
...
+    console.log(runtime.schema,  runtime.resolvers)
+```    
+
+See: `./runtime.ts` for fully functional example.
```
craicoverflow: Maybe this should link to the runtime file?

Suggestion:

See [`./runtime.ts`](https://github.com/aerogear/graphback/blob/84200cd56b9f531db30ea4fb3d8d86555a449b00/templates/apollo-runtime-ts/src/runtime.ts#L30) for a fully functional example.

templates/apollo-runtime-ts/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Running example
+
+The project has been created using `graphback`. Run the project using the following steps. 
+- Start the database
+```
+docker-compose up -d
+```
+
+- create database (Data migration will be supported later by @craicoverflow)
```
craicoverflow: Data migration --> should be database **schema** migration

templates/apollo-runtime-ts/README.md
wtrocki: It is optional so skipping it :D

templates/apollo-runtime-ts/README.md
wtrocki: Done

PR #381, opened 2019-10-10
Db migrations

craicoverflow:
## Motivation

This change enables iterative database schema upgrades.

## Tasks

- [x] Add config option to support schema migrations.
- [x] Add new tables to schema.
- [x] Add new fields to existing table in schema.
- [x] Drop database if migrations is turned off.
- [ ] Add migration execution to runtime.
- [ ] Tests.
- [ ] Documentation.

## CLI Verification Steps

1. Add your schema to the models folder (or a custom location if you have changed it in `graphback.json`.
2. In `graphback.json`, set `db.initialization` to "UpdateDatabase".
3. Run `graphback db`.
4. Run `npm run develop`.
5. Add some data to the database.
6. Add a new field to one of your tables.
7. Add a new table.
8. Run `graphback db`.
9. Run `npm run develop`.
10. Check your database. It should have the new fields and new table(s), along with the old data you added in step 4.
11. In `graphback.json`, set `db.initialization` to "DropCreateDatabase".
12. Run `graphback db`. The schema should be recreated from scratch.

## Runtime Verification

`// TODO`


package.json
```diff
@@ -4,7 +4,9 @@
   "private": true,
   "description": "Mono repository for Graphback project",
   "main": "index.js",
-  "dependencies": {},
+  "dependencies": {
+    "@graphql-inspector/core": "^1.26.0"
```
wtrocki: Wrong place for this. We should have it in graphback (for the moment).
Remove package-log.json commit (maybe we can add it to git ignore.

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -52,14 +53,14 @@ export const createDBResources = async (configInstance: ConfigBuilder): Promise<
...
-    const backend: GraphQLBackendCreator = new GraphQLBackendCreator(schemaText, graphqlCRUD)
+    const schemaContext = new LocalSchemaProvider(folders.migrations, folders.model)
+    const backend: GraphQLBackendCreator = new GraphQLBackendCreator(schemaContext, graphqlCRUD)
 
     const manager = new DatabaseSchemaManager(database, dbConfig);
+
     backend.registerDataResourcesManager(manager);
 
-    await backend.createDatabase()
+    await backend.migrateDatabase(initialization)
```
wtrocki: Side note: Always add some info about breaking API (we have people using this outside cli so db and generate methods are part of the API)

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -77,7 +78,11 @@ Run ${chalk.cyan(`npm run develop`)} to start the server.
 export const createDB = async (): Promise<void> => {
   const configInstance = new ConfigBuilder();
   checkDirectory(configInstance)
-  await dropDBResources(configInstance)
+
+  if (configInstance.config.db.initialization === DatabaseInitializationStrategy.DropCreate) {
```
wtrocki: Dropping resources was intiial idea and also the biggest drawback we had. Since we have very basic migration I think dropping makes no sense anymore in future supported version. It was just quick hack to get us going, but we never actually catched up with doing it properly till today. I have no strong opinion, but do you see any use case where dropping entire database will be an sensible option. 

Dropping databases can have a serious impact if someone missconfigure their project. Once we add migration we will start comming with the cases to make modifications in non-destructive way. Dropping db kinda goes against it in really obvious way.

packages/graphback/package.json
```diff
@@ -46,6 +46,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@graphql-inspector/core": "^1.26.0",
```
wtrocki: Our approach is to pin dependencies (removing `^`)

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -21,17 +29,22 @@ import { logger } from './utils/logger'
 // Datamigration should be the component on it's own
 // TODO rename to backend
 export class GraphQLBackendCreator {
-
   private dataLayerManager: IDataLayerResourcesManager;
   private dbContextProvider: DatabaseContextProvider;
-  private inputContext: Type[]
+  private schemaContext: ISchemaProvider;
+  private inputContext: Type[];
+  private newSchema: string;
```
wtrocki: [Leaking respojnsibility] This fields should be part of the subclass. 
GraphQLBackendCreator is part of the public API so we might just move them to separate class

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -21,17 +29,22 @@ import { logger } from './utils/logger'
...
+  private oldSchema: string;
 
   /**
    * @param graphQLSchema string containing graphql types
    * @param config configuration for backend generator
    */
-  constructor(graphQLSchema: string, config: GraphQLGeneratorConfig) {
-    this.inputContext = createInputContext(graphQLSchema, config)
+  constructor(schemaContext: ISchemaProvider, config: GraphQLGeneratorConfig) {
```
wtrocki: [Question] Do we want migration to be always based on input model or reparsed generated schema. 
If this is input model then it shoud be `InputModelProvider` otherwise `SchemaProvider`.
Using `I` is discouraged by typescript
[Breaking change] Let's add that to breaking change log.

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -21,17 +29,22 @@ import { logger } from './utils/logger'
...
+    this.schemaContext = schemaContext;
+    this.oldSchema = schemaContext.getOldSchemaText();
+    this.newSchema = schemaContext.getNewSchemaText();
```
wtrocki: How about using `current` and `previousVersion`? 
Saying your schema is OLD sounds funny :)

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -88,27 +105,73 @@ export class GraphQLBackendCreator {
...
   public async createClient(): Promise<Client> {
-    const clientGenerator = new ClientGenerator(this.inputContext)
+    const clientGenerator = new ClientGenerator(this.inputContext);
 
-    return clientGenerator.generate()
+    return clientGenerator.generate();
   }
 
+  public async migrateDatabase(
+    initializationStrategy: DatabaseInitializationStrategy
```
wtrocki: I will keep strategy for future dev/prod migration scheme

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -88,27 +105,73 @@ export class GraphQLBackendCreator {
...
+  ): Promise<void> {
+    const context = this.inputContext.filter(
```
wtrocki: Would it be possible to have separate class in migration folder.
Sorry for being anal on API/BreakingChage/Responsibility rules

`GraphQLBackendCreator` is kinda front door to our library so I would keep this clean. 
Ideally here we should instantiate things that do migration.

packages/graphback/src/GraphQLBackendCreator.ts
wtrocki: Actually changing my mind - `DatabaseInitializationStrategy` is just enum - I would expect this to be class that will implement some interface - this is typical strategy pattern. If we want to keep this as flag for different strategies then we can name it `DatabaseMigrationOptions`

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -88,27 +105,73 @@ export class GraphQLBackendCreator {
...
-  public async createDatabase(): Promise<void> {
-    const context = this.inputContext.filter((t: Type) => t.kind === OBJECT_TYPE_DEFINITION && t.name !== 'Query' && t.name !== 'Mutation' && t.name !== 'Subscription')
     try {
       if (this.dataLayerManager) {
-        logger.info("Creating database structure")
-        await this.dataLayerManager.createDatabaseResources(this.dbContextProvider, context);
-        await this.dataLayerManager.createDatabaseRelations(this.dbContextProvider, context);
+        if (initializationStrategy === DatabaseInitializationStrategy.DropCreate || !this.oldSchema) {
+          logger.info('Creating database structure');
+          await this.dataLayerManager.createDatabaseResources(
```
wtrocki: [Question] Do you think that it will be possible to repurpose them to initially create database for dev reasons but not drop the actual database when db is there?

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -88,27 +105,73 @@ export class GraphQLBackendCreator {
...
+        logger.info('Database structure generation skipped.');
       }
     } catch (error) {
       // logger.error(`Error on Database creation ${error}`)
-      throw error
+      throw error;
     }
+  }
+
+  private getSchemaChanges(
```
wtrocki: Move to migration logic.

packages/graphback/src/migrations/changes.ts
```diff
@@ -0,0 +1,4 @@
+export enum CHANGES {
```
wtrocki: [Naming] Too generic name.
[Comments] Lack of comment "contains changes that will be taken into consideration..

Really good start but we will need to probably go thru the all changes and categorize them into:

- Breaking changes we do not support
- Changes that we support but are ambiguous (TYPE_ADDED - but in fact, it was renamed) 

Document and create tasks on how to implement the 20 from 80 % typical cases that we need to support

packages/graphback/src/migrations/schema/LocalSchemaProvider.ts
```diff
@@ -0,0 +1,33 @@
+import { ISchemaProvider } from './ISchemaProvider';
+import { buildSchemaText, removeFiles } from '../../utils';
+import { writeFileSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * Provides old and new schema from a local context
```
wtrocki:
```suggestion
 * Provides old and new schema from a local filesystem
```

packages/graphback/src/migrations/schema/LocalSchemaProvider.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @export
```
wtrocki: Nice!

packages/graphback/src/migrations/schema/LocalSchemaProvider.ts
```diff
@@ -0,0 +1,33 @@
...
+  public getNewSchemaText(): string {
+    return buildSchemaText(this.newSchemaDir);
+  }
+
+  public getOldSchemaText(): string {
+    return buildSchemaText(this.oldSchemaDir);
+  }
+
+  public async updateOldSchema(newSchema: string) {
+    await removeFiles(join(this.oldSchemaDir, '*.graphql'));
```
wtrocki: UUU.. kinda risky business. 
Why we need to do removing part? 
Are we ok to leave dangling files and let people manage their repository?

I'm assuming that all `old` ;) schemas are pushed to repo

packages/graphback/src/utils/graphqlUtils.ts
```diff
@@ -34,3 +38,38 @@ export const filterObjectExtensions = (types: Type[]) => types.filter((t: Type)
 export const createImplementsInterfaceString = (names: string[]) => {
   return `implements ${names.map((name: string) => name).join(' & ')} `;
 }
+
+/**
+ * Collects all GraphQL files in a directory and reads the content into a string.
+ *
+ * @param schemaDir - The directory of the schema file(s)
+ */
+export const buildSchemaText = (schemaDir: string): string => {
```
wtrocki: There can be some implications of this approach that will make validation hard. 
I think in the long run we can threat every file as individual migration. 

If for example, someone will copy the file, made some bogus modifications and forget about them then. 
The fact that we do not control how the models will be processed by migration means that there could be possible issues.

packages/graphback/src/utils/graphqlUtils.ts
```diff
@@ -34,3 +38,38 @@ export const filterObjectExtensions = (types: Type[]) => types.filter((t: Type)
...
+
+  return schemaText.length ? schemaText : undefined;
+}
+
+/**
+ * Builds a GraphQLSchema object from all .graphql files in a directory
+ *
+ * @param schemaDir - The directory of the schema files
+ */
+export const buildSchemaFromDir = (schemaDir: string): GraphQLSchema => {
```
wtrocki: This seems to be unused

wtrocki:
Amazing work! Really really awesome progress. In fact, we have made a lots of investigations that can help us to document how the migration process can look and what are possible problems.

packages/graphback/src/migrations/schema/LocalSchemaProvider.ts
wtrocki: But it might be redundant?

packages/graphback-cli/src/helpers/db.ts
craicoverflow: Someone in developer mode might want their database wiped every time. I was modelling it from Entity Framework:

https://www.entityframeworktutorial.net/code-first/database-initialization-strategy-in-code-first.aspx

But as you said, an accidental misconfiguration could be catastrophic.

Should we add a separate command to drop the database? How will someone in developer mode wipe and start clean if we remove it?

wtrocki: Actually this is a really good perspective. How to prevent pushing this in production is another idea.

PR #398, opened 2019-10-17
feat: add database initialization strategies

craicoverflow:
## Motivation

Adding database initialization strategies is the first step towards building robust schema migrations in Graphback. There are currently 4 initialization strategies:

- `DropCreateDatabaseAlways` - this will always drop and recreate the database every time the application is run.
- `DropCreateDatabaseIfChanges` - if there are changes to the schema then drop the database and create it again. Otherwise don't do anything.
- `UpdateDatabaseIfChanges` - Only update the database if there are changes to the schema.
- `CreateDatabaseIfNotExists` - Create the database if it does not exist. **Not fully working. Knex creates the database on connect so the database always exists**.

These can be executed at runtime or through the CLI.

- `graphback db` - create the database if it doesn't exist.
- `graphback db:recreate` - drop and create the database.
- `graphback db:update` - update the database if there are changes.

## Tasks

- [x] Add `DropCreateDatabaseAlways` strategy
- [x] Add `DropCreateDatabaseIfChanges` strategy
- [x] Add `UpdateDatabaseIfChanges` strategy
- [ ] Add `CreateDatabaseIfNotExists` strategy (requires changes)

## Verification Steps [CLI]

1. Run `graphback db` to create database. This will only create a database if it doesn't exist.
2. Make changes to input schema (add a new field or type).
3. Run `graphback db:update`. The database schema should reflect the new field/type added to the schema.
4. Run `graphback db:recreate`. The database should be dropped and created again.

### DropCreateDatabaseAlways

1. Open `templates/apollo-runtime-ts/src/runtime.ts`.
2. Set initialization strategy to `DropCreateDatabaseAlways`:

```ts
const databaseInitializationStrategy = new DropCreateDatabaseAlways(config);
```

3. Run application: `npm run develop`.
4. The database should drop and recreate itself every time.

### DropCreateDatabaseIfChanges

1. Open `templates/apollo-runtime-ts/src/runtime.ts`.
2. Set initialization strategy to `DropCreateDatabaseIfChanges`:

```ts
const databaseInitializationStrategy = new DropCreateDatabaseIfChanges(config);
```

3. Make changes to the schema.
3. Run application: `npm run develop`.
4. The database should drop and recreate **only if** the schema has been changed.

### UpdateDatabaseIfChanges

1. Open `templates/apollo-runtime-ts/src/runtime.ts`.
2. Set initialization strategy to `UpdateDatabaseIfChanges`:

```ts
const databaseInitializationStrategy = new UpdateDatabaseIfChanges(config);
```

3. Make changes to the schema.
3. Run application: `npm run develop`.
4. The database schema should be updated with the latest changes.


packages/graphback-cli/src/commands/dbRecreate.ts
```diff
@@ -0,0 +1,26 @@
+import chalk from 'chalk';
+import { DropCreateDatabaseAlways } from 'graphback'
+import { ConfigBuilder } from '../config/ConfigBuilder';
+import { createDB, postCommandMessage } from '../helpers'
+
+export const command = 'db:recreate'
```
wtrocki: `db:recreate` is not valid posix argument as `:` has special bash meaning. I would use a recreate-db or something like that.

packages/graphback-cli/src/config/FolderConfig.ts
```diff
@@ -27,4 +25,8 @@ export interface FolderConfig {
      */
     client: string
 
-}
\ No newline at end of file
+    /**
+     * Location to save migrations
+     */
+    migrations: string
```
wtrocki: So our idea is that people will push that to the repository?
I think that this will go away when we use database driven metadata right?
Just checking - no action

craicoverflow: Yep, that was initial idea.

packages/graphback/src/GraphQLBackendCreator.ts
```diff
@@ -60,21 +54,39 @@ export class GraphQLBackendCreator {
...
-    const resolverGenerator = new LegacyResolverGenerator(this.inputContext)
-    backend.resolvers = resolverGenerator.generate(database)
+    const resolverGenerator = new LegacyResolverGenerator(this.inputContext);
+    backend.resolvers = resolverGenerator.generate(database);
 
     return backend;
   }
 
+  public async initializeDatabase(strategy: DatabaseInitializationStrategy) {
+    const typeContext = this.inputContext.filter(
```
wtrocki: Looks like good place for helper.

packages/graphback/src/database/DatabaseConnectionOptions.ts
```diff
@@ -0,0 +1,5 @@
+export interface DatabaseConnectionOptions {
+  client: string;
+  // tslint:disable-next-line: no-any
+  connectionOptions: any;
```
wtrocki: How about our strategies will understand specific implementation (in future) and we will make them explicitly be aware of the engine they work with. Like `PostgressKnexDropAllNoMatterWhatStrategy`. WDYT?

packages/graphback/src/database/initialization/strategies/CreateDatabaseIfNotExists.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export class CreateDatabaseIfNotExists implements DatabaseInitializationStrategy {
+  private schemaManager: DatabaseSchemaManager;
+  constructor(databaseOptions: DatabaseConnectionOptions) {
+    this.schemaManager = new DatabaseSchemaManager(databaseOptions.client, databaseOptions.connectionOptions);
+  }
+
+  public async init(context: DatabaseContextProvider, types: InputModelTypeContext[]): Promise<void> {
+    await this.schemaManager.createDatabaseResources(context, types);
+    await this.schemaManager.createDatabaseRelations(context, types);
```
wtrocki: Relations are tricky as cascade rules and constraints might prefent things. 
For now this approach is ok, but later on we need to think how to properly manage relations.

packages/graphback/src/database/migrations/DataResourcesManager.ts
```diff
@@ -0,0 +1,379 @@
...
+/**
+ * Represents update for data type
+ */
+export interface IDataLayerUpdate {
+  oldType: InputModelTypeContext,
+  newType: InputModelTypeContext
+}
+
+/**
+ * Interface for creating atabase resources .
```
wtrocki: `database`

packages/graphback/src/database/migrations/DataResourcesManager.ts
```diff
@@ -0,0 +1,379 @@
...
+ * For example for relational databases implementors will execute DDL queries into database
+ */
+export interface IDataLayerResourcesManager {
```
wtrocki: Actually this interface is very strongly implying some specific details - as it assumes that we always create relationships (for mongo it might not be case). Let's not change that until we figure out metadata and database part. Just flagging

packages/graphback/src/database/migrations/DataResourcesManager.ts
```diff
@@ -0,0 +1,379 @@
...
+ * Types of database Relations
+ * 1-1, 1:m, m:n
+ */
+enum Relation {
+  oneToOne = 'OneToOne',
+  oneToMany = 'OneToMany',
+  manyToMany = 'ManyToMany',
+}
+
+const createDBConnectionKnex = (
```
wtrocki: Why this is here?

wtrocki:
Workflow test is now broken because of missing handling. See CircleCi tests

packages/graphback/src/database/initialization/strategies/CreateDatabaseIfNotExists.ts
craicoverflow: Fully agree. This was never intended to be the final solution but this PR needed some sort of scope limit.

packages/graphback/src/database/migrations/DataResourcesManager.ts
craicoverflow: This file always existed like this, it just shows as a new file because I moved it. See https://github.com/aerogear/graphback/blob/master/packages/graphback/src/migrations/DataResourcesManager.ts#L51

docs/commands.md
```diff
@@ -33,6 +33,17 @@ Execute `graphback` in your shell for more information
   Create databases resources based on your data model. Currently supports PostgreSQL and sqlite3.
 > **Note** - *Running `graphback db` drops your database and recreates it with tables and relationships.*
 
+## Update Database
+
+```sh
+graphback update-db
```
wtrocki: Question - we can suggest update-db as well in generate output. Good idea?

craicoverflow: Yep good idea!

craicoverflow: Good idea!

PR #402, opened 2019-10-24
fix: VariableFields and inputVariableFields methods

### LINKED ISSUE #401: Client queries and mutations are not generated as expected
# Bug Report

**Current behavior:**

The current implementation of methods that generates client queries and mutations are generating code like the followed:

```
export const findItem = gql`
  query findItem(id: ID!, name: String, title: String!) {
    findItem(fields: {id: $id, name: $name, title: $title}) {
      ...ItemFields
    }
  }

  ${ItemFragment}
`
```
The generated code is incorrect because in the second line the symbol **$** is missing before each field name.

**Expected behavior:**

Generation of code like the followed:

```
export const findItem = gql`
  query findItem($id: ID!, $name: String, $title: String!) {
    findItem(fields: {id: $id, name: $name, title: $title}) {
      ...ItemFields
    }
  }

  ${ItemFragment}
`
```

**Other information**

Probably the error is [here](https://github.com/aerogear/graphback/blob/master/packages/graphback/src/generators/client/targetContext.ts#L9) and [here](https://github.com/aerogear/graphback/blob/master/packages/graphback/src/generators/client/targetContext.ts#L15).
The map function should return something like **`\\$${f.name}...**

vinifarias:
### Description

This Pull Request corrects the implementation of methods that generates client queries and mutations, adding the symbol $ where it was missing.

Fixes #401

### DIFF
--- a/packages/graphback/src/generators/client/targetContext.ts
+++ b/packages/graphback/src/generators/client/targetContext.ts
@@ -6,13 +6,13 @@ const gqlImport = `import gql from "graphql-tag"`
 
 const variableFields = (t: InputModelTypeContext) => {
   return t.fields.filter((f: InputModelFieldContext) => !f.isType && !f.isArray)
-                  .map((f: InputModelFieldContext) => `${f.name}: ${f.type}${f.isNull ? '': '!'}`)
+                  .map((f: InputModelFieldContext) => `\$${f.name}: ${f.type}${f.isNull ? '': '!'}`)
                   .join(', ')
 }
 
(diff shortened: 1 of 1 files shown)


wtrocki:
Amazing change. Thank you so much.
I'm going to verify this tommorow.
Our release going to be next week so if you depend on this bug being fixed early let me know. We can do master release

vinifarias:
We need this bug fixed to expose to the client the queries and mutations in GraphQL Testx project.
But don't worry, we have some other issues to work that doesn't depend on this fix :smile:

PR #404, opened 2019-10-25
fix: Prerelease runtime components

wtrocki:
Prerelease (dev version)


templates/apollo-runtime-ts/package.json
```diff
@@ -15,7 +15,7 @@
     "apollo-server-express": "2.9.4",
     "cors": "2.8.5",
     "express": "4.17.1",
-    "graphback": "0.9.2",
+    "graphback": "^0.10.0-dev1",
```
craicoverflow: Should this be pinned?

PR #417, opened 2019-10-29
fix: broken master build

wtrocki:


wtrocki:
Breaking review rule as master is now broken. 
@craicoverflow Let's make sure that we do proper rebase next time.

PR #419, opened 2019-10-29
Deprecate cli

wtrocki:
REmove graphql-cli plugin as this is no longer supported by official graphql cli


wtrocki:
Published last version with note about depreciation.

PR #425, opened 2019-10-30
[UberPR] Extract codegen packages

wtrocki:
## Motivation

Start by extracting the common input package. 
Extract separate client package.
Remove duplicating elements in graphback

See documentation and README.md for more context.

## Package structure

Input package controls entire graphback ecosystem by building InputModel. 
Then model is being used by number of generators that interact with each other. 
On top of that we have database migration capabilities also controlled by input model and runtime package offering data access layers that can be used in templates.

All of that is going to be controled by GraphQL-CLI and integrate more deeply with GraphQL-Config (thru generator plugin rather than graphback specific namespace).

Screenshot 2019-10-31 at 12 19 53


wtrocki:
Client side gen is not including relationships by default
Working on that now

packages/graphback-codegen-input/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Input package is an abstraction that provides common entry point to all graphback generators:
+- Client generator
+- Schema generator
+- Various resolver generators
+- Various Service layer generators
+- Various DBLayer generators
+
+`InputProcessor` abstraction provides seamless way for providing valid input for generators. 
+This package is not designed to be used directly and it
```
wtrocki: forgot to finish

craicoverflow:
Failed to run both templates.

**apollo-starter-ts**

```sh
~/code/src/github.com/aerogear/graphback.pr/templates/apollo-starter-ts extract-codegen*
❯ graphback generate
TypeError: graphback_1.InputModelProvider is not a constructor
```

packages/graphback-codegen-client/src/transformer/index.ts
```diff
@@ -0,0 +1,19 @@
+import { InputModelTypeContext } from '@graphback/codegen-input';
+import {  createClientDocuments } from './targetContext';
```
craicoverflow:
```suggestion
import {  createClientDocuments } from './targetContext'
```

packages/graphback-codegen-input/README.md
```diff
@@ -0,0 +1,43 @@
+## GraphBack
```
craicoverflow: Is it Graphback or GraphBack? Because we are using both.

packages/graphback/src/index.ts
```diff
@@ -12,16 +11,14 @@ export * from './generators/resolvers'
...
 export * from './database/migrations/DataResourcesManager'
 
 // Logger management
 export * from './utils/logger'
-
-export * from './input/InputTypeVisitor'
-export * from './input/ContextCreator'
-export * from './input/ContextTypes'
-
+export * from '@graphback/codegen-input'
```
craicoverflow: This is getting exported from this file twice (see line 15).

templates/apollo-starter-ts/package.json
```diff
@@ -15,13 +15,13 @@
...
     "express": "4.17.1",
+    "graphback": "0.9.2",
     "graphql": "14.5.8",
     "graphql-subscriptions": "1.1.0",
     "graphql-tag": "2.10.1",
     "knex": "0.20.0",
     "pg": "7.12.1",
-    "graphback": "0.9.2",
-    "sqlite3": "4.1.0"
+    "sqlite3": "^4.1.0"
```
craicoverflow: I think this should be pinned.

templates/apollo-starter-ts/src/resolvers/custom/index.ts
```diff
@@ -0,0 +1,3 @@
+
+
+export const customResolvers = []
```
craicoverflow: Did you mean for this file to be committed?

templates/apollo-starter-ts/src/resolvers/generated/test.ts
```diff
@@ -0,0 +1,25 @@
+import { GraphQLContext } from '../../context'
+
+export const testResolvers = {
```
craicoverflow: Did you mean for this file to be committed?

packages/graphback-codegen-input/src/api/ContextTypes.ts
```diff
@@ -70,14 +71,17 @@ export interface InputModelFieldContext {
  * Context type returned from `InputTypeVisitor`
  */
 export interface InputModelTypeContext {
```
craicoverflow: Naming suggestions:

- TypeContext
- GraphbackModel
- GraphbackType
- InputModel
- GraphbackInputType
- ModelContext
- DataModel

I personally am not a huge fan of prefixing everything with `Graphback` as it just seems like redundant and long without having any extra beneficial information.

What we choose should be applied to all other types also, so `TypeContext`, `FieldContext`, `InterfaceContexr`.

templates/apollo-starter-ts/src/resolvers/generated/test.ts
wtrocki: ups..

packages/graphback-codegen-input/src/api/ContextTypes.ts
wtrocki: Awesome! Would it be possible to get issue and assign that to myself. 
I will action it separately. Do not want to forget about it though

wtrocki:
@craicoverflow Cannot replicate error above. Unit tests also executing generate in CI/CD

wtrocki:
> Should we also rename graphback to @graphback/core, and graphback-cli to @graphback/cli or are we constrained now due to graphback being a dependency in other projects? It just seems like right now we are creating two separate naming conventions.

Very good question. This might require some bigger oversight before we starting publishing stuff. 
I do want to simply create all ecosystem of packages and then see what we going to do.
Cli moves away. There is no core package in this sense as core package is an

`@graphback/codegen-input` 

Renames are hard and this is just the result of lack of future vision from the beginning. 
Graphback needs to stay but it will be an... Runtime package so we going to swap responsibilities and users will not notice.

wtrocki:
@craicoverflow Amazing work!

PR #427, opened 2019-11-01
feat: Extract schema generator to separate package

wtrocki:
## Motivation

Create a separate schema package
This may look like large change but it is really moving things around and changing the import in files.


wtrocki:
@craicoverflow I will add new file that will be simply testing what is required for schema changes.

Issue #430, opened 2019-11-04
Create database migration package


craicoverflow:
Extract the current database migration code from `graphback` into its own separate package.

For now we probably just need a single package, but more may be added:

- `@graphback/db-migrate`

craicoverflow:
Completed in https://github.com/aerogear/graphback/pull/433

wtrocki:
Awesome job!

PR #433, opened 2019-11-05
Extract database migration logic into its own package.

craicoverflow:
See https://github.com/aerogear/graphback/issues/430

## Checklist

- [x] Move database related logic into separate package
- [x] Rename graphback-codegen-core to graphback-core
- [x] Move utils to graphback-core
- [x] Add unit tests
Hahaha 
## Verification

1. `cd examples/runtime-example`
2. `npm run develop`
3. The server should load and the database schema should get migrated each time you change your GraphQL schema.


packages/graphback-db-migrate/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@graphback/db-migrate",
```
wtrocki: Like migrate but WDYT about `db-manage` as less specific.

packages/graphback-db-migrate/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "glob": "7.1.5",
+    "knex": "0.20.0",
+    "pino": "5.13.5"
+  },
+  "devDependencies": {
+    "@types/graphql": "14.2.3",
+    "@types/knex": "0.15.2",
+    "@types/node": "8.10.56",
+    "ava": "2.4.0",
+    "graphql": "14.5.8",
```
wtrocki: Nothing in the package uses graphql.

packages/graphback-db-migrate/src/utils/logger.ts
```diff
@@ -0,0 +1,19 @@
+import * as pino from 'pino'
```
wtrocki: Logging might need to be in the core to get the same approach in entire ecosystem

packages/graphback/src/index.ts
```diff
@@ -4,14 +4,11 @@
...
-// DDL layer
-export * from './database/migrations/DataResourcesManager'
-
 // Logger management
-export * from './utils/logger'
-export * from './database';
+// export * from './utils/logger'
 
 export * from '@graphback/codegen-client'
 export * from '@graphback/codegen-core'
```
wtrocki: Since this package is used everywhere maybe it makes sense to call it core?

packages/graphback/src/utils/index.ts
```diff
@@ -1,4 +1,2 @@
 
-export * from "./graphqlUtils";
-export * from "./fileUtils";
-export * from "./logger"
+// export * from "./logger"
```
wtrocki: Remove this

wtrocki:
Would it be possible to write simple unit/integration test for this package.
We can perform migration on sqllite or use running container on build

packages/graphback-db-migrate/src/utils/logger.ts
craicoverflow: Yeah, logging **was** in `graphback` package but this was causing a circular dependency, as `graphback` imported `@graphback/db-migrate`, and vice-versa. I moved it as this was the only package using it anyway. Separate issue or should this be resolved here?

packages/graphback/src/index.ts
craicoverflow: Renamed to something like `@graphback/core`?

wtrocki: yeah. This is just conversation. No strong opinion as it is used everywhere

packages/graphback-db-migrate/src/utils/logger.ts
wtrocki: Moving it to core is 30 seconds. Creating issue is 2 minutes - and it will not be done anyway. Pick one :)

packages/graphback/src/index.ts
craicoverflow: Yeah I agree with this. I believe I mentioned it somewhere before.

packages/graphback-db-migrate/package.json
craicoverflow: I'm on board with that :D

packages/graphback-db-migrate/package.json
craicoverflow: It does: https://github.com/aerogear/graphback/blob/2344719014711a1beab755dc487029252f85c401/packages/graphback-db-migrate/src/database/migrations/schema/GraphQLSchemaManager.ts#L2

packages/graphback-db-migrate/src/utils/logger.ts
craicoverflow: I can do it now. I just thought it was something we weren't wanting to do yet.

packages/graphback-core/src/graphql/graphqlUtils.ts
```diff
@@ -26,3 +29,24 @@ export const filterObjectExtensions = (types: InputModelTypeContext[]) => types.
 export const createImplementsInterfaceString = (names: string[]) => {
   return `implements ${names.map((name: string) => name).join(' & ')} `;
 }
+
+/**
+ * Collects all GraphQL files in a directory and reads the content into a string.
+ *
+ * @param schemaDir - The directory of the schema file(s)
+ */
+export const buildSchemaText = (schemaDir: string): string => {
```
wtrocki: IMHO this should not be in the core. This is very specific for the migration package and it is going to go away.

packages/graphback-db-manage/src/database/migrations/schema/InMemoryModelProvider.ts
```diff
@@ -0,0 +1,32 @@
...
+import { SchemaProvider } from './SchemaProvider';
+
+/**
+ * Provides old and new schema from application memory
+ *
+ * @export
+ * @class InMemoryModelProvider
+ * @extends {SchemaProvider}
+ */
+export class InMemoryModelProvider implements SchemaProvider {
```
wtrocki: 😍

packages/graphback-db-manage/tests/GraphQLSchemaManager.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  t.assert(invalidChangeTypes.length === 0);
+});
+
+ava('it should update the old schema to match the new one', (t: ExecutionContext) => {
+  const { schemaManager, provider } = setup();
+
+  schemaManager.updateOldSchema();
+
+  t.assert(provider.getCurrentSchemaText() === provider.getPreviousSchemaText());
+  t.snapshot(provider.getPreviousSchemaText());
```
wtrocki: This snapshot makes no sense for me. Snapshots should be used to test outcomes. This is just snapshoting what we provided wich in turn going to be a maintenance nightmare and zero value.

packages/graphback-db-manage/tests/snapshots/GraphQLSchemaManager.test.ts.md
```diff
@@ -0,0 +1,44 @@
...
+        "criticality": {␊
+          "level": "NON_BREAKING"␊
+        },␊
+        "type": "FIELD_ADDED",␊
+        "message": "Field 'age' was added to object type 'User'",␊
+        "path": "User.age"␊
+      }␊
+    ]`
+
+## it should update the old schema to match the new one
```
wtrocki: Not sure I get this case? This is about database so I would like to see how input (2 schemas) affects output (SQL files). Case above is good but to low level

wtrocki:
Let's add test for graphback-cli for this strategy. Let's improve docs of the package in README.md - it is really poor now. We are sorted then

wtrocki:
>  packages/graphback-db-manage/README.md 

Needs work

wtrocki:
All good from my side. 
Follow up work with testing can be done in separate PR! Amazing work man!

PR #436, opened 2019-11-05
Resolver generators

wtrocki:
# Motivation

Drop legacy knex generator and replace it with runtime framework - service/db layers. 
This change breaks our current template format and provides great upgrade path for future

What was done:

- Brought in graphql cli template that contains both client and server-side applications.
Both client and server can be now tested together. We no longer need to use graphiQL

- Fixed schemna generation and formating

- Fixed tons of client side generation issues

- Fixed resolver generation (but there are still some work with relationshipos and subscroptions

- Decomposed runtime layer - graphback as a dependency should be used only for generators and cli tools. BackendGenerator will still contain helper method for developer convenience as not all elements are available in runtime.

## Verification

1. Go to examples/generator-full-stack
2. Use graphback db to generate database (or graphql generate --db)
3. Run example
4. Enjoy queries and mutations working nicely.


packages/graphback-runtime/src/service/DefaultCRUDService.ts
```diff
@@ -26,99 +31,100 @@ export class DefaultCRUDService<T = any, GraphbackContext = any>
...
-        this.logger.log(`deleting object ${inputType.name}`)
+    public async delete(name: string, id: string, context?: GraphbackContext): Promise<string> {
+        this.logger.log(`deleting object ${name}`)
 
-        const result = await this.db.delete(inputType, id, context);
-        if (this.pubSub && inputType.config.subDelete) {
-            const topic = subscriptionTopicMapping(GraphbackOperationType.DELETE, inputType.name);
-            const payload = this.buildEventPayload('deleted', inputType, result);
+        const result = await this.db.delete(name, id, context);
+        if (this.pubSub && context && context.publish) {
```
craicoverflow: What would happen if subscriptions are enabled in `graphback.json`, for example, `subDelete: true`, but `publish` is false in `GraphbackContext`? The subscriptions would be added to the GraphQL schema but would not work update?

It is a question more than anything.

wtrocki: Subscriptions are not implemented yet. Working on this as we speak

wtrocki: Generally, the idea will be that we will need to pass info somehow about subscription being enabled in config.

wtrocki:
I have docs branch for this work. Also continuing with subscription support that needs to be added. 

@craicoverflow I think it will be better to verify that branch and tackle everything else as separate issue.

examples/generator-full-stack-example/README.md
```diff
@@ -0,0 +1,50 @@
...
+yarn install
+```
+
+- Start the database
+
+```sh
+docker-compose up -d
+```
+
+- Generate resources(schema and resolvers) and create database
```
craicoverflow:
```suggestion
Generate resources (schema and resolvers) and create database
```

packages/graphback-runtime/README.md
```diff
@@ -42,7 +42,17 @@ Then developers can create runtime instance:
 ```    
 
 
-Check our example application: https://github.com/aerogear/graphback/tree/master/packages/graphback-runtime/example 
+Check our example application: https://github.com/aerogear/graphback/tree/master/examples/runtime-example 
+
+
+## Usage with Generated resolvers
+
+@graphback/codegen-resolvers package generates an resolver layer that utilizes the crud layer generated by runtime framework.
```
craicoverflow:
```suggestion
@graphback/codegen-resolvers package generates a resolver layer that utilizes the CRUD layer generated by the runtime framework.
```

packages/graphback-runtime/README.md
```diff
@@ -42,7 +42,17 @@ Then developers can create runtime instance:
...
+By default resolver layer will base on the `crudService` field attached to context.
```
craicoverflow:
```suggestion
By default resolver layer will based on the `crudService` field attached to context.
```

packages/graphback-runtime/README.md
```diff
@@ -42,7 +42,17 @@ Then developers can create runtime instance:
...
+Developers can use helper function `createKnexRuntimeContext` that will build default context for resolver layer.
```
craicoverflow:
```suggestion
Developers can use the helper function `createKnexRuntimeContext` that will build default context for resolver layer.
```

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverGen.ts
```diff
@@ -36,44 +35,45 @@ export class LayeredRuntimeResolverGenerator {
       if (resolverElement.config.disableGen) {
         continue;
       }
-
+      const objectName = resolverElement.name.toLowerCase();
       if (resolverElement.config.create) {
         const resolverCreateField = getFieldName(resolverElement.name, GraphbackOperationType.CREATE);
         // tslint:disable-next-line: no-any
```
craicoverflow: We disable this rule in so many places. Do you think we should disable the rule as this line is littered everywhere?

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverGen.ts
```diff
@@ -91,12 +91,12 @@ export class LayeredRuntimeResolverGenerator {
   }
 
   // tslint:disable-next-line: no-any
-  private createSubscriptions(resolverElement: InputModelTypeContext, resolvers: any) {
+  private createSubscriptions(resolverElement: InputModelTypeContext, resolvers: any, objectName: string) {
```
craicoverflow: There is no longer a need to pass the entire `resolverElement` anymore if you pass `config`

```suggestion
private createSubscriptions(config: GraphbackGeneratorConfig, resolvers: any, objectName: string) {
```

packages/graphback-runtime/src/service/DefaultCRUDService.ts
```diff
@@ -1,124 +1,130 @@
...
-        this.logger.log(`querying object ${inputType.name} with filter ${JSON.stringify(filter)}`)
+    public findBy(name: string, filter: any, context?: GraphbackContext): Promise<T[]> {
+        this.logger.log(`querying object ${name} with filter ${JSON.stringify(filter)}`)
 
-        return this.db.findBy(inputType, filter, context);
+        return this.db.findBy(name, filter, context);
     }
 
-    public subscribeToCreate(inputType: InputModelTypeContext, context?: GraphbackContext): AsyncIterator<T> | undefined {
+    public subscribeToCreate(name: string, context?: GraphbackContext): AsyncIterator<T> | undefined {
```
craicoverflow: `context` is not used here. Will it be?

wtrocki: In generic scenario layers can support filtering and other capabilities. So I think that it should be there despite this particular implementation not benefiting from it.

.vscode/launch.json
```diff
@@ -37,7 +49,7 @@
       "args": [
         "generate"
       ],
-      "cwd": "${workspaceRoot}/examples/generator-example",
+      "cwd": "${workspaceRoot}/examples/generator-full-stack-example",
```
craicoverflow: `examples/generator-full-stack-example` could be `examples/generator-full-stack` if you change the name. No need (IMO) to repeat "example".

wtrocki: Perfect!

examples/generator-full-stack-example/graphback.json
```diff
@@ -18,10 +18,10 @@
   },
   "folders": {
     "model": "./model",
-    "resolvers": "./src/resolvers/",
-    "schema": "./src/schema",
+    "resolvers": "./server/src/resolvers/",
```
craicoverflow: Is this file needed now that we have `.graphqlrc`?

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -0,0 +1,111 @@
...
+      subscribe: (_: any, __: any, context: GraphQLContext) => {
+        validateContext(context)
+        return context.crudService.subscribeToCreate("${typeName.toLowerCase()}", context);
+      }
+    }`
+}
+
+export const updatedSub = (typeName: string): string => {
+  return `updated${typeName}: {
+      subscribe  : (_: any, __: any, context: GraphQLContext) => {
```
craicoverflow: I know subscriptions are not working yet, but will they be fixed in separate PR?

packages/graphback-codegen-schema/src/schemaTemplate.ts
```diff
@@ -122,26 +144,24 @@ export const generateSchemaString = (context: TargetContext, customContext?: Cus
...
 
-  ${inputs(inputFields)}
+${inputs(inputFields)}
 
-  ${filters(filterFields)}
-  `
+${filters(filterFields)}`
 
   if (allQueries) {
-    output += `\n\n  ${allQueries}`
```
craicoverflow: I think in future thngs like this could happen in separate PR as it increases PR size tremendously and does not hugely contribute towards PR scope.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -45,8 +43,7 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
     }
 
     // tslint:disable-next-line: no-reserved-keywords
-    public async delete(inputType: InputModelTypeContext,  id: string): Promise<string> {
-        const name = getTableName(inputType.name)
+    public async delete(name: string, id: string, data?: Type): Promise<string> {
```
craicoverflow: Is `data` needed here? I think it is unlikely to be used in `delete`

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -55,35 +52,31 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
...
         const dbResult = await this.db.select().from(name).where('id', '=', id);
         if (dbResult && dbResult[0]) {
             return dbResult[0]
         }
         throw new NoDataError(`Cannot read ${name}`);
     }
 
-    public async findAll(inputType: InputModelTypeContext, ): Promise<Type[]> {
-        const name = getTableName(inputType.name)
+    public async findAll(name: string, ): Promise<Type[]> {
```
craicoverflow:
```suggestion
    public async findAll(name: string): Promise<Type[]> {
```

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverGen.ts
wtrocki: I think we had talked about this in runtime PR. Generally ANY shoud not be used at all. 
This is very very specific exception where we declare runtime layer (and our code) will be actually an generic execution code so we do not benefit from typings at all (or even can do typings here)

.vscode/launch.json
wtrocki: Feel free to rename runtime-example as well in your PR :)

examples/generator-full-stack-example/graphback.json
wtrocki: Good point! It will be needed until we totally remove the CLI from graphback. 
I think we going to keep it for the time being in deprecated state while still allowing people to use it. There will be no templates etc. This is more to discussion on how we polish the release. In this PR I wanted to enable GraphQL Cli but do not close dors on our own cli still used in tests etc.

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
wtrocki: Yes. Subscriptions bring a lots of changes. I prefer to have small dedicated PR to fix them rather than overloading this PR - same with pagination and batching

packages/graphback-codegen-schema/src/schemaTemplate.ts
wtrocki: Yes.. This is on me. I tried to pull some things back from this PR however the way I approached this PR was to make things working on top of the template, which sadly involved fixing issues in the schema and client-side generators.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
wtrocki: Delete uses data for conflict resolution. In offix spec we need data for version (or any marker) to determine if operation triggered conflict.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
wtrocki: How this even compiles?

wtrocki:
Awesome review. I have fixed build issues and done `clean, bootstrap, build` to make sure that things working.

wtrocki:
The sample app is scary but it works :)
Screenshot 2019-11-11 at 21 24 46

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
wtrocki: I wanted to make verification of PR possible, but this meant changing more than usual.

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -0,0 +1,111 @@
+// TODO subscription support
+// TODO support for the database type instead of hardcoded lowercase
+export const createTemplate = (fieldName: string, typeName: string, subscription?: boolean): string => {
+  return `${fieldName}: (_, args, context) => {
```
craicoverflow:
```suggestion
  return `${fieldName}: (_: any, args: any, context: any) => {
```

wtrocki: We need that for codegen.

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -0,0 +1,111 @@
...
+      validateContext(context)
```
craicoverflow: Is there some way this can be done before this function is executed? I don't like it here as it doesn't seem like the  resolver's responsibility. I couldn't think of how to do this though.

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -0,0 +1,111 @@
...
+      return context.crudService.create("${typeName.toLowerCase()}", args.input, context);
+    }`
+}
+
+export const updateTemplate = (fieldName: string, typeName: string, subscription?: boolean): string => {
+  return `${fieldName}: (_, args, context) => {
+      validateContext(context)
+      return context.crudService.update("${typeName.toLowerCase()}", args.id, args.input, context);
```
craicoverflow: Because `context` is passed as `any` `context.crudService` is magically bound. Could we pass `context` as `GraphbackRuntimeContext`?

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
wtrocki: Good idea: Yes.. we can do root resolver or wrap every resolver with it.. 
Generaly this code is repeated everywhere as pattern for adding more stuff in future. 

IMHO we should move this check to runtime and call it something like - validateRuntimeContext. 

As for having a resolver layer for checks.
I think we can simply have this check in service (runtime) for the moment and remove it completely?

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
wtrocki: Separate PR.

PR #451, opened 2019-11-07
fix: do not parse fragments

wtrocki:
Fix for #449


wtrocki:
Fragments are imported into other queries and mutations so parsing makes no sense.

PR #455, opened 2019-11-07
Bugfix for invalid filenames 

wtrocki:


craicoverflow:
I want to verify this, but without more context it will take me time to figure out what to do. Could you provide some simple verification steps?

wtrocki:
We can test it with sample app now. Also snapshots have valid values

PR #461, opened 2019-11-12
chore: cleanup package and build process

wtrocki:
Removing invalid typings and not needed packages. Added extra typescript flag to support latest promise

### DIFF
--- a/packages/graphback/tsconfig.json
+++ b/packages/graphback/tsconfig.json
@@ -9,6 +9,7 @@
     "lib": [
       "esnext.asynciterable",
       "es2015",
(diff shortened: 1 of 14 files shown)


packages/graphback/tsconfig.json
```diff
@@ -9,8 +9,9 @@
     "lib": [
       "esnext.asynciterable",
       "es2015",
+      "es2018.promise",
```
craicoverflow: Is this needed for Knex?

wtrocki: yep

PR #462, opened 2019-11-12
Use runtime layer to validate context

wtrocki:
As in title. Fix for the issue from the previous PR.
This is still quite repetitive but at least the code is coming from runtime package


packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -2,43 +2,43 @@
...
   return `${fieldName}: (_, args, context) => {
-      validateContext(context)
+      validateRuntimeContext(context)
       return context.crudService.findBy("${typeName.toLowerCase()}", args.fields, context);
     }`
 }
 
 export const newSub = (typeName: string): string => {
   return `new${typeName}: {
       subscribe: (_: any, __: any, context: GraphQLContext) => {
```
craicoverflow: Should `GraphQLContext` be `GraphbackRuntimeContext`?

GraphQLContext does not have the `crudDb` or `crudService` properties.

PR #464, opened 2019-11-12
Runtime subscription

wtrocki:
## Motivation 

Support subscriptions for both runtime and generated source code. 
I have added extra parameter to the service layer that can determine runtime options. 
For example developers might decide after to not publish change to subscriptions. 

There are two types here:
GraphbackRuntimeOptions - runtime options and modifiers triggered on resolver level
GraphbackRuntimeContext - global context with applied service etc.

## Verification

```
subscription subtest {
  newNote{
    id
    title
    description
  }
```

### DIFF
--- a/docs/custom.md
+++ b/docs/custom.md
@@ -50,7 +50,7 @@ and will generated a resolver file in custom folder
 ```
 export const likePost = {
   Mutation: {
-    likePost: (_: any, args: any, context: GraphQLContext) => {
+    likePost: (_: any, args: any, context: any) => {
         // Implementation here
     }
   }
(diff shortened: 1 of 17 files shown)


examples/generator-fullstack/server/src/resolvers/index.ts
```diff
@@ -0,0 +1,6 @@
+import { commentResolvers } from './generated/comment'
```
wtrocki: Pushed by mistake. To be removed

wtrocki:
Rebased to avoid conflicts. Ready for the review!

PR #470, opened 2019-11-13
Introduce runtime database schema migrations from GraphQL schema

craicoverflow:
## What

This change enables automatic database schema migrations at runtime and with legacy CLI.

## How it works

`KnexMigrationProvider` (needs to be renamed) is an interface which is applied at runtime from the client. This takes the database client and the local migrations directory as constructor arguments.

## Using

1. Run the example at `examples/runtime-example`.
2. A migration should be added to your migrations directory at `./migrations/<migration-id>`.
3. Check the database. There should be a `gb_schema_migrations` table with the same metadata as in `./migrations/<migration-id>`
4. Check if the user table exists in the database.
5. Update your input model and run the application again. A new migration should be created and applied.

## Tasks

- [ ] Validate input schema
- [x] Filter by valid change type
- [x] Generate SQL scripts
- [x] Persist migration metadata to database
- [x] Persist migration metadata to file system.
- [x] Apply migrations from metadata.
- [x] Create an interface/provider for local and remote migrations
- [ ] Create an interface for remote only migrations
- [x] Tests
- [x] Verified that this works with CLI.
- [x] Error handling
- [x] Rename to `graphql-migrations`


wtrocki:
I like both the architecture and the approach we took here. I will have a couple of questions on the very low-level details.

packages/graphback-db-manage/src/utils/graphqlUtils.ts
```diff
@@ -1,6 +1,8 @@
 import { readFileSync } from 'fs';
 import { sync } from 'glob';
 import { join } from 'path';
+import { GraphbackChange, GraphQLSchemaChangeTypes } from '../changes/ChangeTypes';
```
wtrocki: Maybe we can rename this file

packages/graphback-db-manage/src/migrations/KnexMigrationManager.ts
```diff
@@ -0,0 +1,116 @@
+import { InputModelTypeContext, InputModelFieldContext } from '@graphback/core';
+import { MigrationProvider } from '../providers/MigrationProvider';
+import { GraphbackChange } from '../changes/ChangeTypes';
+import knex from 'knex';
+import { connect } from '../utils/knexUtils';
+import { SchemaMigration } from '../models';
+
+export class KnexMigrationManager {
+  // tslint:disable-next-line:typedef
+  protected primitiveTypesMapping = {
```
wtrocki: This is still hardcoded

packages/graphback-db-manage/src/utils/knexUtils.ts
```diff
@@ -0,0 +1,21 @@
+import * as Knex from 'knex';
+
+export const connect = (
```
wtrocki: It should not be here.

packages/graphback-db-manage/src/validators/InputValidator.ts
```diff
@@ -0,0 +1,17 @@
+import { diff } from '@graphql-inspector/core';
+import { GraphQLSchema } from 'graphql';
+import { invalidChanges } from './IsValidChangeType';
+
+export const validateInput = (newSchema: GraphQLSchema, oldSchema: GraphQLSchema) => {
```
wtrocki: This can go away

wtrocki:
We need to check if we still support relationships.

wtrocki:
@craicoverflow To summarize feedback:

- Rename package
- Document everything carefully in the README.md
- I think we can better clarify current names ProviderManager etc. 
- Reorganize file structure into what is API and what is specific implementatin
- Make it clear that Knex implementation is not depending on specific db.
- Add tests that will involve changes.

Bonus:
- If we can generate very simple README.md file along with the changes that will be awesome
(possibly separate PR)
- If we can provide 2 layers of the tests - One that validates snapshots that are saved
One that actually verifies on database level 
(possibly separate PR)

wtrocki:
Having issues with running runtime example:
```
[wtrocki@wtrockigraph runtime-example (pr/craicoverflow/470-1 ✗)]$ npm run start

> apollo-runtime-template@0.10.0-dev2 start /Users/wtrocki/Projects/gsoc/graphback/examples/runtime-example
> node dist/src/index.js

Knex: run
$ npm install pg --save
Cannot find module 'pg'
Error: Cannot find module 'pg'
```

wtrocki:
Finished testing. I think that there will be couple small fixes required to land it

packages/graphql-migrations/src/migrations/KnexMigrationManager.ts
```diff
@@ -0,0 +1,196 @@
...
+    try {
+      // tslint:disable-next-line: await-promise
+      await this.db.raw(migration.sql_up);
+
+      logInfo(`Executed migration ${chalk.cyan(migration.id.toString())}`)
+    } catch (err) {
+      handleError(err);
+    }
+
+    const now = new Date();
```
wtrocki: As we do not have rollback this can leave entire db in unfinished state meaning that there will be no way to revert or recover other than just wiping the database

packages/graphql-migrations/README.md
```diff
@@ -1,61 +1,73 @@
-## Graphback
+## graphql-migrations
 
-<p align="center">
-  <img width="400" src="https://github.com/aerogear/graphback/raw/master/website/static/img/graphback.png">
-  <br/>
-  Auto generate database structure, <br/>
-  GraphQL Resolvers and Queries from GraphQL types 🚀
-</p>
+Automatically updates your database structure from a GraphQL schema.
```
wtrocki: I would still add a reference to github and also graph back website.

PR #489, opened 2019-11-19
fix: generate database on generator startup

wtrocki:
Quick hacked version to enable migrations at runtime


wtrocki:
After consideration, we can use that for the demo purposes however there is not much value in having migrations for generator example. We can execute them by command-line client.

wtrocki:
There is no need to update the database for generator

PR #496, opened 2019-11-20
Remove model provider approach to loading input schema

craicoverflow:
## What

1. Remove the use of `InputModelProvider` and load the schema from the server/CLI directly.
2. Remove `graphql-migrations` as a dependency of `graphback`. `graphql-migrations` is not imported separately.

## Verification

Everything should work as before.


examples/generator-fullstack/.graphqlrc.yml
```diff
@@ -1,6 +1,4 @@
-schema:
-    ./server/src/schema/index.ts:
-        require: ts-node/register
+schema: './model/*.graphql'
```
wtrocki: GraphQL config is not only for graphback. There will be other projects basing on the entire generated schema. 
We cannot do that as it will break everything

craicoverflow: Sure thing. I've changed it now to load schema from directory (the old way).

Issue #495, opened 2019-11-20
Database migration doesn't work with sqlite


b1zzu:
It seems that SQLite doesn't convert automatically strings to integer, whenn I try to migrate the databse on sqlite I recive this error:
```
insert into `gb_schema_migrations` (`changes`, `id`, `model`, `sql_up`) values ('[{"type":"TYPE_ADDED","path":{"type":"Item"}}]', '1574260178473', '
  type Item {
    id: ID!
    title: String!
  }
', 'create table `item` (`id` integer not null primary key autoincrement, `title` text);') - SQLITE_CONSTRAINT: UNIQUE constraint failed: gb_schema_migrations.id
```

What am I doing:
```js
    const knex = Knex({
      client: "sqlite3",
      connection: { filename: ":memory:" }
    });
    const dir = tmp.dirSync();
    const migrater = new UpdateDatabaseIfChanges(knex, dir.name);
    await migrater.init(this.schema.getSchemaText());
```

My workaround `packages/graphql-migrations/src/migrations/KnexMigrationManager.ts`
```js
  public async createMigration(migration: SchemaMigration): Promise<void> {
    try {
      // convert id to integer
      (migration as any).id = parseInt(migration.id);
      // tslint:disable-next-line: await-promise
      await this.db.table(this.tables.migrations).insert(migration);
    } catch (err) {
      handleError(err);
    }

    return Promise.resolve();
  }
```

Graphback: 0.10.0-dev2

b1zzu:
@enda ping

wtrocki:
@b1zzu Thank you so much. Please use CreateDatabase strategy for the moment. 
We are going to address this issue.

b1zzu:
@wtrocki do you mean the old method?

wtrocki:
Yes. Old method

craicoverflow:
> @wtrocki do you mean the old method?

@d1bzzu 

https://github.com/aerogear/graphback/tree/master/packages/graphql-migrations#drop-and-recreate-database-every-time

const databaseInitializationStrategy = new DropCreateDatabaseAlways('pg', db);

You can use `DropCreateDatabase` initialization strategy which uses the old engine to create the database talbes, relationships etc.

b1zzu:
The `DropCreateDatabaseAlways` works fine

craicoverflow:
@b1zzu https://github.com/aerogear/graphback/pull/503 should have fixed this now.

wtrocki:
[...]

PR #503, opened 2019-11-22
Change `SchemaMigration` ID type to number

craicoverflow:
## Motivation

Fixes https://github.com/aerogear/graphback/issues/495

## What

Change the ID type of `SchemaMigration` to match the ID type in the database.

## Postgres

Postgres migrations should work as before.

## SQLite

1. Install sqlite3 to the runtime example.
2. Change database configuration to:

```
"db": {
  "database": "sqlite3",
  "dbConfig": {
    "filename": "./db.sqlite"
  }
},
```

3. Executing migrations with SQLite should now work.

### DIFF
--- a/packages/graphql-migrations/src/migrations/SchemaMigration.ts
+++ b/packages/graphql-migrations/src/migrations/SchemaMigration.ts
@@ -7,7 +7,7 @@ import { ModelChange } from '../changes/ChangeTypes';
  * @interface SchemaMigration
  */
 export interface SchemaMigration {
-  id?: string
+  id?: Number
   applied_at?: Date
   changes?: string
   model?: string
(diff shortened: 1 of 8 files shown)


wtrocki:
Our tests are not covering everything :( - This will break dataloader code in runtime where we do toString. Can you review our db layer in runtime and made fixes for that

packages/graphql-migrations/src/migrations/SchemaMigration.ts
```diff
@@ -7,7 +7,7 @@ import { ModelChange } from '../changes/ChangeTypes';
  * @interface SchemaMigration
  */
 export interface SchemaMigration {
-  id?: string
+  id?: Number
```
wtrocki: I do question this change without any motivation? 
Why we suddenly started doing it?
What issue can we link here?

wtrocki:
What issue that fixes. I have noticed that running migration with empty folder breaks stuff.
Why we need number vs string and why this matters?

wtrocki:
WDYT about moving runtime to sqllite?

Issue #506, opened 2019-11-22
unify methods exposed by individual generators


wtrocki:
Generators should have the same method name for generating content. 
Currently we use create/generate etc.

`SchemaGenerator` vs `createClient` vs `generateResolvers`

We should support options for all of them.
generateJSResolvers and generateTSResolvers should return typed results.

PR #507, opened 2019-11-22
Api improvements

wtrocki:
Changes tested in graphql-cli already. Mostly renaming to have unified patter for generators and also adding missing types

https://github.com/aerogear/graphback/issues/506

### DIFF
--- a/packages/graphback-codegen-schema/src/createSchema.ts
+++ b/packages/graphback-codegen-schema/src/createSchema.ts
@@ -1,0 +1,26 @@
+import { InputModelTypeContext } from '@graphback/core'
+import { SchemaGenerator, tsSchemaFormatter } from '.';
+import { gqlSchemaFormatter, jsSchemaFormatter } from './schemaFormatters';
+
+export interface SchemaGeneratorOptions {
+    // output format
+    format: 'ts' | 'js' | 'gql'
+}
+
+/**
+ * Create resolvers function that 
+ * @param inputContext 
+ * @param options 
+ */
+export const createSchema = (inputContext: InputModelTypeContext[], options: SchemaGeneratorOptions): string => {
+    if (options.format === 'ts') {
+        return new SchemaGenerator(inputContext, tsSchemaFormatter).generate();;
+    }
+    if (options.format === 'js') {
+        return new SchemaGenerator(inputContext, jsSchemaFormatter).generate();;
+    }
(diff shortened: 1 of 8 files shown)


wtrocki:
There is no way to verify changes as we simply making some minor type changes

packages/graphback-codegen-schema/src/createSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Create resolvers function that 
+ * @param inputContext 
+ * @param options 
+ */
+export const createSchema = (inputContext: InputModelTypeContext[], options: SchemaGeneratorOptions): string => {
+    if (options.format === 'ts') {
+        return new SchemaGenerator(inputContext, tsSchemaFormatter).generate();;
```
craicoverflow:
```suggestion
        return new SchemaGenerator(inputContext, tsSchemaFormatter).generate();
```

packages/graphback-codegen-schema/src/createSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+    if (options.format === 'js') {
+        return new SchemaGenerator(inputContext, jsSchemaFormatter).generate();;
```
craicoverflow:
```suggestion
        return new SchemaGenerator(inputContext, jsSchemaFormatter).generate();
```

packages/graphback-codegen-schema/src/createSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+    if (options.format === 'gql') {
+        return new SchemaGenerator(inputContext, gqlSchemaFormatter).generate();;
```
craicoverflow:
```suggestion
        return new SchemaGenerator(inputContext, gqlSchemaFormatter).generate();
```

packages/graphback-codegen-schema/src/createSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+    throw Error("Invalid format specified. `options.format` supports only `ts` and `js` flags ");
```
craicoverflow:
```suggestion
    throw Error("Invalid format specified. `options.format` supports only `ts` `js` and `gql` flags ");
```

Issue #481, opened 2019-11-18
Duplicate migrations are created when only local migration data exists


craicoverflow:
## What

If there are local migration files that do not exist in the database `schema_migrations` table, a duplicate entry will be created and applied in the database.

## How to fix

The current execution order is as follows:

1. Create migration from current model by comparing to previous model.
2. Check for not applied local/remote migrations.
3. Apply migrations.

PR #508, opened 2019-11-22
Fix failed migration execution

craicoverflow:
Closes https://github.com/aerogear/graphback/issues/481

## What

Migrations were not executing because reading the migration table with Knex was not mapping the ID column back to an number. This meant that local migrations were different to remote and it tried to create the migration twice, which failed.

This change also allows migrations to be executed when only local migration data exists.

## Verification

1. Run 2 migrations using the runtime-example. Everything should work.
2. Delete both migrations from the `gb_schema_migrations` table and delete the changes that were applied in them.
3. Now you should have two migrations on your local filesystem. Run the runtime example again.
4. Both of these migrations should be applied.

### DIFF
--- a/examples/runtime-example/model/runtime.graphql
+++ b/examples/runtime-example/model/runtime.graphql
@@ -4,15 +4,4 @@
 type User {
     id: ID!
     name: String
-    age: String
-}
-
-type City {
-    id: ID!
-    name: String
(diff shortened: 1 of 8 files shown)


wtrocki:
Reverified changes. Really good job.

PR #518, opened 2019-11-25
Fix relationships

wtrocki:
## Motivation

Remove bidirectional relationships that are not supported at the moment and cause validation issues


packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -14,6 +14,11 @@ export const inputVariableFields = (t: InputModelTypeContext) => {
     .join(', ')
 }
 
+export const inputVariableFieldsWithRelations = (t: InputModelTypeContext) => {
```
wtrocki: So this is not easy fix as t is missing relationships

wtrocki:
We need to remove https://github.com/aerogear/graphback/blob/relations-client/packages/graphback-codegen-schema/src/targetSchemaContext.ts#L141-L166

wtrocki:
This is still not mergeable - InputTypes should support relationships so we can actually update the tables.

craicoverflow:
1. `datamodel.graphql` should be updated to reflect relationship change.

```gql
type Note {
  id: ID!
  title: String!
  description: String!
}

type Comment {
  id: ID!
  title: String!
  description: String!
  note: Note!
}
```

How will we query notes and also return related comments?

```gql
query getNotes {
  findAllNotes {
    id
    comments {
      id
    }
  }
}
```

craicoverflow:
Looks good. With the migrations work the following schema will be represented as expected in the database.

```qgl
type Note {
  id: ID!
  title: String!
  description: String!
  comments: [Comment]
}

type Comment {
  id: ID!
  title: String!
  description: String!
  note: Note!
}
```

**Tables**

- note
  - id
  - title
  - description

- comment
  - id
  - title
  - description
  - noteId

But as you said - database queries and mutations will need to be fixed.

The following query is valid according to the generated schema:

```gql
query getComments {
  findAllComments {
    id
    description
    note {
      id
      title
      description
    }
  }
}
```

This throws error `select * from \"note\" where \"commentId\" in ($1) - column \"commentId\" does not exist` as the underlying database query is wrong.

wtrocki:
`noteId` as field name is assumption that will need to be build into core model or relationships. 
Users should be able to change that as well to whatever they feel. 

I think this entire comment should be moved out to https://github.com/aerogear/graphback/issues/539 and we should close this PR - to merge it we will need to provide validation for bidirectional stuff.

wtrocki:
CC @craicoverflow This got outdated as changes were merged in your PR

PR #521, opened 2019-11-25
Fix create mutations and use default export for resolvers

ardatan:
- Do not add fragment definitions' content into operation documents if format is `gql`.
- Fix parsing error in `create` operations
- Use `default` export instead of named exports in resolvers


wtrocki:
Created separate PR

PR #522, opened 2019-11-25
Fix create mutations

wtrocki:
Rebased version from @ardatan


wtrocki:
This works end to end with graphql cli giving us full typesafety

packages/graphback-codegen-resolvers/src/formatters/ApolloTsResolverFormatter.ts
```diff
@@ -46,13 +46,13 @@ const generateTypeResolvers = (context: TargetResolverContext, name: string, opt
   let resolverType = '';
   let typedImports = ''
   if (options.types) {
```
wtrocki: We do not have this done for our own CLI so this code can be tested with graphql cli. 
We can add that later.

wtrocki:
Index is still not valid

wtrocki:
Fixed index issue. Custom resolvers are still generated the old way.

packages/graphback-codegen-client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@graphback/codegen-client",
   "version": "0.10.0-rc2",
-  "description": "GraphQL-Code-Generator compatible client side queries generator",
+  "description": "GraphQL-Code-Generator compatibile client side queries generator",
```
craicoverflow:
```suggestion
  "description": "GraphQL-Code-Generator compatible client side queries generator",
```

PR #532, opened 2019-11-27
add relationship support to mutation

craicoverflow:
This is unfinished and does not work. It is merely a very rough spike to see how to implement a quick fix so that relationships could be generated in mutation templates

@wtrocki

### DIFF
--- a/examples/generator-fullstack/model/datamodel.graphql
+++ b/examples/generator-fullstack/model/datamodel.graphql
@@ -2,14 +2,14 @@ type Note {
   id: ID!
   title: String!
   description: String!
-  ## Relationship
-  comment: [Comment!]!
 }
 
 type Comment {
   id: ID!
   title: String!
   description: String!
+  # Relationship
+  note: Note!
 }
 
 type Query {
@@ -19,3 +19,5 @@ type Query {
 type Mutation {
   likeNote(id: ID!): Note!
 }
+
+
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -7,10 +7,22 @@ export const variableFields = (t: InputModelTypeContext) => {
     .join(', ')
 }
 
+export const generateInputVariableFields = (t: InputModelTypeContext, relatedTypes: InputModelTypeContext[]) => {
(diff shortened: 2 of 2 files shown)


examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -2,14 +2,14 @@ type Note {
...
-  ## Relationship
-  comment: [Comment!]!
 }
 
 type Comment {
   id: ID!
   title: String!
   description: String!
+  # Relationship
+  note: Note!
```
craicoverflow: Defining relationship from this side instead makes it easier to represent in mutation. Not a long term solution though.

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -7,10 +7,22 @@ export const variableFields = (t: InputModelTypeContext) => {
...
+  return t.fields.filter((f: InputModelFieldContext) => !f.isArray && f.type !== 'ID')
+    .map((f: InputModelFieldContext) => {
 
-export const inputVariableFields = (t: InputModelTypeContext) => {
-  return t.fields.filter((f: InputModelFieldContext) => !f.isType && !f.isArray && f.type !== 'ID')
-    .map((f: InputModelFieldContext) => `\$${f.name}: ${f.type}${f.isNull ? '' : '!'}`)
+      if (f.isType) {
+        const relation = relatedTypes.find((r: InputModelTypeContext) => r.name === f.name);
+
+        // TODO: get the relationship constraint type and name
```
craicoverflow: Next we would find the type that of the primary key of the related model and use that in the mutation

examples/generator-fullstack/model/datamodel.graphql
wtrocki: Actually we have it done by bidirectional relationships - they are added to each side if missing. We also have a single unified way to represent OneToOne and OneToMany so it works both ways on the server. The only problem is the resolver for other direction is not defined and client-side do not support stuff at all. This is something I was trying to quickly patch. 
To summarize - relationships are fundamentaly broken because:
- Legacy migration always builds oneToMany relationship
- Schema generator does bidirectional relationships (adding fields that were not defined in model)
- resolvers for birectional relationships are missing
- client side generator is missing the relationships queries

Ways to fix it:

- Agree on how we represent relationships comprehensively 
- Build a reliable model to represent relationships in `InputTypeModelContext` object.

PR #534, opened 2019-11-28
Remove GIF from docs

craicoverflow:
- This change removes the GIF of the terminal executing Graphback CLI commands.


wtrocki:
To do that we need to take last tag and made change on branch. Master has changes that should not be published

PR #538, opened 2019-12-02
Merge migration libraries

craicoverflow:
## What

This change merges both migration libraries into one package. Each API is still exposed as a different "mode" to create a migration.

### Old v New

The old migration API is still exposed publicly through the `runGraphqlMigration` method.

```ts
import { runGraphqlMigration } from 'graphql-migrations';

const changes = runGraphqlMigration(schemaText, dbInitializationStrategy);
```

The new migration API is exposed through the `migrate` method.

```ts
import { migrate } from 'graphql-migrations';

migrate(dbConfig, schemaText).then((ops) => {
   console.log(ops);
});
```

### Change default name transform

In the new library the default name transformation for columns and tables was as "snake case". This has changed to match Graphback's expected transformation.

1. The first letter of columns are lower-cased.
2. Table names are lower-cased.

### Auto-incrementing primary key

The new migration library did not support auto-incrementing IDs. Now it does.

### Prevent changing ID data type

The data type of an ID can no longer be altered in a migration.

### ID validation

- Every type must have an ID
- The ID field name must be "id".

### Resolver Generation

Resolvers now allow bidirectional querying.

**datamodel**
```gql
type Note {
  id: ID!
  title: String!
  description: String!
  ## Relationship
  comments: [Comment!]!
}

type Comment {
  id: ID!
  title: String!
  description: String!
  note: Note!
}
```

**queries**
```gql
query getComments {
  findAllComments {
    id
    description
    note {
      id
      title
      description
    }
  }
}

query getAllNotes {
  findAllNotes {
    id
    comments {
      id
	   title
    }
  }
}
```

## Usage docs

https://github.com/craicoverflow/graphback/blob/spike/graphql-migrate/docs/database-schema-migrations.md

### 1

This use case generates the corresponding tables (many-to-many not supported by Graphback)

```gql
type User {
  id: ID!
  """
  @db.manyToMany: 'users'
  """
  messages: [Message]
}

type Message {
  id: ID!
  """
  @db.manyToMany: 'messages'
  """
  users: [User]
}
```

Generates:

```
- message
- user
- message_users_join_table
  - usersId
  - messagesId
```

### 2

```gql
type Note {
  id: ID!
  title: String!
  description: String
  comments: [Comment]!
}

type Comment {
  id: ID!
  description: String
  note: Note!
}
```

### 3

```gql
type Note {
  id: ID!
  title: String!
  description: String
}

type Comment {
  id: ID!
  description: String
  note: Note!
}
```

### 4

```gql
type Note {
  id: ID!
  title: String!
  description: String!
  ## Relationship
  comment: [Comment]! @OneToMany(field: "noteComment")
}

type Comment {
  id: ID!
  title: String!
  description: String!
  noteComment: Note!
}
```

### 5

```gql
type User {
  id: ID!
  name: String!
  inTransactions: [Transaction!]! @OneToMany(field: "inUser")
  outTransactions: [Transaction!]! @OneToMany(field: "outUser")
}

type Transaction {
  id: ID!
  amount: Int!
  inUser: User! # no need to reference owner as a directive
  outUser: User!
}
```

Document:

```gql
mutation createUser {
  createUser(input:{name:"Davide"}) {
    id
    name
  }
}

mutation createEnda {
  createUser(input:{name:"Enda"}) {
    id
    name
  }
}

mutation createTransaction {
  createTransaction(input: {amount:100, inUserId:1, outUserId:2}) {
    id
    amount
    inUser {
      id
      name
    }
    outUser {
      id
      name
    }
  }
}

query getUsersAndTransactions {
  findAllUsers {
    id
    name
    inTransactions {
      id
      amount
    }
    outTransactions {
      id
      amount
    }
  }
}
```

### 6

```gql
type User {
  id: ID!
  name: String!
  address: Address @OneToOne(field: "user")
}

type Address {
  id: ID!
  loc: String
  user: User @OneToOne(field: "address")
}
```

## Tasks

- [x] Merge migration libraries
- [x] Split libraries into two "modes"
- [ ] Generate database schema mapping object
- [x] Change jest tests to ava
- [x] Remove CircleCI from package
- [ ] Update docs
- [x] Update tsconfig rules
- [x] Change default name transform
- [x] Validation for when type of primary key has changed.
- [x] Relationships
- [x] Make database schema name configurable

## To do (new PR)

- Generate a database schema mapping object
- Use annotations for relationships in Graphback
- SQLite support
- Many-to-many


b1zzu:
Hi @craicoverflow I have this use case:
```graphql
type User {
    id: ID!
    name: String!
    outTransactions: [Transaction!]!
    inTransactions: [Transaction!]!
}

type Transaction {
    id: ID!
    amount: Number!
    inUser: User!
    outUser: User!
}
```
does the graphql-migrations library handle this?

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
 
-- `DropCreateDatabaseAlways` - Drops and creates the database every time the application is run.
-- `UpdateDatabaseIfChanges` - Only update the database when your input schema has been changed.
+### Usage
 
-### Configuration
+The `migrate` method creates and updates your tables and columns to match your GraphQL schema.
 
-Here is an example of how to configure database initialization strategies.
+All the database operations are wrapped in a single transaction, so your database will be fully rolled back to its initial state if an error occurs.
```
wtrocki: This statement will be true only for Postgress. SQLLite, MySQL do not have this feature.
https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
 
 const pubSub = new PubSub();
+
 const runtime = await backend.createRuntime(dbClientProvider, pubSub);
+...
+```
+
+### Migration Options
+
+- `config`: database configuration options.
```
wtrocki: We are missing info what this config means. 
I found that embedding github file with docs would be much better for maintenance than replicating flags here

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
+- `schema`: a GraphQL schema object.
+- `options`:
+  - `updateComments`: overwrite comments on table and columns.
+  - `scalarMap`: Custom scalar mapping. Default: `null`.
```
wtrocki: Post merge Requirement. We need to try out if case with JSON scalar. See if we support saving unstructured data as string which will be important for the cases when migrating from REST

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
+
+### Defining your data model
+
+#### Primary key
+
+Each type must have a primary key. The primary key field must be `id` and the type must be `ID`.
+
+```gql
+type Note {
+  id: ID!
```
wtrocki: While this initial setup is 100% right we might need to have ability to point out different field as id (using annotation or directive)

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
+```
+
+#### Foreign key
+
+To set a foreign key, set a field reference to the related type.
+
+```gql
+type Comment {
+  id: ID!
+  note: Note! # this creates a `noteId` column in the `comment` table.
```
wtrocki: I would go and document customization case - add comment how to create comment that will specify some of the requirements for field names. 

Also note that there is separate documentation that mentions directive that were not changed here.

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
 
-## Limitations
+### Compatibility
 
-Schema migrations are in a very early phase. At present the change types that are allowed is limited to the following:
+The following database providers support full database schema migrations.
 
-- **TYPE_ADDED** - Adding a new GraphQL type to your model will create an associated database table.
-- **FIELD_ADDED** - Adding a field to an existing model will create a new column in your database table.
+- PostgreSQL
```
wtrocki: Note: Going to test SQLLite migrations. This is very important for us to be able to test stuff also to not drop support for the DB that is already supported

examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -1,21 +1,22 @@
...
+  """
+  @db.type: 'string'
+  @db.length: 100
+  """
   title: String!
-  description: String!
-  ## Relationship
-  comment: [Comment!]!
+  """
+  @db.type: 'text'
```
wtrocki: We should have 'text' as default IMHO

examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -1,21 +1,22 @@
...
-
-type Query {
-  getLikedNotes(id: ID!, names: [String]!): Note!
-}
-
-type Mutation {
-  likeNote(id: ID!): Note!
-}
+  """
+  @db.type: 'text'
```
wtrocki: We are missing docs for referencing all possibble db.* values

examples/generator-fullstack/src/schema/generated.ts
```diff
@@ -0,0 +1,79 @@
+
+export const schemaString = `
+
+
+type Note {
```
wtrocki: Shoul not be commited. Let's clean example app completely. 
There is test case that tests migration that we should change instead.

examples/runtime-example/model/runtime.graphql
```diff
@@ -1,7 +1,13 @@
...
+  ## Relationship
+  comment: [Comment!]!
+}
 
-## Model that will be used at server runtime
-
-type User {
-    id: ID!
-    name: String
+type Comment {
```
wtrocki: Same as above. We should revert this file

examples/runtime-example/src/runtime.ts
```diff
@@ -16,13 +17,19 @@ import { loadSchema } from './loadSchema';
...
   const dbClientProvider = new PgKnexDBDataProvider(client);
 
-  const dbInitialization = new UpdateDatabaseIfChanges(client, jsonConfig.folders.migrations);
+  const dbConfig = {
+    client: jsonConfig.db.database,
+    connection: jsonConfig.db.dbConfig
+  };
 
-  await migrate(schemaText, dbInitialization);
+  migrate(dbConfig, schemaWithDirectives).then((ops) => {
```
wtrocki: [Naming] Let's change this to `migrateDB`

packages/graphback-cli/src/helpers/db.ts
```diff
@@ -54,30 +33,35 @@ export const createDBResources = async (configInstance: ConfigBuilder, initializ
...
 
+    const dbConfig = {
+      client: config.db.database,
+      connection: config.db.dbConfig
+    }
+
     const schemaText = loadSchema(folders.model);
 
-    await migrate(schemaText, initializationStrategy);
+    const schemaWithDirectives = applyGeneratorDirectives(schemaText);
```
wtrocki: [RequiresDiscussion] By implementing migration this way we assume that model files are valid schema (and we trying to make them valid). This is very strong design shift from the initial graphback and it will require some refactor in other layers. If we want to keep migration to be based on schema then we should probably pass the generated schema instead? The trick is that we might not know directives that were applied to the schema or scalars etc. 

Building schema this way - without providing resolvers/scalars might end up with validation errors. 
We need to find the best solution for this/document it and make sure that everything in ecosystem follows it.

packages/graphback-cli/src/commands/db.ts
```diff
@@ -10,14 +8,7 @@ export const desc = 'Create database resources'
...
 
 export async function handler() {
-  const configInstance = new ConfigBuilder();
-  const config = configInstance.config;
-
-  const db = await connect(config.db.database, config.db.dbConfig);
-
-  const initializationStrategy = new DropCreateDatabaseAlways(config.db.database, db);
-
-  await createDB(initializationStrategy)
```
wtrocki: We probably mean to print `databaseOperations`?

packages/graphback-cli/src/commands/db.ts
```diff
@@ -10,14 +8,7 @@ export const desc = 'Create database resources'
...
+  await createDB();
```
wtrocki: We probably mean to print databaseOperations?

packages/graphback-cli/tests/workflow-test.ts
```diff
@@ -1,40 +1,31 @@
...
 
   t.true(existsSync(join(basePath, "graphback.json")))
   t.true(existsSync(join(basePath, "db.sqlite")))
-  t.true(existsSync(join(basePath, "client/src/graphql/fragments/Note.ts")))
-  t.true(existsSync(join(basePath, "server/src/schema/generated.ts")))
-  t.true(existsSync(join(basePath, "server/src/resolvers/generated/note.ts")))
-
-
+  t.true(existsSync(join(basePath, "client/src/graphql/fragments/Test.ts")))
+  t.true(existsSync(join(basePath, "src/schema/generated.ts")))
```
wtrocki: Any reason server was flattened?

Master still have it nested:
https://github.com/aerogear/graphback/tree/master/examples/generator-fullstack

packages/graphback-codegen-resolvers/src/api/targetResolverContext.ts
```diff
@@ -169,19 +169,20 @@ function createRelations(inputContext: InputModelTypeContext[]) {
     t.fields.forEach((f: InputModelFieldContext) => {
       if (f.isType) {
```
wtrocki: BUG: (not for this issue) - we need to check if type is not scalar etc.

packages/graphback-codegen-resolvers/src/api/targetResolverContext.ts
```diff
@@ -169,19 +169,20 @@ function createRelations(inputContext: InputModelTypeContext[]) {
...
         if (f.directives.OneToOne || !f.isArray) {
-          let columnName = `${t.name.toLowerCase()}Id`;
+          let columnName = `${f.name}Id`;
           if (f.directives.OneToOne) {
-            columnName = f.directives.OneToOne.field;
+            columnName = `${f.directives.OneToOne.field}Id`;
```
wtrocki: Do we still support directives? 
Is database migration library recomending to mix both?

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
```diff
@@ -72,7 +72,7 @@ export const typeRelation = (relation: string, columnName: string, fieldName: st
   if (relation === 'OneToOne') {
     return `${fieldName}: (parent, args, context) => {
       validateRuntimeContext(context)
-      return context.crudService.batchLoadData("${tableName}", "${columnName}", parent.id, context)
+      return context.crudService.findBy("${tableName}", { id: parent.${columnName} })
```
wtrocki: [Requires Discussion] Missing many2many. What will be our plan for supporting many2many relationship and querying.

packages/graphback-codegen-schema/src/targetSchemaContext.ts
```diff
@@ -144,24 +160,12 @@ export const buildTargetContext = (input: InputModelTypeContext[]) => {
             "type": 'Type',
             "relation": `${f.name}: ${f.type}`
           })
-          relations.push({
```
wtrocki: [Requires Discussion] So our approach is to not generate bidirectional relationships, but we do not provide validation on input schema.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -17,79 +17,79 @@ import { NoDataError } from './NoDataError';
...
-      
-            const resultsById = ids.map((id: string) => dbResult.filter((data: Type) => {
-                return `${data[relationField]}` === id
-            })) 
+    if (dbResult) {
 
-            return resultsById as [Type[]];
-        }
+      const resultsById = ids.map((id: string) => dbResult.filter((data: any) => {
+        return data[relationField] === Number(id)
```
wtrocki: [Requires discussion] this strongly implies that id will be number. Can we do oposite id being string?

examples/generator-fullstack/model/datamodel.graphql
craicoverflow: In Postgres there is no difference between TEXT and VARCHAR. However, there can be a difference in others, such as MySQL. In MySQL a TEXT column cannot be indexed. Since it is possible to customise, should we leave the default as VARCHAR?

packages/graphback-cli/tests/workflow-test.ts
craicoverflow: `initConfig` calls this:

https://github.com/aerogear/graphback/blob/a2bde3d2425dbf9a36b79cfb4802bcb46a8238c0/packages/graphback-cli/src/templates/configTemplates.ts#L83 which in turn returns these default values:
https://github.com/aerogear/graphback/blob/31e499203cb9a4d67ae60d4bc705c0ce240fa7bf/packages/graphback-cli/src/config/ConfigBuilder.ts#L11

which do not match the structure of generator-fullstack.

examples/generator-fullstack/model/datamodel.graphql
wtrocki: So those types are knex types or native types?

packages/graphback-codegen-resolvers/src/api/targetResolverContext.ts
craicoverflow: For now the idea was to recommend to use both. As you said we cannot do all in one PR, There is a lot of changes required to replace relationship directives with annotations.

packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts
craicoverflow: Logging issue for this as I feel that it requires separate discussion.

packages/graphback-codegen-schema/src/targetSchemaContext.ts
craicoverflow: In what scenario would you consider the input schema invalid? I'm not sure I understand. The following different datamodels will produce the same database structure.

```gql
type Note {
  id: ID!
  title: String!
  description: String
  comments: [Comment]!
}

type Comment {
  id: ID!
  description: String
  note: Note!
}
```

```gql
type Note {
  id: ID!
  title: String!
  description: String
}

type Comment {
  id: ID!
  description: String
  note: Note!
}
```

craicoverflow:
> Hi @craicoverflow I have this use case:
> 
> ```graphql
> type User {
>     id: ID!
>     name: String!
>     outTransactions: [Transaction!]!
>     inTransactions: [Transaction!]!
> }
> 
> type Transaction {
>     id: ID!
>     amount: Number!
>     inUser: User!
>     outUser: User!
> }
> ```
> 
> does the graphql-migrations library handle this?
> My expected db structure:
> 
> ```
> transactions:
>   id: primary key ...
>   amount: int ...
>   inUserId: foreign key ...
>   outUserId: foreign key ...
> 
> users:
>   id: primary key ..
>   name: varchar ...
> ```
> 
> how the model could look like:
> 
> ```graphql
> type User {
>   id: ID!
>   name: String!
>   inTransactions: [Transaction!]! @hasMany(ref: "inUser")
>   outTransactions: [Transaction!]! @hasMany(ref: "outUser")
> }
> 
> type Transaction {
>   id: ID!
>   amount: Number!
>  # The key could be avoided if the foreign_key will be genrated using filed name "inUser" and not the type name 
>   inUser: User! @belongsTo(key: "inUserId")
>   outUser: User! @belongsTo(key: "outUserId")
> }
> ```

If you adjust your model to look like this:

```gql
type User {
  id: ID!
  name: String!
  inTransactions: [Transaction!]! @OneToMany(field: "inUser")
  outTransactions: [Transaction!]! @OneToMany(field: "outUser")
}

type Transaction {
  id: ID!
  amount: Int!
  inUser: User! # no need to reference owner as a directive
  outUser: User!
}
```

Then your use case will work. Try it out:

```gql
mutation createUser {
  createUser(input:{name:"Davide"}) {
    id
    name
  }
}

mutation createEnda {
  createUser(input:{name:"Enda"}) {
    id
    name
  }
}

mutation createTransaction {
  createTransaction(input: {amount:100, inUserId:1, outUserId:2}) {
    id
    amount
    inUser {
      id
      name
    }
    outUser {
      id
      name
    }
  }
}

query getUsersAndTransactions {
  findAllUsers {
    id
    name
    inTransactions {
      id
      amount
    }
    outTransactions {
      id
      amount
    }
  }
}
```

packages/graphql-migrations/.gitignore
```diff
@@ -0,0 +1,21 @@
+.DS_Store
```
craicoverflow: TODO:Remove this file

examples/generator-fullstack/model/datamodel.graphql
craicoverflow: Knex types.

packages/graphback-cli/src/commands/db.ts
wtrocki: Proper logging and status update

packages/graphback-cli/tests/workflow-test.ts
wtrocki: Agreement. Let's revert this change. I have posted issue to fix that in CMD

packages/graphql-migrations/src/util/defaultNameTransforms.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function defaultColumnNameTransform(name: string, direction: NameTransformDirection) {
+  if (direction === 'to-db') {
+    return `${name.charAt(0).toLowerCase()}${name.slice(1)}`
+  }
+  return name
+}
+
+export function defaultTableNameTransform(name: string, direction: NameTransformDirection) {
+  if (direction === 'to-db') {
```
wtrocki: Ok so there is very clear way to determine column names and fields and we provide defaults, so if graphback will setup this values beforehand and use it everythwere then we do not need do to anything - we will have our own default way to map data using this functions. Right?

packages/graphql-migrations/src/util/defaultNameTransforms.ts
```diff
@@ -0,0 +1,15 @@
+import { NameTransformDirection } from '../abstract/generateAbstractDatabase'
+
+export function defaultColumnNameTransform(name: string, direction: NameTransformDirection) {
+  if (direction === 'to-db') {
+    return `${name.charAt(0).toLowerCase()}${name.slice(1)}`
```
wtrocki: Why we need that?

What will happen if schema will contain fields like this:

```
type User {
   USER_NAME: String
   USER_PASSWORD: String
   DATA_VALIDATED: Boolean
}
```

wtrocki: I think we simply my want to revert that and accept field names as they are?

docs/database-schema-migrations.md
```diff
@@ -5,51 +5,120 @@ title: Database Schema Migrations
...
+  note: Note! # this creates a `noteId` column in the `comment` table.
+}
+
+type Note {
+  id: ID!
+  title: String!
+}
+```
+
+> NOTE: The field name must match that of the related type. Support for custom field names is coming soon.
```
b1zzu: @craicoverflow if the custom fields works also for OnToOne we can remove this line

examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -1,21 +1,22 @@
 type Note {
   id: ID!
+  """
+  @db.type: 'string'
+  @db.length: 100
```
b1zzu: I'm getting lost, are we using comment-style-annotations or inline-annotations `@OneToMany`?

packages/graphql-migrations/src/util/defaultNameTransforms.ts
craicoverflow: Yes - unless the developer decides to customise a field name with annotations.

packages/graphql-migrations/src/util/defaultNameTransforms.ts
craicoverflow: You're right. Will need to also check how the code-generator does it.

examples/generator-fullstack/model/datamodel.graphql
craicoverflow: Myself and @wtrocki discussed this yesterday and agreed to go full comment-annotations in this PR. The inline directives are a legacy from the current Graphback approach and my plan was to do this in another PR (because this PR is already massive) but we agreed to do it in this PR.

wtrocki:
Published graphql-migrations@0.10.0-dev4

PR #541, opened 2019-12-04
fix: the deleteType mutation is returing the ID and not the Type

b1zzu:
The `deleteType` client mutation is broken because the mutation generated from graphback is returning the ID and not the whole Type
```
type Mutation {
  createTask(input: TaskInput!): Task!
  updateTask(id: ID!, input: TaskInput!): Task!
  deleteTask(id: ID!): ID!
}
```


packages/graphback-codegen-client/tests/snapshots/GraphQLClientCreatorTest.ts.md
```diff
@@ -34,11 +34,7 @@ Generated by [AVA](https://ava.li).
                 ...NoteFields␊
               }␊
             }␊
-          fragment NoteFields on Note {␊
-              id␊
-              title␊
-              description␊
-            }`,
```
b1zzu: I've updated the ava snapshots using `ava --update-snapshots` and also all fragments has been removed, is that correct?

PR #547, opened 2019-12-11
Migration testing

wtrocki:
General idea on how we should be doing testing using testx framework. Going to extend it by more tests


wtrocki:
I think I will simply commit more tests and request permission to push to your fork

wtrocki:
Ready for review

wtrocki:
Generally amount of tests is very small for the moment as there are major blockers in place.
From now on when we fix bugs we will try to resolve or add new unit test to cover this work. 
Any work without unit tests will not be merged as we cannot sustain breaking things anymore.

integration/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "test": "ava",
+    "lint": "echo disabled",
+    "start": "echo disabled",
+    "pretest": "npm run build"
+  },
+  "dependencies": {
+    "apollo-boost": "^0.4.4",
+    "graphback": "0.10.0-rc2",
+    "graphql": "^14.0.0",
+    "graphql-testx": "0.3.0-dev381.bede8a3",
```
b1zzu:
```suggestion
    "graphql-testx": "0.3.0-dev384.3a11247",
```
Just released the a new testx version with the new `Database` interface, we can easily integrate it, it's just a matter of rename a couple of objects and imports, but no cahges to the logic

craicoverflow:
I think the build is failing because the CLI and example apps need to use `graphql-migrations@0.10.0-dev4`

integration/tests/basic-runtime.ts
```diff
@@ -0,0 +1,98 @@
...
+import ApolloClient, { gql } from 'apollo-boost';
+import test, { ExecutionContext } from 'ava';
+import { unlinkSync } from 'fs';
+import { TestxServer } from 'graphql-testx';
+import { InMemoryDatabase } from 'graphql-testx/dist/src/InMemoryDatabase';
+import fetch from "node-fetch";
+import { migrateDB } from '../../packages/graphql-migrations';
+
+// tslint:disable-next-line: no-require-imports
+import Knex = require('knex');
```
craicoverflow:
```suggestion
import * as Knex from 'knex';
```

integration/schemas/relations-schema.graphql.ts
```diff
@@ -0,0 +1,15 @@
+
+export const schema =
+  `type Item {
+    id: ID!
+    name: String
+    title: String!
+    comments: [Comment!]!
```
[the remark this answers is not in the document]
wtrocki: My bad - this is the example we had in cli. Why this will not create a relationship?

integration/tests/basic-runtime.ts
```diff
@@ -0,0 +1,97 @@
...
+test.serial('Graphback runtime relationships - generic OneToMany', async (t: ExecutionContext) => {
+  const server = await createServer(relationsSchema);
+  // TODO relationships not supported in queries and runtime
+  await graphbackRuntimeWorkflow(server, t);
+});
+
+async function graphbackRuntimeWorkflow(server: TestxServer, t: ExecutionContext<unknown>) {
+  // Test migrations
+  const migration = await migrateDB(dbOptions, basicSchema);
+  const dbSchema = await server.getDatabaseSchema();
```
b1zzu: @wtrocki I had this idea of adding more information to the database schema, like the type (varchar, int, float, ..) and things like NOT NULL, UNIQUE, INDEX, ... would this improve your tests?

wtrocki: I think we are ok. We have that info from migration. This one is only because info from migration is soo big that we want to simply see table structure in very simple model for testing and verification purposes.

Issue #552, opened 2019-12-12
Change `graphql-migrations` API


craicoverflow:
## What

`graphql-migrations` has two main APIs: `migrate` and `runGraphqlMigration`. These should be renamed:

- `migrate`: `migrateDB`
- `runGraphqlMigration`: `migrateProd`

PR #559, opened 2019-12-12
feat: change public API methods in `graphql-migrations`

craicoverflow:
https://github.com/aerogear/graphback/issues/552

Change the public API methods in `graphql-migrations`.

- migrate: migrateDB
- runGraphqlMigration: migrateProduction

### DIFF
--- a/docs/database-schema-migrations.md
+++ b/docs/database-schema-migrations.md
@@ -125,7 +125,7 @@ SQLite is not yet fully supported but will be soon.
 
 ### Migrations in Production
 
-`migrateProduction` can be used to perform migrations in a controlled/production environment.
+`migrateDBUsingSchema` can be used to perform migrations in a controlled/production environment.
 
 Options:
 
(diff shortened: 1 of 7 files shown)


wtrocki:
WDYT?

migrateDB -> dev, default
migrateDBUsingSchema -> production/controled env

craicoverflow:
> migrateDBUsingSchema

Are you suggesting just a rename here?

I don't fully understand the choice of `migrateDBUsingSchema`. They both accept the schemaText as the schema source.

wtrocki:
Also `migrateProduction` is not documented. Having a short chapter will be beneficial?
We had some docs before

wtrocki:
migrateProduction it is not reflectin what this is happening.
MigrateProduction is not a destructive way to migrate changes with audit log that uses schema as source and ignores db. migrateDB compares db and schema

craicoverflow:
> migrateProduction it is not reflectin what this is happening.
> MigrateProduction is not a destructive way to migrate changes with audit log that uses schema as source and ignores db. migrateDB compares db and schema

Yeah I see what you mean now.

PR #523, opened 2019-11-25
chore: add one-to-many resolver to runtime package

b1zzu:
## Description

Implementing missing one-to-many relation in runtime package

PR #568, opened 2019-12-17
One to many relationship

craicoverflow:
This change makes it possible to create a database relationship from the 1:N side of the relationship.

Consider the following example:

```gql
type User {
  id: ID
  name: String
  """
  @db.oneToMany: 'sender'
  """
  messages: [Message!]!
}

type Message {
  id: ID!
  text: String
}
```

This will create the following database structure:

```
- user:
   id
   name

- message:
   id
   body
   senderId
```

To create a message with a sender we need to let the `Message` type know about this with a bidirectional relationship. Like this:

```gql
type User {
  id: ID
  name: String
  """
  @db.oneToMany: 'sender'
  """
  messages: [Message!]!
} 

type Message {
  id: ID!
  body: String
  sender: User
}
```

### Queries and Mutations

```gql
mutation createUser {
  createUser(input:{name:"Enda"}) {
    id
    name
  }
}
```

```gql
mutation createMessage {
  createMessage(input:{body:"hello world", senderId: 1}) {
    id
    body
  }
}
```

Query all messages with sender (N:1):

```gql
query allMessages {
  findAllMessages {
    id
    body
    sender {
      id
      name
    }
  }
}
```

This gives us:

```json
{
  "data": {
    "findAllMessages": [
      {
        "id": "1",
        "body": "hello world",
        "sender": {
          "id": "1",
          "name": "Enda"
        }
      }
    ]
  }
}
```

Query all users with messages (1:N):

```gql
query allUsers {
  findAllUsers {
    id
    name
    messages {
      id
      body
    }
  }
}
```

Gives us:

```json
{
  "data": {
    "findAllUsers": [
      {
        "id": "1",
        "name": "Enda",
        "messages": [
          {
            "id": "1",
            "body": "hello world"
          },
          {
            "id": "2",
            "body": "hello world 2"
          }
        ]
      }
    ]
  }
}
```

## Tasks

- [x] Create relationship from 1:M side
- [x] Create data-loader per relationship field to allow querying multiple relationships to the same table at once.
- [x] Documentation
- [ ] Integration tests (blocked by https://github.com/aerogear/graphback/pull/606)


examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -1,22 +1,13 @@
 type Note {
```
craicoverflow: This file will be reverted.

wtrocki:
@craicoverflow Can we reuse some of the code from #523

packages/graphback-codegen-resolvers/src/api/targetResolverContext.ts
```diff
@@ -180,10 +181,11 @@ function createRelations(inputContext: InputModelTypeContext[]) {
           });
         }
         else if (f.annotations.OneToMany || f.isArray) {
-          let columnName = `${f.name}Id`;
+          let columnName = `${lowerCaseFirstChar(t.name)}Id`;
```
wtrocki: I guess this is temporary - when using annotations we need to actually check if there is metadata that affects that file

wtrocki: `getRelationFieldName` method does that

packages/graphback-codegen-schema/src/targetSchemaContext.ts
```diff
@@ -146,6 +146,8 @@ const nullField = (f: InputModelFieldContext) => {
  * subscriptions - generated subscription according to config - new, updated, deleted
  * @param input input visted object from model
  */
+// TODO: function size
```
wtrocki: Function too large in tslint

wtrocki:
There are multiple issues related with this approach. 
```
create table "public"."comment" ("noteId" integer null default null, "description" text null default null, "noteId" integer not null default null, "id" serial primary key) - column "noteId" specified more than once
``` 

Closing and going to start fresh

wtrocki:
I have doubts that this approach will work. Leaving this PR for later discussion.

craicoverflow:
> PR requires more work as this is not working with some use cases - db migration creates schema with duplicates and failes to validate etc. I could not get that to run and it seems that this approach may not work in generic scenarios. More work/discussion needed.

Could you provide some use cases where this was failing? I've made some changes since then so maybe they are working now.

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
```diff
@@ -417,6 +437,48 @@ class AbstractDatabaseBuilder {
...
+      name: annotations.oneToMany || lowerCaseFirstChar(oneToManyRelationship.relation.name),
+      type: oneToManyRelationship.relation,
+      description: oneToManyRelationship.description,
+      args: [],
+      extensions: []
+    }
+
+    const table = this.getRelationTableFromOneToMany(oneToManyRelationship)
+
+    if (!this.hasColumn(table, field)) {
```
craicoverflow: Column already added from the M:1 side so don't add it again.

wtrocki:
As agreed it will be cool to have:

- Integration tests that generate code. 
- Documentation clarification (Do we go into unidirectional and bidirectional relationships. 
Some getting started/tutorial will be cool)_

wtrocki:
Let me know when you would like me to take look into this again

craicoverflow:
> Let me know when you would like me to take look into this again

Hi, you can look now. I am in the process of adding integration tests so they should not affect the main code. Will need to add docs too.

craicoverflow:
> As agreed it will be cool to have:
> 
>     * Integration tests that generate code and do query with relationship.
> 
>     * Documentation clarification (Do we go into unidirectional and bidirectional relationships.
>       Some getting started/tutorial will be cool)_

Currently blocked on completing integration tests by https://github.com/aerogear/graphback/pull/606

This will need to be released and then have another graphql-testx release with the patch before it can be unblocked.

craicoverflow:
> As agreed it will be cool to have:
> 
>     * Documentation clarification (Do we go into unidirectional and bidirectional relationships.

With this change relationships can be created from other side of a `1:m`. Only one side needs to be aware of the relationship to be able to create the foreign key. For this reason it is unidirectional.

That is in database migration. On the codegen side, relationships must be bidirectional purely for the purpose of query/mutation/schema generation. I think we can unify this approach by making it unidirectional also. We have discussed this in the past.

wtrocki:
I guess my request for integration tests were based on limited knowlede. 

I did not cater that there is also issue with client-side codegen that will need to be resolved. 
I do not want to overload this PR. From my point of view I would extend little bit of getting started guide to add more info about relationship model etc. and then merge it. There is a lots of good work here that should land in the master. From what I see there is no Breaking changes in this (but I might be wrong)

craicoverflow:
> I guess my request for integration tests were based on limited knowlede.
> 
> I did not cater that there is also issue with client-side codegen that will need to be resolved.
> I do not want to overload this PR. From my point of view I would extend little bit of getting started guide to add more info about relationship model etc. and then merge it. There is a lots of good work here that should land in the master. From what I see there is no Breaking changes in this (but I might be wrong)

Add it to [docs/gettingstarted.md](https://github.com/aerogear/graphback/blob/master/docs/gettingstarted.md) as a new section?

docs/gettingstarted.md
```diff
@@ -79,3 +79,43 @@ If you use SQLite, you don't need to use Docker.
 Graphback offers runtime layer that will create in memory implementation of the graphql services without code generation.
 Developers can still customize the way that data is queried and cached thanks to runtime architecture.
 For more information about runtime layer please follow [`runtime`](/docs/runtime) documentation
+
+## Creating/updating your database from your data model
+
+Graphback will create or update the database configured in `graphback.json` using the types defined in your data model.
```
wtrocki:
```suggestion
Graphback will create or update the database configured in `graphback.json` file, using the types defined in your data model.
```

docs/gettingstarted.md
```diff
@@ -79,3 +79,43 @@ If you use SQLite, you don't need to use Docker.
 Graphback offers runtime layer that will create in memory implementation of the graphql services without code generation.
 Developers can still customize the way that data is queried and cached thanks to runtime architecture.
 For more information about runtime layer please follow [`runtime`](/docs/runtime) documentation
+
+## Creating/updating your database from your data model
```
wtrocki:
```suggestion
## Creating and updating your database from your data model
```

PR #573, opened 2019-12-18
docs: add versioning to support previous versions of graphback

wtrocki:
No text changes introduced. Just added 0.9.x version to support the previous generation. 

I know we have added some of the breaking changes already into this but they are not that visible and prominent so we can keep it the same anyway.


website/versions.json
```diff
@@ -0,0 +1,3 @@
+[
+  "0.9.x"
```
wtrocki: @darahayes FYI using 0.9.x here as we do not want to maintain patch releases docs.

website/versioned_docs/version-0.9.x/clientsidequeries.md
```diff
@@ -0,0 +1,91 @@
+---
+id: version-0.9.x-clientqueries
+title: Client-side Queries
+original_id: clientqueries
+---
+
+Graphback allows users to generate client-side queries based on the data model. All the queries/mutations/subscriptions
+which are generated through `graphback generate` for the user can also be generated as client-side queries. Users may follow up with using `create-react-app` or `apollo-client` to create client.
```
craicoverflow:
```suggestion
which are generated through `graphback generate` for the user can also be generated as client-side queries. Users may follow up with `create-react-app` or `apollo-client` to create client.
```

website/versions.json
darahayes: Good idea!

PR #596, opened 2019-12-31
fix(docs): update reference

jamesgeorge007:
Made the license badge lead to the respective file.


wtrocki:
Wow. Nice catch! Thank you so much for contribution

PR #603, opened 2020-01-06
Remove directives

wtrocki:
The target of this PR is to clean up docs and code from 2 different mechanisms of mapping data.
Directives were causing some issues in the past mainly as they are required to have valid schema. 

Annotations are extra content that is very flexible and allows us to apply different metadata.

In this PR I'm removing directives and replacing them with annotations.
Please follow the documentation and sample application for testing purposes.

## Verification

I have already made changes inside the model so generation should not include:

- deletion operation
- disable all operations for comment


docs/config.md
```diff
@@ -33,29 +33,61 @@ Changing these flags and performing `generate`, regenerates your `schema` and `r
...
-- `@subCreate`
-- `@subUpdate`
-- `@subDelete`
-- `@disableGen`.
-
-User can use these directives to have more control over individual elements. For example,
+## Changing configuration for the type
+
+Generator config applies the config to all the types in your schema. 
+Graphback allows you to change these for any single type using graphql `annotations`.
```
craicoverflow:
```suggestion
Graphback allows you to change these for any single type using [`graphql-annotations`](https://github.com/Akryum/graphql-annotations).
```

docs/config.md
```diff
@@ -33,29 +33,61 @@ Changing these flags and performing `generate`, regenerates your `schema` and `r
...
+
+https://github.com/Akryum/graphql-annotations
```
craicoverflow:
```suggestion
```

docs/config.md
```diff
@@ -33,29 +33,61 @@ Changing these flags and performing `generate`, regenerates your `schema` and `r
...
+"""
+@crud.create
+"""
+type User {
+  ....
+}
 ```
+ 
+#### Available Options
+All config options can be replicated by specifying `@crud.` prefix
```
craicoverflow:
```suggestion
All config options can be replicated by specifying the `@crud.` prefix
```

docs/config.md
```diff
@@ -33,29 +33,61 @@ Changing these flags and performing `generate`, regenerates your `schema` and `r
...
+@crud.delete: true
+@crud.find: true
+@crud.findAll: true
+@crud.subCreate: true
+@crud.subUpdate: true
+@crud.subDelete: true
+@crud.disableGen: true
+```
+
+User can use these annotations to have more control over individual elements. For example,
```
craicoverflow:
```suggestion
You can use these annotations to have more control over individual elements. For example:
```

docs/config.md
```diff
@@ -33,29 +33,61 @@ Changing these flags and performing `generate`, regenerates your `schema` and `r
...
+```
+"""
+@crud.delete
+"""
 type Note @delete {
```
craicoverflow:
```suggestion
type Note {
```

directive should be removed

docs/releases.md
```diff
@@ -8,6 +8,33 @@ title: Releases
...
+type Note @delete {
+  ...
+}
+```
+
+becomes now:
+
+```
+"""
+  @crud.delete
```
craicoverflow: Any need for the indent here?

examples/generator-fullstack/model/datamodel.graphql
```diff
@@ -12,6 +13,7 @@ type Note {
   comments: [Comment]!
 }
 
+"""@crud.disableGen"""
```
craicoverflow: We should continue to generate resolvers for this type IMO.

packages/graphback-core/README.md
```diff
@@ -33,10 +33,13 @@ Developers can extend functionality by using `InputProcessor` abstraction.
 
 ## Generator config
 
-Each individual type will contain individual config. Developers can pass global configuration that will be applied to elements that do not have any directives.
-Modifier directives directly on types will change default configuration.
+Each individual type will contain individual config. Developers can pass global configuration that will be applied to every type
+
 See `GraphbackCRUDGeneratorConfig` for more information.
 
+Config can be modified directly in the schema by utilizing GraphQL-Annotations:
```
craicoverflow:
```suggestion
Config can be modified directly in the schema by utilizing [`graphql-annotations`](graphql-annotations):
```

packages/graphback-core/README.md
```diff
@@ -33,10 +33,13 @@ Developers can extend functionality by using `InputProcessor` abstraction.
...
+https://github.com/Akryum/graphql-annotations
```
craicoverflow:
```suggestion
```

packages/graphback-core/src/graphql/annotations.ts
```diff
@@ -33,4 +33,10 @@ export const parseFieldAnnotations = (node: FieldDefinitionNode): InputModelFiel
     }
 
     return fieldAnnotations;
+}
+
+
+export const parseTypeAnnotations = (node: TypeDefinitionNode): InputModelFieldAnnotations => {
```
craicoverflow: At the minute it looks like the parsed annotations would not correctly implement the`InputModelFieldAnnotations` interface. Should this return type be removed altogether?

```suggestion
export const parseTypeAnnotations = (node: TypeDefinitionNode) => {
```

packages/graphback-core/tests/mockAnnotations.graphql
```diff
@@ -1,15 +1,17 @@
-type Note @create {
+
+""" @crud.delete: false """
```
craicoverflow: I see the annotations work as one-liners..great! Should we stick with one format though, at least with our examples/tests and templates? We can document both ways to do it, but stick with one of ourself, for nothing other than consistency really.

wtrocki: Which one you preffer. 
I check spec and there is no official format. Which one you like more

wtrocki: I so see value in showcasing both as not everyone will be able to see if both are possible :D

wtrocki:
Follow up will be to refactor core implementation to include relationships

packages/graphback-core/tests/mockAnnotations.graphql
craicoverflow: I find the multiline annotation easier to read because the annotation part has its own newline. If you see examples 2 and 3 below, single-line annotations may become even harder to read when we arguments are introduced.

```gql
"""
@crud.foo: 'bar'
"""
propOne: String
```

```gql
"""@crud.foo: 'bar'"""
propTwo: String
```

```gql
""" @crud.foo: 'bar' """
propTwo: String
```

PR #606, opened 2020-01-08
Add variables for relations to client fragments, subscriptions,…

craicoverflow:
Blocks https://github.com/aerogear/graphback/pull/568

## What

`graphback-codegen-client` was not adding variables for querying or mutating relationship fields in any of: Mutations, queries, subscriptions, fragments.

## Example

```
type Foo {
  id: ID!
  bar: Bar
}
```

Previously this would generate the following client mutation code:

```gql
mutation createFoo($name: String) {
  createFoo(input: {name: $name}) {
    ...FooFields
  }
}
```

This meant that it was not possible to create a relationship between a `Foo` and a `Bar` from the generated client code.

Now the mutation looks like this:

```gql
mutation createFoo($name: String, $barId: Int) {
  createFoo(input: {name: $name, barId: $barId}) {
    ...FooFields
  }
}
```


craicoverflow:
@wtrocki I tried to run `yarn graphql codegen` from our `generator-fullstack` example and got this:

```
❯ yarn graphql codegen
yarn run v1.21.1
$ /home/endaphelan/code/src/github.com/aerogear/graphback/examples/generator-fullstack/node_modules/.bin/graphql codegen
  ✖ Parse configuration
    → Cannot convert undefined or null to object
    Generate outputs
  Cannot convert undefined or null to object

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at normalize (/home/endaphelan/.nvm/versions/node/v10.17.0/lib/node_modules/@graphql-codegen/cli/index.cj
s.js:569:37)
    at Task.task (/home/endaphelan/.nvm/versions/node/v10.17.0/lib/node_modules/@graphql-codegen/cli/index.cj
s.js:621:21)
    at Promise.resolve.then.then.skipped (/home/endaphelan/.nvm/versions/node/v10.17.0/lib/node_modules/@grap
hql-codegen/cli/node_modules/listr/lib/task.js:167:30)
✖ Cannot convert undefined or null to object
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

So I was unable to regenerate `generated-types.tsx` in order to actually verify this PR functionally myself.

It appears our `.graphqlrc.yml` codegen extension is formatted differently to that in the graphql-cli template.

See https://github.com/Urigo/graphql-cli/blob/master/templates/fullstack/.graphqlrc.yml#L5

craicoverflow:
This is a blocker for https://github.com/aerogear/graphback/pull/568.

Integration tests that test mutating and querying relationships cannot be completed without this. Once merged we need to create a Graphback 0.10.1 release and then update graphql-testx with it.

cc @wtrocki @b1zzu

b1zzu:
@craicoverflow releasing graphback to release graphql-testx in order to fix the graphback integration tests is the worst pattern ever :rofl: 

I know this is probably the only approach but we should definitely find a better solution

wtrocki:
@b1zzu Integration tests using pinned graphback. As long as API is not broken we can swap version of graphback without releasing. In fact this is what  I was doing before, but sometimes I needed to release it only because API was changing.

wtrocki:
> So I was unable to regenerate generated-types.tsx in order to actually verify this PR functionally myself.

I think we need to update our version to cli version. I would also look into adding our generator integration  test to their repo

b1zzu:
@wtrocki the graphback integration tests should by default use the local graphback so that we can run the integration tests for each commit

PR #613, opened 2020-01-09
chore: bump knex to 0.20.7

craicoverflow:


wtrocki:
Yes.. This is renovate issue. It doesn't go into those nested folders for some reason, while our lerna does it

PR #622, opened 2020-01-13
Add Pagination support

binayakbehera999:
Added a pagination function to fetch paginated data using the paginate interface which contains the offset and limit values

### DIFF
--- a/examples/runtime-example/package.json
+++ b/examples/runtime-example/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "main": "dist/src/index.js",
   "scripts": {
-    "build": "echo disabled",
+    "build": "tsc",
     "develop": "ts-node src/index.ts",
     "start": "node dist/src/index.js",
     "watch": "tsc -w"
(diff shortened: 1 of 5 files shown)


wtrocki:
Thank you so much for contribution.
Would i be possible to clean changes that are not relevant to the pagination? Change looks great but we will need to cover it by unit test

PS:;We use conventional commits for commiting changes

binayakbehera999:
Thanks for the feedback. Yeah you can clean the changes that aren't relevant. Will definitely use the conventional commits from next time.

wtrocki:
Generally, this change opens many possibilities in the entire ecosystem and we can use it for discussion on how things can be done. I'm worried is that with such an amount of irrelevant changes it will be hard to get things and help you with the implementation of the pagination. Would it be possible to apply my suggestions on this PR or even create new one?

On the technical side. I would recomend focus on better names for the methods and interfaces. Paginate is very generic etc. We might also need pagination with filtering (coresponding to `findBy` method)

wtrocki:
@binayakbehera999 closing as outdated. Feel free to create new PR with the fixes.

PR #644, opened 2020-01-17
New schema generator

wtrocki:
# Motivation

Our schema generator basing on strings operations directly on DSL. 
While this approach has multiple advantages - there are challenges related to supporting every possible graphql type in our own parser rather than utilizing GraphQL-js capabilities. 

In the past, we had so many issues with supports for interfaces. Every time we add new types this contributes back to the higher complexity of the parser and our generator model.
Additionally, it is really hard to test our generators as results are always string-based. 
Generating complex snapshots are very error-prone and it is easy to miss some bugs

## Approach taken 

Generate schema using objects
Return schema (do not operate on stings) so the engine will be able to save it after or reuse it at runtime. 

**THIS PR IS VERY EARLY WIP to show concepts.**

## Testing 

Go to graphback package and run `temp.ts` file.

## Progress

- [ ] Create Transformer architecture
- [x] Basic crud support
- [ ] Subscriptions
- [ ] Start with schema as first element
- [ ] Move helpers to core
- [ ] Introduce unit tests and integration tests


packages/graphback-codegen-schema/src/transformer/SchemaTransformer.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphQLSchema } from "graphql";
+
+interface SchemaTransformer<TransformerConfig = any> {
+
+    /**
+     * Performs transformation on the schema and returns target schemaƒ
+     * @param schema 
+     */
+    transformSchema?(schema: GraphQLSchema, config: TransformerConfig): GraphQLSchema;
```
craicoverflow: Maybe some sort of high level diagram to new generator that shows where this interface fits in?

wtrocki: I will need to commit other files in core. Target of this PR is to focus on schema only, but I'm going to create separate PR for the engine as well.

wtrocki: Working on diagra now and attaching it to the issue

wtrocki: Done. Applied to the issues instead of PR

wtrocki:
Reminder: Let's chat about approach we are taking for marking id field (What is the id field) 
Are we doing some marker for it?

craicoverflow:
> Reminder: Let's chat about approach we are taking for marking id field (What is the id field)
> Are we doing some marker for it?

Right now we don't have a marker as the ID field must always be `id: ID!`. This was a limitation we placed because there did not seem to be a way to allow the ID to be either auto-incrementing or a regular type without causing serious issues when the user would change the ID annotation and update their database.

wtrocki:
Could it be autoincrementing and `id: ID!` by default but allow to use anything as ID and then do not give dam about increments and anything else. For example email could be marked as unique and that will be enough to have it as ID

packages/graphback/src/GraphbackEngine.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.config = config;
+    // Legacy to be removed
+    this.inputContext = graphQLInputContext.createModelContext(schema as string, {});
+  }
+
+  /**
+   * Create backend with all related resources
+   */
+  // FIXME generator options should be moved to plugin config
+  public buildBackend(resolverOptions: ResolverGeneratorOptions): IGraphQLBackend {
```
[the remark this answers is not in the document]
craicoverflow: I see now this method name already existed so maybe it should be kept.

packages/graphback-core/src/plugin/GraphbackPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Helper for plugins to fetch all types that should be processed by Graphback plugins.
+     * To mark type as enabled for graphback generators we need to add `model` annotations over the type. 
+     * 
+     * Returns all user types that have @model in description
+     * @param schema 
+     */
+    protected getUserModels(schema: GraphQLSchema): GraphQLObjectType[] {
```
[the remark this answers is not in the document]
wtrocki: > Would you see a value in this function be moved into a separate file not in this class as a standalone helper function?

If we move to helper files then this refactoring is pointless and we do not have architecture done right. The whole purpose is to have things working with each other and core that drives some information forward.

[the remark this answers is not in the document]
craicoverflow: Having this function in the base `GraphbackPlugin` implies types marked with `@model` will be needed in most cases. But the idea of the plugin architecture is that it is generic enough to fit all use cases.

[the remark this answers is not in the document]
craicoverflow: > Helper for plugins to fetch all types that should be processed by Graphback plugins

What if I want a plugin that does some transformation of non-model type? This JSDoc strongly implies that the plugin engine does not cater for this.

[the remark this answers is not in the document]
wtrocki: Actually I would need to think about it. It was originally as a helper. 
We need some kind of core for the entire ecosystem. 

So we can have 2 layers in the core. Generator plugin and general ModelFacade that gives information for the schema. 

So I would see it that plugin by default will have this facade created and whatever else will need it will simply need to create new opbject for a schema

packages/graphback-codegen-schema/src/plugin/SchemaCRUDPlugin.ts
```diff
@@ -0,0 +1,234 @@
...
+import * as pluralize from "pluralize";
+import { gqlSchemaFormatter, jsSchemaFormatter, tsSchemaFormatter } from '../writer/schemaFormatters';
+import { printSortedSchema } from '../writer/schemaPrinter';
+
+/**
+ * Configuration for Schema generator CRUD plugin
+ */
+export interface SchemaCRUDPluginConfig {
+    // output format for schema string
+    format: 'ts' | 'js' | 'gql'
```
craicoverflow: This config does not seem appropriate for the `SchemaCRUDPluginConfig`. To me this plugin should logically contain the following:

```
// Generate create method
    create?: boolean;
    // Generate update method
    update?: boolean;
    //tslint:disable-next-line
    delete?: boolean;
    // Generate find method
    find?: boolean;
    // Generate findAll method
    findAll?: boolean;
    // Generate subscription for create operation
    subCreate?: boolean;
    // Generate subscription for update operation
    subUpdate?: boolean;
    // Generate subscription for delete operation
    subDelete?: boolean;
    // Disable codegen
    disableGen?: boolean;
```

And we could have another plugin "SchemaFormatterPlugin" which is specifically for formatting the schema as ts, js, gql etc.

I won't change anything before I hear your thoughts on this.

packages/graphback-codegen-schema/src/plugin/SchemaCRUDPlugin.ts
```diff
@@ -0,0 +1,234 @@
...
+        return SCHEMA_CRUD_PLUGIN_NAME;
+    }
+
+    private buildSchemaForModels(models: ModelDefinition[]) {
+        let queryTypes = {};
+        let mutationTypes = {};
+        let subscriptionTypes = {};
+
+        for (const model of Object.values(models)) {
+            if (model.crudOptions.disableGen) {
```
craicoverflow: Missing the check for global CRUD config. is this something I should add or did you plan to add it in this plugin?

PR #651, opened 2020-01-24
fix: fix type error in runtime example & improve runtime-example docs

kingsleyzissou:
- fix the type error in the `runtime.ts` file in the runtime-example app, as discussed in the standup this morning.
- improve the docs to include the correct steps for starting up the server and add instructions for using sqllite 3 in the runtime-example app

### DIFF
--- a/examples/runtime-example/README.md
+++ b/examples/runtime-example/README.md
@@ -66,8 +66,17 @@ The project has been created using `graphback`. Run the project using the follow
   },
   "database": "sqlite3"
 },
+```
+
+- Define your schema in the `model/runtime.graphql` file. Or use the default:
 
 ```
+type User {
+  id: ID!
+  name: String
+}
+```
+
 - Start the server (for the first time)
 ```
 npm run develop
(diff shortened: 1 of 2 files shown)


examples/runtime-example/README.md
```diff
@@ -27,21 +27,66 @@ Then developers can create runtime instance:
 
 See [`./runtime.ts`](https://github.com/aerogear/graphback/blob/master/examples/runtime-example/src/runtime.ts#L32) for a fully functional example.
 
-### Running example
+### Running example (postgres)
```
wtrocki:
```suggestion
### Running example using Postgres database
```

examples/runtime-example/README.md
```diff
@@ -27,21 +27,66 @@ Then developers can create runtime instance:
...
+- Start the server (for the first time)
+```
+npm run develop
+```
+Or, if using yarn
+```
+yarn develop
+```
+
+If the server is being re-run, modify the `src/runtime.ts` and comment out the `migrateDb` function, since it will not be possible to re-migrate the database with sqllite3.
```
wtrocki: Also we need to change the interface from `PGKnex` to `Knex`

kingsleyzissou: Updated this too

PR #653, opened 2020-01-24
Resolver generator plugin

craicoverflow:
## What

Plugin to generate resolver code from the schema and write content to file-system.

## Verification

1. Run `packages/graphback/src/temp.ts` or run the VSCode `Temp` debugger option to execute the plugin engine outside of the Graphback regular workflow.
2. Custom and default resolvers should be generated in `generator-fullstack` example.

## TODO

- Refactor method to write custom resolvers to file.
- Relationships
- JavaScript file support
- GraphQL-JS formatting.

### DIFF
--- a/packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
+++ b/packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
@@ -1,9 +1,6 @@
 import { getFieldName, GraphbackCoreMetadata, GraphbackCRUDGeneratorConfig, GraphbackOperationType, GraphbackPlugin, ModelDefinition } from '@graphback/core';
-import { writeFileSync } from 'fs';
-import { GraphQLField, GraphQLList, GraphQLNamedType, GraphQLNonNull, GraphQLObjectType, GraphQLOutputType, GraphQLSchema, isListType, isNonNullType, isWrappingType } from 'graphql';
-import { join, resolve } from 'path';
-import * as prettier from 'prettier';
-import { generateResolverTemplate, resolversIndexTemplate, resolversRootIndexTemplate } from './ApolloTypeScriptResolverFormatter';
+import { GraphQLObjectType, GraphQLSchema } from 'graphql';
+import { getCustomTypeResolverKeys } from '../util/getCustomResolverFieldNames';
 import { blankResolver, blankSubscription, createTemplate, deletedSubscriptionTemplate, findAllTemplate, findTemplate, newSubscriptionTemplate, updatedSubscriptionTemplate, updateTemplate } from './resolverTemplates';
 import { writeTypeScriptResolvers } from './writeResolvers';
 
@@ -26,6 +23,8 @@ export interface ResolverGeneratorPluginOptions {
     }
 }
 
+const PLUGIN_NAME = 'RESOLVER_GENERATOR';
+
 export class ResolverGeneratorPlugin extends GraphbackPlugin {
     private options: ResolverGeneratorPluginOptions;
     constructor(options: ResolverGeneratorPluginOptions) {
@@ -35,7 +34,7 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
     }
 
     public getPluginName() {
-        return 'RESOLVER_GENERATOR';
+        return PLUGIN_NAME;
     }
 
     public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
@@ -62,8 +61,6 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
             };
 
             outputResolvers[graphqlType.name] = typeResolvers;
-
-            // TODO relationships
         }
 
         return outputResolvers;
@@ -78,9 +75,9 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
         for (const { graphqlType } of models) {
             const modelResolvers = generatedResolvers[graphqlType.name];
 
-            const queries = this.createCustomQueries(graphqlType, queryType, modelResolvers.Query);
-            const mutations = this.createCustomQueries(graphqlType, mutationType, modelResolvers.Mutation);
-            const subscriptions = this.createCustomSubscriptionResolvers(graphqlType, subscriptionType, modelResolvers.Subscription);
+            const queries = this.createCustomResolvers(graphqlType, queryType, Object.keys(modelResolvers.Query));
+            const mutations = this.createCustomResolvers(graphqlType, mutationType, Object.keys(modelResolvers.Mutation));
+            const subscriptions = this.createCustomSubscriptionResolvers(graphqlType, subscriptionType, Object.keys(modelResolvers.Subscription));
 
             const typeResolvers = {
                 Query: queries,
@@ -155,8 +152,8 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
         return subscriptions;
     }
 
-    private createCustomQueries(graphqlType: GraphQLObjectType, resolverType: GraphQLObjectType, generatedResolvers: any) {
-        const customKeys = getCustomResolverKeys(graphqlType, resolverType, generatedResolvers);
+    private createCustomResolvers(graphqlType: GraphQLObjectType, resolverType: GraphQLObjectType, generatedResolvers: string[]) {
+        const customKeys = getCustomTypeResolverKeys(graphqlType, resolverType, generatedResolvers);
 
         const resolvers = {};
         for (const key of customKeys) {
@@ -166,8 +163,8 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
         return resolvers;
     }
 
-    private createCustomSubscriptionResolvers(graphqlType: GraphQLObjectType, subscriptionType: GraphQLObjectType, generatedResolvers: any) {
-        const customKeys = getCustomResolverKeys(graphqlType, subscriptionType, generatedResolvers);
+    private createCustomSubscriptionResolvers(graphqlType: GraphQLObjectType, subscriptionType: GraphQLObjectType, generatedResolvers: string[]) {
+        const customKeys = getCustomTypeResolverKeys(graphqlType, subscriptionType, generatedResolvers);
 
         const resolvers = {};
         for (const key of customKeys) {
@@ -176,31 +173,4 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
 
         return resolvers;
     }
-}
-
-function getCustomResolverKeys(graphqlType: GraphQLObjectType, resolverObject: GraphQLObjectType<any, any, { [key: string]: any; }>, generatedResolvers: any) {
-    const typeResolverFields = getTypeResolvers(graphqlType, resolverObject);
-    const modelKeys = typeResolverFields.map((field: GraphQLField<any, any>) => field.name);
-    const generatedKeys = Object.keys(generatedResolvers);
-
-    return arrayDiff(generatedKeys, modelKeys);
-}
-
-// tslint:disable-next-line: no-reserved-keywords
-export function getBaseType(type: GraphQLOutputType): GraphQLNamedType {
-    if (isWrappingType(type)) {
-        return getBaseType(type.ofType);
-    } else {
-        return type;
-    }
-}
-
-function getTypeResolvers(graphqlType: GraphQLObjectType, resolverType: GraphQLObjectType) {
-    const resolverFields = Object.values(resolverType.getFields());
-
-    return resolverFields.filter((field: GraphQLField<any, any>) => getBaseType(field.type).name === graphqlType.name);
-}
-
-function arrayDiff(defaultKeys: string[], allKeys: string[]): string[] {
-    return allKeys.filter((key: string) => defaultKeys.indexOf(key) === -1);
 }
--- a/packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
+++ b/packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
@@ -2,7 +2,7 @@ import { getFieldName, GraphbackCoreMetadata, GraphbackCRUDGeneratorConfig, Grap
 import { GraphQLObjectType, GraphQLSchema } from 'graphql';
 import { getCustomTypeResolverKeys } from '../util/getCustomResolverFieldNames';
 import { blankResolver, blankSubscription, createTemplate, deletedSubscriptionTemplate, findAllTemplate, findTemplate, newSubscriptionTemplate, updatedSubscriptionTemplate, updateTemplate } from './resolverTemplates';
-import { writeTypeScriptResolvers } from './writeResolvers';
+import { writeTypeScriptResolvers } from './writeTypeScriptResolvers';
 
 export interface ResolverGeneratorPluginOptions {
     format: 'ts' | 'js'
@@ -39,22 +39,30 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
     }
 
     public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
-        const generatedResolvers = this.generateResolvers(metadata.getModelDefinitions());
-        const customResolvers = this.generateCustomResolvers(metadata.getSchema(), metadata.getModelDefinitions(), generatedResolvers);
+        const schema = metadata.getSchema()
+        const models = metadata.getModelDefinitions();
+        if (models.length === 0) {
+            this.logWarning("Provided schema has no models. Returning original schema without generating resolvers.")
+
+            return schema;
+        }
+
+        const generatedResolvers = this.generateResolvers(models);
+        const customResolvers = this.generateCustomResolvers(schema, models, generatedResolvers);
 
         if (this.options.format === 'ts') {
             writeTypeScriptResolvers({ generated: generatedResolvers, custom: customResolvers }, this.options);
         } else {
-            console.log('Not implemented');
+            throw new Error("Not implemented");
         }
 
         return metadata.getSchema();
     }
 
-    private generateResolvers(modelDefinitions: ModelDefinition[]) {
+    private generateResolvers(models: ModelDefinition[]) {
         const outputResolvers = {};
 
-        for (const { graphqlType, crudOptions } of modelDefinitions) {
+        for (const { graphqlType, crudOptions } of models) {
             if (crudOptions.disableGen) {
                 continue;
             }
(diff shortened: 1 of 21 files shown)


packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
+import { getModelTypesFromSchema, GraphbackCoreMetadata, GraphbackPlugin, TypeScriptResolversVisitor } from '@graphback/core';
+import { buildASTSchema, GraphQLField, GraphQLFieldMap, GraphQLObjectType, GraphQLOutputType, GraphQLSchema, isListType, isNonNullType, parse, printSchema, visit } from 'graphql';
+import { ResolverGeneratorOptions } from '../api/ResolverGeneratorOptions';
+import * as pluralize from "pluralize";
+
+export interface ResolverDefinition {
```
wtrocki: We do not need those types as generally resolvers are not type safe. 
TypeSafety will be coming externally from codegen

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export interface ResolverFunctionMap {
+    [name: string]: string;
+}
+
+export class ResolverGeneratorPlugin extends GraphbackPlugin {
+    private resolverDefinition: ResolverDefinition;
+    constructor(config?: ResolverGeneratorOptions) {
+        super();
```
wtrocki: In my mind plugins will need to deal with writing files themselves. 
For example, resolver plugin will accept some layout config that specifies how to spread files etc.

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+    public getPluginName() {
+        return 'RESOLVER_GENERATOR';
+    }
+
+    public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
+        return metadata.getSchema();
+    }
+
+    public createResolvers(schema: GraphQLSchema) {
+        const models = getModelTypesFromSchema(schema);
```
wtrocki: Not needed. That is in metadata already

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
craicoverflow: This is purely for building a collection of resolvers which can then be piped to a file at the very end.

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+        const queryType = schema.getQueryType();
+        const queryFields = queryType.getFields();
+
+        for (const model of models) {
+            this.resolverDefinition[model.name] = {
```
wtrocki: We do not base on query fields. This is not possible as we do not know what queryField is custom and what is not. We should base on CRUD config from metadata

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        console.log(queryNames, y, x === y)
+
+        if (queryNames[`findAll${pluralizedModelName}`]) {
+            console.log('hello');
+        }
+    }
+}
+
+export function getBaseType(graphqlType: GraphQLOutputType) {
```
wtrocki: If we apply suggestions those helpers might be not needed

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+            generated: []
+        }
+    }
+
+    public getPluginName() {
+        return 'RESOLVER_GENERATOR';
+    }
+
+    public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
+        return metadata.getSchema();
```
wtrocki: MAIN issue: Plugins are stateless by definition. They should hold config only.

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public createResolvers(schema: GraphQLSchema) {
```
wtrocki: There is no need to write this function. We have it written already here: 
https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverGen.ts

We can just transform it to string format

packages/graphback/src/temp.ts
```diff
@@ -41,6 +41,9 @@ interface aestInterface {
     comment: [Comment]!
   }
   
+  """
```
wtrocki: Any reason why this is annotated as model?

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface ResolverFunctionMap {
+    [name: string]: string;
+}
+
+export class ResolverGeneratorPlugin extends GraphbackPlugin {
+    private resolverDefinition: ResolverDefinition;
+    constructor(config?: ResolverGeneratorOptions) {
+        super();
+        this.resolverDefinition = {
```
wtrocki: We need to filter out custom/CRUD resolves

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
+        return metadata.getSchema();
+    }
+
+    public createResolvers(schema: GraphQLSchema) {
+        const models = getModelTypesFromSchema(schema);
+        const queryType = schema.getQueryType();
+        const queryFields = queryType.getFields();
```
wtrocki: There is value in this as we will need to compare it with crud config.

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,86 @@
+import { getModelTypesFromSchema, GraphbackCoreMetadata, GraphbackPlugin, TypeScriptResolversVisitor } from '@graphback/core';
+import { buildASTSchema, GraphQLField, GraphQLFieldMap, GraphQLObjectType, GraphQLOutputType, GraphQLSchema, isListType, isNonNullType, parse, printSchema, visit } from 'graphql';
+import { ResolverGeneratorOptions } from '../api/ResolverGeneratorOptions';
```
wtrocki: Let's clone it.

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,214 @@
...
+        const resolvers = this.generate(metadata.getModelDefinitions());
+
+        this.write(resolvers);
+
+        return metadata.getSchema();
+        // metadata.getGraphQLTypesWithModel
+    }
+
+    private write(resolvers: { generated: any; custom: any; }) {
+        const resolversDir = join(process.cwd(), this.options.resolverPath);
```
wtrocki: Process.cwd() might not be the project root. What will be your prefference:
- Pass absolute filepath
- Always pass reference against project root and enforce

`join(process.cwd(), this.options.resolverPath)` => `resolve(this.options.resolverPath)

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -38,57 +38,95 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
...
-        return resolvers;
+        const outputResolvers = {};
+        for (const { graphqlType } of models) {
+            const modelResolvers = generatedResolvers[graphqlType.name];
+
+            const queries = this.createCustomResolvers(graphqlType, queryType, modelResolvers.Query);
+            const mutations = this.createCustomResolvers(graphqlType, mutationType, modelResolvers.Mutation);
+            const subscriptions = this.createCustomSubscriptionResolvers(graphqlType, subscriptionType, modelResolvers.Subscription);
+
+            const typeResolvers = {
```
wtrocki: We can add formatter here for graphql-js

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -38,57 +38,95 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
...
+        }
+
+        for (const group of Object.keys(modules)) {
+            let exportName: string;
+            if (group === 'custom') {
+                exportName = 'customResolvers';
+            }
+            const indexOutput = resolversIndexFileTemplate(modules[group], exportName);
+
+            const formattedIndex = prettier.format(indexOutput, { semi: false, parser: "babel" });
```
wtrocki: Love it!

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+
+function getCustomResolverKeys(graphqlType: GraphQLObjectType, resolverObject: GraphQLObjectType<any, any, { [key: string]: any; }>, generatedResolvers: any) {
+    const modelQueries = getTypeResolvers(graphqlType, resolverObject);
+    const modelKeys = modelQueries.map((query: GraphQLField<any, any>) => query.name);
+    const generatedKeys = Object.keys(generatedResolvers);
+
+    return arrayDiff(generatedKeys, modelKeys);
+}
+
+export function isWrapperType(t: GraphQLOutputType): t is GraphQLNonNull<any> | GraphQLList<any> {
```
wtrocki: This function is now defined in grapql-js :P

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,245 @@
...
+        const resolvers = {};
+        for (const key of customKeys) {
+            resolvers[key] = blankResolver();
+        }
+
+        return resolvers;
+    }
+
+    private createCustomSubscriptionResolvers(graphqlType: GraphQLObjectType, queryType: GraphQLObjectType, generatedResolvers: any) {
+        const customKeys = getCustomResolverKeys(graphqlType, queryType, generatedResolvers);
```
wtrocki: We call this multiple times

packages/graphback-codegen-resolvers/src/plugins/writeTypeScriptResolvers.ts
```diff
@@ -0,0 +1,81 @@
...
+function writeCustomResolvers(resolvers: any, options: ResolverGeneratorPluginOptions) {
+    const customResolversPath = resolve(options.resolverPath, 'custom');
+
+    if (!existsSync(customResolversPath)) {
+        mkdirSync(customResolversPath, { recursive: true });
+    }
+
+    const importNames = [];
+    const resolverBase = Object.values(resolvers);
+    // TODO: Refactor to remove need for 3-level nested for loop
```
wtrocki: This should be addressed before merge

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,210 @@
...
+        if (crudOptions.create) {
+            const fieldName = getFieldName(graphqlType.name, GraphbackOperationType.CREATE);
+            // tslint:disable-next-line: no-any
+            mutations[fieldName] = createTemplate(objectName, crudOptions.subCreate)
+        }
+        if (crudOptions.update) {
+            const fieldName = getFieldName(graphqlType.name, GraphbackOperationType.UPDATE);
+            mutations[fieldName] = updateTemplate(objectName, crudOptions.subUpdate);
+        }
+        if (crudOptions.update) {
```
wtrocki: Update is added twice

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,210 @@
...
+            const fieldName = getFieldName(graphqlType.name, GraphbackOperationType.UPDATE);
+            mutations[fieldName] = updateTemplate(objectName, crudOptions.update);
+        }
+        if (crudOptions.delete) {
+            const fieldName = getFieldName(graphqlType.name, GraphbackOperationType.DELETE);
+            mutations[fieldName] = updateTemplate(objectName, crudOptions.delete);
```
wtrocki: Wrong method is used here

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,210 @@
...
+
+        return mutations;
+    }
+
+    private createQueries(graphqlType: GraphQLObjectType, crudOptions: GraphbackCRUDGeneratorConfig) {
+        const queries = {};
+        const objectName = graphqlType.name.toLowerCase();
+
+        if (crudOptions.find) {
+            const fieldName = getFieldName(objectName, GraphbackOperationType.FIND, 's');
```
wtrocki: Les add TODO for pluralize

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
```diff
@@ -0,0 +1,210 @@
...
+
+        return queries;
+    }
+
+    private createSubscriptions(graphqlType: GraphQLObjectType, crudOptions: GraphbackCRUDGeneratorConfig) {
+        const objectName = graphqlType.name.toLowerCase();
+
+        const subscriptions = {};
+        if (crudOptions.create && crudOptions.subCreate) {
+            const fieldName = `new${graphqlType.name}`;
```
wtrocki: We need CRUD name format comming from core. Needs todo to use core helper yo get method name

packages/graphback-codegen-resolvers/src/plugins/resolverTemplates.ts
```diff
@@ -0,0 +1,79 @@
+const defaultResolverArgs = `_, args, context`;
```
wtrocki: I will use parent instead of _. We will use it later

packages/graphback-codegen-resolvers/src/plugins/writeTypeScriptResolvers.ts
wtrocki: Wrong responsibility. This function should write strings but it is doing much more. We should do that beforehand

packages/graphback-codegen-resolvers/src/plugins/writeTypeScriptResolvers.ts
```diff
@@ -0,0 +1,81 @@
...
+                const resolverTemplate = generateBlankResolverTemplate(resolverType, resolverField, typeResolvers[resolverField]);
+
+                writeFileSync(join(customResolversPath, `${resolverField}.ts`), formatFile(resolverTemplate));
+
+                importNames.push(resolverField);
+            }
+        }
+    }
+
+    const indexOutput = resolversIndexTemplate(importNames.map((name: string) => ({ importAs: name, importFrom: name })), `customResolvers`);
```
wtrocki: This code can be problematic when placed here. We should just pass strings

packages/graphback-codegen-resolvers/src/util/getCustomResolverFieldNames.ts
```diff
@@ -0,0 +1,20 @@
+import { getBaseType } from '@graphback/core';
+import { GraphQLField, GraphQLObjectType } from "graphql";
+
+export function getCustomTypeResolverFieldNames(graphqlType: GraphQLObjectType, resolverObject: GraphQLObjectType, generatedResolverKeys: string[]) {
```
wtrocki: We need just graphqlType.name

packages/graphback-codegen-resolvers/src/util/getCustomResolverFieldNames.ts
```diff
@@ -0,0 +1,20 @@
...
+    const typeResolverFields = getTypeResolvers(graphqlType, resolverObject);
+    const modelKeys = typeResolverFields.map((field: GraphQLField<any, any>) => field.name);
+
+    return findUniqueKeys(generatedResolverKeys, modelKeys);
+}
+
+function getTypeResolvers(graphqlType: GraphQLObjectType, resolverType: GraphQLObjectType) {
+    const resolverFields = Object.values(resolverType.getFields());
+
+    return resolverFields.filter((field: GraphQLField<any, any>) => getBaseType(field.type).name === graphqlType.name);
```
wtrocki: There is helper in graphql-js to get base type

wtrocki: See client generator.

wtrocki: Filtering by type is not needed. If we generate resolver using crud config and someone provides different resolver it should be considered custom. We might validate it

packages/graphback-codegen-resolvers/src/plugins/ResolverGeneratorPlugin.ts
craicoverflow: Current subscription resolvers don't use helper to get method name:

https://github.com/aerogear/graphback/blob/e7d8c75679925489f165562facb62fa8f2f18260/packages/graphback-codegen-resolvers/src/templates/LayeredResolverTemplates.ts#L43

wtrocki: Yes. But then we have this contract between schema resolvers and client not written anywhere. I have added issue to build our own  GraphQLCRUD spec.

PR #656, opened 2020-01-25
Cli fix

wtrocki:
Create folder if not exist


wtrocki:
Released in graphback 0.10.4

PR #657, opened 2020-01-27
Plugin for client side query generation

wtrocki:
## Motivation

Migrate client side to use plugin architecture.
Simplify queries generation etc.

## Progress

- [x] Create plugin wireframe
- [x] Templates
- [x] File storage implementation
- [x] Unit tests
- [x] Formatting


packages/graphback-codegen-client/src/plugin/templates/gqlTemplates.ts
```diff
@@ -0,0 +1,202 @@
...
+        return field.name;
+      }
+    }
+    else {
+      return field.name;
+    }
+  });
+}
+
+export const fragment = (t: ModelDefinition) => {
```
craicoverflow: could we just pass the graphql type to these functions? I don't see `t` data being used in them.

packages/graphback-codegen-client/src/plugin/templates/gqlTemplates.ts
```diff
@@ -0,0 +1,202 @@
...
+      name: t.graphqlType.name,
+      implementation: fragment(t)
+    }
+  })
+}
+
+export const createQueries = (types: ModelDefinition[]) => {
+  const queries = []
+
+  types.forEach((t: ModelDefinition) => {
```
craicoverflow: A `for...of` loop would be faster.

wtrocki:
Storage is now supported
Screenshot 2020-01-28 at 13 18 11

wtrocki:
Found couple formatting issues in generated content apart from that all should be ok.

packages/graphback-codegen-client/src/plugin/ClientCRUDPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+        const schema = metadata.getSchema()
+        const models = metadata.getModelDefinitions();
+        if (models.length === 0) {
+            this.logWarning("Provided schema has no models. No Client side queries will be generated")
+
+            return schema;
+        }
+        let documents: ClientTemplates;
+        let outputFormat
+        if (this.pluginConfig.outputFormat === 'ts') {
```
craicoverflow: This `if..else` block is kind of hard to read, and really I don't need to read it. It could be abstracted into single helper function => `createClientDocuments(models, 'ts)`...

packages/graphback-codegen-client/src/plugin/templates/fragmentFields.ts
```diff
@@ -0,0 +1,54 @@
...
+    let resultString = '';
+  
+    for (const element of resultArray) {
+      if (element instanceof Object) {
+        const key = Object.keys(element)[0];
+        resultString += `   ${key} {
+${printReturnFields(element[key], '   ')}
+   }`
+      } else {
+        resultString += `${shift}   ${element}\n`;
```
craicoverflow: Causes a blank newline at end of fragment. We should leave this though and have a common formatter than can be used by plugins.

packages/graphback-codegen-client/src/plugin/templates/gqlTemplates.ts
```diff
@@ -0,0 +1,195 @@
...
+} `
+}
+
+export const createFragments = (types: ModelDefinition[]) => {
+  return types.reduce((data: ClientTemplate[], model: ModelDefinition) => {
+    data.push({
+      name: model.graphqlType.name,
+      implementation: fragment(model.graphqlType)
+    })
+    data.push({
```
craicoverflow: What is the purpose of multiple fragment types for a model? Is one for mutations and one (expanded) for queries?

wtrocki:
Forgot about disablegen. My bad

craicoverflow:
> Forgot about disablegen. My bad

So have I!

packages/graphback-codegen-client/src/plugin/templates/fragmentFields.ts
wtrocki: What do you mean by common formatter. What fix I can apply here?

wtrocki: Going to fix it outside \n stays here :D

PR #662, opened 2020-01-29
Migrated to ESLint.

cecobask:
Migrated the project from tslint to eslint.


packages/graphback-cli/src/helpers/transformOpenApiSpec.ts
```diff
@@ -79,3 +45,37 @@ async function processSingleDefinition(model: string, isYaml: boolean, openApiCo
...
     }
 }
+
+/**
+ * OpenAPI command line tool
+ * 1. Reads all the OpenAPI yaml and json files from the model folder
+ * 2. Generates GraphQL Schema
+ * 3. Renames original OpenAPI definitions so they will not be processed anymore
+ */
+export const transformOpenApiSpec = async () => {
```
wtrocki: Was this some kind of rule that was failing? Order of functions

cecobask: Eslint complained that the function shouldn't be used before it is defined

wtrocki:
Thank you so much for contribution. I needed to change base as we have large portion of changes of separate branch

wtrocki:
Closing for #664

PR #665, opened 2020-01-30
fix: Cli command using new framework

wtrocki:


packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -1,182 +1,189 @@
...
 }
 
-export const deleteMutation = (t: InputModelTypeContext, ) => {
+export const deleteMutation = (t: GraphQLObjectType) => {
   const fieldName = getFieldName(t.name, GraphbackOperationType.DELETE)
 
-  return `mutation ${fieldName}($id: ID!) {
-    ${fieldName}(id: $id)
+  return `mutation ${fieldName} ($input: ${t.name}Input!) {
+  ${fieldName} (input: ${t.name}Input!) {
```
craicoverflow: Generated schema input types do not contain an ID field. This would produce the following:

```
NoteInput
  title
  description
```

This will try to delete a non-unique item (multiple notes can have the same description and title). Same goes for createMutations.

wtrocki: Yes.. This is TODO we need to add to the schema, but it depends on the ID mapping logic from the core so I left that without TODO now

wtrocki: Target here is do not do half backed generation engine. The only thing outstanding is just field, relation and id mapping that will allow us to fix all other generators

PR #671, opened 2020-01-30
fix: generate subscriptions when both config values are true

craicoverflow:
## What

Subscriptions should not be generated if the CRUD config value for that operation type is `false`.

Example: `@crud.delete: false` and `@crud.subDelete: true` should not create the delete subscription.

From documentation: 

> **Note**: For subscriptions, the user needs to change the value of the respective operations to true. For example, changing subDelete to true won't work unless, delete is true.

## Verification

```gql
""" 
@crud.delete: false 
"""
type Note {
  id: ID!
  title: String
}
```

Run `generate` with the model above. Neither a `deletedNote` subscription or a `deleteNote` mutation will be generated.


wtrocki:
I was strange that thie ended up in custom methods. But then custom method calculation is based on crud so..

PR #672, opened 2020-01-30
Resolver generator plugin improvements

craicoverflow:
## What

This change brings in JavaScript resolver generation in Apollo format and did a tidy up of the package folder.

- [x] Clean up folder structure
- [x] JavaScript formatter
- [ ] Support for custom resolvers using default CRUD names.


craicoverflow:
@wtrocki just adding JavaScript resolver generation in this PR.

packages/graphback-codegen-resolvers/src/formatters/apollo/jsResolverFormatter.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const importNames = resolverNames.map((name: string) => `${name}Resolvers`);
+
+    return `${imports}
+
+    exports.${exportName} = [${importNames.join(', ')}];`;
+}
+
+export const rootResolversIndexJS = (groups: string[] = ['generated', 'custom']) => {
+    return `${groups.map((name: string) => `const { ${name}Resolvers } = require('./${name}')`).join('\n')}
```
wtrocki: This is what was killing me with the previous implementation. 
Too much logic in template. itself that is hard to maintain. If you take look on the other templates I have been doing for loops outside and using concat. NOt that this is bad, but we target simplicity.

PR #673, opened 2020-01-30
Remove invalid code from core and schema`

wtrocki:
Also unified mapping for CRUD names. It also fixed entire runtime to work with the new model. Not sure what to do with production-based migrations


packages/graphback-core/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "include": [
-    "./src/**/*.ts"
```
wtrocki: Commited by mistake. To be removed

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -0,0 +1,44 @@
+import { GraphbackOperationType} from '../api';
```
wtrocki: @craicoverflow This is the only change in this PR. I have added one place for mapping every CRUD element. This needs to be extended and used in plugins

wtrocki:
What is outstanding for this is to fix database migration.

wtrocki:
@craicoverflow Looking for feedback on what we do with production migration. 
There is full flow ongoing refactoring for production migration required that could take time. We could remove it for the moment and create task to restore them if needed. WDYT?

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -0,0 +1,45 @@
+import { GraphbackOperationType } from './GraphbackOperationType';
+
+// TODO Talk with Enda. Should mapping be dynamic and swapable or we going strong with CRUD spec
```
wtrocki: So question. Are we going strong with CRUD model (I believe we should to keep things simple or allowing peole to override it if needed and provide custom names. This is very simple change in terms of code but it has tons of side impacts on docs etc.

wtrocki:
@craicoverflow Looking for feedback on what we do with production migration. 
There is full flow ongoing refactoring for production migration required that could take time. We could remove it for the moment and create task to restore them if needed. WDYT?

wtrocki:
Production migration will require day to carefully test it and make sure that it works. InputType context is used there with some extra flags so there will be need for separate vistor etc. I totally forgot about this and thanks for typescript I got nicely reminder in form of 71 compilation errors

craicoverflow:
> @craicoverflow Looking for feedback on what we do with production migration.
> There is full flow ongoing refactoring for production migration required that could take time. We could remove it for the moment and create task to restore them if needed. WDYT?

I think we could remove this for the time being as it is not in use anywhere.

PR #678, opened 2020-02-03
fix: check both operation types

craicoverflow:
## What

The `if` condition does not get entered entered when the action is `GraphbackOperationType.FIND`.

### DIFF
--- a/packages/graphback-core/src/crud/mappingHelpers.ts
+++ b/packages/graphback-core/src/crud/mappingHelpers.ts
@@ -26,7 +26,7 @@ import { GraphbackOperationType } from './GraphbackOperationType';
  */
 export const getFieldName = (typeName: string, action: GraphbackOperationType): string => {
   let finalName = upperCaseFirstChar(typeName);
-  if (action === GraphbackOperationType.FIND_ALL && GraphbackOperationType.FIND) {
+  if (action === GraphbackOperationType.FIND_ALL || action === GraphbackOperationType.FIND) {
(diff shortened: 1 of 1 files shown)


packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -93,10 +93,10 @@ export function getRelationFieldName(field: any, type: any) {
   return fieldName;
 }
 
-function lowerCaseFirstChar(text: string) {
+export function lowerCaseFirstChar(text: string) {
```
wtrocki: I have seen them as internal implementation. Generally if someone will use this in any plugin it will mean that they do not implement the mappings right, but I guess you have some code that will use it as well for table names etc. so Let's merge

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -26,7 +26,7 @@ import { GraphbackOperationType } from './GraphbackOperationType';
  */
 export const getFieldName = (typeName: string, action: GraphbackOperationType): string => {
   let finalName = upperCaseFirstChar(typeName);
-  if (action === GraphbackOperationType.FIND_ALL && GraphbackOperationType.FIND) {
+  if (action === GraphbackOperationType.FIND_ALL || action === GraphbackOperationType.FIND) {
```
wtrocki: A person who wrote it must have been extremely tired or drunk

PR #681, opened 2020-02-03
feat: Improved Plugin execution engine

wtrocki:
The motivation here is to split resource generation and schema operations.


packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -41,9 +41,11 @@ export class GraphbackCoreMetadata {
 
     /**
      * Get Graphback Models - GraphQL Types with additional CRUD configuration
+     * 
+     * @param forceRefresh - refresh models calculated from schema
```
wtrocki: This implementation is still flawed. I think we have this optimalization to get models only once, but if we will have plugin that creates model then this will not be included. Same if we add new field (for relationships etc.)

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
```diff
@@ -40,12 +40,17 @@ export class GraphbackPluginEngine {
 
     public execute(): GraphbackCoreMetadata {
```
wtrocki: This it the core of the plugin. I has some flaws at the moment:

init accepts and caches metadata for create resources, but it might happen that schema changes might affect work we have done in init. 

Generally, I'm thinking that we should get rid of init completely.
Then we can have transformSchema and generateResouces always fetching freshModels (no caching as we do now)


So it should be something like:

```
    this.metadata = plugin.transformSchema(this.metadata);	            
}

   this.metadata = plugin.generateResources(this.metadata);
}
```
WDYT?

wtrocki: As for the unit testing issue. Individual plugin can still have plublic methods that we will use for testing (apart from generating resources.

packages/graphback-codegen-client/src/ClientCRUDPlugin.ts
```diff
@@ -41,28 +41,27 @@ export const CLIENT_CRUD_PLUGIN = "ClientCRUDPlugin";
...
     constructor(pluginConfig?: ClientGeneratorPluginConfig) {
         super()
         this.pluginConfig = Object.assign({ outputFormat: 'ts' }, pluginConfig);
 
     }
-    // TODO change interface to return metadata instead of schema
-    public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
-        const schema = metadata.getSchema()
+
+    public init(metadata: GraphbackCoreMetadata) {
```
craicoverflow: So we have this as a public method, but purely for testing purposes? I guess I find this a bit unusual. A method should not be public just so that tests can use it IMO.

I would also never expect a method named `init` to return data.

wtrocki: This can be used in instrumentation tests on top that I'm building where we do not want to polute filesystem. We have been doing a lots of testing with filesystem but that is kinda tricky. I'm going to use this pattern for testing, but later maybe we can figure out something else - like tests can extend plugin etc :)

packages/graphback-codegen-client/src/ClientCRUDPlugin.ts
```diff
@@ -41,28 +41,27 @@ export const CLIENT_CRUD_PLUGIN = "ClientCRUDPlugin";
  */
 export class ClientCRUDPlugin extends GraphbackPlugin {
```
craicoverflow: I find the naming of the functions a bit confusing for this plugin.

`init` => this doesn't reflect what's happening. I believe `getDocuments` would be a better name as it is doing: 

1. Getting the models from the schema.
2. Generate documents for those models.
3. Returning documents.

`generateResources` should be called `createResources`. This function is delegating the docment generation and executes the persistence of the documents to the filesystem. I believe `createResources` is more clear.

`createDocuments` should be called `generateDocuments`. It is doing the generation.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -52,8 +58,18 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
         return mergeSchemas({ schemas: [modelsSchema, schema] });
     }
 
+    public generateResources(metadata: GraphbackCoreMetadata): void {
+        const schemString = this.transformSchemaToString(metadata.getSchema());
```
craicoverflow:
```suggestion
        const schemaString = this.transformSchemaToString(metadata.getSchema());
```

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -52,8 +58,18 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+        if (!existsSync(this.pluginConfig.outputPath)) {
+            mkdirSync(this.pluginConfig.outputPath, { recursive: true });
+        }
+        const location = resolve(this.pluginConfig.outputPath, `schema.${this.pluginConfig.format}`);
```
craicoverflow:
```suggestion
        const location = resolve(this.pluginConfig.outputPath, `generated.${this.pluginConfig.format}`);
```

Should be `generated.ts`

craicoverflow:
This architecture looks fine to me. I can see that we are incrementally changing it as we discover the different requirements, which is fine.

Spotted some implementation-level bugs.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
wtrocki: Hmm.. Actually previously we were hardcoding folder name to ./schema/generated.ts since now we do not enforce anything like it we can have just `schema.ts` in whatever location we want

PR #682, opened 2020-02-03
feat: Remove not needed init method in plugin

wtrocki:


packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
```diff
@@ -59,25 +59,25 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
         return PLUGIN_NAME;
     }
 
-
-    public init(metadata: GraphbackCoreMetadata): void {
+    public init(metadata: GraphbackCoreMetadata) {
```
craicoverflow: This should be `transformSchema`

packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
```diff
@@ -48,7 +48,7 @@ const PLUGIN_NAME = 'CRUD_RESOLVER_GENERATOR';
 export class ResolverGeneratorPlugin extends GraphbackPlugin {
 
     private options: ResolverGeneratorPluginOptions;
-    private outputResolves: OutputResolvers;
```
craicoverflow: `outputResolves` => `outputResolvers`

PR #685, opened 2020-02-04
Declarative plugins based on graphql-config and example app

wtrocki:
## Plugin based architecture

Plugins can be declared inside graphql.config file

## Verification

```
yarn graphql codegen
```

- execute cli commands:
	- config
    - generate
    - db 
    - openapi (https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore.yaml)


examples/generator-fullstack/.graphqlrc.yml
```diff
@@ -3,16 +3,12 @@ schema:
...
-        password: postgres
-        database: users
-        host: localhost
-        port: 55432
-      database: pg
-    graphqlCRUD:
+    ##  Input schema
+    model: ./model
+    ## Global configuration for CRUD generator
+    crudSpec:
```
craicoverflow: Hmm, `crudConfig`? Not sure if this is a spec.

packages/graphback-cli/src/components/generate.ts
```diff
@@ -0,0 +1,271 @@
...
+    fsWriteFile(path, data, err => {
+      if (err) {
+        reject(err);
+      }
+      resolve();
+    })
+  })
+}
+
+export function globPromise(glob: string, options: import('glob').IOptions = {}) {
```
craicoverflow: import glob at top of file is preferred.

craicoverflow: Why import glob at function level (and twice)?

packages/graphback-cli/src/components/generate.ts
```diff
@@ -0,0 +1,271 @@
...
+    globAsync(glob, options, (err, data) => {
+      if (err) {
+        reject(err);
+      }
+      resolve(data);
+    })
+  })
+}
+
+export async function createSchemaFile(cwd: string, generatedSchema: string, config: GenerateConfig) {
```
craicoverflow: I thought plugins will be responsible for file persistence? It looks like schema, resolver (...etc) generation is still happening in CLI here.

craicoverflow: I guess this is still working off the old Graphback architecture. Will file-writing logic be removed from here in favour of plugins?

wtrocki: Yes.. This is old. I pushed original file for reference to be able to migrate graphql-cli easly and have diff

packages/graphback-cli/src/components/generate.ts
wtrocki: This is kinda strange codestyle that is all over the CLi. I will replace that soon

examples/generator-fullstack/.graphqlrc.yml
wtrocki: maybe crudOptions?

packages/graphback-cli/src/components/generate.ts
craicoverflow: I've seen this happen when using VS Code's extract function.

wtrocki:
FYI - @ardatan we reusing a lots of complex code that was in graphql-cli. 
When merging that we can finally document the graphback specific configuration and make it clear for the users what is used where. Plugins also will allow us to define plugins for certain boilerplates that already exist.

wtrocki:
FYI - @ardatan we reusing a lots of complex code that was in graphql-cli. 
When merging that we can finally document the graphback specific configuration and make it clear for the users what is used where. Plugins also will allow us to define plugins for certain boilerplates that already exist.

examples/generator-fullstack/.graphqlrc.yml
```diff
@@ -21,12 +17,31 @@ extensions:
...
+          format: 'ts'
+          outputPath: ./client/src/graphql
+        graphback-resolvers:
+          format: ts
+          outputPath: ./server/src/resolvers
+          types:
+            resolverRootLocation: ../../generated-types
+            resolverRootType: Resolvers
+    ## Knex DB Migration config
+    dbmigrate:
```
wtrocki: I really struggling to find good name here. we can incorporate that for dbmigration and then have separate db configuration in the application.

@craicoverflow Looking for opinion on this as I would like to avoid renaming it.

examples/generator-fullstack/.graphqlrc.yml
```diff
@@ -3,16 +3,12 @@ schema:
...
-      dbConfig:
-        user: postgresql
-        password: postgres
-        database: users
-        host: localhost
-        port: 55432
-      database: pg
-    graphqlCRUD:
+  # Graphback configuration
+  graphback:
```
wtrocki: @ardatan I have renamed generate to `graphback` as this will reflect configuration of the graphback generators that will be documneted on our website. We will have the same format everywhere. If we want to align with each cli command then we will probably need two keys now:

- Generate
- Migrate 

However, that will be more misleading IMHO

examples/generator-fullstack/.graphqlrc.yml
wtrocki: generally we had this feedback that db is confusing and we did not documented how to swap DB here. DBMigrate is more descriptive about intention but then it will be strange to use that inside example app

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -70,21 +62,60 @@ export const askForClient = async (): Promise<boolean> => {
...
+          "graphback-schema": {
+            "format": "gql",
+            "outputPath": "./src/schema"
+          },
+          "graphback-resolvers": {
+            "format": "ts",
+            "outputPath": "./src/resolvers"
+          }
+        },
+        "dbmigrations": {
```
craicoverflow: Mixed use of `dbmigrations` and `dbmigrate`

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -70,21 +62,60 @@ export const askForClient = async (): Promise<boolean> => {
...
+  const configPath = `${process.cwd()}/.graphqlrc.yml`
 
   const dockerComposePath = `${process.cwd()}/docker-compose.yml`
-  const config = {}
   const [dbConfig, dockerCompose] = getConfig(database)
-  config['db'] = {
-    dbConfig: JSON.parse(dbConfig),
-    database: database
+  const graphqlConfig: IGraphQLConfig = {
+    schema: './src/schema/*.gql',
```
craicoverflow: Why are we now preferring .gql files over .graphql files?

packages/graphback-cli/src/templates/resources/models/Default.graphql
```diff
@@ -1,12 +1,14 @@
+
+# """ @model """
 # type Note {
 #   id: ID!
 #   title: String!
 #   description: String!
 #   ## Relationship
 #   comment: [Comment]!
```
craicoverflow:
```suggestion
#   comments: [Comment]!
```

Should be comment**s**

packages/graphback-codegen-client/src/index.ts
```diff
@@ -1,2 +1,5 @@
 export * from './ClientDocuments'
-export * from './ClientCRUDPlugin'
\ No newline at end of file
+export * from './ClientCRUDPlugin'
+
+// Required for plugins
+export { ClientCRUDPlugin as Plugin } from './ClientCRUDPlugin'
```
craicoverflow: I wonder is `Plugin` too generic. `ClientCRUDPlugin as GraphbackPlugin`?

packages/graphback/src/GraphbackGenerator.ts
```diff
@@ -36,13 +36,33 @@ export class GraphbackGenerator {
...
+      } else {
+        pluginName = pluginLabel;
+      }
+      try {
+        // tslint:disable-next-line: non-literal-require
+        const plugin = require(pluginName)
+        if (plugin.Plugin) {
+          const config = this.config.plugins[pluginLabel]
+          // FIXME - plugins missing basic validation
+          pluginEngine.registerPlugin(new plugin.Plugin(config))
```
craicoverflow: Love it!

packages/graphback/src/GraphbackGenerator.ts
```diff
@@ -36,13 +36,33 @@ export class GraphbackGenerator {
...
-    const resolverPlugin = new ResolverGeneratorPlugin(this.config.plugins?.ResolversCRUD);
-    const clientCRUDPlugin = new ClientCRUDPlugin(this.config.plugins.ClientCRUD)
-    pluginEngine.registerPlugin(schemaCRUDPlugin, resolverPlugin, clientCRUDPlugin);
+    const pluginEngine = new GraphbackPluginEngine(this.schema, { crudMethods: this.config.crud })
+
+    for (const pluginLabel of Object.keys(this.config.plugins)) {
+      let pluginName;
+      if (pluginLabel.startsWith('graphback-')) {
+        // Internal graphback plugins needs rename
+        pluginName = pluginLabel.replace('graphback-', '@graphback/codegen-')
```
craicoverflow: We need to document that custom plugins should not start with "graphback-"

packages/graphback/src/GraphbackGenerator.ts
```diff
@@ -36,13 +36,33 @@ export class GraphbackGenerator {
    */
   // FIXME generator options should be moved to plugin config
```
craicoverflow: Can this be removed now?

packages/graphback-cli/src/templates/configTemplates.ts
craicoverflow: `documents` config falls under `./client` but `schema` config does not fall under `server` (as it does in template config. For this reason `graphback config` does not work out of the box.

examples/generator-fullstack/server/tsconfig.json
```diff
@@ -7,7 +7,7 @@
       "dom"
     ],
     "resolveJsonModule": true,
-    "noImplicitAny": false,
+    "noImplicitAny": true,
```
craicoverflow: Generated resolvers do not use `any` for arguments. Server fails to start:

```
"⨯ Unable to compile TypeScript:src/resolvers/generated/note.ts(5,17): error TS7006: Parameter 'parent' implicitly has an 'any' type.src/resolvers/generated/note.ts(5,25): error TS7006: Parameter 'args' implicitly has an 'any' type.src/resolvers/generated/note.ts(5,31): error TS7006: Parameter 'context' implicitly has an 'any' type.src/resolvers/generated/note.ts(9,20): error TS7006: Parameter 'parent' implicitly has an 'any' type.src/resolvers/generated/note.ts(9,28): error TS7006: Parameter 'args' implicitly has an 'any' type.src/resolvers/generated/note.ts(9,34): error TS7006: Parameter 'context' implicitly has an 'any' type.src/resolvers/generated/note.ts(16,18): error TS7006: Parameter 'parent' implicitly has an 'any' type.src/resolvers/generated/note.ts(16,26): error TS7006: Parameter 'args' implicitly has an 'any' type.src/resolvers/generated/note.ts(16,32): error TS7006: Parameter 'context' implicitly has an 'any' type.src/resolvers/generated/note.ts(27,18): error TS7006: Parameter 'parent' implicitly has an 'any' type.src/resolvers/generated/note.ts(27,26): error TS7006: Parameter 'args' implicitly has an 'any' type.src/resolvers/generated/note.ts(27,32): error TS7006: Parameter 'context' implicitly has an 'any' type."
```

examples/generator-fullstack/server/src/index.ts
```diff
@@ -20,40 +18,40 @@ async function start() {
...
-  const context = createKnexRuntimeContext(db, pubSub);
-  const apolloConfig = {
-    schema,
-    context
-  }
 
-  const apolloServer = new ApolloServer(apolloConfig)
+  const apolloServer = new ApolloServer({
+    typeDefs: loadSchemaFiles(join(__dirname, '/schema/')),
+    resolvers: loadResolversFiles(join(__dirname, '/resolvers/')),
```
craicoverflow: Getting an error and believe it is something to do with how resolvers are loaded here.
```
Error: "customResolvers" defined in resolvers, but not in schema
```

examples/generator-fullstack/.graphqlrc.yml
craicoverflow: Does the config parent strictly have to include the migrate/migrations keyword? Could we have something like:

I don't see how `DBMigrate` is more beneficial really.

- dbOptions
- dbSettings
...

examples/generator-fullstack/server/src/index.ts
wtrocki: Format of custom resolvers might have changed

examples/generator-fullstack/.graphqlrc.yml
wtrocki: It is really defining where this data is used. We do not use it for any other reason than migration

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: Yes. Some of the changes were reverted by mistake. But I guess we need to rename this field. Can you tell me your prefference?

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: That is a bug. We need to change type name to graphql instead of gql

wtrocki: Graphback config will not work out of the box as we never know what folder structure people will have. Adjusting it to our example app makes no sense at all.

packages/graphback-codegen-client/src/index.ts
wtrocki: The purpose for this is to be generic. This is how we know that package defines plugin. Better docs needed.

packages/graphback/src/GraphbackGenerator.ts
wtrocki: I will add TODO

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: Do you have any proposition for this? I came up with some assumption that we will be running this in the server side root and client is just optional

packages/graphback-cli/src/templates/configTemplates.ts
craicoverflow: Of the two, I prefer `dbmigrations` (perhaps it should be `dbMigrations`)?

`dbmigrate` reads more like an action, not a config.

packages/graphback-cli/src/templates/configTemplates.ts
craicoverflow: Ah okay that's fair enough then.

examples/generator-fullstack/.graphqlrc.yml
wtrocki: Coming back to this. Do you think we should be back to db as name?

examples/generator-fullstack/server/tsconfig.json
wtrocki: All should be strongly typed. This example in GraphQL CLI is actually very very focused on strong typing as this comes from graphql-code-generator.

craicoverflow: Well then we should probably change the default resolver args to use `any`

https://github.com/aerogear/graphback/blob/60da66c1e911881ab1fbac9c9854ec4b4be18ada/packages/graphback-codegen-resolvers/src/output/resolverTemplates.ts#L1

examples/generator-fullstack/.graphqlrc.yml
craicoverflow: I prefer `dbMigrations` over `dbMigrate` and am fine with using that.

The db config data is not just used in migrations. Migrations might happen once while these settings are used ever time the server runs for making a connection to the database.

Issue #668, opened 2020-01-30
Generate model objects in resolver generator


wtrocki:
More info here:
https://github.com/aerogear/graphback/issues/468

This will allow us to decompose the runtime layer as the default choice. 
Developers then could manually change models as they will not be overwritten once created.

wtrocki:
Done in #639

PR #687, opened 2020-02-04
Database mapping abstraction

craicoverflow:


craicoverflow:
@wtrocki Sample mapping output:

```json
{
   "modelMap":[
      {
         "idField":"title",
         "typeName":"Note",
         "tableName":"note",
         "fieldMap":{
            "id":"id",
            "title":"title",
            "description":"overview",
            "comments":"comments"
         }
      },
      {
         "idField":"description",
         "typeName":"Comment",
         "tableName":"note_comment",
         "fieldMap":{
            "id":"id",
            "description":"overview"
         }
      }
   ]
}
```

packages/graphback-core/src/annotations/parser.ts
```diff
@@ -2,9 +2,9 @@ import { FieldDefinitionNode, GraphQLField, GraphQLObjectType, TypeDefinitionNod
 import { parseAnnotations } from "graphql-metadata";
 
 export const parseDbAnnotations = (node: TypeDefinitionNode | FieldDefinitionNode | GraphQLObjectType | GraphQLField<any, any>): any => {
-    return node.description ? parseAnnotations('db', String(node.description)) : undefined;
+    return node.description ? parseAnnotations('db', String(node.description)) : {};
```
wtrocki: [Question] Motivation for this change?

packages/graphback-core/src/db/getPrimaryKey.ts
```diff
@@ -0,0 +1,21 @@
+import { GraphQLObjectType, isScalarType } from "graphql";
+import { parseDbAnnotations } from '../annotations/parser';
+import { getBaseType } from '../utils/getBaseType';
+
+export function getPrimaryKey(graphqlType: GraphQLObjectType): string | undefined {
```
wtrocki: Unit tests will be handy

packages/graphback-core/src/db/getPrimaryKey.ts
```diff
@@ -0,0 +1,21 @@
...
+  const fields = Object.values(graphqlType.getFields());
+
+  let primaryKey: string;
+  for (const field of fields) {
+    const dbConfig: any = parseDbAnnotations(field);
+    const baseType = getBaseType(field.type);
+
+    if (dbConfig.primary) {
+      primaryKey = field.name;
+    } else if (isScalarType(baseType) && baseType.name === 'ID') {
```
wtrocki: [Possible bug] It always takes the last and still continues to look for another id. 
We can detect if there are two and pick one with `id` if not fail migration completely as this is clearly bug on the user side.

packages/graphback-core/src/db/databaseMapper.ts
```diff
@@ -0,0 +1,56 @@
+import { GraphQLField, GraphQLFieldMap, GraphQLObjectType, GraphQLSchema } from "graphql";
+import { parseDbAnnotations } from '../annotations/parser';
+import { getUserModels } from '../crud';
+import { defaultTableNameTransform } from './defaultNameTransforms';
+import { getPrimaryKey } from './getPrimaryKey';
+
+export interface DatabaseMap {
```
wtrocki: Very missleading and generic name

packages/graphback-core/src/db/databaseMapper.ts
```diff
@@ -0,0 +1,56 @@
...
+  modelMap: ModelMap[]
+}
+
+export interface ModelMap {
```
wtrocki: Lack of comments

packages/graphback-core/src/db/databaseMapper.ts
```diff
@@ -0,0 +1,56 @@
...
+  typeName: string
+  tableName: string
+  id: string
+  fieldMap: any
+}
+
+export const getModelTableMappings = (schema: GraphQLSchema): DatabaseMap => {
```
wtrocki: Can we use array of model types here. It will help us to make this clearly about the CRUD types. If someone do not map element as model it should not be included (equivalent to db.disable

wtrocki: This is kinda complex topic so happy to have chat about this.

packages/graphback-core/src/db/defaultNameTransforms.ts
```diff
@@ -1,19 +1,30 @@
-import { NameTransformDirection } from '../abstract/generateAbstractDatabase'
+export type NameTransformDirection = 'from-db' | 'to-db'
+
+export type NameTransform = (
```
wtrocki: Generic name, No comments. Hard to get sense of what this does

packages/graphback-core/src/annotations/parser.ts
craicoverflow: Saves having to check if the returned `annotations` is defined after returning.

`if (annotations && annotations.name)` => `if (annotations.name)`.

The only place we really cared about having an undefined annotation is with the `@model` marker.

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -62,15 +65,10 @@ export const getInputTypeName = (typeName: string): string => {
   return `${typeName}Input`;
 }
 
+export function getUserModels(schema: GraphQLSchema) {
```
wtrocki: [Possible bugs] This method is kinda hard to grasp as we doing completely separate processing here on top of schema. 
Maybe we should simply pass models here

packages/graphback-core/src/db/defaultNameTransforms.ts
craicoverflow: Agreed, I moved this from `graphql-migrations`.

packages/graphback-core/src/db/databaseMapper.ts
craicoverflow: Yeah I had this changed already but not pushed!

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -0,0 +1,55 @@
...
+    data?: any
+    fieldMap?: any
+}
+
+export interface TableID {
+    name: string
+    value: any
+}
+
+export const mapDataToTable = (data: any, modelMap: ModelTableMapping): DataMapping => {
```
craicoverflow: Would `mapDTO` be a better name here?

wtrocki: DTO stands for DataTransferObject and it will be confusing

wtrocki: Name is ok

wtrocki:
While mapping database objects is ok. My initial suggestion was to go with the id and relationship mapping. ID and Relationship mapping is the most important one here as we need to know name of the field when:

##### Generation phase 
- Generating schema input type
- Generating client 

##### Generation phase 
- Batching request

The actual mapping of the table, fieldNames is less an blocker here.
As for table mapping I think we should pass table name to the crud engine and simply have only fieldMappings (those can be both generated or supported inside CRUD layer)

packages/graphback-codegen-resolvers/src/output/createResolvers.ts
```diff
@@ -53,53 +53,49 @@ export function generateCustomCRUDResolvers(schema: GraphQLSchema, models: Model
...
+        const fieldName = getFieldName(modelName, GraphbackOperationType.CREATE);
         // tslint:disable-next-line: no-any
-        mutations[fieldName] = createTemplate(objectName, crudOptions.subCreate)
+        mutations[fieldName] = createTemplate(modelName, crudOptions.subCreate)
     }
     if (crudOptions.update) {
-        const fieldName = getFieldName(graphqlType.name, GraphbackOperationType.UPDATE);
-        mutations[fieldName] = updateTemplate(objectName, crudOptions.update);
+        const fieldName = getFieldName(modelName, GraphbackOperationType.UPDATE);
+        mutations[fieldName] = updateTemplate(modelName, crudOptions.update);
```
wtrocki: I think we need to first do https://github.com/aerogear/graphback/issues/668 to visualize how this will work, as passing string to CRUD layer might be confusing. 

WE can have:
- New crud layer for each Type that accepts field mapping and id in constructor.
- Generic CRUD layer that will accepts string as Model name but then Service layer and crud layer need to map it to the table. 
- Pass the table name 

Some discussion needs to happen there

packages/graphback-core/src/db/createModelTableMappings.ts
```diff
@@ -0,0 +1,85 @@
...
+ * 
+ * - typeName: Original GraphQLObjectType name
+ * - tableName: Name of datase table
+ * - id: Indicates the primary key field
+ * - fieldMap: Object of key-value mapping between GraphQL fields and database columns.
+ */
+export interface ModelTableMapping {
+  typeName: string
+  tableName: string
+  id: string
```
wtrocki:
```suggestion
  idField: string
```

wtrocki:
Build is still failing

PR #689, opened 2020-02-04
fix: add missing @types/glob in runtime examples

machi1990:
Without this the example was failing with `src/loadSchema.ts:2:26 - error TS7016: Could not find a
declaration file for module 'glob'`

/cc @wtrocki

### DIFF
--- a/examples/runtime-example/package.json
+++ b/examples/runtime-example/package.json
@@ -24,6 +24,7 @@
     "pg": "7.17.1"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


wtrocki:
Thank you so much for contribution 💯 !! 
We using mono repository structure where some packages will be hoisted in root node_modules. This fix allows compiling example when directly installing dependencies in runtime example. That is why we did not detected that in our CICI 
process and on local machines

PR #699, opened 2020-02-07
Resolver refactor

wtrocki:
## Motivation

Current resolver generator logic is really hard to maintain:

- There is no way to reuse any of the codebase as we do not expose methods and abstractions
- Formatting to Apollo/GraphQLJs layout is mixed up
- Prettier is applied in wrong stage making it impossible to pick right format
(we do not need prettier I think)
- Index files are not needed and are tricky to build

I'm adding model generator and going to refactor implementation to be easier to maintain. I will squash commits soon


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -123,18 +123,16 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
                 const fieldBaseType = getBaseType(field.type);
 
                 return !isObjectType(fieldBaseType);
             }).reduce((fieldObj: any, current: any) => {
-                // FIXME read id annotation instead of hardcoding id!
-                if (current.name !== 'id') {
-                    fieldObj[current.name] = { type: current.type, description: '' };
-                }
+                const fieldType = current.type;
+                fieldObj[current.name] = { type: getNullableType(fieldType), description: '' };
```
craicoverflow: You only want to force the input field to be nullable when it is the primary key field.

wtrocki: The same object is being used in filtering so id cannot be not null

craicoverflow: Should we have a separate object for filtering and mutations then?

wtrocki: It can be added later when building use cases
I want to start simple with this and continue extending it with docs etc.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -16,24 +18,31 @@ import { NoDataError } from './NoDataError';
 export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext>{
 
   protected db: Knex;
+  protected baseType: GraphQLObjectType;
+  protected tableName: string;
 
-  constructor(db: Knex) {
+  constructor(baseType: GraphQLObjectType, db: Knex) {
     this.db = db;
+    this.baseType = baseType;
```
craicoverflow: I wonder what the implications of only having a single model passed to the provider. Will we need information about relationship models etc which may not be possible to get from single object?

wtrocki: Object will contain relationships information as it will be embeeded into the type.getFields()[0].getType(). 

In fact we will need to check if the type is not custom scalar or union so we should always based on this. 
There are couple questions about relationships. 

If the relationship type has no model annotation it should probably be ignored.

packages/graphback-runtime/src/createKnexRuntimeContext.ts
```diff
@@ -1,23 +0,0 @@
-import { getModelTableMappings } from '@graphback/core';
-import { GraphQLSchema } from 'graphql';
-import { PubSubEngine } from 'graphql-subscriptions';
-import { CRUDService, PgKnexDBDataProvider } from '.';
-import { GraphbackRuntimeContext } from './api/GraphbackRuntimeContext';
-
-/**
- * Create context object required for the graphback runtime layer 
- */
-export const createKnexRuntimeContext = (db: any, pubSub: PubSubEngine, schema: GraphQLSchema): GraphbackRuntimeContext => {
```
wtrocki: We need to revert this

PR #700, opened 2020-02-10
Resolve GraphQL↔️Database ID mapping at Database Provider Layer

craicoverflow:
## What

This change maps the primary key field from the input data to the database table. This enables custom IDs to be used.

New schema mutation definition:

```gql
# New update spec
updateNote(input: NoteInput): Note
# Old update spec
updateNote(id: ID!, input: NoteInput): Note
```

The problem with the old specification is that the primary key field needed to be type `id: ID`. With the new specification we have just delegated resolving the primary key until we get to the database layer.

⬇️
`KnexDBDataProvider` has the entire schema-database table mapping passed in the constructor. This is used to dynamically figure out the ID field for the entity.
⬇️
```ts
const { id, data: updateData } = getUpdateArgs(idField, data); // get ID key/value from table mapping

// tslint:disable-next-line: await-promise
const updateResult = await this.db(tableName).update(updateData).where(id.field, '=', id.value);
```
## Tasks

- [x] Update CRUD and Database Provider interface arguments to not accept ID as parameter in UPDATEs. 
- [x] Set type-table mapping in Data Provider. 
- [ ] Map ID data in update method to update an entity by its ID.

## Verification

Execute a mutation to update your model using the new specification:

```gql
mutation updateNote {
  updateNote(input:{description:"foobar", title:"barfoo", id: 2}) {
    id
    description
    title
  }
}
```

## Notes

This is only implemented for the UPDATE method.


packages/graphback-codegen-schema/src/writer/schemaPrinter.ts
```diff
@@ -42,13 +42,15 @@ export function printSortedSchema(schema: GraphQLSchema) {
     }).join('\n\n');
 
     schemaString += '\n\n';
-    if (schema.getQueryType()) {
+    if (schema.getQueryType() && Object.keys(schema.getQueryType().getFields()).length) {
```
wtrocki: Revert this

packages/graphback-core/src/db/getPrimaryKey.ts
```diff
@@ -1,19 +1,19 @@
...
   for (const field of fields) {
     const dbConfig: any = parseDbAnnotations(field);
     const baseType = getBaseType(field.type);
 
     if (dbConfig.primary) {
-      return field.name;
+      return field;
     } else if (isScalarType(baseType) && baseType.name === 'ID') {
-      primaryKey = field.name;
+      primaryKey = field;
```
wtrocki: we need to add `break;`

examples/generator-fullstack/server/src/index.ts
```diff
@@ -32,10 +33,13 @@ async function start() {
 
   const pubSub = new PubSub();
 
+  const typeDefs = loadSchemaFiles(join(__dirname, '/schema/'));
+  const schema = buildSchema(typeDefs.join('\n'));
+
   const apolloServer = new ApolloServer({
-    typeDefs: loadSchemaFiles(join(__dirname, '/schema/')),
+    typeDefs,
```
wtrocki: should use schema probably.

examples/generator-fullstack/server/src/index.ts
```diff
@@ -32,10 +33,13 @@ async function start() {
 
   const pubSub = new PubSub();
 
+  const typeDefs = loadSchemaFiles(join(__dirname, '/schema/'));
+  const schema = buildSchema(typeDefs.join('\n'));
```
wtrocki: Wrong method. Apollo is using `buildExecutableSchema`

packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
```diff
@@ -78,12 +78,10 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
 
         const generatedResolverGroup = createOutputResolvers(generatedResolvers, this.pluginConfig);
         const customResolverGroup = createCustomOutputResolvers(customResolvers, this.pluginConfig);
-        const rootResolverIndex = createRootResolversIndex(this.pluginConfig.format);
 
         return {
             generated: generatedResolverGroup,
-            custom: customResolverGroup,
-            index: rootResolverIndex
+            custom: customResolverGroup
```
wtrocki: Revert this change please.

packages/graphback-codegen-resolvers/src/output/createResolvers.ts
```diff
@@ -1,5 +1,5 @@
 import { getFieldName, getTableOrColumnName, GraphbackCRUDGeneratorConfig, GraphbackOperationType, ModelDefinition } from '@graphback/core';
```
wtrocki: This entire file no longer exist. I have rewritten entire resolvers.

craicoverflow:
Closing in favour of https://github.com/aerogear/graphback/pull/702

PR #703, opened 2020-02-11
Use column name transform

craicoverflow:
This change uses the `this.transformColumnName` instead of `lowerCaseFirstChar` to transform the column name. when creating mapping a one-to-many relationship.

`this.transformColumnName` should be used at all times for column name transformation as it ensures a common transformer is used. This is configurable at the point of instantiation.


craicoverflow:
Resolved

PR #705, opened 2020-02-11
feat: Runtime support

wtrocki:
Remove complexity when generating context creator file. 

File content looks as follows: 
```ts

import {
  createKnexCRUDRuntimeContext,
  KnexRuntimeContextConfig
} from "@graphback/runtime"

export const createCRUDResolversRuntimeContext = (
  options: KnexRuntimeContextConfig
) => {
  const { schema, db, pubSub } = options

  const notePubSubContext = {
    pubSub,
    publishCreate: true,
    publishUpdate: true,
    publishDelete: true
  }
  const commentPubSubContext = {
    pubSub,
    publishCreate: true,
    publishUpdate: true,
    publishDelete: true
  }

  return {
    Note: createKnexCRUDRuntimeContext("Note", schema, db, notePubSubContext),

    Comment: createKnexCRUDRuntimeContext(
      "Comment",
      schema,
      db,
      commentPubSubContext
    )
  }
}

```


wtrocki:
We have added also support for runtime case. Runtime example now will use new subsystem

wtrocki:
Build failing because of linting issues (we should fix it on feature branch to not conflict)

examples/generator-fullstack/server/src/graphql.ts
```diff
@@ -16,7 +16,7 @@ export const createApolloServer = async () => {
 
     const typeDefs = loadSchemaFiles(join(__dirname, '/schema/')).join('\n')
```
craicoverflow: I thought this would only be temporary? :P

Issue #648, opened 2020-01-21
Add support for custom id's


wtrocki:
Could it be autoincrementing and id: ID! by default but allow to use anything as ID and then do not care about increments and anything else. For example email could be marked as unique and that will be enough to have it as ID

wtrocki:
`default: id: ID!`

Without annotation:
`email: ID!`

Validation for multiple ID's error!

""" db.id """
`email: String!`

craicoverflow:
What order do we choose?

If our model has:

```gql
id: ID!
```
and

```gql
@db.primary
email: String
```

which should we choose as the primary key used in database generation and mapping?

I think we should prefer the `@db.primary` field over `id: ID!` field. This is an explicit request by the user to make this the primary key so it should be treated as such.

^^ @wtrocki

wtrocki:
Fully agree. Explicit over implicit case

craicoverflow:
This is now merged in Graphback, but `graphql-migrations` does not have this support yet (it is restricted to `id: ID!`. Closing this issue however as this is related to mapping and not migrations.

PR #711, opened 2020-02-12
Map ID fields at database level to enable custom ID fields

craicoverflow:
See https://github.com/aerogear/graphback/issues/648

This change dynamically maps the ID field/value to the database column at runtime. This means we can have custom IDs as we don't need to know the ID field during generation of queries and mutations.

## Tasks

- [x] Create model-table mapping abstraction
- [x] Map ID field
- [x] Use ID mapping in CRUD Database operations.
- [x] Remove `read` from CRUD spec.
- [x] Add unit tests

## Verification

Create a `Note`, then update it:

```gql
mutation createNote {
  createNote(input: {title: "test", description: "test"}) {
    id
    title
  } 
}

mutation updateNote{
  updateNote(input: {id: 1, title: "test2", description: "test"}) {
    id
    title
  }
}
```


craicoverflow:
Ready for review @wtrocki - just working on adding tests.

examples/generator-fullstack/server/src/schema/schema.graphql
```diff
@@ -1,6 +0,0 @@
-## NOTE: Schema should be generated by schema plugin generator
```
wtrocki: Why?

packages/graphback-core/src/db/buildModelTableMap.ts
```diff
@@ -0,0 +1,68 @@
...
+ * - typeName: Original GraphQLObjectType name
+ * - tableName: Name of datase table
+ * - id: Indicates the primary key field
+ * - fieldMap: Object of key-value mapping between GraphQL fields and database columns.
+ */
+export interface ModelTableMap {
+  typeName: string
+  tableName: string
+  idField: string
+  fieldMap?: any
```
wtrocki: Maybe we can strongly type this to something like 
```suggestion
  fieldMap?: {[key:string]: string }
```

packages/graphback-core/src/db/buildModelTableMap.ts
```diff
@@ -0,0 +1,68 @@
...
+    // TODO: Map relationship fields
+
+    return obj;
+  }, {});
+}
+
+export function getTableName(model: GraphQLObjectType): string {
+  let mappedTableName = defaultTableNameTransform(model.name, 'to-db');
+
+  const dbAnnotations = parseDbAnnotations(model);
```
wtrocki: I think that will be really hard to document and very confusing. 
We are basically going to say that DB migrations are optional and only to be used when
we do have some db, but then people will need to use some subset of the db annotations to do mapping (even when their data source is REST etc.)

Not sure what we can do with this and if that is doable in current PR, but raising as possible issue

packages/graphback-core/src/db/buildModelTableMap.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (dbAnnotations.name) {
+    mappedTableName = dbAnnotations.name;
+  }
+
+  return mappedTableName;
+}
+
+export function getColumnName(field: GraphQLField<any, any>): string {
+  let mappedColumnName = defaultColumnNameTransform(field.name, 'to-db');
```
wtrocki: Why we doing this here. Can we skip transform by default when migration is happening. 
For example when we currently have only `model` annotation on plain type - What transforms this will produce? Do we need transforms at all. It seems to be overly complicated.

I kinda feel that we still need to control default transform but `  return lowerCaseFirstChar(name);` is not the best default value

packages/graphback-core/src/db/getPrimaryKey.ts
```diff
@@ -1,19 +1,20 @@
...
     const baseType = getBaseType(field.type);
 
     if (dbConfig.primary) {
-      return field.name;
+      primaryKey = field;
+      break;
     } else if (isScalarType(baseType) && baseType.name === 'ID') {
-      primaryKey = field.name;
+      primaryKey = field;
     }
```
wtrocki: Missing else. Missing warning when more than one ID is present

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -20,29 +20,37 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
...
   }
 
   public async create(data: Type): Promise<Type> {
-    // TODO refactor to use new knex api
-    const [id] = await this.db(this.tableName).insert(data);
-    const dbResult = await this.db.select().from(this.tableName).where('id', '=', id)
+    const { data: createData } = getDatabaseArguments(this.tableMap, data);
+
+    // tslint:disable-next-line: await-promise
+    const [id] = await this.db(this.tableName).insert(createData);
```
wtrocki: I do not think that this will work with anything else than `id`. 
We need to specify returning as in docs: http://knexjs.org/#Builder-insert

wtrocki: Ok. Ignore me. That will not work with MySQL. We need to update CRUD tests to use non id and mapping to tests this use case

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
```diff
@@ -20,29 +20,37 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
...
     if (dbResult && dbResult[0]) {
       return dbResult[0]
     }
     throw new NoDataError(`Cannot create ${this.tableName}`);
   }
 
   public async update(data: Type): Promise<Type> {
-    // TODO refactor to use id mapping
-    const updateResult = await this.db(this.tableName).update(data).where('id', '=', (data as any).id);
+    const { idField, data: updateData } = getDatabaseArguments(this.tableMap, data);
```
wtrocki: Lack of validation when `id` is missing

packages/graphback-core/src/db/buildModelTableMap.ts
craicoverflow: We are doing this here because our migration engine does a transformation of field names. IF for example you provide a model with `MyField: String` then `graphql-migrations` transforms this to column `myField`. In this way, we also need to do the same transformation here to find the default transformation value.

The migration package uses it from core:

https://github.com/aerogear/graphback/blob/0f1f48f238516eaa99f390721dd27c58ac0539a6/packages/graphql-migrations/src/migrate.ts#L1

So it is centrally controlled already.

packages/graphback-core/src/db/buildModelTableMap.ts
craicoverflow: I don't see how it is confusing? database migrations are optional, but the `@db.name` annotation is simply a way to map to your existing column.

craicoverflow: The other option is to introspect an existing database and generated a schema from that.

examples/generator-fullstack/server/src/schema/schema.graphql
craicoverflow: Mistakenly removed.

packages/graphback-core/src/db/buildModelTableMap.ts
wtrocki: Lets keep it simple and remove transformation by default

packages/graphback-core/src/db/buildModelTableMap.ts
craicoverflow: Agreement made to remove default transformations and let people bring their own with annotations only.

packages/graphback-runtime/src/data/KnexDBDataProvider.ts
craicoverflow: Resolved in https://github.com/aerogear/graphback/pull/711/commits/06e6bde642b6c785a5eb9f9890447b9c3ebb3558

packages/graphback-core/src/db/buildModelTableMap.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const dbAnnotations = parseDbAnnotations(model);
+  if (dbAnnotations.name) {
+    tableName = dbAnnotations.name;
+  }
+
+  return tableName;
+}
+
+export function getColumnName(field: GraphQLField<any, any>): string {
```
wtrocki: [Optional] Some docs

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -1,55 +1,36 @@
...
 
-    const dataMap: DataMapping = {
-        table: modelMap.tableName,
-        fieldMap
-    }
-
-    if (!fieldMap) {
-        return dataMap;
-    }
+    // TODO: Map fields to custom db names
```
wtrocki: What this means. Is this what we do here?

craicoverflow: This PR is about mapping custom ID fields. In future we want to map custom data fields with the `@db.name` annotation.

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -1,55 +1,36 @@
...
-    const modelMap = getModelMappingByName(modelName, modelsMap);
-    const fieldMap = modelMap.fieldMap;
-
-    const mappedData = {};
-    for (const key of Object.keys(fieldMap)) {
-        if (fieldMap[key] && data[fieldMap[key]]) {
-            const newKey = fieldMap[key];
-            mappedData[key] = data[newKey];
-        }
+function getTableId(idField: string, data: any = {}): TableID {
```
wtrocki: I question value of this function. It do not check if idField is undefined. Or handle situation when data not exist (which means there is no chance to do query or mutation).

packages/graphback-core/tests/buildModelTableMapTest.ts
```diff
@@ -0,0 +1,92 @@
...
+});
+
+ava('should use default ID field', (t: ExecutionContext) => {
+    const schema = buildSchema(`
+    """
+    @model
+    """
+    type User {
+        id: ID!
+        email: String!
```
wtrocki:
```suggestion
        email: String!
        secondaryID: ID!
```

packages/graphback-core/tests/buildModelTableMapTest.ts
```diff
@@ -0,0 +1,92 @@
...
+    t.snapshot(JSON.stringify(modelTableMap, undefined, 1))
+});
+
+ava('should use default ID field', (t: ExecutionContext) => {
+    const schema = buildSchema(`
+    """
+    @model
+    """
+    type User {
+        id: ID!
```
wtrocki:
```suggestion
        primary: ID!
```

craicoverflow: This case is covered here: https://github.com/aerogear/graphback/blob/35d426e4d203cfe6f92da3f6d58d2618473dfee7/packages/graphback-core/tests/getPrimaryKeyTest.ts#L48

PR #732, opened 2020-02-17
Offix plugin

wtrocki:
Creating for early feedback.
This plugin will be able to provide some basic version field support on top of the graphback. Stretch goal will be to provide diff queries for better handling for deletes (for conflict)


wtrocki:
Released. Ready to be integrated with showcase. 
There are numerous fixes and improvements here that need to be applied.

wtrocki:
Ready for review

wtrocki:
We going to merge that temporarily to the graphback to be able to build plugin faster and then decide if that should go to offix or datasync repository.

.vscode/settings.json
```diff
@@ -11,7 +11,7 @@
         "**/.factorypath": true,
         "**/.nyc_output": true,
         "**/node_modules": true,
-        "**/.vscode": true,
+        "**/.vscode": false,
```
craicoverflow: Should this file even be committed?

docs/migrations/migrations.md
```diff
@@ -15,4 +15,5 @@ the chance to adjust the schema and underlying data access layer.
 ## Graphback supported migrations
 
 - OpenAPI (RESTful server migrations)
-- Database migrations (WIP)
+- Various plugins available as part of the GraphQL compose
```
craicoverflow:
```suggestion
- Various plugins available as part of GraphQL compose
```

.vscode/settings.json
wtrocki: Ok.. Yes. This file should be gone.

docs/migrations/migrations.md
craicoverflow: This point needs more information to be useful.

wtrocki: There is link bellow. GraphQL compose can do db/mongose/json migrations

docs/plugins/creating-plugin.md
```diff
@@ -75,6 +75,9 @@ If your plugin relies on other plugins it should be placed on the bottom of the
 
 ```
 
+> NOTE: Graphback cli includes graphback core plugings. When adding your own plugins you need to run graphback cli from your npm scripts
```
craicoverflow:
```suggestion
> NOTE: Graphback CLI includes Graphback core pluging. When adding your own plugins you need to run Graphback CLI from your npm scripts
```

docs/migrations/migrations.md
wtrocki: There is no point to duplicate info - we just put reference if people want that and we have now the solution for db based access

docs/plugins/graphback-client.md
```diff
@@ -28,8 +28,8 @@ apolloClient.mutate(createUserQuery)
 
 Plugin allows to create files in many formats to suit needs of your application.
 
-Please reffer to generator-fullstack app for fully functional example:
-https://github.com/aerogear/graphback/tree/master/examples/generator-fullstack
+Please reffer to ts-apollo-fullstack app for fully functional example:
```
craicoverflow:
```suggestion
Please refer to ts-apollo-fullstack app for fully functional example:
```

packages/graphback-codegen-offix/src/OffixPlugin.ts
```diff
@@ -0,0 +1,99 @@
...
+    generateDeltaQueries: boolean
+
+    /**
+     * Delta resolvers format
+     */
+    deltaResolverFormat: 'ts' | 'js' | 'graphql',
+
+}
+
+export const SCHEMA_CRUD_PLUGIN_NAME = "SchemaCRUD";
```
craicoverflow: Should be updated to match the plugin name `OFFIX_PLUGIN_NAME` or similar

wtrocki: Good point. We actually ended up not using this at all so this fields do not matter.

packages/graphback-codegen-offix/src/OffixPlugin.ts
```diff
@@ -0,0 +1,99 @@
...
+                    lastSync: 'String!'
+                })
+                // TODO generate delta resolvers
+            }
+        })
+
+        return schemaComposer.buildSchema()
+    }
+
+    public createResources(metadata: GraphbackCoreMetadata): void {
```
craicoverflow: Can this be removed completely if it is not used?

wtrocki: Sadly not. It is defined as abstract in the top class additionally it will be used when I add resolver generation for diff queries. This plugin is just to get us sense how integration will work. We will have it refined later on.

wtrocki:
@cecobask FYI. This PR contains an example custom plugin that we have build to extend schema. There are a couple missing elements in it that we will need to implement later.

packages/graphback-codegen-offix/src/OffixPlugin.ts
wtrocki: I could get that as non abstract but then when implementing the class there will be no methods created - there needs to be some abstract method.

wtrocki:
ts-apollo-fullstack do not have runtime.ts file. Do you mean runtime example?

craicoverflow:
> ts-apollo-fullstack do not have runtime.ts file. Do you mean runtime example?

I attache the wrong screenshot. Happens in graphql.ts

Issue #727, opened 2020-02-17
Relationship support


craicoverflow:
Incorporate relationship support into the new plugin architecture.

# Types

We have discussed making two relationship types available:

- `@oneToMany` (can be applied to opposite model to make a `@manyToOne`)
- `@oneToOne`

## Implementation

We should process the pre-process the GraphQL types to build relationship information which will be added as a `ModelDefinition` property. This way the relationship models will be available as part of `GraphbackCoreMetadata` which is then supplied to each plugin.

Needed by:

- Generated schema models
- Input Types
- Resolvers

craicoverflow:
```gql
""" @model """
type Comment {
  id: ID!
  description: String
  author: String
  """
  @db.oneToOne: 'comment'
  @db.name: 'note_id'
  """
  note: Note
}
```

I noticed that `graphql-migrations` does not handle custom relationship field names. This model will create `note_id` and `noteId` in the comment table.

wtrocki:
That is a serious bug that I will consider separate to the spike, bu really good catch! 
We can work without mapping field for the moment and add that gradually

craicoverflow:
> That is a serious bug that I will consider separate to the spike, bu really good catch!
> We can work without mapping field for the moment and add that gradually

True! Created https://github.com/aerogear/graphback/issues/745 for it.

craicoverflow:
**Update**

@wtrocki

For a flexible and simple relationship mapping system to work, we really only need two fields specified in annotations, along with the relationship type:

```
@oneToMany | @oneToOne | @manyToOne
    field: The field name on the opposite model to map to.
    column/key: The database column name on the opposite model to map to.
```

The `field` and `column` can be optional depending on what other metadata is provided. See **Cases** below for some examples.

- Users could use `field` only if they do not care about the underlying database implementation.
- Users can additionally specify `column` if they want to customise the database column name or map to an existing column table.
- `column` on its own would not work as it is not mapped to  GraphQL field.

### Cases

✅ `field` + `column` annotations exist with our without opposite field. We can generate the opposite field since we have all information about it.

✅ `field` annotation exists with or without opposite field. Here we can assume the column name is `${fieldName}Id`.

❌ only `column` annotation on `@oneToMany` (without `field`) and opposite field exists but without annotation. This is invalid because we do not know which field is associated with the custom colum name on the `M:1` side.

❌ `column` annotation exists and opposite field does not exist. This is invalid as we do not have enough information to know which GraphQL field to map to.

[...]

PR #736, opened 2020-02-18
Relationships

craicoverflow:
## Motivation

https://github.com/aerogear/graphback/issues/727

## Details

**oneToMany**
- field: graphql field to map to
- key: (optional) database column name

```gql
""" @model """
type Note {
  id: ID!
  title: String!
  description: String
  """
  @oneToMany field: 'note', key: 'note_id'
  """
  comments: [Comment]!
  likes: Int
}

""" @model """
type Comment {
  id: ID!
  description: String
  author: String
  note: Note
}
```

**oneToOne**
- key: (optional) database column name

```gql
""" @model """
type User {
  id: ID!
  name: String
}

"""
@model
"""
type Address {
  id: ID!
  postcode: String
  """
  @oneToOne key: 'user_id'
  """
  user: User
}
```

## Tasks

- [x] Build relationship metadata abstraction model
- [x] Validate relationships
- [x] Add relation fields to input types.
- [x] Add relationship resolvers.
- [x] Runtime resolvers
- [ ] Unit tests

## Verification

Run these mutations and queries in order:

```gql
mutation addNote {
  createNote(input:{description:"note1", title:"NOTE ONE"}) {
    id
    title
    description
  }
}

mutation addComment {
  createComment(input:{description:"Comment for note 1", noteId:1}) {
    id
    note {
      id
    }
  }
}

query findNotes {
  findAllNotes {
    id
    comments {
      id
      description
    }
  }
}

query findComments {
  findAllComments {
    id
    note {
      id
    }
  }
}
```

## Notes

Not testes with GraphQL format.


craicoverflow:
Sample output:

```json
[
   {
      "parent":"Note",
      "parentField":"comments",
      "relationshipKind":"oneToMany",
      "relationType":"Comment",
      "relationField":"note"
   },
   {
      "parent":"Comment",
      "parentField":"note",
      "relationshipKind":"oneToOne",
      "relationType":"Note",
      "relationField":"comments",
      "foreignKey":{
         "name":"noteId",
         "type":"ID!"
      }
   }
]
```

packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
```diff
@@ -92,7 +92,7 @@ export class ResolverGeneratorPlugin extends GraphbackPlugin {
         const generatedResolvers = createResolverTemplate(resolvers, pluginConfig);
 
         return {
-            fileName: `resolvers.${pluginConfig.format}`,
+            fileName: `${pluginConfig.resolversFileName}.${pluginConfig.format}`,
```
wtrocki: 🙉

packages/graphback-codegen-resolvers/src/templates/createResolvers.ts
```diff
@@ -1,8 +1,29 @@
-import { getFieldName, getSubscriptionName, GraphbackCRUDGeneratorConfig, GraphbackOperationType, ModelDefinition } from '@graphback/core';
+import { getFieldName, getSubscriptionName, GraphbackCRUDGeneratorConfig, GraphbackOperationType, ModelDefinition, RelationshipMetadata, getPrimaryKey } from '@graphback/core';
 import { GraphQLObjectType } from 'graphql';
 import { GeneratorResolversFormat } from '../GeneratorResolversFormat';
-import { createTemplate, deletedSubscriptionTemplate, deleteTemplate, findAllTemplate, findTemplate, newSubscriptionTemplate, updatedSubscriptionTemplate, updateTemplate } from './resolverTemplates';
- 
+import { createTemplate, deletedSubscriptionTemplate, deleteTemplate, findAllTemplate, findTemplate, newSubscriptionTemplate, oneToOneTemplate, updatedSubscriptionTemplate, updateTemplate, oneToManyTemplate } from './resolverTemplates';
+
+export function createRelationshipResolvers(relationships: RelationshipMetadata[]) {
+    if (!relationships.length) { return undefined; }
```
wtrocki:
```suggestion
    if (!relationships.length) { 
        return 
    }
```

packages/graphback-codegen-resolvers/src/templates/createResolvers.ts
```diff
@@ -1,8 +1,29 @@
...
+
+    const resolvers = {};
+    for (const relationship of relationships) {
+        let resolverOutput: string;
+        const relationTypeName = relationship.relationType.name;
+        const relationIdField = getPrimaryKey(relationship.relationType);
```
wtrocki: Method name here is confusing. Are we simply repurposing PrimaryKey method to get foreignKey?

packages/graphback-codegen-resolvers/src/templates/createResolvers.ts
```diff
@@ -1,8 +1,29 @@
...
+
+        if (relationship.kind === 'oneToMany') {
+            resolverOutput = oneToManyTemplate(relationTypeName, relationship.foreignKey.name, relationIdField.name)
+        } else {
```
wtrocki: My prefference will be to handle other kinds explicitly.

wtrocki:
@craicoverflow Missing docs updates: https://graphback.dev/docs/crud/relationships

wtrocki:
This is also good use case for the integration tests

packages/graphback-codegen-resolvers/src/templates/resolverWrapper.ts
```diff
@@ -45,11 +45,23 @@ function createResolverFileTemplate(outputResolvers: string[], options: Resolver
...
+
     const queries = mapResolverKeyValueTemplates(typeResolvers.Query);
+    delete typeResolvers.Query;
+
     const subscriptions = mapResolverKeyValueTemplates(typeResolvers.Subscription);
+    delete typeResolvers.Subscription;
 
     const outputResolvers: string[] = [];
 
+    for (const [typeName, resolverObj] of Object.entries(typeResolvers)) {
```
wtrocki: Why not use model relationships here?

packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
craicoverflow: Should this be reverted?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -123,17 +123,27 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
-            fields: () => (modelFields.filter((field: GraphQLField<any, any>) => {
-                const fieldBaseType = getBaseType(field.type);
-
-                return !isObjectType(fieldBaseType);
-            }).reduce((fieldObj: any, current: any) => {
+            fields: () => (modelFields.reduce((fieldObj: any, current: any) => {
                 const fieldType = current.type;
-                fieldObj[current.name] = { type: getNullableType(fieldType), description: '' };
+
+                const relationshipField = findRelationship(current.name, model.relationships);
```
wtrocki: Let's refactor this into core so we can simply answer the mapping question - what is the field name of the relationship same as we do for ID etc. Not sure why we need so much extra code here

packages/graphback-codegen-resolvers/src/templates/createResolvers.ts
craicoverflow: We are getting the primary key field for the relation type.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -45,11 +52,17 @@ export class GraphbackCoreMetadata {
...
         //Contains map of the models with their underlying CRUD configuration
         this.models = [];
+        const relationships = [];
         //Get actual user types 
         const modelTypes = this.getGraphQLTypesWithModel();
         for (const modelType of modelTypes) {
-            const model = this.buildModel(modelType)
-            this.models.push(model)
+            const relationshipMetadata = this.buildRelationships(modelType);
+            relationships.push(...relationshipMetadata);
```
wtrocki:
```suggestion
            relationships.push(relationshipMetadata);
```

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -68,11 +81,134 @@ export class GraphbackCoreMetadata {
...
+                const modelPrimaryKey = getPrimaryKey(modelType);
+
+                const manyToOne: RelationshipMetadata = {
+                    parent: relationType,
+                    parentField: dbAnnotations.oneToMany,
+                    kind: 'manyToOne',
+                    relationType: modelType,
+                    relationField: f.name,
+                    foreignKey: {
+                        name: transformForeignKeyName(dbAnnotations.oneToMany, 'to-db'),
```
wtrocki: to-db can be ignored? Optional?

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -68,11 +81,134 @@ export class GraphbackCoreMetadata {
...
+                    }
+                };
+
+                relationships.push(oneToMany, manyToOne);
+            }
+
+            if (dbAnnotations.oneToOne) {
+                const relationType = getBaseType(f.type) as GraphQLObjectType;
+
+                // ignore types that are do not have `@model` marker
```
wtrocki: This is very important thing to document

packages/graphback-codegen-resolvers/src/ResolverGeneratorPlugin.ts
wtrocki: No.. Just the other guy who write code needs to slow down and start reading their own code. So many silly mistakes there :D

packages/graphback-codegen-resolvers/src/templates/createResolvers.ts
wtrocki: Which translateds to getForeginKey name

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
craicoverflow: This wouldn't work. Here we are merging two arrays with spread operator. Without it you would get an array inside an array.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
craicoverflow: I'll make it optional I think.

wtrocki:
Really good prototype. I will work on this to solve:

- Bidirectionality
- Universal mapping helpers in core
- Idenfity what we really need from relationship metadata to minimize amount of the computation we do after (like gettingRelationField etc.)
- simplify plugin extension (especially input type)
- Clear out docs around db migration/plugin core (we have this separted now)
- OneToOne markers? Documentation to explain how to build relationships
- Some testing suite that involves entire ecosystem (integration tests)

wtrocki:
Needs rebase

wtrocki:
Rebase is needed

wtrocki:
@craicoverflow I believe that we have integration tests that were trying to use relationships - however they are only for runtime at the moment. As relationships are quite large feature we might simply have some interation tests in place to test it - similar concept to what we have in datasync repo

packages/graphback-codegen-resolvers/src/templates/resolverTemplates.ts
```diff
@@ -54,12 +54,17 @@ export const deletedSubscriptionTemplate = (modelName: string): string => {
...
+}
 
-export const blankSubscription = `{
-    subscribe: (parent, args, context) => {
-        // Implementation here
-    }
-}`
+export const oneToOneTemplate = (modelName: string, relationField: string, idField: string) => {
+  return `(parent, args, context) => {
+    return context.${modelName}.findBy({ ${idField}: parent.${relationField} })
```
wtrocki: Absolutely love that!

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -55,17 +57,25 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
         if (models.length === 0) {
             this.logWarning("Provided schema has no models. Returning original schema without any changes.")
 
             return schema;
-        }
+        };
+
+        schema = this.buildSchemaModelRelationships(schema, models);
+
+        // Print and rebuild to keep field and type descriptions
```
wtrocki: Do we still need this workaround? If that related with your print schema?
I think we might put that info into issue that we have already as we need to be clear about what is happening.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -55,17 +57,25 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+        // WORKAROUND!
+        schema = buildSchema(printSchema(schema));
 
         const modelsSchema = this.buildSchemaForModels(models);
 
-        return mergeSchemas({ schemas: [modelsSchema, schema] });
+        schema = mergeSchemas({ schemas: [modelsSchema, schema] });
```
wtrocki: Generally this is very convoluted API - we do merge schemas while using graphql compose that should deal with it

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -123,17 +132,17 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
         return new GraphQLInputObjectType({
             name: inputName,
-            fields: () => (modelFields.filter((field: GraphQLField<any, any>) => {
-                const fieldBaseType = getBaseType(field.type);
-
-                return !isObjectType(fieldBaseType);
-            }).reduce((fieldObj: any, current: any) => {
-                const fieldType = current.type;
-                fieldObj[current.name] = { type: getNullableType(fieldType), description: '' };
+            fields: () => (modelFields.filter(canInputField).map((field: GraphQLField<any, any>) => {
```
wtrocki: I found it hard to get this. `canInputField` is not a very descriptive filter.
Offix plugin rebuilds input types (and there will be multiple plugins getting it.
It could be issue to improve the way that we build this and in this PR let's improve naming in core

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable max-lines */
 import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
-import { getBaseType, getFieldName, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, getTableName, getInputTypeName } from '@graphback/core'
+import { getFieldName, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, getInputTypeName, buildGeneratedRelationshipsFieldObject, getInputFieldName, canInputField, getInputFieldType, buildModifiedRelationshipsFieldObject } from '@graphback/core'
```
wtrocki: This crazy big amount of helpers makes me worried. From someone taking a lot on those it will be realy hard to get them around. Plus we could have them namespaces as it is not clear what they involved. I will do wrap this into namespace and create follow up issue to create wrapper class that will give us them. There is no need to have them in core. We can simply have relationship object that offer those helpers

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
```diff
@@ -0,0 +1,334 @@
...
+            throw new Error(`${modelName}.${field.name} relationship field maps to ${relationType.name}.${relationField.name} (${relationFieldBaseType.name} type) which should be ${modelName} type.`);
+        }
+
+        const manyToOneMetadata = parseRelationshipAnnotation(relationField.description);
+
+        if (manyToOneMetadata && oneToManyMetadata.key && manyToOneMetadata.key && oneToManyMetadata.key !== manyToOneMetadata.key) {
+            throw new Error(`${modelName}.${field.name} and ${relationType.name}.${relationField.name} 'key' annotations are different. Ensure both are the same, or remove one so that it can be generated.`);
+        }
+
+        // this.validateManyToOneField(relationType.name, relationField);
```
wtrocki: Why this is commented out?

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
```diff
@@ -0,0 +1,334 @@
...
+
+        if (!isObjectType(relationFieldBaseType) || relationFieldBaseType.name !== modelName) {
+            throw new Error(`${modelName}.${field.name} relationship field maps to ${relationType.name}.${relationField.name} (${relationFieldBaseType.name} type) which should be ${modelName} object type.`);
+        }
+    }
+
+    private validateField(modelName: string, field: GraphQLField<any, any>) {
+        const relationshipAnnotation = parseRelationshipAnnotation(field.description);
+        if (!relationshipAnnotation) {
+            throw new Error(`${modelName}.${field.name}`)
```
wtrocki: So I guess we made a call to not build validation into the engine and each plugin will simply respond to throw errors - this will mean that we meed to close created issue and add some note to the docs

packages/graphback-core/src/relationships/relationshipHelpers.ts
```diff
@@ -0,0 +1,155 @@
+import { parseMarker } from 'graphql-metadata';
```
wtrocki: This place is kinda evil. I will at least namespace those fields.
Any reason why we cannot have this as part of the metadata implementation so we can get that enclosed and available thru single object on model rather than allowing plugin creators to use them directly.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -98,8 +97,37 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
   }
 
   public async batchRead(relationField: string, ids: string[]): Promise<Type[][]> {
-    // TODO
-    return [];
+    let result: any;
+
+    const { idField } = getDatabaseArguments(this.tableMap);
```
wtrocki: Note: Create issue to verify mongodb relationships. Did you swapped datasources

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -150,40 +155,35 @@ export class LayeredRuntimeResolverCreator {
...
-      //}
-
-      //// tslint:disable-next-line: no-any
-      //// TODO - this is very wrong
-      //resolvers[modelName][field.name] = (parent: any, args: any, context: any) => {
-      //return this.service.findBy(field.type.toLowerCase(), { [foreignId]: parent.id }, context);
-      //};
-      //}
-      //}
+  private createRelations(relationships: FieldRelationshipMetadata[]) {
```
wtrocki: Note: Create issue to test runtime relationships.

wtrocki:
Need rebase

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: If I remove this line and, the generated does not keep the new (generated) descriptions, but keeps the ones from the input model.

This issue appears in both `printSchema` and `printSortedSchema`. I will create an issue.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Fully agree, I was thinking this myself

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
craicoverflow: It should have been removed.

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
craicoverflow: Yeah I thought about it and this way seemed better to me. Separating validation from execution means scanning the data model twice. It also means separating closely related operations.

It would be very hard to implement relationship validation into the engine because a user could swap the SchemaCRUDPlugin for their own that uses directives. This makes the pre-validation...invalid.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: 100% agree with this.

Issue #714, opened 2020-02-12
REST based datasource


wtrocki:
WE can create new data source that will allow us to fetch data from the restful layer. See CRUDService for more details.

wtrocki:
Done. Rest DataSource is very generic and missing caching but it is something that will give people overview how things might work

PR #739, opened 2020-02-18
Dev/rest

NomadXD:
REST based datasource #714
Added 2 new files
 * RESTDataProvider.ts
* RESTDataProviderTest.ts


package.json
```diff
@@ -56,5 +56,8 @@
   "bugs": {
     "url": "https://github.com/aerogear/graphback/issues"
   },
-  "homepage": "https://github.com/aerogear/graphback#readme"
+  "homepage": "https://github.com/aerogear/graphback#readme",
+  "resolutions": {
```
wtrocki: Out of curiosity. Why this was required?

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,127 @@
+import { GraphbackDataProvider } from './GraphbackDataProvider';
+const fetch = require('node-fetch');
+
+export class RESTDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext>{
+    // Base url of the endpoint. When creating an RESTDataProvider instance, a string is passed into the constructor.
+    protected baseUrl : string;
+
+    //constructor
+    public constructor(baseUrl:string) {
```
wtrocki:
```suggestion
    public constructor(type: GraphqlObjectType, baseUrl:string) {
```

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param data - JSON object which carries the data needed for the create operation
+     * @param context - Context carries the path needed to create a valid endpoint url
+     * 
+     * valid url => (baseUrl+path) 
+     * baseUrl is defined at instance creation via the constructor. path is passed at runtime via 
+     * the context parameter.
+     * 
+     */
+    async create(data: Type, context?: GraphbackContext): Promise<Type> {
+        const url = this.baseUrl+`/${context!==undefined ? context : ""}`
```
wtrocki:
```suggestion
        const url = `${this.baseUrl}/${this.type.name.toLowerCase()}`   
```

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+     * baseUrl is defined at instance creation via the constructor. path is passed at runtime via 
+     * the context parameter.
+     * 
+     */
+    async create(data: Type, context?: GraphbackContext): Promise<Type> {
+        const url = this.baseUrl+`/${context!==undefined ? context : ""}`   
+        const res = await fetch(url, {
+            method: 'post',
+            body:    JSON.stringify(data),
+            headers: { 'Content-Type': 'application/json' },
```
wtrocki: It should allow peopel to pass headers as well.

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param data - JSON object which carries the data needed for the update operation
+     * @param context -Context carries the path needed to create a valid endpoint url
+     * 
+     * valid url => (baseUrl+path) 
+     * baseUrl is defined at instance creation via the constructor. path is passed at runtime via 
+     * the context parameter.
+     * 
+     */
+    async update(data: Type, context?: GraphbackContext): Promise<Type> {
+        const url = this.baseUrl+`/${context!==undefined ? context : ""}`+`/${data['id']}`
```
wtrocki: Same pattern as above

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface Project {
+    category_id: number;
+    name: string;
+}
+
+const test = _test as TestInterface<Context>;
+
+test.beforeEach(async t => {
+    const provider = new RESTDataProvider('http://64.225.124.94:7000/api');
```
wtrocki: I would use something public for testing. SWAPI?

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
```diff
@@ -0,0 +1,62 @@
...
+    const result = await t.context.provider.create(project_category,'category_manage/category')
+    console.log("create",result)
+    t.assert(result['data']['affectedRows'] === 1)
+})
+
+
+// test for delete.
+test('delete a project category',async t => {
+    const data = {category_id:16};
+    const result = await t.context.provider.delete(data,'category_manage/category')
```
wtrocki: Context here comes from resolvers and it will not have ability to pass some urls.
We will eather do:

- Use type name and some follow up rest
- Let people to specify array of the methods that will be mapped (but then why we need this)

examples/generator-fullstack/.graphqlrc.yml
```diff
@@ -1,63 +1,31 @@
-schema: ./server/src/schema/*.graphql
-documents: ./client/src/graphql/**/*.graphql
+## GraphQL Config Generated by Graphback
```
wtrocki: git reset --soft on this and remove it from commit as anything from it should not be commited. Also rebase is needed.

package.json
NomadXD: @wtrocki when I run yarn install from root after cloning, yarn resolved two versions of `@types/react .`  And `@types/react-dom`  also has a dependency to `@types/react`  but it depends on any of the versions of `@types/react` . So yarn tries to resolve 2 versions of `@types/react` and a conflict occurs. So package.json specifies which version of `@types/react` to resolve.
[https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes](stackoverflow)

I tried some other workarounds like deduplicating yarn.lock but nothing worked for me. Anyway it's a issue with my local machine. I was able to build it in my server without resolutions in package.json. So you are correct. I shouldn't have commited that line. Sorry

wtrocki: No need to be sorry. I would say it is valid scenario. Create PR with just this change. I will merge it.

packages/graphback-runtime/src/data/RESTDataProvider.ts
NomadXD: @wtrocki So we can add headers to data object passed into the method and access headers from that. Right ?

packages/graphback-runtime/src/data/RESTDataProvider.ts
NomadXD: @wtrocki Will fix it.

packages/graphback-runtime/src/data/RESTDataProvider.ts
wtrocki: Constructor might work (but not always as we might know token later)

package.json
NomadXD: Just want to clarify something about that pull request. What do you mean by "just this change ?"

1. Do i have to checkout to a new branch, remove resolutions from package.json , commit, push and pull request ? 

2. Or checkout to new branch cherry pick a commit with package.json unedited and without any other commits ? So it's like right after yarn install without the resolution part in package.json.  "Just this change" sounds like this scenario but it doesn't make sense because it's like a PR with fresh built.

wtrocki: Please create separate PR and copy paste this 3 lines of code into it. No need for got magic. This PR on the other hand needs some git work. I would squash commits and then redo this single commit to contain only relevant changes

NomadXD: Okay just copied 3 lines and created a pull request. What about the changes you requested from yesterday's review ? Shall I create another pull request with all the suggested changes ? So it will be easier for you to merge that PR instead of this ?

wtrocki:
Can be closed now as we have other PR

PR #742, opened 2020-02-19
resolutions added in package.json

NomadXD:
@wtrocki Only 3 lines added in package.json.


.gitignore
```diff
@@ -12,6 +12,7 @@ types/
 # Dependencies
 node_modules/
 package-lock.json
+examples/
```
wtrocki:
```suggestion
```

PR #748, opened 2020-02-19
REST datasource  

NomadXD:
REST based datasource #714

changes requested in previous PR

### DIFF
--- a/packages/graphback-runtime/src/data/RESTDataProvider.ts
+++ b/packages/graphback-runtime/src/data/RESTDataProvider.ts
@@ -1,0 +1,129 @@
+import { GraphbackDataProvider } from './GraphbackDataProvider';
+import { GraphQLObjectType } from 'graphql';
+const fetch = require('node-fetch');
+
+export class RESTDataProvider<Type = any, GraphbackContext = any> implements GraphbackDataProvider<Type, GraphbackContext>{
+    // Base url of the endpoint. When creating an RESTDataProvider instance, a string is passed into the constructor.
+    protected baseUrl : string;
+    protected baseType : GraphQLObjectType;
+    protected headers : any;
+
+    //constructor
+    public constructor(baseType: GraphQLObjectType, baseUrl:string, headers:any) {
+      this.baseUrl = baseUrl;
+      this.baseType = baseType;
+      this.headers = headers;
+    }
+
+    /** 
+      @param data - JSON object which carries the data needed for the update operation
+     * @param context -passed by the resolvers
+     * 
+     * valid url => (baseUrl+path) 
+     * baseUrl is defined at instance creation via the constructor. path is created from baseType name and other passed values.
+     * 
+     */
+    async create(data: Type, context?: GraphbackContext): Promise<Type> {
+        console.log(this.baseType.name.toLocaleLowerCase())
+        const url = `${this.baseUrl}/${this.baseType.name.toLowerCase()}/`   
+        const res = await fetch(url, {
+            method: 'post',
+            body:    JSON.stringify(data),
+            headers: this.headers,
+        })
+        const json = await res.json();
+        return json;
+    }
+
+
+    /**
+     * 
+     * @param data - JSON object which carries the data needed for the update operation
+     * @param context -passed by the resolvers
+     * 
+     * valid url => (baseUrl+path) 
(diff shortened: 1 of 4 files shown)


packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+     * 
+     * @param data - JSON object which carries the data needed for the update operation
+     * @param context -passed by the resolvers
+     * 
+     * valid url => (baseUrl+path) 
+     * baseUrl is defined at instance creation via the constructor. path is created from baseType name and other passed values.
+     * 
+     */
+    async update(data: Type, context?: GraphbackContext): Promise<Type> {
+        const url = this.baseUrl+`/${this.baseType.name.toLocaleLowerCase}`+`/${data['id']}`
```
wtrocki: We need to use mapping for id same as in knex. `id` as id is not guaranteed here

wtrocki: Also it looks like we forgot about calling method, which makes me think about unit tests handling this case :D

```suggestion
        const url = this.baseUrl+`/${this.baseType.name.toLocaleLowerCase()}`+`/${data['id']}`
```

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+ @param data - JSON object which carries the data needed for the update operation
+     * @param context -passed by the resolvers
+     * 
+     * valid url => (baseUrl+path) 
+     * baseUrl is defined at instance creation via the constructor. path is created from baseType name and other passed values.
+     * 
+     * 
+     */
+    async delete(data: Type, context?: GraphbackContext): Promise<Type> {
+        const url = this.baseUrl+`/${this.baseType.name.toLocaleLowerCase()}/${data['id']}`
```
wtrocki:
```suggestion
        const resourceName = this.baseType.name.toLocaleLowerCase();
        const id = ${data['id']
        const url = this.baseUrl+`/${resourceName}/${id}` 
```

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
```diff
@@ -0,0 +1,76 @@
...
+test('create a person',async t => {
+    const person = {name:'JBossss',job:'Software engineer'}
+    const result = await t.context.provider.create(person)
+    console.log("create",result)
+    t.assert(result.name === person.name)
+})
+
+
+//test for delete.
+// test('delete a person with id',async t => {
```
wtrocki: ?

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
```diff
@@ -0,0 +1,76 @@
...
+const modelType = schema.getType('Users') as GraphQLObjectType
+
+test.beforeEach(async t => {
+    const provider = new RESTDataProvider(modelType,'https://reqres.in/api',{'Content-Type': 'application/json'});
+    t.context = { provider };
+});
+
+/**
+ * 
+ * Currently I just have 3 endpoints in my hosted server. So I tested only
```
wtrocki: SWAPI is standard for API. https://swapi.co/documentation
Let's use that

packages/graphback-runtime/src/data/RESTDataProvider.ts
NomadXD: @wtrocki honestly, I didn't look much into how the data is passed. I will see more about how data is passed and the CRUD service is used at run time and implement a mapping.

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
NomadXD: Because I was using https://reqres.in/  API and they didn't allow the delete operation.

packages/graphback-runtime/tests/data/RESTDataProviderTest.ts
NomadXD: Yeah I was gonna use that for testing. But they only provided the read operation. Thats why I used the other one.

wtrocki:
Good work overall. build is currently failing. Ping me again for final review

PR #749, opened 2020-02-19
dbmigrations extension

### LINKED ISSUE #737: Db migrations as separate extension in graphqlrc.yml configuration
When doing this we can stop basing on the model and use output schema that will be decorated.

cecobask:
- I changed how the schema is loaded. Initially was loaded from model directory, now - from graphqlrc.yml schema path.

- Separated dbmigrations plugin from graphback extension, into its own extension.

Am I going in the right direction with this? Fix #737


packages/graphback-cli/src/components/db.ts
```diff
@@ -13,57 +14,62 @@ const handleError = (err: { code: string; message: string; }): void => {
...
+      throw new Error(`You should provide a valid '${dbmigrationsExtension}' config to migrate schema`);
     }
 
     if (!graphbackConfig.model) {
-      throw new Error(`'${extensionName}' config missing 'model' value that is required`);
+      throw new Error(`'${graphbackExtension}' config missing 'model' value that is required`);
     }
 
-    const models = new GlobSync(`${graphbackConfig.model}/*.graphql`)
+    const models = new GlobSync(`${graphbackConfig.model}/*.graphql`);
```
wtrocki: So I had this idea that if we migrate dbMigrationConfig then we do not need graphbackConfig for migration. 

Why? We can use the top-level schema as input instead. config.getSchema() method will be much better here as db migration package is not specific to graphback and it can be used outside graphback to do database migrations

packages/graphback-cli/src/components/db.ts
```diff
@@ -13,57 +14,62 @@ const handleError = (err: { code: string; message: string; }): void => {
...
-    databaseOperations = await migrateDB(graphbackConfig.dbmigrations, schemaText, migrateOptions);
+    const schemaText = loadSchema(project.schema.toString());
+    databaseOperations = await migrateDB(dbMigrationConfig, schemaText, migrateOptions);
   } catch (err) {
     handleError(err)
   }
 
   return databaseOperations;
-}
+};
```
wtrocki: Webstorm? :>

packages/graphback-cli/src/components/db.ts
```diff
@@ -13,57 +14,62 @@ const handleError = (err: { code: string; message: string; }): void => {
...
 
     const migrateOptions: MigrateOptions = {
       //Do not perform delete operations on tables
       operationFilter: removeNonSafeOperationsFilter
-    }
+    };
 
-    const schemaText = loadSchema(graphbackConfig.model);
-    databaseOperations = await migrateDB(graphbackConfig.dbmigrations, schemaText, migrateOptions);
+    const schemaText = loadSchema(project.schema.toString());
```
wtrocki: This is no longer needed. We can get schema from the project directly.

packages/graphback-cli/src/components/db.ts
```diff
@@ -13,57 +14,62 @@ const handleError = (err: { code: string; message: string; }): void => {
...
+    databaseOperations = await migrateDB(dbMigrationConfig, schemaText, migrateOptions);
```
wtrocki: migrateDb actually should allow both text and compiled schema. 

```
export async function migrateDB(
  config: Knex.Config,
  schema: string | GraphQLSchema,
  options: MigrateOptions = {},
): Promise<MigrationResults> {
```
This way api will be much simpler and we will not need to transform schema back and forth.
CC @craicoverflow

packages/graphback-cli/src/components/db.ts
wtrocki: There is example for schema loading from config in generate

packages/graphback-cli/src/components/db.ts
cecobask: I will remove graphbackConfig from this file. Wasn't sure whether I need to keep it, but now I know

packages/graphback-cli/src/components/db.ts
cecobask: It was a suggestion from WS, yes :D Does it clash with any ESLint rule? I didn't get a warning about that.

wtrocki:
World class job!  Minor fixes suggested. There will be some corrections in docs as well. 
Thank you so much for getting this done in such short timeframe

packages/graphback-cli/src/components/db.ts
wtrocki: It looks like eslint is not giving out about it but we do not use them :D

cecobask:
I couldn't get it to work with project.loadSchema(). I was running into this issue:

wtrocki:
This error might mean something wrong with the graphqlrc.yml file or generated schema is empty. We are missing config yml file changes for templates.

wtrocki:
Awesome work @cecobask. I will need to update docs so merging this PR and adjusting changes in separate PR!

docs/db/migrations.md
```diff
@@ -46,7 +46,6 @@ migrateDB(dbConfig, schemaText, {
   - `scalarMap`: Custom scalar mapping. Default: `null`.
   - `mapListToJson`: Map scalar lists to JSON column type by default.
   - `debug`: display debugging information and SQL queries.
-  - `removeDirectivesFromSchema`: removes all directives from the GraphQL schema.
```
wtrocki: @craicoverflow FYI we have killed this as it was just leftover from using CRUD markers as directives

docs/intro/config.md
```diff
@@ -31,7 +31,7 @@ Typical configuration looks as follows
 schema: ./server/src/schema/*.graphql
 ## Documents are not used by graphback but it is required by graphql-config and can be used by other extensions
 ## Usually it should match the Graphback client generator plugin
-documents: ./client/src/graphql/**/*.graphql
+documents: ./client/src/graphql/*.graphql
```
wtrocki: WHY this change?

templates/ts-apollo-fullstack/.graphqlrc.yml
```diff
@@ -1,38 +1,35 @@
...
-        outputPath: ./server/src/resolvers
-    ## Knex DB Migration config
+        outputPath: ./src/resolvers
+      graphback-client:
+        format: graphql
+        outputPath: ./client/src/graphql
   dbmigrations:
-    ## See knex.js for db specific config format
-    client: pg
     connection:
```
wtrocki:
```suggestion
     ## See knex.js for db specific config format
    connection:
```

docs/intro/config.md
cecobask: That's the way `graphback config` command generated the graphqlrc.yml file. I wasn't sure either.

wtrocki: Ok.. So if we check generated files the original format is right - bug in config :D

wtrocki:
This will be merged after build will pass

wtrocki:
Rebase needed

wtrocki:
Resolved in separate PR. Let's test template apps next time

PR #761, opened 2020-02-21
Add react example to Graphback fullstack template

kingsleyzissou:


wtrocki:
Rebased. going to improve it over the comming days

wtrocki:
Spent some time trying to refine this and I think there is simply too much work involved:

- Todo model can be easily broken - Our client app needs to be dynamically generated on some specific object
- App is in JS
- Offix makes things very very specific and it will make absolutely hard to use it in generic context. Removing offix is too much work. It will be easier for us to integrate with existing boilerplate react typescript app etc.

wtrocki:
To be clear. App seems amazing.. I just did not anticipated how much work will be needed to get this running

wtrocki:
If we have master detail with pluggable offix we will be much ready. no sense to spent effort to migrate app, but we can use this wireframes later

PR #766, opened 2020-02-21
Load schema specified in graphqlrc.yml

wtrocki:


wtrocki:
@cecobask I needed to squash commits as there were multiple conflicts in the same file.
Also there were many issues in the PR: Runtime template were not updated.
I have pushed those commits on top so you can check what was there

wtrocki:
I think we should move those configs into graphback so people are able to read them without specifying them directly in the template apps

PR #770, opened 2020-02-22
Mysql support

udith-shalinda:


packages/graphback-runtime-knex/src/MysqlDBKnexDataProvider.ts
```diff
@@ -0,0 +1,50 @@
+import { getDatabaseArguments } from '@graphback/core';
+import { GraphQLObjectType } from 'graphql';
+import * as Knex from 'knex';
+import { KnexDBDataProvider } from './KnexDBDataProvider';
+import { NoDataError } from './NoDataError';
```
craicoverflow: Getting a build error on this line:

```
src/MysqlDBKnexDataProvider.ts(5,29): error TS2307: Cannot find module './NoDataError'.

It should be imported from `@graphback/runtime`
```

packages/graphback-runtime-knex/src/MysqlDBKnexDataProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        const { data: createData } = getDatabaseArguments(this.tableMap, data);
+
+        // check availability of the table
+        const result = await this.db(this.tableName).columnInfo().then();
+
+        if(Object.keys(result).length == 0){
+            //creating table
+            const tableRes:any = await this.db.schema.createTable(this.tableName, function (table) {
+                table.increments();
```
craicoverflow: This is the data persistence layer and the database structure should never be modified here.

wtrocki:
Work for the issues was to verify if database migrations supporting MySQL. 
What I see here is that we tried to replicate lack of the database migrations inside data source provider - which seems to try to work around the issue? I would suggest opening a conversation on the issue.

wtrocki:
As team we made call that bringing mysql support will require a lots of work and we do not want to support it for the moment. Thank you so much for contribution

Issue #738, opened 2020-02-18
Add limit/offset to findAllX crud call


wtrocki:
We need to add limit/offset as first step before we apply a pagination container.

ssd71:
I would love to take this!

ssd71:
Can you give me some top tips to get started?

wtrocki:
Check `GraphbackDataProvider`

ssd71:
Thanks! I'll get started right away

ssd71:
I looked around a bit, and I think that to add pagination we would have to add offset in all of the data provider implementations (Is this right?) and to the interface itself. Sorry if I'm getting it wrong
```
diff --git a/packages/graphback-runtime/src/data/GraphbackDataProvider.ts b/packages/graphback-runtime/src/data/GraphbackDataProvider.ts
index 146d1db..f85dab1 100644
--- a/packages/graphback-runtime/src/data/GraphbackDataProvider.ts
+++ b/packages/graphback-runtime/src/data/GraphbackDataProvider.ts
@@ -52,7 +52,7 @@ export interface GraphbackDataProvider<Type = any, GraphbackContext = any> {
      * @param id id of the object
      * @param context context object passed from graphql or rest layer
      */
-    findAll(context?: GraphbackContext): Promise<Type[]>;
+    findAll(offset?: Number, context?: GraphbackContext): Promise<Type[]>;
 
     /**
      * Implementation for reading objects with filtering capabilities
```

wtrocki:
We could start with separate methods that only some implementations will apply and then introduce pagination container for it. Generally concept here was to introduce pagination gradually as there is separat issue to add pagination that might be little bit too much as good first issue.

wtrocki:
See https://github.com/aerogear/graphback/pull/622

ssd71:
I understand. So I will proceed with adding both offset & limit in database (data provider implementations) first and then open a draft PR, would that be okay?

wtrocki:
Yep. We are totally against asking community to do complex implementation. Ideally we can just extend mongo and knex providers just for find all queries

ssd71:
Understood! Will make a PR as soon as I can!

ssd71:
This is what I got working so far
```
diff --git a/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts b/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
index 722e03a..c394369 100644
[...]

PR #788, opened 2020-02-25
Add limit and offset to findAll methods in DB layer

ssd71:
Addresses #738 
This PR consists of the following changes:

- Create a findMore function in the Graphback Data Provider class that can be used to implement pagination
- Implement above function in Knex, REST and MongoDB data providers
- Write tests for forementioned functionality


packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -81,6 +81,15 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
     throw new NoDataError(`Cannot find all results for ${this.tableName}`);
   }
 
+  public async findMore(offset: number, limit: number): Promise<Type[]> {
```
wtrocki: Changed my mind. If we keep those args optional we can extend existing function without breaking API.

wtrocki:
Also it might need change in resolvers https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-resolvers/src/templates/resolverTemplates.ts#L23

After that we will be able to test it

ssd71:
I have modified the findAll resolver to use findMore when limit and offset args are specified in the query. Please check

packages/graphback-codegen-resolvers/src/templates/resolverTemplates.ts
```diff
@@ -20,7 +20,11 @@ export const deleteTemplate = (modelName: string): string => {
 
 export const findAllTemplate = (modelName: string): string => {
   return `(parent, args, context) => {
-      return context.${modelName}.findAll();
+      if (args.offset && args.limit){
+        return context.${modelName}.findMore(args.offset, args.limit);
```
wtrocki: Hi
This is not going to work. This is executed at the time of the code generation. We should always use findAll like this
`return context.${modelName}.findAll(args);`

packages/graphback-runtime/src/data/GraphbackDataProvider.ts
```diff
@@ -54,6 +54,15 @@ export interface GraphbackDataProvider<Type = any, GraphbackContext = any> {
...
     findAll(context?: GraphbackContext): Promise<Type[]>;
 
+    /**
+     * Implementation for finding all objects with support for pagination
+     * 
+     * @param limit Limit on number of objects to return 
+     * @param offset Number of objects to exclude starting from the first
+     * @param context context object passed from graphql or rest layer 
+     */
+    findMore(offset: number, limit: number, context?: GraphbackContext): Promise<Type[]>
```
wtrocki: Let's remove this method and use find All with new type:

`GraphbackPage: {limit, offset}`

packages/graphback-codegen-resolvers/src/templates/resolverTemplates.ts
ssd71: Understood

packages/graphback-runtime/src/data/GraphbackDataProvider.ts
ssd71: On it

ssd71: Where do you think I should put the GraphbackPage type?

ssd71:
I have added the `GraphbackPage` type to `GraphbackDataProvider` and have modified all implementations to use it such that if no page is specified, findAll will return all records. Please check. 


Also I am facing some difficulties in writing tests for the MongoDB provider implementation. It seems that the database records persist between tests. I am confused if it was meant to be like that.

wtrocki:
> Also I am facing some difficulties in writing tests for the MongoDB provider implementation. It seems that the database records persist between tests. I am confused if it was meant to be like that.

Yes. I seen that as well so adjusted tests to workaround that but if we want to do things like limit it might be challenging. The best way might be to use the serial interface for tests or use after each to wipe out collection

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -72,9 +72,10 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
 
   }
 
-  public async findAll(): Promise<Type[]> {
+  public async findAll(page?: GraphbackPage): Promise<Type[]> {
     //tslint:disable-next-line: await-promise
-    const dbResult = await this.db.select().from(this.tableName);
+    const dbResult = page ? (await this.db.select().from(this.tableName).limit(page.limit).offset(page.offset)):
```
wtrocki:
```suggestion
    let query = this.db.select().from(this.tableName)
    if(page){
    	query = query.limit(page.limit).offset(page.offset)
    }
    const dbResult = await query:
```

wtrocki:
let's see if we can improve unit tests and extend the schema to accept those 2 arguments. We will also need to modify docs "CRUD Spec" to cover new changes

ssd71:
I understand. I think we should write some tests first, to preserve some sanity :P
Could you point me at something I can get started with?

ankitjena:
> Edit: Is it the graphback-codegen-schema?
@ssd71 Yes

ssd71:
I ran the tests for `graphback-codegen-schema` and I got `Warning - SchemaCRUD: Provided schema has no models. Returning original schema without any changes.` and the resulting generated graphql schema didn't have any CRUD fields (it simply returned the input schema i.e. mock.graphql)

I checked `mock.graphql` and the Note and Comment types didn't have the `"""@model"""` annotation. On adding these in and then generating the schema I got all the CRUD fields in resulting schema.

wtrocki:
Good catch. We missed that. As for schema please see https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L257

wtrocki:
When all done we should be able to start server and test things

ssd71:
I tried some things and I came up with this:
```
type Query {
      .
      .
      findAllNotes(limit: Int!, offset: Int!): [Note]!
      .
      .
      findAllComments(limit: Int!, offset: Int!): [Comment]!
      .
      .
}
```
Is this acceptable? I thought that making another 'Page' gql type would be overkill

wtrocki:
@ssd71 Awesome! This actually more clever. Rease why I asked for top level type is that there is problem as those two values should be provided together - without that knex will fail on making query.

If we want to do it without wrapper then code should change to figure out defaults. If offset is passed we should pick some default limit like 10. If Limit is passed offset should be 0.
 
I think that we can drop `!` as well!

Amazing work!

ssd71:
Would you like me to change the Data Provider implementations? I believe there should be a better way to do it, but I can't see default arguments anywhere on the gql layer

wtrocki:
@ssd71 If we agree that this format is better then yes - we need to drop `!` and supply defaults

ssd71:
I have done the following:
- Add limit and offset args to findAllX queries in schema generator
- Add default arguments for limit and offset in all Data Provider implementations
- Write some tests for this (as best as I could, anyway :P)

While running the tests, I encountered a problem: `CRUDService.ts`in `graphback-runtime` calls findAll like so:
```
return this.db.findAll(context);
```
Which conflicts with the current spec of findAll:
```
findAll(page?: GraphbackPage, context?: GraphbackContext): Promise<Type[]>;
```

This resulted in the findAll test of `DefaultCRUDServiceTest.ts` in `graphback-runtime-knex` to fail as in that file, findAll is being passed a string. This doesn't fit well with checking the fields of the page argument to fill in default values for missing fields. In the meantime, I have implemented some roundabout way to pass that test by interpreting strings as empty argument and returning all objects:
In `KnexDBDataProvider.ts`
```
if(page && (page.limit || page.offset)){
      if (!page.limit) {
        page.limit = 10;
      }
      if (!page.offset) {
        page.offset = 0;
      }
    	query = query.limit(page.limit).offset(page.offset)
    }
```

I also noticed that there are no CRUD Service Tests for Mongo services.

wtrocki:
@ssd71 This is by far the best community contribution that we recieved in Graphback so far. Thank you so much for not only resolving such challenging issues but also managing to do this end to end.

I will take it from there and play with the implementation and try to see if there are some issues or corner cases and get that merged. 
I will also merge that without squashing commits so you are getting all 27 commits as contributions to graphback! Thank you so much again - this is world class job and manifest of your skills!

ssd71:
Thank you so much! It was an absolute pleasure contributing, thank you @wtrocki  and @ankitjena and everyone else for making this such a wonderful community! :)

packages/graphback-runtime/src/data/RESTDataProvider.ts
```diff
@@ -109,15 +109,17 @@ export class RESTDataProvider<Type = any, GraphbackContext = any> implements Gra
...
-    public async findAll(context?: GraphbackContext): Promise<Type[]> {
+    public async findAll(page?: GraphbackPage): Promise<Type[]> {
         // eslint-disable-next-line prefer-template
         const url = this.baseUrl+`/${this.baseType.name.toLocaleLowerCase()}/`
         const res = await fetch(url)
         const json = await res.json()
+        console.log((json));
 
-        return json
+        return page ? {...json, data: json.data.slice((page.offset || 0), (page.offset || 0) + (page.limit || 10))} : json;
```
wtrocki: I will remove this code.

wtrocki:
@ssd71 Closing the PR as I see a couple of bugs that I need to fix.

wtrocki:
Merged

PR #789, opened 2020-02-25
fix: Offix version bump

wtrocki:
Add Versioning changes to offix. When updating we need to bump version.
There is still bug about updates. I think offix need to have by definition an soft updates so we can introduce `_deleted` field


craicoverflow:
It looks like you haven't included the version bump commit/file.

Issue #791, opened 2020-02-26
Type error in apollo full stack template


NVS16:
The two state variables in the App component of the apollo full stack template are not initialized to empty strings which raises the following error - 
Type 'undefined' is not assignable to type 'string'.  TS2322

PR #792, opened 2020-02-26
fix: initialized state variables

NVS16:
It addresses the issue #791 
The two state variables were initialized to empty strings.


wtrocki:
Thank you so much for your contribution!

PR #795, opened 2020-02-27
fix: runtime was not using dbmigrations config

craicoverflow:
The runtime app was trying to load config from `graphback` config extension.


wtrocki:
CC @cecobask Some small fix added

PR #798, opened 2020-02-27
fix: incorrect return value from offix monogdb data provider

kingsleyzissou:
- Fix: `OffixMongoDBDataProvider` was returning the previous value rather than the updated value in the update query.


packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -39,8 +39,8 @@ export class OffixMongoDBDataProvider<Type = any, GraphbackContext = any> extend
       data.version = data.version + 1;
       // TODO use findOneAndUpdate to check consistency afterwards
       const result = await this.db.collection(this.collectionName).updateOne({ _id: new ObjectId(idField.value) }, { $set: data });
-      if (result) {
-        return queryResult[0];
+      if (result.result.ok) {
```
wtrocki:
```suggestion
      if (result.result?.ok) {
```

Issue #757, opened 2020-02-21
Graphback Config should ask for different databases


wtrocki:
When doing config we should utilize our datasources intergration:
See
https://github.com/aerogear/graphback/pull/746

wtrocki:
We need to ask for 
- MongoDB (and provide default docker-compose
- Postgres (Nothing changes here)

We not going to support other databases.

Install different runtime package based on the choice (currently installs only @graphback/runtime - if should @graphback/runtime-mongo or @graphback/runtime-knex

DataSources documentation should be also updated

wtrocki:
@craicoverflow This is blocker for relase. If not done by commmunity we will need to do it ourselves.

NVS16:
Can I work on this issue?

wtrocki:
@NVS16 Perfect. Copy some files from our example app:
https://github.com/aerogear/datasync-starter/blob/master/server/docker-compose.yml

NVS16:
Ok. Working on it.

ankitjena:
> Install different runtime package based on the choice (currently installs only @graphback/runtime - if should @graphback/runtime-mongo or @graphback/runtime-knex

@wtrocki We need to have changes in the templates too since they have this as default
```
import { createKnexPGCRUDRuntimeServices } from "@graphback/runtime-knex"
```

wtrocki:
Yes. There is issue to document that but I'm currently occupied by finishing code changes

wtrocki:
See: https://github.com/aerogear/graphback/issues/756

Swapping datasource involves changes in the template which is tricky to do. 
What we can do is to have some file in template that initializes proper datasource. Like datasources.ts and replace it with the cli based on choice. 
This will make it hard to build new templates but we are not woried about it

NVS16:
I had some doubts regarding writing the documentation and wanted your help.

```
I guess we need to at least add some info here:
https://graphback.dev/docs/intro/existingproject
and here:
https://graphback.dev/docs/db/datasources
```

in datasources - We add that we now support mongodb (and remove sqlite3 and mysql) 
in the first link - 
We have to add what a user needs to manually do to change his code so as to comply with postgres or mongo, whatever he chooses.
As far as I can see the user will have to modify these two files in the template - 
1. db.ts
2. graphql.ts 
(Are there anymore files to change?)
I believe I can use the files here - 
https://github.com/aerogear/datasync-starter/blob/master/server/src/db.ts
[...]

PR #800, opened 2020-02-27
Added option for mongodb in graphback config

NVS16:
Addresses #757 
Updates - 
1. Added the option of MongoDB (and removed the option of sqlite3) in graphback config command 
2. Added the files mongodb.yml and mongodb.json for docker-compose of mongodb 
3. Added - Installing different runtime package based on the choice - @graphback/runtime-mongo or @graphback/runtime-knex 
Command `graphback db` failed with - 
`knex: Unknown configuration option 'client' value MongoDB. Note that it is case-sensitive, check documentation for supported values.` 
@ankitjena pointed out the reason for it.
Working on it.

### DIFF
--- a/packages/graphback-cli/src/components/config.ts
+++ b/packages/graphback-cli/src/components/config.ts
@@ -13,11 +13,12 @@ export async function installDependencies(database: string): Promise<void> {
   await execa('npm', ['i'])
   if (database === 'pg') {
     await execa('npm', ['i', '-S', 'pg'])
-  } else if (database === 'sqlite3') {
-    await execa('npm', ['i', '-S', 'sqlite3'])
+    await execa('npm', ['i', '-S', '@graphback/runtime-knex'])
+  } else if (database === 'MongoDB') {
+    await execa('npm', ['i', '-S', 'mongodb'])
+    await execa('npm', ['i', '-S', '@graphback/runtime-mongo'])
   }
   //Using opiniated layer
-  await execa('npm', ['i', '-S', '@graphback/runtime-knex'])
 }
 
 function postSetupMessage(): string {
--- a/packages/graphback-cli/src/components/config.ts
+++ b/packages/graphback-cli/src/components/config.ts
@@ -18,7 +18,6 @@ export async function installDependencies(database: string): Promise<void> {
     await execa('npm', ['i', '-S', 'mongodb'])
     await execa('npm', ['i', '-S', '@graphback/runtime-mongo'])
   }
-  //Using opiniated layer
 }
 
 function postSetupMessage(): string {
--- a/packages/graphback-cli/src/components/db.ts
+++ b/packages/graphback-cli/src/components/db.ts
@@ -35,10 +35,10 @@ export const createDBResources = async (cliFlags: { project?: string }): Promise
       process.exit(0)
     }
 
-    // if (dbMigrationConfig.client === 'sqlite3') {
-    //   logInfo(`SQLLite database will need to be recreated from scratch with every migration. Please manually remove/backup existing SQLLite database file before migration`);
-    //   await execa('touch', [dbMigrationConfig.connection.filename])
-    // }
+    if (dbMigrationConfig.client === 'sqlite3') {
+      logInfo(`SQLLite database will need to be recreated from scratch with every migration. Please manually remove/backup existing SQLLite database file before migration`);
+      await execa('touch', [dbMigrationConfig.connection.filename])
+    }
 
     const migrateOptions: MigrateOptions = {
       //Do not perform delete operations on tables
--- a/packages/graphback-cli/src/components/db.ts
+++ b/packages/graphback-cli/src/components/db.ts
@@ -12,7 +12,7 @@ const handleError = (err: { code: string; message: string; }): void => {
     logError(err.message)
   }
   process.exit(0)
-};
+}; 
 
 export const createDBResources = async (cliFlags: { project?: string }): Promise<any> => {
   let databaseOperations: any;
@@ -35,10 +35,10 @@ export const createDBResources = async (cliFlags: { project?: string }): Promise
       process.exit(0)
     }
 
-    if (dbMigrationConfig.client === 'sqlite3') {
-      logInfo(`SQLLite database will need to be recreated from scratch with every migration. Please manually remove/backup existing SQLLite database file before migration`);
-      await execa('touch', [dbMigrationConfig.connection.filename])
-    }
+    // if (dbMigrationConfig.client === 'sqlite3') {
+    //   logInfo(`SQLLite database will need to be recreated from scratch with every migration. Please manually remove/backup existing SQLLite database file before migration`);
+    //   await execa('touch', [dbMigrationConfig.connection.filename])
+    // }
 
     const migrateOptions: MigrateOptions = {
       //Do not perform delete operations on tables
--- a/packages/graphback-cli/src/templates/configTemplates.ts
+++ b/packages/graphback-cli/src/templates/configTemplates.ts
@@ -15,7 +15,7 @@ const getConfig = (database: string) => {
   if (database === 'pg') {
     return [readFileSync(`${configFilesPath}/postgres.json`, 'utf8'), readFileSync(`${dockerFilesPath}/postgres.yml`, 'utf8')]
   } else if (database === 'MongoDB') {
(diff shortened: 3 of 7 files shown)


wtrocki:
Ok.. So when choosing mongo we need to remove GraphQL migrate section. Mongo do not need it at all as it is no SQL dB.

packages/graphback-cli/src/components/config.ts
```diff
@@ -13,11 +13,12 @@ export async function installDependencies(database: string): Promise<void> {
   await execa('npm', ['i'])
   if (database === 'pg') {
     await execa('npm', ['i', '-S', 'pg'])
-  } else if (database === 'sqlite3') {
-    await execa('npm', ['i', '-S', 'sqlite3'])
+    await execa('npm', ['i', '-S', '@graphback/runtime-knex'])
+  } else if (database === 'MongoDB') {
+    await execa('npm', ['i', '-S', 'mongodb'])
```
wtrocki: WOW WE FOUND BUG. Both knex and mongo runtime packages have now Pg and mongo. They do not need them and we can have them as peer dependencies

packages/graphback-cli/src/components/db.ts
```diff
@@ -35,10 +35,10 @@ export const createDBResources = async (cliFlags: { project?: string }): Promise
       process.exit(0)
     }
 
-    if (dbMigrationConfig.client === 'sqlite3') {
-      logInfo(`SQLLite database will need to be recreated from scratch with every migration. Please manually remove/backup existing SQLLite database file before migration`);
-      await execa('touch', [dbMigrationConfig.connection.filename])
-    }
+    // if (dbMigrationConfig.client === 'sqlite3') {
```
wtrocki: Let's keep this for our unit tests..We do not support SQL lite but we will use it for testing

packages/graphback-cli/src/templates/resources/config/mongodb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "user": "mongodb",
+  "password": "mongo",
+  "database": "users",
```
wtrocki: Graphback?

wtrocki:
Ok so adding this will not be enough. We also need to change documentation what needs to be done once config is applied

wtrocki:
If we find some open source GraphQL example and try to add graphback using config that will clear it

wtrocki:
Tests are failing due to some code issue. check yarn test

packages/graphback-cli/src/components/config.ts
NVS16: Ok. Will add mongodb and pg as peer dependencies in mongo and knex runtime packages respectively.

packages/graphback-cli/src/components/db.ts
NVS16: Yes. I realised that a later.

packages/graphback-cli/src/templates/resources/config/mongodb.json
NVS16: Name the database graphback?

NVS16:
Ok I will have a look at tests and get back to you.

NVS16:
Ok, when graphback config is applied, we need to change files like db.ts, graphql.ts in the templates, according to user choice. But changing multiple files like that on graphback config, will that be good approach. Any changes in file structure later on may lead to rewrites?

packages/graphback-cli/src/templates/resources/config/mongodb.json
wtrocki: Yes, but we do not need this now since migrate is gone

wtrocki:
Hmm.. Graphback config makes no sense to be executed on templates. It will be executed on new projects. What this mean that we can:

- document how to then manually add some source code
- generate some source code ourselves basing on question: What language you use: ts/js

wtrocki:
We are missing docs updates to merge this. I think we can merge that PR after we test this in any *non graphback template* project. Unit tests need fix. Follow up can resolve problem of adding some files

wtrocki:
I will wait till build pases and integrate that back to master. Thank you for your time

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -23,7 +25,8 @@ const getConfig = (database: string) => {
 
 const databases = [
   'PostgreSQL',
-  'sqlite3'
+  'MongoDB'
+  // 'sqlite3'
```
wtrocki: Remove that as well

wtrocki:
```suggestion
```

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -91,15 +94,20 @@ export const createConfig = async (database: string, client: boolean) => {
...
-      dbmigrations: {
-        "connection": JSON.parse(dbConfig),
-        "client": database
+        }
       }
     }
   };
 
+  if(database === "pg") {
+    graphqlConfigJSON.extensions.graphback['dbMigrations'] = {
```
wtrocki: Was there bug for postgress? Why this code

NVS16: No I was adding the dbmigrations option when the database option was pg or in fact not mongo db. If it is not pg it will not send dbmigrations option

packages/graphback-cli/src/templates/configTemplates.ts
NVS16: Ok will remove that.

NVS16:
Hey I am running into quite unusual error on running graphback init/config -
Error: ENOENT: no such file or directory, open '/home/nvs/Work/gsoc_2020/aerogear/graphback/packages/graphback-cli/dist/templates/resources/models/Default.graphql'
Even though those files are there in the dist folder
Any ideas regarding this?

wtrocki:
I'm on different branch now and cannot move, but that is not happening for myself. Also init should no longer ask for model or use it. Maybe you need to rebase?

NVS16:
Yes. My branch seems messed up somehow. I can't get rid of that error.
I have made all the changes asked for in another branch and it works just fine. 
Should I make a PR on that branch?

NVS16:
Hey, the issues I was having with this branch were resolved.

wtrocki:
DOne some testing and I found bug. When I choose postgres dbmigrations are not added. That is serious issue as we support that for pg. Also build is still failing so no way to merge this

NVS16:
Thats strange. I will check and get back to you.

NVS16:
In my local machine the migrations are getting added for postgres. 
And here it shows circle ci build passed all the tests for the latest commit...

NVS16:
I look more into the postgres thing..

NVS16:
I will test the postgres dbmigrations a bit more to see if it fails...

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -91,15 +94,18 @@ export const createConfig = async (database: string, client: boolean) => {
             "format": "ts",
             "outputPath": "./src/resolvers"
           }
-        },
-      },
-      dbmigrations: {
```
wtrocki: That is the source of the problem. We want to keep it for pg

packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -91,15 +94,18 @@ export const createConfig = async (database: string, client: boolean) => {
...
-        "client": database
+        }
       }
     }
   };
 
+  if(database === "pg") {
+    graphqlConfig.extensions.graphback.dbmigrations = {
+      "connection": JSON.parse(dbConfig),
+      "client": database
```
NVS16: I am adding it back here for pg. 
If database is pg it is supposed to get added back. 
Will it not work?

wtrocki:
It will work

NVS16:
> Hmm.. Graphback config makes no sense to be executed on templates. It will be executed on new projects. What this mean that we can:
> 
> * document how to then manually add some source code
> * generate some source code ourselves basing on question: What language you use: ts/js

Regarding the documentation, can you point me to a direction, I am not sure where to get started with this.

wtrocki:
I guess we need to at least add some info here:
https://graphback.dev/docs/intro/existingproject
and here:
https://graphback.dev/docs/db/datasources

PR #803, opened 2020-02-28
Fix small issues in limits implementation

wtrocki:


packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -64,8 +64,9 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
 
-  public async findAll(): Promise<Type[]> {
-    const data = await this.db.collection(this.collectionName).find({}).toArray();
+  public async findAll(page?: GraphbackPage): Promise<Type[]> {
+    const data =  page ? (await this.db.collection(this.collectionName).find({}).skip(page.offset || 0).limit(page.limit || 10).toArray()):
```
wtrocki: This is really bad stuff. It should be handled differently.

PR #829, opened 2020-03-02
PluginEngine gql-compose migration

cecobask:
Started to play around with graphql-compose package.

### DIFF
--- a/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
+++ b/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
@@ -1,22 +1,20 @@
-import { buildSchema, GraphQLObjectType, GraphQLSchema, print } from 'graphql';
+import { GraphQLObjectType, GraphQLSchema } from 'graphql'
 import { parseAnnotations, parseMarker } from 'graphql-metadata'
-import { SchemaComposer } from 'graphql-compose';
 import { RelationshipMetadataBuilder, FieldRelationshipMetadata } from '../relationships/RelationshipMetadataBuilder'
 import { getModelTypesFromSchema } from './getModelTypesFromSchema'
 import { GraphbackCRUDGeneratorConfig } from './GraphbackCRUDGeneratorConfig'
 import { GraphbackGlobalConfig } from './GraphbackGlobalConfig'
 import { ModelDefinition } from './ModelDefinition';
-import { SchemaCompType } from './SchemaComposer';
 
 const defaultCRUDGeneratorConfig = {
-    "create": true,
-    "update": true,
-    "findAll": true,
-    "find": true,
-    "delete": true,
-    "subCreate": true,
-    "subUpdate": true,
-    "subDelete": true,
+  "create": true,
+  "update": true,
+  "findAll": true,
+  "find": true,
+  "delete": true,
+  "subCreate": true,
+  "subUpdate": true,
+  "subDelete": true,
 }
 
 /**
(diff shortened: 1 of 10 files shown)


cecobask:
Hi guys, when running the tests with schema object from schemaComposer.build(), I get a mismatch from the expected snapshot. The difference is that directives are apparent. Are directives supposed to be stripped down from the schema when testing?

image

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
```diff
@@ -52,11 +53,11 @@ export class GraphbackPluginEngine {
         if (this.plugins.length === 0) {
             throw new Error("GraphbackEngine: No Graphback plugins registered")
         }
-        //We need to apply all required changes to the schema we need 
+        //We need to apply all required changes to the schema we need
         //This is to ensure that every plugin can add changes to the schema
         for (const plugin of this.plugins) {
-            const newSchema = plugin.transformSchema(this.metadata);
-            this.metadata.setSchema(newSchema);
+          const newSchema = schemaComposer.merge(this.metadata.getSchema());
```
wtrocki: We need to create schema in composer

cecobask: In the following way?
https://graphql-compose.github.io/docs/intro/quick-start.html#bonus-track

wtrocki:
> Are directives supposed to be stripped down from the schema when testing?

That is the bug that we want to workaround..We should create issue in graphql compose maybe?

cecobask:
Is that a step in the right direction?

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -61,13 +62,13 @@ export class GraphbackCoreMetadata {
...
-     * 
+     * To mark type as enabled for graphback generators we need to add `model` annotations over the type.
+     *
      * Returns all user types that have @model in description
-     * @param schema 
+     * @param schema
      */
     public getGraphQLTypesWithModel(): GraphQLObjectType[] {
-        const types = getModelTypesFromSchema(this.schema)
+        const types = getModelTypesFromSchema(this.schemaComposer.buildSchema())
```
wtrocki: Is this currently stripping out the descriptions/markers? 
I think we need to figure out why descriptions/comments are missing when printing the schema

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
```diff
@@ -21,26 +21,20 @@ export class GraphbackPluginEngine {
     private plugins: GraphbackPlugin[];
     private metadata: GraphbackCoreMetadata;
 
-    public constructor(schema: GraphQLSchema | string, config: GraphbackGlobalConfig) {
+    public constructor(schemaComposer: SchemaComposer<any>, config: GraphbackGlobalConfig) {
```
wtrocki: More typings: NO <any>

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
```diff
@@ -49,17 +43,17 @@ export class GraphbackPluginEngine {
...
-        //We need to apply all required changes to the schema we need
-        //This is to ensure that every plugin can add changes to the schema
-        for (const plugin of this.plugins) {
-          const newSchema = schemaComposer.merge(this.metadata.getSchema());
-          this.metadata.setSchema(newSchema.buildSchema());
-        }
-
-        return this.metadata;
-    }
+    // public createSchema(): GraphbackCoreMetadata {
```
wtrocki: Why this is commented out?

wtrocki:
Changes look ok but it is hard to say anything as code is not functional at the moment

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
cecobask: Yes, it is. I shall open an issue at graphql-compose.

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
cecobask: Should I revert the function signature to be `public constructor(schema: GraphQLSchema | string, config: GraphbackGlobalConfig)`?

packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
cecobask: I will fix that.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
wtrocki: Let's find what is causing this to know if that is something to do with merge that we are doing using GraphQL tools. if you create sample app that just use compose with comment does it also remove them?

wtrocki:
I see couple challenges - our model is basing on schema types.  Moving to composer will require large refactor if we return different types as model. On the other hand we need schema updates for every plugin. Which means converting to schema is unavoidable. 

It looks like we need to see why we getting issues with composer - possibly some side example outside graphback codebase like gist. 
If we make call that we use composer we can still.have helper in model like - get editable schema etc.

This is chicken or egg problem and from all options moving to some custom schema types might be problematic in future

craicoverflow:
We could try writing our own custom implementation to update the schema objects.

I had started experimenting with this, it is possible to create a new schema with the same type references as the old one.

https://gist.github.com/craicoverflow/61469b104513ea4e7a0e6fb452e38658

This is a rough draft, proof of concept that shows a away to add a new field to a type. 

This could be an approach to try. These schema transfomer methods could be in the form of helpers that can be used across all plugins.

wtrocki:
Amazing idea @craicoverflow. I think we proven here that we might not need to swap to graphql-compose. We can still use compose inside. 

The biggest challenge we have now is that we use everything:

- graphql-js config and schema creation
- graphql-toolkit merge function that does some magic
- Graphql compose

Mixture of those gives us bugs but we do not know what specifically is causing it. If we have one generic approach that will be awesome. Compose seems nice api for extending every type. Using graphql-js will always mean that we will need to create new objects - compose gives this very easy wrapper on top of that.

We can have helpers returning us some editale elements and then still return immutable schema as result of the plugin.

@cecobask sorry for moving direction but after modifying your code and playing with it I do not see current approach solving the issue

wtrocki:
We have decided to make some small cleanup from kinda messy situation we have at the moment. IN https://github.com/aerogear/graphback/pull/835 we removed some of the hacks we had and accepted that schema looks different. 

This will allow us to evaluate graphql-compose vs plain graphql-js operations
Please rebase on top of that change and then we can see some results of the spikes

wtrocki:
Requires rebase

wtrocki:
Outdated to the point it will probably make sense to start from beggining

PR #830, opened 2020-03-03
WIP: Integrate graphql-serve with graphback

ssd71:
- Complete overhaul of the graphql-serve package
- Provides a `gqlserve` command that creates a gql server that listens on an available port
- The command needs two files:
    - A graphql-config file
    - A model graphql schema

To run the gqlserve command, you need a .graphqlrc.yml or equivalent file in the root folder and a *.graphql file inside a model folder. Same can be found in the runtime template folder.

Work In progress:

- [ ] Add tests
- [ ] Integrate gql serve with graphback-cli?
- [ ] Fix errors with SQLite in-memory database
- [ ] Fix lint errors and tests
- [ ] General code cleanup


packages/graphql-serve/.graphqlrc.yml
```diff
@@ -0,0 +1,22 @@
+schema: ./server/src/schema/*.graphql
```
wtrocki: Why this file here?

packages/graphql-serve/bin/printSchema.ts
```diff
@@ -1,22 +1,22 @@
-#!/usr/bin/env node
+// #!/usr/bin/env node
```
wtrocki: Uncomment

packages/graphql-serve/package.json
```diff
@@ -13,35 +13,35 @@
...
     "graphql": "14.6.0",
     "graphql-migrations": "0.10.2",
     "graphql-tag": "2.10.3",
-    "knex": "0.20.7",
+    "knex": "0.20.10",
     "knex-cleaner": "1.3.0",
     "sqlite3": "4.1.1"
   },
   "bin": {
-    "print-schema": "./dist/bin/printSchema.js"
```
wtrocki: Print schema is also cool feature

packages/graphql-serve/src/GraphbackServer.ts
```diff
@@ -92,37 +94,31 @@ export class GraphbackServer {
...
+export async function buildGraphbackServer(graphbackConfigOpts?: any) {
+  const app = express();
 
-  const sub = new PubSub();
+  app.use(cors());
 
-  const service: GraphbackCRUDService = serviceBuilder
-    ? await serviceBuilder(data, sub)
-    : new CRUDService(data, sub);
+  if (!graphbackConfigOpts) graphbackConfigOpts = await getGraphbackConfig(path.join(__dirname, ".."));
```
wtrocki: What is this about? Why we loading config from package? Should we have some defaults here and let people override it by config

packages/graphql-serve/src/db.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+};
+
+/**
+ * Creates knex based database using migration configuration
+ * For production use please use different source of the configuration
+ */
+export const createDB = () => {
+
+    const dbmigrations = migrateConfig;
```
wtrocki: Not needed

wtrocki:
Great job! It needs some general love but concept looks ok. We can reuse command in graphql cli

wtrocki:
Also the way I see it is that - we can read config in root folder and this will use model value in config. Config can be also optional then user just passes the model file and we assume that all crud options are enabled

packages/graphql-serve/.graphqlrc.yml
wtrocki: See comment bellow. We do not need it here. We can use json for defaults

ssd71:
As of 0bac83c :
- General refactoring
- Made it optional to have a graphql config
- Rewrote printSchema

@wtrocki Please check

packages/graphql-serve/.graphqlrc.yml
wtrocki: This is still pushed?

packages/graphql-serve/bin/serve.ts
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/node
+import { buildGraphbackServer, getGraphbackServerConfig } from "../src";
```
wtrocki: I would prefer these files to simply call some method rather than having implementation in the bin folder

packages/graphql-serve/.graphqlrc.yml
ssd71: I will remove it in the next push. It is not needed anymore

packages/graphql-serve/bin/serve.ts
ssd71: Should I also move the arg parser stuff to another file?

Issue #831, opened 2020-03-03
UnhandledPromiseRejectionWarning: error: syntax error at or near "primary"


craicoverflow:
I am getting the following error when I run the runtime example twice with database migrations on.

```
(node:91221) UnhandledPromiseRejectionWarning: error: syntax error at or near "primary"
    at Connection.parseE (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:614:13)
    at Connection.parseMessage (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:413:19)
    at Socket.<anonymous> (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:129:22)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:275:11)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23)
(node:91221) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:91221) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:91221) UnhandledPromiseRejectionWarning: error: syntax error at or near "primary"
    at Connection.parseE (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:614:13)
    at Connection.parseMessage (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:413:19)
    at Socket.<anonymous> (/home/ephelan/code/aerogear/graphback/node_modules/pg/lib/connection.js:129:22)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:275:11)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23)
warning.js:32
```

It creates the database the first time, but running it again fails with above error.

[...]

PR #833, opened 2020-03-03
Map `increments` to `integer` in non-primary fields

craicoverflow:
Fixes https://github.com/aerogear/graphback/issues/831

Non-primary columns were getting created with the SERIAL data type, which is intended as a primary key field. Issues arose when running a migration against the database

```
alter table "public"."note" alter column "ownerId" type serial primary key using ("ownerId"::serial primary key) - syntax error at or near "primary"
```

This fix maps `increments` (SERIAL) to `integer` for foreign key fields.

## Other

- Add filter to prevent table deletion in runtime template.
- enable `@oneToOne` and `@oneToMany` markers in `graphql-migrations`. migrations will use `key` as the preferred way to create a foreign key field.


packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
```diff
@@ -232,7 +234,7 @@ class AbstractDatabaseBuilder {
 
       // Object
     } else if (isObjectType(fieldType)) {
-      columnName = annotations.name || `${field.name}Id`
+      columnName = relationshipMarker?.key || annotations.name || `${field.name}Id`
```
wtrocki: That could be nightmare to document. Do we need both annotations and marker support at the time?

wtrocki:
Do we have any tests that will simulate creation and later migration in this package? Trying to figure out how to protect from similar issues in near future

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
craicoverflow: We discussed supporting the existing annotation but just documenting the new one in Graphback.

craicoverflow:
> Do we have any tests that will simulate creation and later migration in this package? Trying to figure out how to protect from similar issues in near future

I don't believe we do - our tests use SQLite which works only for a single migration.

Issue #828, opened 2020-03-02
Integrate graphql-serve with graphback


ssd71:
The graphql-serve package uses outdated version of the graphback packages. It needs to be ported to the newer versions of said packages.

Work In progress:

- [x] Add serve and print schema commands
- [x] Update Readme
- [ ] Integrate gql serve with graphback-cli
- [ ] Update docs
- [ ] Add tests
- [ ] Improve CLI
- [ ] General code cleanup

ssd71:
I am working on this issue

ssd71:
Some of the work has been done in #830 
I will continue to work on some of these in another PR

wtrocki:
Package has been set to private true. Once we fix readme we can remove it in order to release it

craicoverflow:
This looks to be done, but feel free to reopen if I am wrong.

PR #840, opened 2020-03-03
chore: Update serve README.md

ssd71:
Updated graphql-server readme
Relevant Issue:  #828


packages/graphql-serve/README.md
```diff
@@ -30,354 +42,80 @@ yarn add graphql-serve
...
-  findAllItems {
-    ...ItemFields
-  }
-}
-
-fragment ItemFields on Item {
-  id
-  name
-}
+$ printSchema --model=models
```
wtrocki: I to think that we need to have this command as subcommand of serve

ssd71: I will make another PR for that

PR #843, opened 2020-03-04
Offix unit test

wtrocki:
Add unit tests for offix plugin.
Change way we use graphql-compose


wtrocki:
CC @cecobask I have fixed issues based on recommendation made upstream

packages/graphback-codegen-offix/src/OffixPlugin.ts
```diff
@@ -64,13 +62,18 @@ export class OffixPlugin extends GraphbackPlugin {
...
-                });
+            try {
+                const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name))
+                if (inputType) {
+                    inputType.addFields({
+                        version: 'Int'
+                    });
+                }
+            } catch (e) {
+                // ignore as we are not guaranteed to have this input type
```
wtrocki: Not convinced I'm properly handling this

packages/graphback-codegen-offix/src/OffixPlugin.ts
```diff
@@ -80,14 +83,15 @@ export class OffixPlugin extends GraphbackPlugin {
                 schemaComposer.Query.addFieldArgs(diffQuery, {
                     lastSync: 'String!'
                 })
-                // TODO generate delta resolvers
             }
         })
 
-        return schemaComposer.buildSchema()
+        return buildSchema(schemaComposer.toSDL())
```
wtrocki: `schemaComposer.toSDL()` returns defaults scalars so this hack allows us to get the right schema.

packages/graphback-codegen-offix/tests/OffixPluginTest.ts
```diff
@@ -0,0 +1,27 @@
...
+    "subCreate": true,
+    "subUpdate": true,
+    "subDelete": true
+  }
+
+  const schemaGenerator = new OffixPlugin({ outputPath: './tmp' })
+  const metadata = new GraphbackCoreMetadata({
+    crudMethods: defautConfig
+  }, buildSchema(schemaText))
+  const schema = schemaGenerator.transformSchema(metadata)
```
wtrocki: We are missing here GraphQL Schema extension. I wasn't sure if we want to add that as it will not longer be unit test. We have it already in dependencies so I guess we can extend that later to see how input types will be extended.

packages/graphback-codegen-offix/tests/mock.graphql
```diff
@@ -0,0 +1,27 @@
+"""@model"""
+type Note {
+  id: ID!
+  title: String!
+  description: String!
+  ## Relationship
+  """
+  @db.oneToMany: 'noteComment'
```
craicoverflow:
```suggestion
  @oneToMany field: 'noteComment'
```

PR #845, opened 2020-03-04
Relationship tests

craicoverflow:


wtrocki:
Unit tests not only covering extensive amount of cases, but this also shows how amazing and mature our implementation became. Thank you so much for this masterpiece. Good testament of your skills. I guess this is no longer draft so approving it

craicoverflow:
Going to merge this. I noticed that the client generator plugin may need to be updated to work with new relationships but will do this in a separate PR.

PR #846, opened 2020-03-04
chore: Serve refactor

ssd71:
This patch contains the following:
- Move printSchema functionality to gqlserve
- Update Readme accordingly
- Refactor serve command


packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
...
 
 Options:
-  --version    Show version number                                     [boolean]
-  --model, -m  Specify the models directory                             [string]
-  --port, -p   Specify the port on which to listen on                   [number]
-  -h, --help   Show help                                               [boolean]
+  --version     Show version number                                    [boolean]
+  --model, -m   Specify the models directory                            [string]
+  --port, -p    Specify the port on which to listen on                  [number]
+  --schemaOnly  Only show GraphQL schema generated from data models    [boolean]
```
craicoverflow: I'm not sure about using camel case in CLI arguments. I think kebab case would be better. @wtrocki what do you think? Is there any compliance issues with camel case?

packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
 Usage: gqlserve [options]
 
 Options:
-  --version    Show version number                                     [boolean]
-  --model, -m  Specify the models directory                             [string]
-  --port, -p   Specify the port on which to listen on                   [number]
-  -h, --help   Show help                                               [boolean]
+  --version     Show version number                                    [boolean]
+  --model, -m   Specify the models directory                            [string]
+  --port, -p    Specify the port on which to listen on                  [number]
```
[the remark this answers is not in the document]
ssd71: My bad! I'll fix that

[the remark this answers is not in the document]
ssd71: Please take a look now @craicoverflow

packages/graphql-serve/README.md
ssd71: @craicoverflow I have no issues with making it kebab case

wtrocki: I will follow the same format like graphback

packages/graphql-serve/README.md
```diff
@@ -60,14 +60,15 @@ type Note {
...
 $ gqlserve --model=models --port=8080
 ```
 
-For yarn:
+
+If you only need to see the generated GraphQL Schema, use the `--schemaOnly` option:
+
 ```
-$ yarn gqlserve --model=models --port=8080
+$ gqlserve --schemaOnly
```
wtrocki: This should be command like print schema.

ssd71: Should we also make serve a command? This way it'll have two commands serve and print schema

wtrocki: Ok

wtrocki: Ok

ssd71:
> Can we also get that integrated with graphback cli

On it

packages/graphback-cli/package.json
```diff
@@ -54,6 +54,7 @@
   },
   "dependencies": {
     "@graphql-toolkit/code-file-loader": "0.9.7",
+    "graphql-serve": "0.11.0-rc5",
```
ssd71: Is this alright to do?

packages/graphback-cli/src/index.ts
```diff
@@ -10,6 +11,8 @@ if (require.main === module) {
   // eslint-disable-next-line no-unused-expressions
   yargs
     .commandDir('commands')
+    .command(serve)
+    .command(printSchema)
```
ssd71: I only added the commands from graphql-serve package, Is this okay?

packages/graphback-cli/package.json
craicoverflow: Yeah it is because the module is private. You will need to make it public so that it can be published.

packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
...
+  --schemaOnly  Only show GraphQL schema generated from data models    [boolean]
+  -h, --help    Show help                                              [boolean]
 
 Examples:
-  gqlserve --model=model  generate schema from models in the model directory
-  gqlserve --port=8000    server listens on port 8000
+  gqlserve --model=modeldir  generate schema from models in the "modeldir"
+                             directory
+  gqlserve --port=8000       server listens on port 8000
+  gqlserve --schemaOnly      prints schema and quits
```
craicoverflow: Should be 
```
  gqlserve serve print-schema      prints schema and quits
```

packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
...
+  --version     Show version number                                    [boolean]
+  --model, -m   Specify the models directory                            [string]
+  --port, -p    Specify the port on which to listen on                  [number]
+  --schemaOnly  Only show GraphQL schema generated from data models    [boolean]
+  -h, --help    Show help                                              [boolean]
 
 Examples:
-  gqlserve --model=model  generate schema from models in the model directory
-  gqlserve --port=8000    server listens on port 8000
+  gqlserve --model=modeldir  generate schema from models in the "modeldir"
```
craicoverflow: Should be
```
  gqlserve serve --model=modeldir  generate schema from models in the "modeldir"
```

packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
...
+                             directory
+  gqlserve --port=8000       server listens on port 8000
```
craicoverflow: Should be
```
gqlserve serve --port=8000       server listens on port 8000
```

packages/graphql-serve/README.md
```diff
@@ -76,45 +77,24 @@ $ gqlserve --help
 Usage: gqlserve [options]
```
craicoverflow:
```suggestion
Usage: gqlserve serve [options]
```

craicoverflow: `gqlserve` on its own makes more sense than `gqlserve serve`. But currently the latter is what we have. Might it make more sense to remove the `serve` subcommand?

ssd71: I would do that, but currently the `graphback-cli` also uses the commands  i.e. it also has the serve and print-schema commands, let me look into it

ssd71:
I have made the necessary changes(with the exception of making graphql-serve public) @craicoverflow please take a look

wtrocki:
@ssd71 build failing

packages/graphback-cli/src/index.ts
wtrocki: This is a very bad pattern. See how commands are created here - we use folders. 
As we have mentioned before we should use the same patterns for all commands and simply wrap them.

wtrocki: TL;DR - let's create commands the same way they are created in this package

ssd71:
@wtrocki the build is failing because graphql-serve package is not public as pointed out by @craicoverflow 

Is it okay to make it public now?

packages/graphql-serve/bin/gqlserve.ts
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/node
```
wtrocki: This is very hacky way of creating commands. Please follow standard we have in the graphback-cli - using cmd wrappers so we can reuse actual command services while keeping commands lightweight in code

packages/graphql-serve/src/serveHandler.ts
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/node
```
wtrocki: THis is not needed. General refactor needed for this class

craicoverflow:
> @wtrocki the build is failing because graphql-serve package is not public as pointed out by @craicoverflow
> 
> Is it okay to make it public now?

Why have you not made it public? You can.

packages/graphql-serve/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "graphql-serve",
   "version": "0.11.0-rc5",
-  "private": true,
```
ssd71: This didn't work, I think

ssd71:
@wtrocki I have pushed somewhat palatable code this time, please take a look, and the build fails even after 5ec24d7 

One other thing, as @craicoverflow  said:

> `gqlserve` on its own makes more sense than `gqlserve serve`. But currently the latter is what we have. Might it make more sense to remove the `serve` subcommand?

I totally agree with him, but I need some more time to figure out how to do it

ssd71:
I will change the readme after the commands and cli is finalized

wtrocki:
Lets continue working on top of this PR. I will help with review once we clear data sources problem

wtrocki:
I think model should be positional parameter rather than -m.
Serve should not print schema. That should be separate command. 
`Listening at: 58528` Let's provide full url so people can click on it

wtrocki:
For example http://localhost:58528/graphql

wtrocki:
Once these minor issues will be fixed we can write blog post announcing graphql-serve. That will be awesome

PR #849, opened 2020-03-05
Add Relationship fields in client side fragments

craicoverflow:
## What

Client-side generated fragments would only include expanded fields for 1:M types, whereas we need to have expanded fields for M:1 and 1:1.


packages/graphback-codegen-client/src/templates/fragmentFields.ts
```diff
@@ -9,45 +10,43 @@ import { getNamedType, GraphQLObjectType, isCompositeType, isWrappingType } from
...
 export function buildReturnFields(t: GraphQLObjectType, level?: number) {
-    const fieldsMap = t.getFields();
-    if (level > 1) {
-      throw new Error("Function supports only 1 nested level")
-    }
-  
-    //tslint:disable-next-line: typedef
-    return Object.keys(fieldsMap).reduce((data, key) => {
-      const field = fieldsMap[key];
-      if (isWrappingType(field.type)) {
```
craicoverflow: @wtrocki I removed `isWrappingType` because I couldn't understand what it is for in this context.

wtrocki: I have the same with our own relationship helpers ;)
Helpers are evil and some large amount of knowledge is lost because we have them. If we remove wrapping type we will apply NonNull and break stuff.

wtrocki:
We haven't verified that before and even now it will be hard to verify those without actuall app implementing this or integration tests. 
I would be cool to see app copied (prefferable) or integrations tests for this and then merge

packages/graphback-codegen-client/src/templates/fragmentFields.ts
craicoverflow: Hmm I am not so sure about this. `getNamedType` unwraps a type so `modelType` is actually unwrapped, so there is no possibility of applying non-null.

Also fragments only used the `key` which never includes `!`

packages/graphback-cli/tests/workflow-test.ts
```diff
@@ -23,8 +35,19 @@ test('Test cli workflow', async (t: ExecutionContext) => {
 
   t.true(existsSync(join(basePath, "db.sqlite")));
   t.true(existsSync(join(basePath, "client/src/graphql/fragments/Note.graphql")));
+  t.true(existsSync(join(basePath, "client/src/graphql/fragments/NoteExpanded.graphql")))
```
wtrocki: I would move this into integration test folder in future

packages/graphback-codegen-client/src/templates/fragmentFields.ts
wtrocki: This was to detect array.

wtrocki:
Needs rebase as we have merged previous changeset

PR #850, opened 2020-03-05
More CURD methods for full stack template 

udith-shalinda:


wtrocki:
@udith-shalinda You are still behind master somehow:

Screenshot 2020-03-05 at 16 26 20

templates/ts-apollo-fullstack/client/src/generated-types.tsx
```diff
@@ -14,22 +14,22 @@ export type Scalars = {
 export type Comment = {
```
wtrocki: Do we actually need this file? We can use typescript to load queries

udith-shalinda: Since the queries are written in there,we need it

templates/ts-apollo-fullstack/client/src/components/comment/CreateComment.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const CreateComment = ({noteId,addCommentState}:createCommentProps) => {
+    const [createComment] = useCreateCommentMutation();
+    const [newCommentTitle, setNewCommentTitle] = useState("");
+    const [newCommentDescription, setNewCommentDescription] = useState("");
+
+    return(
+        <div>
+            <fieldset>
+                <legend>Create New Comment</legend>
```
wtrocki: While this work is there any reason why we do not use react elements? It will be much easier to style that after

templates/ts-apollo-fullstack/client/src/components/comment/CreateComment.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    onChange={(e) => setNewCommentTitle(e.target.value)} />
+                </p>
+                <p>
+                    <label htmlFor="description">Description: </label>
+                    <input
+                    name="description"
+                    value={newCommentDescription}
+                    onChange={(e) => setNewCommentDescription(e.target.value)} />
+                </p>
+                <input type='submit' />
```
wtrocki: That should be react button etc.

udith-shalinda: ok.I'll change them

wtrocki:
Of we apply some react components we can add some styles easily

templates/ts-apollo-fullstack/client/src/generated-types.tsx
```diff
@@ -14,22 +14,22 @@ export type Scalars = {
...
    __typename?: 'Comment',
   id: Scalars['ID'],
-  title: Scalars['String'],
```
wtrocki: Was that generated ?

udith-shalinda: The file was already there.And the queries for the comments wasn't working(Because it uses old comment model).So I had to change them.

wtrocki: So this file is generated using graphql code generator. If we see value we can bring codegen for client side queries.

wtrocki:
Needed to squash because of git issues

Issue #864, opened 2020-03-07
gqlserve print-schema does not quit after printing schema


ssd71:
I have confirmed this is due to the MongoDB memory server not stopping, I am working on a PR to solve this issue

ssd71:
Closed by #865

PR #865, opened 2020-03-07
fix: Separate serve and print-schema Handlers

ssd71:
This PR addresses #864


ssd71:
#864 is fixed by 90ac933

wtrocki:
Amazing changes.

wtrocki:
Amazing work. This is live and published as graphql-serve@0.10.0

wtrocki:
I have used 0.10.0 to match previous version of the graphback. 0.11.0 will kick in along with the rest of the packages once we do proper release (we use single version for entire stack)

Issue #753, opened 2020-02-20
Graphback template build script is disabled.


wtrocki:
As here https://github.com/aerogear/graphback/blob/master/templates/ts-apollo-fullstack/server/package.json#L6-L8

wtrocki:
Blocker for release

wtrocki:
After that is done we need to push templates for testing

PR #874, opened 2020-03-09
chore: enable build script

craicoverflow:
See https://github.com/aerogear/graphback/issues/753

This change enables the build script for the fullstack client app.

### DIFF
--- a/templates/ts-apollo-fullstack/client/package.json
+++ b/templates/ts-apollo-fullstack/client/package.json
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "echo disabled",
+    "build": "react-scripts build",
     "test": "echo disabled",
     "lint": "echo disabled",
     "build:react": "react-scripts build",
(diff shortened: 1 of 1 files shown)


templates/ts-apollo-fullstack/client/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "echo disabled",
+    "build": "react-scripts build",
```
wtrocki: This should be typescript build

craicoverflow: Are you sure about that? Because I tried that first and it did nothing. Would `tsc` compile the HTML and CSS too?

craicoverflow: Hmm I wonder if it is because these are tsx files. I will check.

wtrocki: Actually, we might remove build completely from client, but put build into server etc. Trick is to remove echo cmd that look quite unprofessional when app is bootstrapped

PR #880, opened 2020-03-09
fix: remove serve from grahback-cli

wtrocki:


wtrocki:
This is causing circular dependency issues

wtrocki:
GraphQL CLI should be configured to work with any graphql version. Adding cli to the app ends up with the graphql conflict if version is different

Issue #851, opened 2020-03-05
Replace uses of `getBaseType` with `getNamedType`


craicoverflow:
`getBaseType` in `@graphback/core` does the same thing as [`getNamedType`](https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L518)

We should use `getNamedType` from the official reference implementation library and remove `getBaseType`.

NVS16:
Can I work on this?

craicoverflow:
> Can I work on this?

Absolutely!

NVS16:
I have made a preliminary PR. Do I need to make any other changes?

PR #889, opened 2020-03-10
Replaced all the occurrences of getBaseType graphback-core

NVS16:
Addresses #851 
Replaced all the occurrences of getBaseType from graphback-core with getNamedType.


craicoverflow:
Thanks a lot @NVS16, We will get to this soon.

PR #893, opened 2020-03-11
Client fixes

wtrocki:
Fix multiple issues with client side code.
Generate proper find query.
Add codegen as it is being used in the current app.

This PR is enormously large because the last commits pushing the code changes for the generated code. I needed to do that in order to make some changes but since now all is fixed it should be ok. Please review this using commit ranges.

## Verification

Generate code (again)
Generate codegen (again) 
See there is no changes in the codebase

Start server
Start client 

Do all possible operations with the client.


templates/ts-apollo-fullstack/codegen.yml
```diff
@@ -0,0 +1,13 @@
+schema: ./server/src/schema/schema.graphql
+generates:
+  ./client/src/generated-types.tsx:
+    config:
+      withComponent: false
```
wtrocki: @darahayes @kingsleyzissou 

Codegen resolvers nicely problem of hooks vs hoc by giving people flags. 
What we get from codegen is fully featured data layer. No need to do anything apart from using hooks.

darahayes: Yeah perhaps generating the hooks could be a config option.

wtrocki: Yep.. That is already there - what we need to do is to just call offline mutate in generated-types.ts - which is small extension to existing apollo plugin. It will be much easier than writing that from scratch, but also tie us to codegen

wtrocki:
Note: Still working on fixing client side app. Not ready for review

wtrocki:
Screenshot 2020-03-11 at 15 13 32

This is working currently with sample app

wtrocki:
@darahayes It is actually very easy now to swap generated hooks to offix hooks and test it. I have done that quickly and it works (without testing conflict and proper error handling)

packages/graphback-core/tests/snapshots/RelationshipMetadataBuilderTest.ts.md
```diff
@@ -79,7 +77,9 @@ Generated by [AVA](https://ava.li).
          "start": 89,␊
          "end": 204␊
         }␊
-       }␊
+       },␊
+       "name": "sentMessages",␊
```
craicoverflow: Is this just a result of AST order being changed in GraphQL-JS?

templates/ts-apollo-fullstack/README.md
```diff
@@ -32,6 +32,9 @@ yarn start:server
 ```
 
 - Generate React components
```
craicoverflow: Not very accurate description as to be that sounds like it will generate React TSX components.

packages/graphback-core/tests/snapshots/RelationshipMetadataBuilderTest.ts.md
wtrocki: Yes. result.of downgrade

craicoverflow:
After running `graphback generate`:

Screenshot from 2020-03-11 17-05-08

After running `yarn codegen:client`:

Screenshot from 2020-03-11 17-06-18

Issue #904, opened 2020-03-12
Bug Report: Query, Subscription, Mutation types defined in input model are not added to schema


craicoverflow:
This is a bug in 0.11.x

Input:

```graphql
type Query {
  test: String
}
```

Schema:

```graphql
type Query {
  findAllNotes(limit: Int, offset: Int): [Note]!
  findNotes(fields: NoteInput, limit: Int, offset: Int): [Note]!
  findAllComments(limit: Int, offset: Int): [Comment]!
  findComments(fields: CommentInput, limit: Int, offset: Int): [Comment]!
}
```

This is a critical bug. I am almost certain it is because of [this](https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L71-L74).

`config` from the original schema (containing custom query) is overwritten by `modelConfig` because we are only using the spread operator to merge at the top level.

PR #912, opened 2020-03-12
Add custom root type fields to generated schema

craicoverflow:
Fixes https://github.com/aerogear/graphback/issues/904

There was a bug where user defined root type fields in the data model were not getting added to the final schema.

This takes a different approach so that we do not need to do any schema stitching.

## Verification

Define some custom root fields in a Query, Mutation and Subscription object in data model.

```gql
type Query {
  """hello"""
  hello: String
}

type Mutation {
  hey: String
}

type Subscription {
  test: String
}
```

And run `graphback generate`.

Inspect the generated schema which should contain these custom fields as well as the standard CRUD fields.


packages/graphback-codegen-schema/tests/snapshots/GraphQLSchemaCreatorTest.ts.md
```diff
@@ -123,6 +125,7 @@ Generated by [AVA](https://ava.li).
       createTest(input: TestInput): Test!␊
       updateTest(input: TestInput): Test!␊
       deleteTest(input: TestInput): Test!␊
+      likeNote(id: ID!): Note!␊
```
wtrocki: It is sometimes as easy as checking the outputs.
I rarely did that before by simply skipping those files but they are actually the most important ones to review first.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -298,6 +298,32 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
         return queryTypes;
     }
 
+    protected createCustomRootTypes(inputRootType: GraphQLObjectType) {
```
wtrocki: What this function does?

wtrocki: This sounds like something we should not do. 
What I see that this code will strip out all directives from types. 
Is that desired?

wtrocki: Generally, we can go back and use mergeType from toolkit - writing our own helper in plugin is risky business and then this also will mean complexity for other plugins doing schema

craicoverflow: Actually this is a bad name now that I think about it. This function simply receives an object input and maps the fields.

So it will receive a query object, and maps the fields into a field object in the same format that is used to create the CRUD objects (see https://github.com/aerogear/graphback/blob/124a3873e2300353bd239cc9de1a9d2eac419eb4/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L282)

Issue #810, opened 2020-03-01
Change plugin engine to operate on graphql compose schema


wtrocki:
Currently for creating schema we pass immutable schema object. We should 
operate on graphql compose schema and generate the immutable schema.after all plugins will finish. See core package for more details.

Interesting places:
GraphbackMetadata - schema field (this in immutable)
GraphbackPluginEngine - we passing this schema over
Each plugin has some hacks.

The most interesting place to look on is OffixPlugin (to be Renamed to DataSync plugin)

wtrocki:
This is the implementation that we need to use: 
https://graphql-compose.github.io/

wtrocki:
@craicoverflow Can you add more information what is currently missing.
On top of my head issues with the composer are:
- composer adding null to params
- composer removing descriptions

Maybe we can investigate why this happens.

craicoverflow:
schemaComposer.buildSchema() will throw an error if the data model does not have any root query defined.

This and the two issues you mentioned can be worked around (hacked to work) but this is far from ideal.

> We should operate on graphql compose schema

Is this a good idea? We would be moving away from the reference implementation. As we saw in https://github.com/graphql-compose/graphql-compose/pull/237 there can be divergences which causes issues.

Obviously it seems like the best approach right now other than writing something ourself.

wtrocki:
Considering how good and responsive GraphQL-Compose team is we might actually log issue for them to not remove comments etc. but we should make sure first if the problems are not on our side

wtrocki:
I think that we good really good deal here already thanks to graphql compose contributor. Do you think we can do more? Maybe docs for plugin creation?

craicoverflow:
> I think that we good really good deal here already thanks to graphql compose contributor. Do you think we can do more? Maybe docs for plugin creation?

We could have a dummy plugin described in the docs that use graphql-compose or something. We don't want to push it too hard as the only way to do things.

wtrocki:
I think offix plugin can become that example. Descriptions are back so things should be quite easy from now on. I might create some helpers in core but we will see how it goes once we start refining offix plugin

ssd71:
So we will create new plugin for this, or modify Offix?

wtrocki:
@ssd71 For what specifically. There is PR that @cecobask done that was partially resolving the problem. https://github.com/aerogear/graphback/pull/829

ssd71:
I misunderstood, sorry about that! I will try some stuff and get back to you

[...]

PR #913, opened 2020-03-12
Use schema composer in Plugin Engine

ssd71:
Work-in-progress
Relevant issue: #810 

This PR contains:
- Modify Plugin Engine to use schemaComposer to merge schemas

The tests fail because descriptions aren't copied, I am still looking into this


wtrocki:
@ssd71 I think we need to chat about this approach.

Generally we can apply this both ways:

1) Have every plugin extending schema creating GraphQL compose instance making edits and then converting things back to normal schema.

2)Have merge in engine like in this PR - however this is very inefficient due to the fact that most of the plugins will not change schema at all and we going to be doing pointless merges that can take time for large schemas

While this is good change. I think we might need to explore other approach by simply refactoring Schema plugin to use compose first.

ssd71:
I'll open another PR using first approach, can we talk about approaches on the issue itself

PR #914, opened 2020-03-12
Addition of Soft Delete to the MongoDB

jaskeerat789:


wtrocki:
while this works this feature is not complete
- queries still fetch deleted element
- lack of indexing on deleted field
- lack of tests

jaskeerat789:
The last two points are clear to me and i am working on them.

jaskeerat789:
sir indexing is done when the database is created.., could you pls specify the file which is responsible for the creation of the database.

wtrocki:
@jaskeerat789 We do not have anything that creates collections for mongo. 
Basing on my understanding collections are also created then the element is inserted. This is definitely a challenge.

jaskeerat789:
> The index is updated automatically on every insert operation. You don't have to do this manually but have it in mind. This is what makes insert operations expensive. The more indexes you have the longer it takes to insert a document and update all of them.

this from the official docs

jaskeerat789:
should I make the DB migration package? @wtrocki @craicoverflow

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -63,8 +63,22 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
 
+  public async softDelete(data: Type): Promise<Type> {
```
craicoverflow: I don't see this function being used anywhere other than tests

craicoverflow:
> I guess we might need some form of mongo related dbmigration package that will simply apply indexes?? @craicoverflow WDYT?

From some quick Googling it seems there might be a few reasons to have MongoDB migrations.

craicoverflow:
> should I make the DB migration package? @wtrocki @craicoverflow

No, not yet. This requires further planning and discussion on our side to see what is needed. We may not need to create our own since there may already be another open source project that does this.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
jaskeerat789: I don't get it... where was it suppose to be used??

jaskeerat789:
@wtrocki @craicoverflow Sir pls let me know what changes need to be done in this PR

jaskeerat789:
Actually I am unable to understand the flow of the package. I define whose functions in `MongoDBDataProvider` which directly control the DB. 
What does `OffixMongoDBDataProvider` do?  what is offix is general?

Cause of this, I am unable to understand the point that @wtrocki  and @craicoverflow made in the previous comments. To see where are the other functions from `MongoDBDataProvider` are used, I tried using VSCode's references feature, but couldn't find the file where the `softDelete` function needs to be used

craicoverflow:
@jaskeerat789 thanks a lot for the PR! Right now this is not a focus of our roadmap so we will revisit it at a later stage - but going to close for now.

PR #915, opened 2020-03-12
Port plugins to schemaComposer

ssd71:
WIP
Issue: #810


ssd71:
I have modified Schema plugin so that it exclusively uses schema composer internally to add to the schema and then converts to immutable schema afterwards and returns that

wtrocki:
Really good work and definitely good path to swap things one by one

ssd71:
Finished with the SchemaPlugin at d27b17e

wtrocki:
It is amazing how simple and easy it was.
@craicoverflow wdyt? Solid merge.for me but will need to test it first

craicoverflow:
> It is amazing how simple and easy it was.
> @craicoverflow wdyt? Solid merge.for me but will need to test it first

I've tested and it works perfect! Great stuff @ssd71

PR #921, opened 2020-03-16
Switch to jest

kingsleyzissou:
- Switched all the tests from AVA to Jest
- Tests can be run from root level or package level
- Had to add a build stage for circleci


wtrocki:
@kingsleyzissou Can you point me to the problem you have mentioned. I forgot where to look. On local machine all tests are passing. Proposed structure works fine. It looks like this is also going to resolve couple other package oriented issues. We can now rely on the coverage. Great timing for that changes as there is no major PR created that will affect it

.circleci/config.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - run:
           name: run lint
           command: yarn lint
+      - run:
```
wtrocki: I was assuming that test will do prebuild of the packages. Good catch

.vscode/launch.json
```diff
@@ -152,5 +152,22 @@
       ],
       "cwd": "${workspaceRoot}/templates/ts-apollo-fullstack",
     },
+    {
```
wtrocki: ❤️

packages/graphback-cli/tests/workflow-test.ts
```diff
@@ -25,30 +24,52 @@ const model = {
...
-  t.true(existsSync(join(basePath, "client/src/graphql/queries/findNotes.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/createNote.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/createComment.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/deleteComment.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/deleteNote.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/updateComment.graphql")));
-  t.true(existsSync(join(basePath, "client/src/graphql/mutations/updateNote.graphql")));
-  t.true(existsSync(join(basePath, "src/schema/schema.graphql")));
-  t.true(existsSync(join(basePath, "src/resolvers/resolvers.ts")));
+  expect(existsSync(join(basePath, "db.sqlite"))).toBe(true);
```
wtrocki: Oh.. This must have been so hard. 
We can improve this by snapshotting folder structure.

packages/graphback-core/tests/__snapshots__/EngineTests.ts.snap
```diff
@@ -0,0 +1,21 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
wtrocki: Sadly Snapshots in Jest are much less readable :(. Nothing to do here anyway

.circleci/config.yml
kingsleyzissou: I removed the pretest script, it didn't seem to be working

kingsleyzissou:
@wtrocki it's the graphback-cli tests. It doesn't cause the tests to fail, but the tests don't run. Here's a screenshot from circleci:

image

packages/graphback-core/tests/__snapshots__/EngineTests.ts.snap
kingsleyzissou: I can try see if I can enable the markdown too.

packages/graphback-cli/tests/workflow-test.ts
kingsleyzissou: This is the test that didn't really run, so can't confirm if it actually works. Also, that part was done by the jest-codemods.

It was some of the other tests that I had to do manually

wtrocki:
Oh yes. This was happening before so not something new, due to db migration logic. I will fix that in separate PR.

PR #958, opened 2020-03-23
fix: Website tweaks

wtrocki:
Screenshot 2020-03-23 at 17 22 51


wtrocki:
Pushed to web already to fix current uglyness. Happy to incorporate other fixes

PR #963, opened 2020-03-24
Improvement

urvashigupta7:
The init image example on the website was not looking good. The width of the image was not proper.

## Before
Screenshot from 2020-03-24 20-25-31

## After
Screenshot from 2020-03-24 20-25-20


urvashigupta7:
@wtrocki @craicoverflow

wtrocki:
We are missing after screenshot

wtrocki:
@urvashigupta7 Linux terminal header makes this look very unprofessional. 
I was looking for something that looks very simple like here:
https://www.prisma.io Just groups of commands etc.

urvashigupta7:
Prisma.io shows different code snippets. We can do that but it will take time

wtrocki:
I meant something simple. There is so much happening on this screen now. 
I having something like https://hyper.is/ or http://erikosterberg.com/terminaljs/  where we typing a command and then **simplified** response from init like:

```
npx graphback-cli init
Select template: Apollo TypeScript
Your application was created

yarn install 
yarn start
```

wtrocki:
How this will look when we revert to previous photo?

wtrocki:
Another nice terminal: https://github.com/Kirkhammetz/termly.js

Adding this to the website should not take more than 10 minutes and they come with nice styles and even animations!

urvashigupta7:
Screenshot from 2020-03-25 23-28-12

@wtrocki I have changed the init image. This is how it looks now.

PR #964, opened 2020-03-24
Add bidirectional 1:1 relationship querying

craicoverflow:
Closes https://github.com/aerogear/graphback/issues/930

## Overview

This PR makes bidirectional one-to-one relationship querying possible by generating a "virtual" relationship field on the other model.

This is completely optional and backwards compatible. It is still possible to have unidirectional relationships in the same way as before. 

The addition of a `field` key on `oneToOne` annotations signifies a bidirectional 1:1 relationship.

## Verification

Define your data model (below) and run `graphback generate`, then `graphback db` and start your server.

```gql
"""
@model
"""
type User {
  id: ID!
  name: String
}

"""@model"""
type Address {
  id: ID!
  description: String
  """
  @oneToOne field: 'address'
  """
  resident: User!
}
``` 

Check your database structure; only the `address` table should have a foreign key.

Run these mutations and queries in order:

```gql
mutation createUser {
  createUser(input:{name:"John Doe"}) {
    name
    id
  }
}

mutation createAddress {
  createAddress(input:{description:"Address for John Doe", residentId:1}) {
    id
    description
  }
}

query findAddresses {
  findAllAddresses {
    id
    description
    resident {
      id
      name
    }
  }
}

query findUsers {
  findAllUsers {
    id
    name
    address {
      id
      description
    }
  }
}
```

## Tasks

- [x] Add unique constraint to one-to-one fields in graphql-migrations
- [x] Generate "virtual" one to one field to enable querying from both models in relationship.
- [x] Generate additional resolver for virtual relationship fetching
- [x] Add tests
- [x] Update integration tests
- [x] Update documentation

### DIFF
--- a/packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
+++ b/packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
@@ -22,7 +22,7 @@ import getKnexColumnType from '../util/getKnexColumnType';
 import { AbstractDatabase } from './AbstractDatabase'
 import getColumnTypeFromScalar, { TableColumnTypeDescriptor } from './getColumnTypeFromScalar'
 import { OneToManyRelationship } from './RelationshipTypes'
-import { Table } from './Table'
+import { Table, TableUnique } from './Table'
 import { ForeignKey, TableColumn } from './TableColumn'
 
 const ROOT_TYPES = ['Query', 'Mutation', 'Subscription']
@@ -169,6 +169,9 @@ class AbstractDatabaseBuilder {
     this.currentTable = undefined
     this.currentType = undefined
 
+    const uniques = this.getUniqueColumnConstraints(table.name, table.columns);
+    table.uniques.push(...uniques);
+
     this.database.tables.push(table)
     this.database.tableMap.set(type.name, table)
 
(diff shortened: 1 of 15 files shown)


packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
```diff
@@ -234,6 +238,10 @@ class AbstractDatabaseBuilder {
 
       // Object
     } else if (isObjectType(fieldType)) {
+      if (relationshipMarker?.virtual) {
```
wtrocki: Why we need virtual relationship marker? It seems overly complicated and reasons are not obvious to me

craicoverflow: Because in a database a one-to-one relationship is unidirectional always - only one table in a 1:1 will have a foreign key field.

So we need a way to add a 1:1 field to the other model without generating a foreign key field.

This is a virtual one-to-one, as it does not exist in the database.

`@oneToOne key: 'note_id'` ⬅️ this generates a 'note_id' column in the table for this model.

`@oneToOne key: 'note_id', virtual: true` ⬅️ this does not generate a `note_id` column for this model.

craicoverflow:
Closing as this needs a different approach.

Issue #867, opened 2020-03-08
serve Runtime does not support pagination w/findAll


ssd71:
The runtime generated when using graphql-serve shows pagination args in schema but providing them in a query does not create any noticeable difference in query result

ssd71:
I believe this is fixed now

PR #998, opened 2020-03-30
Fix findall pagination

ssd71:
This PR does the following:
- Update CRUD services to use pagination in findall
- Update Runtime Resolver Creator to use pagination in findall

Relevant issue: #867

### DIFF
--- a/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
+++ b/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
@@ -112,13 +112,25 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
 
   private usePage(query: Knex.QueryBuilder, page?: GraphbackPage, defaultLimit: number = 10, defaultOffset: number = 0) {
     if (page) {
-      page.limit = page.limit || defaultLimit
-      page.offset = page.offset || defaultOffset
-      
-      return query.offset(page.offset).limit(page.limit);
-    } else {
-      return query;
+      if (!(page.hasOwnProperty("offset"))) {
+        // If no offset is supplied
+        page.offset = defaultOffset
+      } else {
+        if (page.offset < 0) {
+          throw new Error("Please use an offset of greater than or equal to 0 in queries")
+        }
+      }
+      query.offset(page.offset)
+      // console.log('page desu: ', page)
+      if (page.hasOwnProperty("limit")) {
+        if (page.limit <= 0) {
+          throw new Error("Please use a limit of greater than 0 in queries")
+        }
+        query = query.limit(page.limit)
+      }
     }
+
+    return query;
   }
 
 }
--- a/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
+++ b/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
@@ -120,7 +120,7 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
           throw new Error("Please use an offset of greater than or equal to 0 in queries")
         }
       }
-      query.offset(page.offset)
+      query = query.offset(page.offset)
       // console.log('page desu: ', page)
       if (page.hasOwnProperty("limit")) {
         if (page.limit <= 0) {
--- a/packages/graphback-runtime/src/GraphbackPage.ts
+++ b/packages/graphback-runtime/src/GraphbackPage.ts
@@ -1,0 +1,7 @@
+/**
+ * Interface represents GraphbackPage type used for findAll queries
+ */
+export interface GraphbackPage {
+    limit?: number,
+    offset?: number,
+};
--- a/packages/graphback-runtime/src/service/CRUDService.ts
+++ b/packages/graphback-runtime/src/service/CRUDService.ts
@@ -6,6 +6,7 @@ import { GraphbackDataProvider } from "../data/GraphbackDataProvider";
 import { defaultLogger, GraphbackMessageLogger } from '../utils/Logger';
 import { GraphbackCRUDService } from "./GraphbackCRUDService";
 import { GraphbackPubSub } from "./GraphbackPubSub"
+import { GraphbackPage } from "../GraphbackPage"
 
 /**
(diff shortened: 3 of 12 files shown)


packages/graphback-runtime/src/GraphbackPage.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Interface represents GraphbackPage type used for findAll queries
+ */
+export interface GraphbackPage {
```
wtrocki: This is already defined: https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime/src/data/GraphbackDataProvider.ts#L9-L12

packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -74,10 +75,10 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
         return result;
     }
 
-    public findAll(context?: any): Promise<T[]> {
-        this.logger.log(`querying object ${this.modelName}`)
+    public findAll(page?: GraphbackPage, context?: any): Promise<T[]> {
```
wtrocki: This will not work without change in resolvers plugin and general spec. It worked as context was any

packages/graphback-runtime/src/service/GraphbackCRUDService.ts
```diff
@@ -48,7 +49,7 @@ export interface GraphbackCRUDService<Type = any, GraphbackContext = any> {
      * @param id id of the object
      * @param context context object passed from graphql or rest layer
      */
-    findAll(context?: GraphbackContext): Promise<Type[]>;
+    findAll(page?: GraphbackPage, context?: GraphbackContext): Promise<Type[]>;
```
wtrocki: Same should happen to `findBy`

wtrocki:
@ssd71 @craicoverflow This shows me that while in fact we have capabilities in the graphback db layer. Service and rest of the system is not working properly:

https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime/src/data/GraphbackDataProvider.ts#L62-L72

We need to produce integration tests that include pagination

wtrocki:
Same here - we seem to ignore context in some cases etc:
https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-resolvers/src/templates/resolverTemplates.ts#L21-L32

wtrocki:
We need to add and test findBy as well and then we are done

packages/graphback-runtime/src/GraphbackPage.ts
ssd71: I decided to extract it out and keep it separate from GraphbackDataProvider as I am using Page type in a number of places

ssd71:
Added pagination for findBy, will add integration tests next

ssd71:
Added integration tests for testing pagination, seems to work now

wtrocki:
Amazing work @ssd71! We going to review this and integrate soon

wtrocki:
Just coincidence that 0 maps to false in javascript.

ssd71:
I have found that [this](https://github.com/aerogear/graphback/blob/ebafdc1fa7469308f441fb374261ad4b556990ab/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L142) line is where it cannot distinguish between `page.limit` being `undefined` and 0
One solution would be to check for undefined using `Object.hasOwnProperty(key)`
This allows to pass a zero limit but mongo driver just returns all the objects on being passed a zero limit. We could either return an empty set or the complete set whichever makes more sense for us

WDYT

wtrocki:
If limit === 0 then return new Error()

ssd71:
Sounds great! But when do we use default limit? Should we remove default limits?

wtrocki:
Default limit will make sense only when we introduce sorting,

ssd71:
As of 76f1d31 if no limit is given, It will give the complete set of results

craicoverflow:
> How about `If limit <== 0 then return new Error()`
> 
> Full set when no limits are present

Try to see this from a client-side scenario:

A page has a table of items loaded with `findAllNotes`. There is a textbox that allows the user to enter the number of items to load per page, which is used for setting the `limit` argument.

Should we be assuming this is an error than the client will then have to handle?

In my opinion we should just return an empty data set and we should allow values of `0` for both limit and offset.

Knex and raw SQL allows for LIMIT 0 to return an empty set, so there is no reason that we should not do the same.

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -112,13 +112,25 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
...
-      page.offset = page.offset || defaultOffset
-      
-      return query.offset(page.offset).limit(page.limit);
-    } else {
-      return query;
+      if (!(page.hasOwnProperty("offset"))) {
+        // If no offset is supplied
+        page.offset = defaultOffset
+      } else {
+        if (page.offset < 0) {
```
craicoverflow, line 119: See https://github.com/aerogear/graphback/pull/998#issuecomment-607256467

wtrocki:
> Knex and raw SQL allows for LIMIT 0 to return an empty set, so there is no reason that we should not do the same.

I get your point. 
If the developer builds API that will make a request to the server to return no results we should tell him that he does something wrong :D 0 is special case of the number of wrong values that can be send like `-3000` :)

ssd71:
> A page has a table of items loaded with findAllNotes. There is a textbox that allows the user to enter the number of items to load per page, which is used for setting the limit argument.

I agree with @craicoverflow but I also think the client should avoid making that zero limit request in the first place

craicoverflow:
> > A page has a table of items loaded with findAllNotes. There is a textbox that allows the user to enter the number of items to load per page, which is used for setting the limit argument.
> 
> I agree with @craicoverflow but I also think the client should avoid making that zero limit request in the first place

It ideally should avoid making it, but should that be for the client to handle?

craicoverflow:
Approved. This detail does not have to be finalised now.

wtrocki:
Awesome work! Thank you so much @ssd71

PR #1001, opened 2020-03-30
Small improvements to CLI config command

craicoverflow:
## Tasks

- [x] If user does not select a model then do not generate a blank one.
- [x] Remove generation of `dbmigrations` from `.graphlrc`

### DIFF
--- a/packages/graphback-cli/src/components/config.ts
+++ b/packages/graphback-cli/src/components/config.ts
@@ -25,8 +25,8 @@ function postSetupMessage(): string {
 Graphback configuration successfully bootstrapped :rocket:
 
 Next Steps:
-1. Review your configuration in ${chalk.white('.graphqlrc')} file
-2. Edit the ${chalk.white('*.graphql')} files inside your model folder.
+1. Review your configuration in your ${chalk.green('.graphqlrc')} file
+2. Edit the ${chalk.green('*.graphql')} files inside your model folder
 3. Run ${chalk.cyan('graphback generate')} to generate your schema and resolvers
 `
 }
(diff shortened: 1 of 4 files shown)


packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -99,13 +99,6 @@ export const createConfig = async (database: string, client: boolean) => {
     }
   };
 
-  if(database === "pg") {
```
craicoverflow: @wtrocki I have removed this from generation as `graphback config` is for people with existing systems.

wtrocki: I could not find information how to add this so If we remove it here we need extra chapter in https://graphback.dev/docs/intro/existingproject

packages/graphback-cli/src/components/config.ts
```diff
@@ -25,9 +25,9 @@ function postSetupMessage(): string {
 Graphback configuration successfully bootstrapped :rocket:
 
 Next Steps:
-1. Review your configuration in "graphqlrc.yaml" file
-2. Edit the *.graphql files inside your model folder.
-3. Run  generate command to generate your schema and resolvers
+1. Review your configuration in ${chalk.white('.graphqlrc')} file
```
wtrocki: It might be my terminal settings, but green could be more prominent

packages/graphback-cli/src/templates/configTemplates.ts
craicoverflow: It is there in **Next Steps**:

> 4. If you are using a PostgreSQL datasource you can migrate your database to match your schema by running graphback db. See Database Migrations for more.

packages/graphback-cli/src/components/config.ts
craicoverflow: You are right - changed to green now

packages/graphback-cli/src/templates/configTemplates.ts
wtrocki: Right I could not find exacy format of the dbmigrations anywhere - we removed it from config and it is not documented in actual migrations doc

PR #1011, opened 2020-03-31
fix(resolvers): export generated code

craicoverflow:
## What

The generated resolvers JavaScript file was not exporting the resolver object.


wtrocki:
We might need snapshots for those as well

Issue #977, opened 2020-03-26
`graphback.model` configuration should be glob format


craicoverflow:
If a user places their data model in `./src/datamodel.graphql` they will need to specify the following config:

```yml
graphback:
  model: ./src
```

This will process all `.graphql` files in all subdirectories of `src`. Users may not want this and may have specific reasons to exclude certain files from their generated schema.

This also means that the generated schema is also processed as an input model and custom queries, mutations etc from data model are overwritten by ones in generated schema. This is a :bug: 

It would be really nice if the model config is a glob, so it could be like this:

```yml
graphback:
  model: ./src/*.graphql
```

wtrocki:
It could be:

- Folder
- Glob
- Array

Underlying implementation has simple means to do all of them in the single line of the code

craicoverflow:
> It could be:
> 
>     * Folder
> 
>     * Glob
> 
>     * Array
> 
> 
> Underlying implementation has simple means to do all of them in the single line of the code

Absolutely - if we can cater for everyone's preference it will be even better.

PR #1067, opened 2020-04-08
fix: accept globs and array of globs for model option in config

darahayes:
fixes: https://github.com/aerogear/graphback/issues/977

For backwards compatibility specifying a dir like `./model` will still work as it did before, but it also means you can supply globs or arrays. For example, the following are both valid now.

```yaml
extensions:
  # Graphback configuration
  graphback:
    ##  Input schema
    model: ./model/**/*.graphql
```

```yaml
extensions:
  # Graphback configuration
  graphback:
    ##  Input schema
    model:
      - ./model/comment.graphql
     - ./model/note.graphql
```

There is also a custom [`loadSchema` function in graphql-serve](https://github.com/aerogear/graphback/blob/master/packages/graphql-serve/src/loadSchema.ts#L12-L24) which I am more than happy to update too if you think it's worth it

### DIFF
--- a/docs/intro/config.md
+++ b/docs/intro/config.md
@@ -33,8 +33,9 @@ documents: ./client/src/graphql/**/*.graphql
 extensions:
   # Graphback configuration
   graphback:
-    ##  Location to file that is contains GraphQL schema that will be processed by Graphback
-    model: ./model
+    ##  Path to file that contains GraphQL schema that will be processed by Graphback
+    ## Can be an array of paths too
+    model: ./model/**/*.graphql
     ## Global flags for CRUD methods that will be used in plugins
     crud:
       create: true
--- a/packages/graphback-cli/src/components/generate.ts
+++ b/packages/graphback-cli/src/components/generate.ts
@@ -36,9 +36,23 @@ export const generateUsingPlugins = async (cliFlags: CliFlags) => {
     throw new Error(`'${graphbackExtension}' config 'plugins' section is empty. No code will be generated`);
   }
 
+  let { model } = graphbackConfig
+
+  // Formats the model path or list of paths to ensure
+  // 1. only paths from inside the project are allowed
+  // 2. if a directory is supplied *.graphql is appended
+
+  if (typeof model === 'string' && existsSync(model) && lstatSync(model).isDirectory()) {
+    model = join(config.dirpath, model, '/*.graphql')
+  } else if (typeof model === 'string') {
+    model = join(config.dirpath, model)
+  } else if (Array.isArray(model)) {
+    model = model.map((path: string) => join(config.dirpath, path))
+  }
+
   const runGeneration = async () => {
     try {
-      const schemaDocument = await project.loadSchema(join(config.dirpath, graphbackConfig.model, '/**/*.graphql'));
+      const schemaDocument = await project.loadSchema(model);
       const generator = new GraphbackGenerator(schemaDocument, graphbackConfig);
       generator.generateSourceCode();
     } catch (e) {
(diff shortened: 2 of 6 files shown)


packages/graphback-cli/src/components/generate.ts
```diff
@@ -49,15 +50,37 @@ export const generateUsingPlugins = async (cliFlags: CliFlags) => {
...
+ * Formats the model path or list of paths to ensure
+ *   * only paths from inside the project are allowed
+ *   * if a directory is supplied *.graphql is appended
+ * 
+ * @param baseDir the base directory that should be appended to the path or paths
+ * @param model a path or a list of paths
+ */
+function getModelPath(baseDir: string, model: string | Array<string>): string | Array<string> {
+  if (typeof model === 'string' && existsSync(model) && lstatSync(model).isDirectory()) {
+    return join(baseDir, model, '/**/*.graphql')
```
craicoverflow, line 79: Should we assume that

 **(a)** a basic `./model` directory means search `./model` and all of its subdirectories, or 
**(b)** should specifying `./model` mean only look in the `model` folder itself?

From a declarative point and just looking at the config without referring to the documentation it should be **(b)**.

This may break compatibility but if is wrong it's better to do this now before v1. What do you think?

I found this issue when trying out an example app on GitHub. The folder structure caused Graphback's generation to pick up `datamodel.graphql` and `generatedSchema.graphql` getting picked up as the data model because `./src` was the only model path that worked, causing some information to be lost from the final schema.

```
- src
  - datamodel.graphql
  - schema
      - generatedSchema.graphql
```

craicoverflow, line 79: Thanks to graphql-config schema loaders, we should be able to use multiple input forms. This would mean that using `./model` will always assume `.graphql` file. This is absolutely fine - I think our documentation and templates should focus on specifying glob format so that this clear.

darahayes: I would argue that the particular case you describe is a misconfiguration on the user's part but at the same time, I'm totally happy to go with option **b**. While it technically breaks compatibility, None of the current templates place things in subdirectories.

wtrocki:
To not go into a discussion with details about I think that proposed changes are removing couple really troubling issues and will help us to maintain backward compatibility and better flexibility for model creation.

packages/graphback-cli/src/components/generate.ts
wtrocki: If we phrase `./model` as corner case and backwards compatibility mode we should not care much as users will specify exactly what they looking for.

templates/ts-apollo-runtime-backend/.graphqlrc.yml
```diff
@@ -4,7 +4,7 @@ extensions:
   # Graphback configuration
   graphback:
     ##  Input schema
-    model: ./model
+    model: ./model/**/*.graphql
```
wtrocki, line 7: Needs doc update

packages/graphback-cli/src/components/generate.ts
craicoverflow: Let's go with **a** then, which is what you have :) Just docs updates are needed now.

packages/graphback-cli/src/components/generate.ts
```diff
@@ -35,9 +36,23 @@ export const generateUsingPlugins = async (cliFlags: CliFlags) => {
...
+  } else if (typeof model === 'string') {
+    model = join(config.dirpath, model)
+  } else if (Array.isArray(model)) {
+    model = model.map((path: string) => join(config.dirpath, path))
+  }
+
   const runGeneration = async () => {
     try {
-      const schemaDocument = await project.loadSchema(join(config.dirpath, graphbackConfig.model, '/**/*.graphql'));
+      const schemaDocument = await project.loadSchema(model);
```
darahayes, line 55: I noticed I didn't include this in the first draft of the PR :)

docs/intro/config.md
```diff
@@ -33,8 +33,9 @@ documents: ./client/src/graphql/**/*.graphql
 extensions:
   # Graphback configuration
   graphback:
-    ##  Location to file that is contains GraphQL schema that will be processed by Graphback
-    model: ./model
+    ##  Path to file that contains GraphQL schema that will be processed by Graphback
+    ## Can be an array of paths too
+    model: ./model/**/*.graphql
```
darahayes, line 38: @craicoverflow is this good enough or should I add some actual sentence to the doc explaining it?

wtrocki: It is self explanatory for me

darahayes:
@craicoverflow thanks for the review, I've updated it now and I think it's a bit cleaner. I've also updated the docs and other places where we were using `./model` as a value

docs/intro/config.md
craicoverflow: I think it is self explanatory :)

craicoverflow:
@darahayes I have just noticed that the runtime template fails to run because it loads the model the old way:

https://github.com/aerogear/graphback/blob/4a10783e73dd1d30522289cb8831ebbcdc712226/templates/ts-apollo-runtime-backend/src/loadSchema.ts#L12-L24

This can probably be quick fixed for now - I was in the process I removing the custom `loadSchema` from the template [here](https://github.com/aerogear/graphback/pull/1069/files#diff-132b30b3d9176017b87eba1eb7895816R25).

The model should not be loaded in the runtime template at all, so this can probably get resolved later in https://github.com/aerogear/graphback/issues/1071

PR #1069, opened 2020-04-08
Tidy up template applications

craicoverflow:
The template apps had become a little bit less maintained and this is just an effort to clean up what we can to make them simpler to adopt for users.

## Tasks

- [x] Removed async loading of graphql-config files causing everything in the templates to be wrapped in an async function.
- [x] Removed unused dependencies and moved some to devDevepenncies
- [x] Use environment variables to load database configuration when used purely for data source configuration. The examples were using `dbmigrations` for this which was incorrect usage.
- [x] Organised and simplified use of util files in templates


templates/ts-apollo-fullstack/server/.env
```diff
@@ -0,0 +1,6 @@
+DB_CLIENT=pg
```
wtrocki, line 1: 😍

templates/ts-apollo-fullstack/server/src/index.ts
```diff
@@ -1,30 +1,25 @@
...
 import cors from 'cors';
 import express from 'express';
-import http from 'http';
-
+import dotenv from 'dotenv'
 import { createApolloServer } from './graphql';
 
-async function start() {
-  const app = express();
+dotenv.config()
```
wtrocki, line 7: This should be done at very very top of the imports. If not it can cause bugs where environment will not be set.

templates/ts-apollo-fullstack/server/src/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { loadConfigSync } from 'graphql-config';
+import Knex from 'knex';
+
+export function getProjectConfig() {
```
wtrocki, line 4: I talked about that with our community. We really need that stuff exposed in the CLI so we can avoid asking people to have this util copied into their code. This is very generic and it should not change at all.

templates/ts-apollo-fullstack/server/src/index.ts
craicoverflow: Added now in https://github.com/aerogear/graphback/pull/1069/commits/9f8acb4cadc0a42cf77178ce7fd203e5eebd2448

templates/ts-apollo-fullstack/server/src/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  const config = loadConfigSync({
+    rootDir: process.cwd(),
```
wtrocki, line 6: this is also where also loaders are supplied so we can have more control how we load config and schema

templates/ts-apollo-fullstack/server/src/utils.ts
craicoverflow: Agreed.

templates/ts-apollo-runtime-backend/src/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  return config.getDefault()
+}
+
+export function createDB() {
+  let port;
+  if (process.env.DB_PORT) {
+    port = parseInt(process.env.DB_PORT, 10)
+  }
+
+  return Knex({
```
wtrocki, line 26: I would like to have db created in separate file (not hidden in utils)

templates/ts-apollo-runtime-backend/src/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { loadConfigSync } from 'graphql-config';
+import Knex from 'knex';
+
+export function getProjectConfig() {
```
wtrocki, line 4: If this will be moved to cli then utils file can be removed

templates/ts-apollo-fullstack/server/src/utils.ts
wtrocki: The only issue is graphql-cli feature to recognize projects. If we load that directly we might not support project use cases, but we do not support that here anyway.

wtrocki: Ignore previous convoluted comments. I think your proposal makes sense.

templates/ts-apollo-runtime-backend/src/utils.ts
craicoverflow: Can you explain why it should be in the CLI or how that would work? Do you mean core?

wtrocki: We will need it somewhere but what we have currently is graphql-config only on the cli. Moving this code anywhere will mean moving it to CLI as CLI should be imported into a template anyway to execute generate. Moving to core will mean bringing some dependencies there that are not relevant for the rest but leaving decision for yourself. Generally WDYT about this idea?

wtrocki:
I'm happy to merge that. Config improvements might require quick chat as I'm not sure if I explained the problem properly.

templates/ts-apollo-runtime-backend/src/utils.ts
craicoverflow: CLI should only be a devDependency, so it shouldn't be used in the template code though, right?

craicoverflow:
> I'm happy to merge that. Config improvements might require quick chat as I'm not sure if I explained the problem properly.

I think loading the config is not an issue, given we can just use graphql-config directly.

Issue #1024, opened 2020-04-01
[Schema] Separate input types for query fields and mutation fields


craicoverflow:
Currently only a single input type is generated for each model and this is used for queries and mutations. This assumes all fields are non-required to be able to cater for both cases.

We need to generate a separate `<ModelName>Fields` input type specifically for queries, and update the `<ModelName>Input` input type to map the required constraint from the model.

wtrocki:
We should have input types that actual match the require fields etc. to enable better validation

We should have 2 input types:

Used for filtering (without required fields)
Used for mutations (with required fields same as in original type)

Closing https://github.com/aerogear/graphback/issues/844 as duplicate

craicoverflow:
```gql
type Note {
  id: ID!
  title: String
  description: String!
}
```

Two input types are generated:

```gql
type NoteData {
  id: ID # not required as this can be auto-generated
  title: String
  description: String!
}

type NoteFilter {
  id: ID
  title: String
  description: String
}
```

`NoteData` respects the required level of the fields - except for ID. This is because ID can be an auto-generated field and the input type is used for both update and create mutations. Updates require the ID for updating an item by its ID. Creates do not require IDs as they can be auto-generated by the database.

One solution for this is to have a separate input type for create and update mutations, but when we action https://github.com/aerogear/graphback/issues/1031 the CRUD Spec will change again anyway, so that update mutations would accept two types:

`updateNote(filter: NoteFilter!, data: NoteData!)`

At this point ID can be commpletely removed from NoteData as it would no longer be used.

craicoverflow:
Outdated by https://github.com/aerogear/graphback/issues/1031

PR #1081, opened 2020-04-09
feat(schema): generate separate input type for fields

craicoverflow:
Closes https://github.com/aerogear/graphback/issues/1024

Generate a `<Model>Filter` and `<Model>Data` input type for filtering and mutations.

## Tasks

- [x] Generate separate input types for filtering fields
- [x] Use filter type in client documents
- [x] Add unit tests
- [ ] Update documentation

## Verification

Generate a new schema. Each type should you have 2 input types - one for filtering and one for data input.


packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -35,7 +35,7 @@ export const findAllQuery = (t: GraphQLObjectType) => {
 export const findQuery = (t: GraphQLObjectType) => {
   const fieldName = getFieldName(t.name, GraphbackOperationType.FIND)
 
-  return `query ${fieldName}($fields: ${t.name}Input!, $limit: Int, $offset: Int) {
+  return `query ${fieldName}($fields: ${t.name}Fields!, $limit: Int, $offset: Int) {
```
wtrocki, line 38: In terms of names WDYT about:

`${t.name}Data` for creation
`${t.name}Filter for query and mutation filtering

craicoverflow: Agreed - data is more clear and fields was always wrong :D

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
```diff
@@ -10,20 +10,27 @@ type Comment {
...
-  deleteNote(input: NoteInput): Note!
-  createComment(input: CommentInput): Comment!
-  updateComment(input: CommentInput): Comment!
-  deleteComment(input: CommentInput): Comment!
+  createNote(data: NoteData!): Note!
+  updateNote(data: NoteData!): Note!
+  deleteNote(filter: NoteFilter!): Note!
+  createComment(data: CommentData!): Comment!
+  updateComment(data: CommentData!): Comment!
+  deleteComment(filter: CommentFilter!): Comment!
```
wtrocki, line 33: @craicoverflow Missing rename

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
```diff
@@ -10,20 +10,27 @@ type Comment {
   note: Note
 }
 
-input CommentInput {
+input CommentData {
+  id: ID
+  text: String
```
wtrocki, line 15: Missing mapping to required fields

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
```diff
@@ -36,25 +43,30 @@ type Note {
...
-  deletedNote(input: NoteInput): Note!
-  newComment(input: CommentInput): Comment!
-  updatedComment(input: CommentInput): Comment!
-  deletedComment(input: CommentInput): Comment!
+  newNote(data: NoteData): Note!
+  updatedNote(data: NoteData): Note!
+  deletedNote(filter: NoteFilter): Note!
+  newComment(data: CommentData): Comment!
+  updatedComment(data: CommentData): Comment!
+  deletedComment(filter: CommentFilter): Comment!
```
wtrocki, line 71: Missing rename

templates/ts-apollo-fullstack/server/src/resolvers/resolvers.ts
```diff
@@ -9,85 +9,85 @@ export default {
   Note: {
     comments: (parent, args, context) => {
       return context.Comment.batchLoadData("noteId", parent.id, context)
-    }
+    },
```
wtrocki, line 12: Invalid `,`

templates/ts-apollo-fullstack/server/src/resolvers/models.ts
```diff
@@ -4,15 +4,15 @@ export const models = [
     pubSub: {
       publishCreate: true,
       publishUpdate: true,
-      publishDelete: true
-    }
+      publishDelete: true,
+    },
```
wtrocki, line 8: Invalid generate

wtrocki:
Missing docs change, Breaking change note etc.

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
craicoverflow: `filter` is valid here, no? It should not be `data` since data is used to insert the data, whereas with a delete we are using the data to do a WHERE clause.

Maybe `filter` is misleading in all cases, `where` might be better?

wtrocki: Filter is very confusing. For maintaining consistency we need delete to have more than just id (once again - offix cases) and we do not know what field is being used. Filter is very confusing as it suggest you can filter deltes (where it is all id based)

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
craicoverflow: The data model has not changed and this is the correct mapping - text and description are not required in the model. ID is not required to allow both update and create using the input. To fix this we would have to create a separate input type for each when advanced CRUD updates are available with `updateComment(filter, data)`

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
craicoverflow: the delete argument is used for querying, so it should be filter.

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
wtrocki: This is where we should probably do an investigation and see how other systems do it.

wtrocki: I will suggest to rename it to data and we can see later how to make it work

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
wtrocki: It might be misscommunication. Requires chat

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
wtrocki: So then what is the point of having this change merged? We have two exactly the same input types?

wtrocki: Requires discussion

templates/ts-apollo-fullstack/server/src/resolvers/resolvers.ts
craicoverflow: Nothing in code has changed to make this happen, I investigated this and it appears this is down to the Prettier formatter:

https://github.com/aerogear/graphback/blob/9df2aa574299db464b7f77293bb493f1f799c522/packages/graphback-codegen-resolvers/src/templates/formatter.ts#L16

https://prettier.io/blog/2020/03/21/2.0.0.html

Prettier v2 has changed the default value for `trailingCommas` to 'all'

https://github.com/prettier/prettier/issues/68

craicoverflow: Is there any negative points to keeping it using the es5 standard? This is a generate file anyway.

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
craicoverflow: No - this is a bad example because `Comment` does not have any non-ID required fields, so the data input type does not have required fields.

You can see that `NoteData` maps required fields from `Note` because it has a required field  - Comment does not.

https://github.com/aerogear/graphback/blob/52f591e054281477c9fbd623a706690d134b2aee/templates/ts-apollo-fullstack/server/src/schema/schema.graphql#L48

With our current limited CRUD capabilities the only way to resolve this is to have a separate input type for update and delete.

- CreateCommentData - maps input fields EXCEPT for primary key
- UpdateCommentData - maps input fields WITH primary key.

We could do this now, but the work would become obsolete when we change the CRUD implementation/spec to:

`updateComment(where, data)`

craicoverflow: So what do you think? Create 3 input types for now and remove one at a later stage? I suppose this makes the most sense as it fixes the current implementation.

templates/ts-apollo-fullstack/server/src/resolvers/resolvers.ts
wtrocki: Knowing it was prettier we can ignore that

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
wtrocki: 1) Apply my suggestions and merge this knowing that this is still not getting us in the place we want to be.

2) Work on top of this issue to figure out our desired CRUD schema (possible options we can take) and propose it in demo/doc

3) Alpaca farming

Issue #1088, opened 2020-04-09
Cannot create a user record with only an id


lastmjs:
Here is my model:

```graphql
"""
@model
"""
type User {
  id: ID!
}
```

And here's the generated schema:

```graphql
## NOTE: This schema was generated by Graphback and should not be changed manually

type Mutation {
  createUser(input: UserInput): User!
  updateUser(input: UserInput): User!
}

type Query {
  findAllUsers(limit: Int, offset: Int): [User]!
  findUsers(fields: UserInput, limit: Int, offset: Int): [User]!
}

"""@model"""
type User {
  id: ID!
}

input UserInput {
  id: ID
}
```

I would expect to be able to execute a mutation like this:

```graphql
mutation {
  createUser {
    id
  }
}
```

But when I do, I get:

```json
{
  "errors": [
    {
      "message": "No result from database: Cannot create user",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createUser"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: No result from database: Cannot create user",
            "    at PgKnexDBDataProvider.<anonymous> (/home/lastmjs/development/single-shuffle/node_modules/@graphback/runtime-knex/src/PgKnexDBDataProvider.ts:28:15)",
            "    at Generator.next (<anonymous>)",
            "    at fulfilled (/home/lastmjs/development/single-shuffle/node_modules/tslib/tslib.js:104:62)",
            "    at processTicksAndRejections (internal/process/task_queues.js:93:5)"
          ]
        }
      }
    }
  ],
  "data": null
}
```

There seems to be something wrong with default values, perhaps the postgres SQL query for having no values, which needs to be something like `insert into public.user values(default);`, is not being constructed correctly.

craicoverflow:
Hey @lastmjs! Yeah, that's a bug! Thanks for letting us know. I've a fix up in https://github.com/aerogear/graphback/pull/1106 - we will get it released ASAP.

craicoverflow:
```gql
createNote(input: NoteInput): Note
```

Graphback's resolvers pass `args.input` to the database, which is used to create the item.

```gql
mutation {
  createUser {
    id
  }
}
```

By using the mutation this way, `input` is undefined. If you want to create a user without supplying data you can do:

```gql
mutation createUser {
  createUser(input: {}) {
    id
  }
}
```

[...]

PR #1106, opened 2020-04-14
fix: allow creating item in Postgres when all fields are automatically filled by database

craicoverflow:
Fixes https://github.com/aerogear/graphback/issues/1088

When a mutation is sent with no fields, Graphback tries to insert `undefined` into the database.

This is a problem because if the fields are automatically added (by a DEFAULT value or if it is an incrementing field), Graphback does not construct the insert query correctly and the operation fails.

This maps `undefined` mutation data to an empty object.

## Verification

See https://github.com/aerogear/graphback/issues/1088

### DIFF
--- a/packages/graphback-core/src/db/dataMapper.ts
+++ b/packages/graphback-core/src/db/dataMapper.ts
@@ -1,38 +1,38 @@
 import { ModelTableMap } from './buildModelTableMap';
 
 export interface TableDataMap {
-    idField?: TableID
-    table?: string
-    data?: any
-    fieldMap?: any
+  idField?: TableID
+  table?: string
+  data?: any
+  fieldMap?: any
 }
 
 export interface TableID {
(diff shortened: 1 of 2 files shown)


wtrocki:
@craicoverflow This is symptom of the other issue - Generally data should not be undefined at all. GraphQL validation will now allow creating object. Currently, it does because we have the same inputype for filtering and creation

Also ID mapping should not be responsible for returning data at all (for Mongo we actually mapping some items in the data like _id to id etc.
Most of our providers do not use this at all.

craicoverflow:
> Works locally. Approving as a temporary fix. Explanation above

I'm not going to merge this, I feel that https://github.com/aerogear/graphback/pull/1081 . https://github.com/aerogear/graphback/pull/1081 fixes this by ensuring `data` is required.

PR #1130, opened 2020-04-17
fix: Mongo batching with string as key

wtrocki:
Verification possible in AtHomeCrisisCommunity


wtrocki:
We save connections without indexing and as strings for Mongo which is not effective. This simple patch is partially solving issue by checking if is are type of ObjectID and then transforming them to strings. If they are strings nothing will happen

wtrocki:
@craicoverflow This ideally would require integration tests but since we will need to refine this anyway I will hold of with creating ones

wtrocki:
@craicoverflow There is bug for that as well. Generally if we use filter then objects are assigned to the wrong item. We need to return empty array for every batch ID that is why filter is not perfect here - we do want to return empty array if no batch results were found

Issue #1120, opened 2020-04-16
Disabling generation on single objects affects others 


wtrocki:
```
"""
@crud.create: false 
@crud.update: false
@crud.delete: false
"""
type DistributionCentre {
  id: ID!
  name: String
  address: String
  address2: String
  city: String
  lat: String
  long: String
}

```

<img width="759" alt="Screenshot 2020-04-16 at 09 53 23" src="https://user-images.githubusercontent.com/981838/79435889-21793600-7fc8-11ea-8306-06c73b2e5c14.png">

PR #1131, opened 2020-04-17
fix: Problem with CRUD config being global

wtrocki:
Using Object.assign without understanding how it works :)

See https://github.com/aerogear/graphback/issues/1120

### DIFF
--- a/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
+++ b/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
@@ -75,7 +75,7 @@ export class GraphbackCoreMetadata {
     private buildModel(modelType: GraphQLObjectType, relationships: FieldRelationshipMetadata[]): ModelDefinition {
         let crudOptions = parseAnnotations('crud', modelType.description)
         //Merge CRUD options from type with global ones
-        crudOptions = Object.assign(this.supportedCrudMethods, crudOptions)
+        crudOptions = Object.assign({}, this.supportedCrudMethods, crudOptions)
 
         return { graphqlType: modelType, relationships, crudOptions };
     }
--- a/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
+++ b/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
@@ -28,7 +28,7 @@ export class GraphbackCoreMetadata {
 
     public constructor(globalConfig: GraphbackGlobalConfig, schema: GraphQLSchema) {
         this.schema = schema;
-        this.supportedCrudMethods = Object.assign(defaultCRUDGeneratorConfig, globalConfig.crudMethods)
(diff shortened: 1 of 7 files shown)


craicoverflow:
I don't believe this fixes https://github.com/aerogear/graphback/issues/1120 - in fact the original usage of `Object.assign` is perfectly valid and both have the same result:

`Object.assign({}, obj1, obj2)` merges obj2 into obj1 (overriding values with same key), and this is merged into an empty object.

`Object.assign(obj1, obj2)` merges obj2 into obj1 also overriding values with same key.

This has nothing to do with model-level CRUD generation, it simply for assuming global CRUD defaults and overriding them if ones are defined in the config file.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -28,7 +28,7 @@ export class GraphbackCoreMetadata {
 
     public constructor(globalConfig: GraphbackGlobalConfig, schema: GraphQLSchema) {
         this.schema = schema;
-        this.supportedCrudMethods = Object.assign(defaultCRUDGeneratorConfig, globalConfig.crudMethods)
+        this.supportedCrudMethods = Object.assign({}, defaultCRUDGeneratorConfig, globalConfig.crudMethods)
```
craicoverflow, line 31: This can be reverted

wtrocki: This will work only if a single metadata object is present. 
Having multiple metadata objects will still fail. 
Will add test for that case.

wtrocki: Unit tests now cover this case as well

packages/graphback-core/tests/MetadataTest.ts
```diff
@@ -0,0 +1,33 @@
+//tslint:disable-next-line: match-default-export-name no-implicit-dependencies
+import { readFileSync } from 'fs';
+import { buildSchema } from 'graphql';
+import { GraphbackCoreMetadata } from '../src'
+import { existedOperationTypeMessage } from 'graphql/validation/rules/UniqueOperationTypes';
+
+const schemaText = readFileSync(`${__dirname}/mock.graphql`, 'utf8')
```
craicoverflow, line 7: I'm personally not a fan of loading the mock data model from a separate file because it is difficult to quickly see what the model looks like from the test when updating or reviewing. Because this file that has not changed it is not included in the PR.

Including the schema in the test makes it easier to read

https://github.com/aerogear/graphback/blob/dbc498831d241913fe9262449e1ae93a00045a1a/packages/graphback-core/tests/RelationshipMetadataBuilderTest.ts#L69

wtrocki: There are multiple instances of that so I will create issue to get that done

PR #1136, opened 2020-04-17
feat: add mongodb template

craicoverflow:
This is a clone of the existing runtime template and modified to work with MongoDb out of the box.
Once this is merged I will update templates-release template


templates/ts-apollo-mongodb-runtime-backend/README.md
```diff
@@ -0,0 +1,79 @@
+## Graphback Runtime Template
```
craicoverflow, line 1: Need to change this

templates/ts-apollo-mongodb-runtime-backend/.env
```diff
@@ -0,0 +1,6 @@
+DB_USER=mongodb
```
wtrocki, line 1: This is from datasync example right? We use there some dodgy Red Hat Mongo for compatibility. 
Normal Mongo do not require any authentication etc. We might just not need it for upstream context

templates/ts-apollo-mongodb-runtime-backend/.graphqlrc.yml
```diff
@@ -0,0 +1,24 @@
+## GraphQL Config Generated by Graphback
+## Please review configuration and adjust it for your own project
+schema: ./src/schema/*.graphql
+documents: ./client/src/graphql/**/*.graphql
```
wtrocki, line 4: This might not be required as there is no client queries.
Generally what is the point of having this config file for runtime example? 
All parameters can be setup remotely. For postgress we needed it for migration purposes.

templates/ts-apollo-mongodb-runtime-backend/.env
craicoverflow: No this is the exact config from `graphback config`

templates/ts-apollo-mongodb-runtime-backend/.graphqlrc.yml
craicoverflow: Yeah forgot about this.

templates/ts-apollo-mongodb-runtime-backend/src/runtime.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Method used to create runtime schema
+ */
+export const createRuntime = async () => {
+  const db = await connectDB();
+
+  const projectConfig = loadConfigSync({
+    extensions: [
+      () => ({ name: 'graphback' }),
+      () => ({ name: 'dbmigrations' })
```
wtrocki, line 17: Not needed here

templates/ts-apollo-mongodb-runtime-backend/src/runtime.ts
```diff
@@ -0,0 +1,32 @@
...
+    ]
+  }).getDefault()
+
+  // const projectConfig = getProjectConfig();
+  const graphbackConfig = projectConfig.extension('graphback');
+  const model = projectConfig.loadSchemaSync(path.resolve(graphbackConfig.model));
+
+  const pubSub = new PubSub();
+  const runtimeEngine = new GraphbackRuntime(model, graphbackConfig);
```
wtrocki, line 26: Ah.. We require graphback config for runtime - where what we use is just crud. I think it is worth to make this small (breaking change) here to enable runtime to work without config file. 
This can be separate issue, but generally it will reduce number of steps required to create runtime

wtrocki:
@craicoverflow Sorry that I used that PR to point out some existing issues.. 
We have outstanding discussion to remove `graphback db`. The remaining part was that it was needed for runtime cases, however if runtime will perform migration in the code (on start) this will actually give developers much better workflow and help us with fixing couple issues with scalars. I'm leaving up to you how we deal with this but it is great opportunity to clear things up and minimize some tech debt we have.

templates/ts-apollo-mongodb-runtime-backend/.env
wtrocki: Do we create .env file there?

templates/ts-apollo-mongodb-runtime-backend/src/runtime.ts
craicoverflow: We still need the config for:

- global CRUD configuration
- plugins

wtrocki: Yep.. But why not pass it over the JS/TS. Same with plugins - as runtime do not care about them at all we can just bundle them and pass them programmatically. This will be dream come true as it will make entire runtime setup just 2 method calls (setup services and setup runtime) nothing else will be required including some strange config setup.

wtrocki: It might be too much for this issue but this is great place to talk about this and explore these options as we have context of he code.

templates/ts-apollo-mongodb-runtime-backend/.env
craicoverflow: no I coped the database values from the docker-compose file

craicoverflow:
> @craicoverflow Sorry that I used that PR to point out some existing issues..
> We have outstanding discussion to remove `graphback db`. The remaining part was that it was needed for runtime cases, however if runtime will perform migration in the code (on start) this will actually give developers much better workflow and help us with fixing couple issues with scalars. I'm leaving up to you how we deal with this but it is great opportunity to clear things up and minimize some tech debt we have. This problem is even more present in Mongo where we do not have db migration at all.
> 
> EDIT:
> 
> Maybe some quick wins in this PR and follow up issue to cover this?

FYI `graphback db` is not needed for runtime cases as runtime template has programmatic migrations.

wtrocki:
> FYI graphback db is not needed for runtime cases as runtime template has programmatic migrations.

This makes it even more appealing to kill all config stuff now

wtrocki:
Also in generation scenario. Config should not be used inside the application at all. We use it to reuse migration to connect to the app but that should change and we might have small duplication of app and migration config

Issue #1089, opened 2020-04-09
Do not remove custom directives from the model


lastmjs:
It looks like custom directives that I place into my model are removed in the generated schema. The definitions are not removed, which is nice, but not very useful without keeping the directives themselves. It would be great if custom directives placed in the model could be kept in the generated schema. This was a major pain point with Prisma and implementing permissions with custom directives. They were removed, causing me to have to write a parser to place them back in so that I could have my directives executed against the generated schema, which I imagine is desirable for a variety of use cases.

wtrocki:
Critical issue. Going to be resolved in the next release

craicoverflow:
We are using `printSchema` to convert the schema object to a string before writing to a file. It [seems to be expected behaviour](https://github.com/graphql/graphql-js/issues/869#issuecomment-309899380) that custom directives are not added.

We should use something other than `printSchema` for this.

wtrocki:
GraphQL-Toolkit has printSchema with directives. 
https://gist.github.com/arjun-kava/0bd38e0c8feec360284677b1ee31728a

Generally if we were using GraphQL-Compose for that purpose (as everywhere else) we will be fine

craicoverflow:
As I said, GraphQL-Toolkit changes order completely and adds a `query` root type.

I tested with GraphQL-Compose and it has the same issue.

craicoverflow:
Yeah I already tested these - GraphQL-Toolkit changes print order and adds a root `query` type. GraphQL-Compose also loses directives.

craicoverflow:
Directives are getting removed from fields in SchemaComposer. When we add the schema string or object to the constructor as we do now (see example below) then the final schema will have directive definitions, but no directives in use.

```ts
import { SchemaComposer } from 'graphql-compose'
import { loadSchemaSync } from '@graphql-toolkit/core';

const schema = loadSchemaSync(`directive @test(
  reason: String = "No longer supported"
) on FIELD_DEFINITION | ENUM_VALUE

type ModifyMe {
  id: ID! @test(reason: "asjdk")
}

type Query {
  hello: ModifyMe
}`, { loaders: [] })

const schemaComposer = new SchemaComposer(schema);

const schemaStr = schemaComposer.toSDL({ exclude: ['ID', 'String', 'Int', 'Boolean', 'Float'] })

console.log(schemaStr)
```

This can be resolved by _adding_ typedefs string separately:

```ts
const schemaString = `directive @test(
  reason: String = "No longer supported"
[...]

PR #1147, opened 2020-04-20
fix: Print schema with directives

craicoverflow:
Fixes https://github.com/aerogear/graphback/issues/1089

This fix prints custom directives to the final schema

### DIFF
--- a/packages/graphback-codegen-offix/package.json
+++ b/packages/graphback-codegen-offix/package.json
@@ -37,7 +37,7 @@
     "@graphback/core": "0.11.4",
     "@graphql-toolkit/common": "0.10.4",
     "@graphql-toolkit/core": "0.10.4",
-    "graphql-compose": "7.14.2",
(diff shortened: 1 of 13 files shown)


wtrocki:
@craicoverflow Strange that we do not have any snapshot for verifying this schema. Would it be possible to add one (SDL string) to see outputs for example model

craicoverflow:
> @craicoverflow Strange that we do not have any snapshot for verifying this schema. Would it be possible to add one (SDL string) to see outputs for example model

Yeah it seems like an important one to have! Of course!

craicoverflow:
@wtrocki there is a snapshot but it did not have any custom directives - and it used `printSchema` this stripping out custom directives. Added in https://github.com/aerogear/graphback/pull/1147/commits/981da21886503f3b11131b303fe2a65c1f61d974

wtrocki:
Perfect

PR #1151, opened 2020-04-20
chore(release): bump to version 0.12.0

craicoverflow:
Update version to Graphback 0.12.0


docs/releases.md
```diff
@@ -8,6 +8,31 @@ title: Releases
 This file contains changes and migration steps for the Graphback project. 
 Please follow individual releases for more information.
 
+# 0.12.0
+
+##### New Features
```
wtrocki, line 13: Not really the format we were using but it is ok to change it

craicoverflow: Essentially the same but using a title - easier to read.

PR #1160, opened 2020-04-21
fix(Apollo+Postgres+Runtime): add missing dependency

craicoverflow:
The Apollo PostgreSQL Runtime template is missing `@graphback/runtime-knex` which is required to create data sources and start the server.


wtrocki:
It is really cool that we can push that without relase

PR #1176, opened 2020-04-25
test: make sure that `GraphQLClientCreatorTest.ts` compiles.

machi1990:
The test was failing because the property `fragmentOnly` was missing from a required type for the constructor of `ClientGeneratorPluginConfi
g`. The command `yarn test` errored the below message:

```
packages/graphback-codegen-client/tests/GraphQLClientCreatorTest.ts:34:39 - error TS2345: Argument of type '{ format: "graphql"; outputPath: string; }' is not assignable to parameter of type 'ClientGeneratorPluginConfig'.

Property 'fragmentOnly' is missing in type '{ format: "graphql"; outputPath: string; }' but required in type 'ClientGeneratorPluginConfig'.

 34   const plugin = new ClientCRUDPlugin({ format: 'graphql', outputPath: './tmp' })
```

/cc @wtrocki @craicoverflow


wtrocki:
@machi1990 This is strange as master and PR that introduced that change was building.

machi1990:
> @machi1990 This is strange as master and PR that introduced that change was building.

@wtrocki Yes, strange. 

Saw your commit, that's another way to go abut it. 

Thanks for chiming in.

PR #1178, opened 2020-04-25
update GraphQLResolverCreatorTest snapshots

machi1990:
- Delete phantom snapshots that were generated by Ava
- Update Jest snapshots to take into account latest changes in master. 
- Also explicit set the `trailingComma` to `all` prettier formatting options which seem to be projects default see https://github.com/aerogear/graphback/blob/a8813d451487a91114c51db5ee246fafc0446c07/.prettierrc#L3 

/cc @wtrocki


machi1990:
> * Delete phantom snapshots that were generated by Ava

Seems like there are couple other files that were generated by Ava and the test runner itself is still there as shown by a quick search on GH https://github.com/aerogear/graphback/search?q=ava&unscoped_q=ava e.g 

Are we using jest or ava or both? 

@wtrocki @craicoverflow

machi1990:
Great, I think we can go ahead and remove all `Ava` references?

wtrocki:
Yes. Anything related to ava should be gone

machi1990:
> Yes. Anything related to ava should be gone

Will prepare a PR some time later today.

wtrocki:
@machi1990 keep it for the week. No need to do boring stuff on weekend

PR #1181, opened 2020-04-25
chore: remove trailing Ava references by deleting phantom Ava snapshots and removing the Ava dependency

machi1990:
Also update old jest snapshots

See this comment https://github.com/aerogear/graphback/pull/1178#issuecomment-619357364


templates/ts-apollo-fullstack/.graphqlrc.yml
```diff
@@ -1,40 +1,27 @@
 ## GraphQL Config Generated by Graphback
 ## Please review configuration and adjust it for your own project
-schema: ./server/src/schema/*.graphql
```
wtrocki, line 3: I think this file was commited by mistake (should be reverted?)

machi1990: Good catch.

machi1990: I noticed a couple of other `templates/apollo...` generated files , should they be ignored?

PR #1194, opened 2020-04-27
fix: correct pass resolvers to ApolloServer

machi1990:
See https://github.com/aerogear/graphback/commit/673fce28a700d4ccc5ef72642a6226e446e61bb6#r38764338


machi1990:
Thanks @wtrocki

PR #1196, opened 2020-04-27
docs(datamodel): add many-to-many documentation

craicoverflow:
Add information about how to define a many-to-many relationship.

### DIFF
--- a/docs/db/migrations.md
+++ b/docs/db/migrations.md
@@ -154,6 +154,44 @@ type Note {
 
 This maps to `comment.note_id` in the database.
 
+##### ManyToMany
+
+To create a many-to-many relationship, add a model for your join table and use two one-to-many relationship mappings to create the relationship.
+
+```graphql
+""" 
+@model 
+"""
+type Note {
+  id: ID!
+  title: String!
+  description: String
+  """
+  @oneToMany field: 'note'
+  """
+  authors: [NoteAuthor]
+}
+
+"""
+@model
+"""
+type NoteAuthor {
+  id: ID!
+}
+
+"""
+@model
+"""
+type User {
+  id: ID!
+  name: String
+  """
+  @oneToMany field: 'author'
+  """
+  notes: [NoteAuthor]
+}
+```
+
 #### Default field value
 
 ```graphql
--- a/docs/intro/datamodel.md
+++ b/docs/intro/datamodel.md
@@ -120,3 +120,41 @@ type Profile {
   user: User!
 }
 ```
+
+### ManyToMany
+
+To create a many-to-many relationship, add a model for your join table and use two one-to-many relationship mappings to create the relationship.
+
+```graphql
+""" 
+@model 
+"""
+type Note {
+  id: ID!
+  title: String!
+  description: String
+  """
+  @oneToMany field: 'note'
+  """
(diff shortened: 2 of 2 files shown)


docs/db/migrations.md
```diff
@@ -154,6 +154,44 @@ type Note {
...
+  """
+  @oneToMany field: 'note'
+  """
+  authors: [NoteAuthor]
+}
+
+"""
+@model
+"""
+type NoteAuthor {
```
wtrocki, line 178: So generally people will get a lots of pointless boilerplate generated for this - entire crud and will need to manually remove all CRUD capabilities for each type like that.

docs/db/migrations.md
```diff
@@ -154,6 +154,44 @@ type Note {
...
+  id: ID!
+}
+
+"""
+@model
+"""
+type User {
```
wtrocki, line 185: While I do see this working on the db level. It is hard for me to get how this can be queried with Graphback now.

docs/db/migrations.md
```diff
@@ -154,6 +154,44 @@ type Note {
```
wtrocki, line 154: Would it be possible to move that docs from db. Contents here focusing on building schema relationships and many people struggling to find it.

docs/db/migrations.md
wtrocki:
```
getUser(...){
 notes: {
   note: {
      ??
   }
 }
}
```

I need to test this in production app we have - there are couple examples where this relation makes sense. There is also problem with lack of client side generation capabilities on the client to support 3 levels of relations now

wtrocki:
I would like to see how this works with the Mongo sample app first before we merge to see, but looks great for me

docs/db/migrations.md
craicoverflow: So, remove it from database section completely? It is use for migrations though.

wtrocki: hmm.. That is actually good question. I see this as way to design schema. 

Then db migrations can literally reference this information explaining how db is generated for individual use cases etc. We are missing that at the moment

craicoverflow: Yep I considered this but then there are variations between the two. For example:

- `@db.name` annotation is used in migrations, but will do nothing in data model, i.e. we do not auto-map so it is misleading to put it there.
- Migrations require the `id: ID` field, but Graphback does not require this in runtime or generation, just migrations does.

wtrocki: Ah.. let me think about it then

wtrocki:
This doesn't work that nice on both postgress and mongo:
- only 2 levels of client side queries are generated
- on mongo we are getting complex queries that are not indexed properly
- tons of extra crud methods on relation object - need to disable every single crud to not take it

While docs are ok.. This type of relation is not production ready
what we can do:
- inspect indexes for performance.in separate issue
- create flag for model to include it but do not generate crud methods for it

wtrocki:
@craicoverflow we can mege.if you agree with suggested approach

wtrocki:
We have plans to address this as part of the Mongodata provider that will give performance. I have created separate issue 
https://github.com/aerogear/graphback/issues/1228

PR #1211, opened 2020-04-29
New CRUD specification 

craicoverflow:
**Note: This is a rough, prototype implementation**

## Verification

Create some notes and attach some comments (using existing mutation resolver format).

Next, perform some filtering on the note, like so:

```graphql
query findNotes {
  findNotes(filter:{title:{contains:"title 1"}}) {
    items {
      id
      title
      description
      comments(filter:{description:{contains:"2"}, or:{description:{eq:"test"}}}) {
        id
        description
        text
      }
    }
  }
}
```

## Tasks

- [x] find a single record 
- [x] filter records with advanced querying
- [x] filter relationships
- [x] add to runtime
- [ ] client documents
- [ ] unit tests


packages/graphback-cli/src/components/db.ts
```diff
@@ -42,7 +42,8 @@ export const createDBResources = async (cliFlags: { project?: string }): Promise
 
     const migrateOptions: MigrateOptions = {
       //Do not perform delete operations on tables
-      operationFilter: removeNonSafeOperationsFilter
+      operationFilter: removeNonSafeOperationsFilter,
+      schemaTransformer: removeNonModelTypesFromSchema
```
craicoverflow, line 46: this is not working but leaving it in cos I might try to make it work in future

packages/graphback-runtime-knex/src/knexQueryMapper.ts
```diff
@@ -0,0 +1,108 @@
...
+    case 'not':
+      return `${method}Not`;
+    case 'or':
+      return `${clause}${method.charAt(0).toUpperCase()}${method.slice(1)}`;
+    default:
+      return method;
+  }
+}
+
+function where(builder: Knex.QueryBuilder, filter: any, clause?: string) {
```
craicoverflow, line 60: Similar to graphql-server-crud with some changes

https://github.com/charlie0077/graphql-server-crud/blob/35c04ab06c6dd05d2c4cc35b50a2d06aeacada81/src/knexTransform.js#L61

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -67,23 +67,23 @@ export class LayeredRuntimeResolverCreator {
...
         }
       }
       if (resolverElement.crudOptions.find) {
         const findField = getFieldName(modelName, GraphbackOperationType.FIND);
         //tslint:disable-next-line: no-any
         resolvers.Query[findField] = (parent: any, args: any, context: any) => {
           const page = { limit: args.limit, offset: args.offset };
-          
+
           return this.services[modelName].findBy(args.fields, page, context)
```
craicoverflow, line 87: Need to do

packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -16,136 +16,140 @@ import { GraphbackPubSub } from "./GraphbackPubSub"
...
-    private publishConfig: GraphbackPubSub;
-    private modelName: string;
-
-    public constructor(modelType: GraphQLObjectType, db: GraphbackDataProvider, subscriptionConfig: GraphbackPubSub, logger?: GraphbackMessageLogger) {
-        this.db = db;
-        this.pubSub = subscriptionConfig.pubSub;
-        this.logger = logger || defaultLogger;
-        this.publishConfig = subscriptionConfig;
-        this.modelName = modelType.name;
+  private db: GraphbackDataProvider;
```
craicoverflow, line 19: A lot of stuff auto-formatted with my editor :(

packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -16,136 +16,140 @@ import { GraphbackPubSub } from "./GraphbackPubSub"
...
     }
 
-    //tslint:disable-next-line: no-reserved-keywords
-    public async delete(data: T, context?: any): Promise<T> {
-        this.logger.log(`deleting object ${this.modelName}`)
+    return result;
+  }
 
-        const result = await this.db.delete(data, context);
+  public findOne(filter: any, context?: any): Promise<T> {
```
craicoverflow, line 78: Get a single record

packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -16,136 +16,140 @@ import { GraphbackPubSub } from "./GraphbackPubSub"
...
-            this.logger.log(`Cannot subscribe to events for ${this.modelName}`)
+    return this.pubSub.asyncIterator(deleteSubKey)
+  }
 
-            throw Error(`Missing PubSub implementation in CRUDService`);
-        }
-        const deleteSubKey = this.subscriptionTopicMapping(GraphbackOperationType.DELETE, this.modelName);
 
-        return this.pubSub.asyncIterator(deleteSubKey)
+  public batchLoadData(relationField: string, id: string | number, filter: any, context?: any) {
```
craicoverflow, line 129: Filtering on batches

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
```diff
@@ -103,7 +103,7 @@ class AbstractDatabaseBuilder {
     for (const key of Object.keys(this.typeMap)) {
       const type = this.typeMap[key]
       // Tables
-      if (isObjectType(type) && !type.name.startsWith('__') && !ROOT_TYPES.includes(type.name)) {
+      if (isObjectType(type) && !type.name.startsWith('__') && !ROOT_TYPES.includes(type.name) && isModelType(type)) {
```
craicoverflow, line 106: hard-coded this for now, to prevent errors in migrations

packages/graphql-migrations/src/plugin/SchemaTransformer.ts
```diff
@@ -0,0 +1,27 @@
+import { GraphQLSchema } from 'graphql';
+import { filterNonModelTypes } from '@graphback/core'
+import { SchemaComposer } from 'graphql-compose'
+
+export interface SchemaTransformerPlugin {
+  transform(schema: GraphQLSchema): GraphQLSchema
+}
+
+export const removeNonModelTypesFromSchema: SchemaTransformerPlugin = {
```
craicoverflow, line 9: This implementation will not work, so need to change it

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
```diff
@@ -7,23 +7,53 @@ type Comment {
...
   description: String
-  noteId: ID
+}
+
+type CommentListResult {
+  items: [Comment]!
+  nextCursor: String
+}
+
+input IDInput {
```
craicoverflow, line 37: Static input type for each scalar

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
wtrocki: Actually missing context. Why migrations are affected by the filtering work? Is that we provide some extra types? Migration obviously ignores input types right ?

packages/graphql-migrations/src/plugin/SchemaTransformer.ts
wtrocki: Seems like separate issue to me ;D

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
```diff
@@ -7,23 +7,53 @@ type Comment {
   description: String
 
   """@manyToOne field: 'comments', key: 'noteId'"""
-  note: Note
+  note(filter: NoteFilterUniqueInput): Note
```
wtrocki, line 10: Excited to see this!

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
wtrocki: This is where our previous print mechanism were much better. 
Printing this across large schema is minor issue but stands up when seeing this for the first time

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
craicoverflow: We are now creating object types to wrap the list of models for returning. These do not have an `id: ID` field, but migrations expects this on all types.

packages/graphql-migrations/src/plugin/SchemaTransformer.ts
craicoverflow: https://github.com/aerogear/graphback/pull/1211#discussion_r417421475

Related to this

templates/ts-apollo-fullstack/server/src/schema/schema.graphql
craicoverflow: Ah that should not be there (can't filter on a 1:1 record) - it should only be on a list.

wtrocki:
@craicoverflow It will be amazing to try a new CRUD engine in our new project. 
Would you be looking into releasing dev version at some point of the day for testing?

Do we plan to support text Search features?

craicoverflow:
> @craicoverflow It will be amazing to try a new CRUD engine in our new project.
> Would you be looking into releasing dev version at some point of the day for testing?
> 
> Do we plan to support text Search features?

Great idea! I will fix tests first.

Can you specify what you mean by text-search? Do you mean on a single column or across searching all columns with a single query?

wtrocki:
AWS AppSync text search. I'm looking for something that will be able to give me capability to find results by text but that can be done by `or` and `like` 
if there is like then we are fine

craicoverflow:
> AWS AppSync text search. I'm looking for something that will be able to give me capability to find results by text but that can be done by `or` and `like`
> if there is like then we are fine

Yes we have this in this CRUD prototype. It is the `contains` filter

wtrocki:
Yep.. For typical text search data needs to be indexed and this requires apache luciene like solution

wtrocki:
@craicoverflow @darahayes This is great oportunity to talk about subscriptions. 
I have noticed that we did not covered it in docs. 

Would we do subscriptions filtering as well - we needed to remove subscriptions from the project as getting all the changes were not particularily useful for us.

packages/graphback-core/src/plugin/GraphbackCRUDGeneratorConfig.ts
```diff
@@ -1,21 +1,22 @@
...
+  create?: boolean;
+  //Generate update method
+  update?: boolean;
+  // Generate delete method
+  delete?: boolean;
+  //Generate findOne method
+  findOne?: boolean;
+  findAll?: boolean;
+  //Generate find method
+  find?: boolean;
```
darahayes, line 15: Potentially silly question. If `findOne` fetches one object based on some filter then what's the difference between `find` and `findall`?

craicoverflow: findAll is going away, I just kept it in for now to limit the amount of things that break.

craicoverflow: find now does the exact same.

craicoverflow:
I've added some tests for the CRUDService, which will make this easier to refactor.

craicoverflow:
Screenshot from 2020-05-01 17-10-41

Build is still failing but not sure why yet. It says it can't find the dependency but build works fine locally and this is included in the package.json

packages/graphback-core/src/plugin/GraphbackCRUDGeneratorConfig.ts
wtrocki: I would kill it. Ideally we want next release to be uber breaking so people do stuff once

wtrocki:
Amazig work. I had quick look while pair programming. Ideally we should remove findAll etc. from every layer.

wtrocki:
Going to release this as dev

wtrocki:
@craicoverflow We will work with @Eunovo on client side documents

PR #1213, opened 2020-04-30
Implement New CRUD spec for mongo

ssd71:
The mquery syntax is mostly similar to mongodb native driver queries with some sugar added, All of find, delete, update queries are still present, with the exception of a insert query. So with this, MongoDataProvider uses mquery for all operations except 'create'. I also noticed that we don't test the batchRead operation, which I also changed to use mquery, so if possible I would like to write a test for it just to be safe?

### DIFF
--- a/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
+++ b/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
@@ -76,6 +76,9 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements 
     throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
 
+  /**
+   * @deprecated
+   */
   public async findAll(page?: GraphbackPage): Promise<Type[]> {
     const query = this.mcollection.find({});
     const data = await this.usePage(query, page);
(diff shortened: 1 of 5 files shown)


packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -92,11 +93,11 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     const { idField } = getDatabaseArguments(this.tableMap, filter);
     // TODO MongoDB should use advanced filter with JSON scalar defined as InputType
     if (filter[idField.name]) {
-      const query = this.db.collection(this.collectionName).
+      const query = mquery(this.db.collection(this.collectionName)).
         find({ _id: new ObjectId(filter[idField.name]) });
       dbResult = await this.usePage(query, page);
     } else {
-      const query = this.db.collection(this.collectionName).find(filter);
+      const query = mquery(this.db.collection(this.collectionName)).find(filter);
```
wtrocki, line 100: Me thinking maybe we can have mquery as property:

`mquery(this.db.collection(this.collectionName))` => `this.mcollection `

ssd71: Agreed :+1:  it's a bit of a mouthful

wtrocki:
Changed base to @craicoverflow branch. This might break things

wtrocki:
Overall I do see this being quite nice.and fluent api. We can try it in new crud

ssd71:
Only used in findBy as of a95b07a
Will add other implementations in time,
also tests to be written

wtrocki:
I would skip tests for now. What we want is early alpha of crud spec to play with in real project to see possible issues or extensions

wtrocki:
I have been using DataSync starter for testing.

ssd71:
Mongo now uses advanced filtering in both findBy and batchLoading

ssd71:
Seems like there is some issue with Input Fields in #1211 
There is no note_id input field in [this](https://github.com/aerogear/graphback/blob/8dbb1d593ea3d6142fd777daf82432f139393065/packages/graphback-codegen-schema/tests/__snapshots__/GraphQLSchemaCreatorTest.ts.snap#L25) test snapshot, Is that intentional?
I was having issues testing it out with relationships because of this

wtrocki:
Yep.  That is a blocker bug for us.
@craicoverflow is off for 3 days so I will check why this is happening

wtrocki:
Merging this into @craicoverflow PR in order to organize work

Issue #1071, opened 2020-04-08
Improve runtime API


craicoverflow:
Here is how to set up a runtime Graphback API:

```ts
const graphbackConfig = await getGraphbackConfig();

const schemaText = loadSchema(graphbackConfig.model);

const pubSub = new PubSub();

const runtimeEngine = new GraphbackRuntime(schemaText, graphbackConfig);
const models = runtimeEngine.getDataSourceModels();
const services = createKnexPGCRUDRuntimeServices(models, buildSchema(schemaText), db, pubSub);
const runtime = runtimeEngine.buildRuntime(services);

const apolloServer = new ApolloServer({
    typeDefs: printSchema(runtime.schema),
    resolvers: runtime.resolvers,
})
```

This is extremely convoluted and needlessly difficult to configure - nobody will want to use this. Much of this can be moved into our library, it could be as simple as this:

```ts
import { graphback } from 'graphback';

const pubSub = new PubSub();

const { typeDefs, resolvers, models } = graphback(pubSub);

const db = Knex(...);

const services = createKnexPGCRUDRuntimeServices(models, typeDefs, db, pubSub);

const apolloServer = new ApolloServer({
    typeDefs,
    resolvers,
    context: context => {
        ...context,
        models
    }
})
```

lastmjs:
I absolutely agree with the complexity here. I'm just trying to setup the runtime now, and it is quite confusing, and I'm looking at multiple pieces of documentation and examples that seem to say slightly different things. I think the API could even be simplified beyond what you have suggested @craicoverflow. Couldn't the Knex db initialization be abstracted away? Why not provide the db details in the `.graphqlrc.yml`? We could provide an API that allows people to replace this db if they want to, but not require them to configure it by default. My dream for this project (I am coming from Prisma and had hoped they would do this, but they aren't going to) is to be able to write a simple data model schema in GraphQL SDL, import maybe one thing, pass it in, and have it work. To really be able to persuade others, this needs to be dead simple to start with so that it is obvious how much better a GraphQL-centered backend is than the status quo.

Great job so far! I'm very optimistic about the project.

darahayes:
@craicoverflow I wonder how would database migrations fit into this? Is it something that the user looks after at an application level?

I've got a very rough prototype based on the API shown above. Going to post a draft PR for it soon to get your thoughts on it.

craicoverflow:
Related https://github.com/aerogear/graphback/issues/1283

[...]

PR #1215, opened 2020-04-30
refactor: minor changes to runtime initialization

darahayes:
This PR introduces some minor refactoring of the runtime and core packages, without making any changes to functionality. There are more areas that can be improved but I figured it would be nice to do things in smaller easy to digest PRs.

* GraphbackGenerator `initializePlugins` refactored into pure `loadPlugins` function. Instead of passing the plugin engine into it just to call `registerPlugin` we just return an array of initialized plugins which can then be passed into `GraphbackPluginEngine`. I've left the `registerPlugin` function just in case.
* `GraphbackRuntime` no longer extends `GraphbackGenerator`. The only reason it was doing that was to call `initializePlugins` which is not longer needed.
* `GraphbackRuntime` used to call `this.getMetaData()` in multiple places which would create the plugin engine and `require` and create the plugins each time. This seemed wasteful and unnecessary. Instead the plugin engine is created once. the `getMetaData` function is still there but it just returns the already created metadata.
* Renamed `GraphbackGeneratorConfig` to `GraphbackConfig` and moved to its own file. This is still used both by the generator and the runtime classes.

### DIFF
--- a/packages/graphback/src/index.ts
+++ b/packages/graphback/src/index.ts
@@ -1,4 +1,5 @@
 export * from '@graphback/core'
 export * from '@graphback/runtime'
+export * from './GraphbackConfig'
 export * from './GraphbackGenerator'
 export * from './GraphbackRuntime'
--- a/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
+++ b/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
@@ -4,13 +4,21 @@ import { GraphbackGlobalConfig } from './GraphbackGlobalConfig';
 import { GraphbackPlugin } from './GraphbackPlugin';
 
 /**
+ * options for the GraphbackPluginEngine
+ */
+export interface GraphBackPluginEngineOptions {
+    schema: GraphQLSchema | string
+    config: GraphbackGlobalConfig
+    plugins: GraphbackPlugin[]
+}
+/**
  * Allows to execute chain of plugins that create resources. 
  * Common use case is to decorate GraphQL schema with additional 
  * actions and generate files like resolvers and database access logic
  * 
  * Usage:
  * ```js
- * const engine = GraphbackPluginEngine(schema);
+ * const engine = GraphbackPluginEngine({ schema });
  * engine.registerPlugin(plugin);
  * printSchema(engine.createResources());
  * ```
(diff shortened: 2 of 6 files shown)


wtrocki:
Like the changes, but @craicoverflow has final call on this

packages/graphback/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from '@graphback/core'
 export * from '@graphback/runtime'
+export * from './GraphbackConfig'
 export * from './GraphbackGenerator'
```
craicoverflow, line 4: I am questioning why this is here (I know it was here already). I don't think we should expose the code generator class to the application using our runtime.

@wtrocki do you think this can be moved? It is only used by the CLI, maybe it should be in `@graphback/core`.

packages/graphback/src/index.ts
```diff
@@ -1,4 +1,5 @@
...
 export * from './GraphbackRuntime'
```
craicoverflow, line 5: I think eventually like to hide this class from the application consuming it and have a helper function do the class creation. You will probably see from the runtime templates how much work is involved in configuring Graphback runtime from the template but most of this can happen from our package.

https://github.com/aerogear/graphback/issues/1071

packages/graphback/src/index.ts
darahayes: You mean you're questioning why the Generator class is exposed?

Yeah definitely as I go deeper into this I'll have opinions on how the API should change but I figured I'd start with something that doesn't make those kinds of changes 😄

packages/graphback/src/index.ts
darahayes: Yeah I agree. The something like a generic version of the `createRuntime` function in the actual runtime template itself. To be honest, the whole thing could be a function the whole way down but I won't start a flame war 😅

packages/graphback/src/index.ts
wtrocki: considering that we can load different plugins moving to core makes no sense. I would question value of this and probably suggest to remove it

craicoverflow: Yeah, I was questioning why it is exposed.

wtrocki:
It is strange that this build was passing. I see issues with unit tests containing wrong API. seems like CI is not running all tests

wtrocki:
Yep. EngineTests were not executed.

wtrocki:
Also, change was merged but there was no update in docs or note about breaking API

wtrocki:
https://circleci.com/gh/aerogear/graphback/4060?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

darahayes:
My bad, I ran the tests locally too and I didn't see any failures. I will fix it!

PR #1216, opened 2020-04-30
Client side query expansion fix

Eunovo:
@wtrocki


Eunovo:
Alright, enjoy your holidays!

packages/graphback-codegen-schema/tests/GraphQLSchemaCreatorTest.ts
```diff
@@ -69,3 +70,28 @@ test('Test snapshot config js', async () => {
   const schema = schemaGenerator.transformSchema(metadata)
   expect(printSchema(schema)).toMatchSnapshot();
 });
+
+test('Test one side relationship schema query type generation', async () => {
```
wtrocki, line 74: This is good but we already have lots of coverage on this with unit tests.

packages/graphback-codegen-client/tests/__snapshots__/GraphQLClientCreatorTest.ts.snap
```diff
@@ -867,3 +867,97 @@ export const deletedComment = gql\`
...
+query findAllNotes($limit: Int, $offset: Int) {
+    findAllNotes(limit: $limit, offset: $offset) {
+      ...NoteExpandedFields
+    }
+  }
+
+fragment NoteExpandedFields on Note {
+   id
+   title
+   description
```
wtrocki, line 907: This exactly shows the issue?

packages/graphback-codegen-client/tests/__snapshots__/GraphQLClientCreatorTest.ts.snap
```diff
@@ -867,3 +867,97 @@ export const deletedComment = gql\`
...
+query findNotes($fields: NoteInput!, $limit: Int, $offset: Int) {
+    findNotes(fields: $fields, limit: $limit, offset: $offset) {
+      ...NoteExpandedFields
+    }
+  }
+
+fragment NoteExpandedFields on Note {
+   id
+   title
+   description
```
wtrocki, line 888: Issue here - we should see nested comments - we do not see them

wtrocki:
@Eunovo I do not see any functional code changes apart from tests. Issue is clearly visible. I have explained source in the gh issue itself

Eunovo:
@wtrocki The test cases are just to help me understand the problem. I'm proposing that we generate the expanded fields differently by using the relationships in the model definitions. In the test case that reveals the issue, I observed that models contain the relationships but the types do not. So, by changing the code to use the relationships the problem will be solved. Can I go ahead?

packages/graphback-codegen-schema/tests/GraphQLSchemaCreatorTest.ts
Eunovo: I was just checking that `transformSchema` does work as expected

Eunovo:
[This is the function I want to change](https://github.com/Eunovo/graphback/blob/fb6b91e6a059b2b4b0b41aacb33efc1ce87f1d07/packages/graphback-codegen-client/src/templates/fragmentFields.ts#L12)

wtrocki:
sounds like plan

wtrocki:
I think that models should have this field already. Bug is related to models somehow getting cached so when we get models after schema plugin we should be able to see new relation added. This meams that there is not need for special function

Eunovo:
I have already confirmed that the SchemaPlugin works as expected. I don't see how the schema would be cached from these [lines](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts#L65-#L68) and these lines in [Metadata](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts#L38-#L40). Unless there is some other interaction I'm not seeing?

wtrocki:
While this resolves the issue I think we have misunderstood. 

If schema plugin adds a field to the model why the client-side plugin needs this helper. Why we cannot process relation field the same way knowing that we will have it in the model. I see the relation in schema plugin snapshot, but why it is not included in models that client plugin is working with?
If we answer this question we will have much better fix for this problem.

-1 one for adding this helper as it will be needed in every plugin.

wtrocki:
Rebase needed

Issue #1146, opened 2020-04-20
Remove graphql-config from runtime templates


craicoverflow:
As graphql-config is only needed for code generation,it can be removed from our runtime templates and documentation.

PR #1219, opened 2020-05-01
experimental: new graphback runtime api

darahayes:
Relates to https://github.com/aerogear/graphback/issues/1071

TL;DR this is where we're at

```js
import { graphback } from 'graphback'
import { CRUDService as service } from '@graphback/runtime'
import { PgKnexDBDataProvider as provider } from '@graphback/runtime-knex'


const db = new Knex(...);
const pubSub = new PubSub(...);

const options = {
  db,
  pubSub,
  service,
  provider,
}

const { schema, resolvers } = graphback(graphbackOptions)

// do your own db migrations

  typeDefs: printSchema(schema),
  resolvers
})
```

It's super prototypal but the whole idea is that you can pass in the uninitialised DataProvider and CrudService classes. I eventually see the CrudService class going away which would make the API even more intuitive. I didn't do it here but I think we can deprecate/remove the various functions like `createKnexPGCRUDRuntimeServices`, `createKnexCRUDRuntimeServices` for the more generic `createRuntimeServices` function I added.

Check the changes to the runtime-backend template to see how the API looks. This is by no means complete. It's more of an idea to get feedback on.

Things that I still haven't really figured out yet are:

* how plugins should work with this API (hopefully we can continue to just load them and nothing else is needed)
* how resolver composability will work. (wrapping, chaining etc)
* where database migrations fit in. (Possibly the user just does them)


packages/graphback-runtime/src/createRuntimeServices.ts
```diff
@@ -0,0 +1,49 @@
+import { GraphQLObjectType, GraphQLSchema } from "graphql"
+import { PubSubEngine } from 'graphql-subscriptions';
+import { GraphbackDataProvider } from './data/GraphbackDataProvider';
+import { CRUDService } from './service/CRUDService';
+import { GraphbackModelConfig } from "./GraphbackModelConfig";
+
+export type RuntimeServicesOptions<Provider extends GraphbackDataProvider, Service extends CRUDService> = {
+  schema: GraphQLSchema,
+  models: GraphbackModelConfig[]
+  db: any
```
wtrocki, line 10: If we pass db then we do not need provider and also if we pass provider then we do not need db

packages/graphback-runtime/src/createRuntimeServices.ts
```diff
@@ -0,0 +1,49 @@
+import { GraphQLObjectType, GraphQLSchema } from "graphql"
+import { PubSubEngine } from 'graphql-subscriptions';
+import { GraphbackDataProvider } from './data/GraphbackDataProvider';
+import { CRUDService } from './service/CRUDService';
+import { GraphbackModelConfig } from "./GraphbackModelConfig";
+
+export type RuntimeServicesOptions<Provider extends GraphbackDataProvider, Service extends CRUDService> = {
+  schema: GraphQLSchema,
+  models: GraphbackModelConfig[]
```
wtrocki, line 9: This should be gone as it was only codegen artifacts and it makes no sense for runtime. Runtime reads our annotations and markers and have access to the global crud options

packages/graphback-runtime/src/createRuntimeServices.ts
```diff
@@ -0,0 +1,49 @@
...
+  db: any
+  pubSub: PubSubEngine
+  provider: new (...args: any[]) => Provider,
+  service: new (...args: any[]) => Service;
```
wtrocki, line 13: If we remove service and simply pass pubsub we will be sorted. Removing service though will require specific resolver wrappers as functions for pub sub and security later. Any customisation from users could be done by wrapping code into their apps. Is that making sense? would you like to use that kind of API yourself

packages/graphback/src/GraphbackRuntime.ts
```diff
@@ -26,6 +27,18 @@ export class GraphbackRuntime {
       config: { crudMethods: this.config.crud }
     });
     this.metadata = pluginEngine.createSchema();
+
+    const models = this.metadata.getModelDefinitions();
+    this.modelConfigs = models.map((model: ModelDefinition) => {
```
wtrocki, line 32: Note: This was hack to keep similar code between codegen and runtime.

When moving to runtime we do not need that config at all as it duplicates .graphqlrc.yml.config. In codegen we did not needed that.

packages/graphback-runtime/src/createRuntimeServices.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export function createRuntimeServices<Provider extends GraphbackDataProvider, Service extends CRUDService> 
+  (servicesConfig: RuntimeServicesOptions<Provider, Service>) {
+
+  const { models, schema, db, pubSub, provider, service } = servicesConfig
+  if (!models || models.length === 0) {
+    throw new Error(`No models provided`)
+  }
+
+  return models.reduce((services: any, model: GraphbackModelConfig) => {
```
wtrocki, line 32: Similar to above we do not need to have external models. this is very convoluted - we can base on schema itself

templates/ts-apollo-runtime-backend/src/runtime.ts
```diff
@@ -1,39 +1,39 @@
...
     //Do not perform delete operations on tables
     operationFilter: removeNonSafeOperationsFilter
   };
 
-  const pubSub = new PubSub();
-  const runtimeEngine = new GraphbackRuntime(model, graphbackConfig);
-  const models = runtimeEngine.getDataSourceModels();
-  const services = createKnexPGCRUDRuntimeServices(models, model, db, pubSub);
-  const runtime = runtimeEngine.buildRuntime(services);
+  const graphbackOptions = {
```
wtrocki, line 29: I see this as just provider and pubsub only. 
eventually format as compared to generation runtime is going to work with GraphQL-Tools resolvers only

wtrocki:
Suggestion for the moment would be to work on this PR to:

- remove services.
- remove resolver codegen
- remove docs and usages for codegen
- remove codegen usage

Merge this to CRUD feature branch and release change for testing. 
Considering that most of the team is off for a while I will take look and try to apply this changes here.

wtrocki:
Merging into feature branch to continue work and further refactor

packages/graphback/src/graphback.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Create a Graphback Runtime
+ * 
+ * @param options 
+ */
+export function graphback<Provider extends GraphbackDataProvider, Service extends CRUDService>(options: GraphbackOptions<Provider, Service>) {
+  const { db, pubSub, provider, service } = options
+
+  const projectConfig = loadConfigSync({
```
craicoverflow, line 26: There was a separate discussion previously where the runtime templates will no longer use `graphql-config`, as it is mainly used for code generation, which is not a factor in a runtime Graphback API.

So eventually we can drop support for graphql-config completely, and the user can provide a JSON configuration object in `options`.

https://github.com/aerogear/graphback/issues/1146

wtrocki: Hmm.. yaml will be used for generate but if we drop resolves generation we will simply have schema plugin so for runtime use cases having ability to define plugins in the code with defaults will be much better than using config - while config can still stay I do not need to be used by default

PR #1235, opened 2020-05-06
Crudspec feature

wtrocki:
Feature branch based on @craicoverflow PR. I have made mistake to integrate other changes together and this is the second attempt. Initially I have merged @ssd71 Mongo changes into this but we have decided to change approach and removed the changes from the branch initially.

We will PR against it with all new changes as this contains the new approach for the graphback

## Verification

Create some notes and attach some comments (using existing mutation resolver format).

Next, perform some filtering on the note, like so:

```graphql
query findNotes {
  findNotes(filter:{title:{contains:"title 1"}}) {
    items {
      id
      title
      description
      comments(filter:{description:{contains:"2"}, or:{description:{eq:"test"}}}) {
        id
        description
        text
      }
    }
  }
}
```

## Tasks

- [x] find a single record 
- [x] filter records with advanced querying
- [x] filter relationships
- [x] add to runtime
- [x] client documents
- [x] unit tests


packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -1,7 +1,8 @@
 import { buildModelTableMap, getDatabaseArguments, ModelTableMap } from '@graphback/core';
 import { GraphQLObjectType } from 'graphql';
 import * as Knex from 'knex';
-import { GraphbackDataProvider, GraphbackPage, NoDataError, AdvancedFilter } from '@graphback/runtime';
+import { GraphbackDataProvider, GraphbackPage, NoDataError, AdvancedFilter } from '../../graphback-runtime/src';
```
ssd71, line 4: Please correct me if I'm wrong but shouldn't this stay '@graphback/runtime'?

wtrocki: Good catch!

wtrocki:
# Review
## Schema 

Focus on schema related elements only

### Questions

1) We are generating helpers for each specific objects.
Ho we will handle scalars. For example I want to issue query for date between some range?

2) Create Input still accepts id as parameter. Is that intentional?

3) Create is missing relationship ID. I cannot create object with relationship

4) Nexcursor seems to be very strongly related to the specifics of the databases that supports stuff. I assumed that we did not wanted to implement this and simply investigate that at some point later. Is pagination working?

## Querying

1) Pagination never returns cursor or anyvalue? Seems to be that we are mixing cursor based pagination with pages (skip, limit). Limit seems to be working for both but returns no cursor

wtrocki:
Made couple edits on the comment above. Please ignore email and check github instead

wtrocki:
Plan to release this
- address rough corners
- build set of integration tests 
- update docs 

Let's collaborate on this

craicoverflow:
> ## Schema Questions
> 1. We are generating helpers for each specific scalar objects.
>    How we will handle scalars. For example, I want to issue query for date between some range?

We still need to add a filter object for DateTime. Before that we need to decide on a DateTime scalar we should include as part of Graphback. This is still a todo for now. When we use a DateTime scalar like [graphql-iso-date](https://github.com/excitement-engineer/graphql-iso-date) it would be easier to work with and calculate ranges etc.

> 2. Create Input still accepts id as a parameter. Is that intentional?

[Our docs](https://graphback.dev/docs/intro/datamodel#primary-key) describe how to specify a custom primary key such as email. For this reason it is still left in, but you are right we should remove any `id: ID` field from the create input.

This is actually really confusing for a developer. If they specify a custom primary key such as email they will not want to also add `id: ID` as it seems counter intuitive. Perhaps we should not make it required but automatically add it to each model in the schema crud plugin? This way it is effectively an invisible primary key.

> 3. Create is missing relationship ID. I cannot create object with relationship

Needs to be done.

> 4. Nexcursor seems to be very strongly related to the specifics of the databases that supports stuff. I assumed that we did not wanted to implement this and simply investigate that at some point later. Is pagination working?

Yeah this is simply not implemented and will be removed. I believe the existing pagination still works, but we need to add page information to the result type, in place of nextCursor.

> 5. I do not see support for `between` from the schema.
>    Is that in progress?

Yep https://github.com/aerogear/graphback/pull/1248

> ## Querying
> 1. Does pagination never return cursor or anyvalue? Seems to be that we are mixing cursor based pagination with pages (skip, limit). Limit seems to be working for both but returns no cursor

We decided not to use cursor based pagination, field should be removed.

wtrocki:
> We still need to add a filter object for DateTime. Before that we need to decide on a DateTime scalar we should include as part of Graphback. This is still a todo for now. When we use a DateTime scalar like graphql-iso-date it would be easier to work with and calculate ranges etc.

But how about any other scalar. It could be nice to detect scalars and automatically match them into string type of filter. For some use cases (like JSON) this will not make sense but it will allow using any scalar out of the box (like date)

wtrocki:
As there is 5060 combinations of filter we need to agree on what kind of filters make sense for testing and integration.

wtrocki:
@craicoverflow 
Quick question on fetching a single item.

We have a discrepancy between code API and Schema.
Schema has:
`getComment(id: ID!): Comment` 

but code allows specifying multiple fields from the object.

craicoverflow:
> @craicoverflow
> 1)
> Quick question on fetching a single item.
> 
> We have a discrepancy between code API and Schema.
> Schema has:
> `getComment(id: ID!): Comment`
> 
> but code allows specifying multiple fields from the object.

The main reason for this was to leave the interface flexible. Although our default CRUD findOne fetches by a single key, if someone required a slightly different methopd like `getNote(id: ID!, version, Int!)` this would mean they don't have to create a new special CRUD service just for it.

That is a special case though, I am happy to adapt the code to always get by the single ID argument.

> 1. I'm not convinced about multiple input types.

AFAIK each input type is slightly different?

> We are using the same input types for creation and subscriptions (subscriptions might not work here)

I haven't touched subscriptions yet, as this was the way it used to work I just left it the same

craicoverflow:
> But how about any other scalar. It could be nice to detect scalars and automatically match them into string type of filter.

That seems sensible and easy to do.

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -115,27 +115,27 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
     throw new NoDataError(`No results for ${this.tableName} and id: ${JSON.stringify(ids)}`);
   }
 
-  private usePage(query: Knex.QueryBuilder, page?: GraphbackPage, defaultLimit: number = 10, defaultOffset: number = 0) {
-    if (page) {
-      if (!(page.hasOwnProperty("offset"))) {
```
craicoverflow, line 115: this failed for me giving "page.hasOwnProperty does not exist". It is worth checking if the same issue exists on Mongo provider @wtrocki

wtrocki: yep. Page is not required so we just simply need to check for it.
Sorry for introducing this bug. I have separate issue to test pagination.

docs/crud/crudspec.md
```diff
@@ -15,8 +15,8 @@ For every model object in your schema we can provide the folowing operations:
 - UPDATE
```
wtrocki, line 15: This got oudated. Instead of investing time here  I will prefer to start moving things to CRUDSpec instead and link it in our docs.

docs/releases.md
```diff
@@ -10,7 +10,18 @@ Please follow individual releases for more information.
 
 # 0.14.0
 
+## Migrating to 0.14.0
```
wtrocki, line 13: Pointing this as area to review specifically. Did I missed anything?

docs/crud/crudspec.md
craicoverflow: Agreed, so did not bother touching it.

docs/releases.md
```diff
@@ -10,7 +10,18 @@ Please follow individual releases for more information.
...
+
+Version 0.14.0 removed support codegeneration for resolvers. 
+If you used one of the Codegen templates you will need to migrate 
+
+1. Remove your generated resolvers and model file. They are not longer needed as resolvers are generated at runtime.
+Check configuration of the resolvers plugin to see where they are located.
+2. Remove @graphback/codegen-resolvers from your package and the config.
+3. Use Graphback runtime Getting Started Guide
```
wtrocki, line 21: This is going to be updated. I will refactor getting started once we deal with runtime issues

docs/releases.md
craicoverflow: I think with such major changes the upgrade guide should have more detail. We know that the runtime API is confusing right now and so are the docs as a consequence of this. So the upgrade guide should be more thorough as it is big change.

craicoverflow: If you agree then I can do something for this.

packages/graphback-runtime-knex/src/createKnexServices.ts
```diff
@@ -1,13 +1,13 @@
...
 import * as Knex from 'knex';
-import { CRUDService, GraphbackPubSubModel } from "@graphback/runtime";
+import { CRUDService, GraphbackPubSubModel } from "@graphback/runtime"
 import { PgKnexDBDataProvider } from './PgKnexDBDataProvider';
 import { KnexDBDataProvider } from './KnexDBDataProvider';
 
 /**
  * Helper function for creating array of datasources based on the model files that will 
- * be able to connect to Postgres database using knex
+ * be able to connect to Postgress database using knex
```
craicoverflow, line 10: Hmm why whas this changed _to_ Postgress

wtrocki, line 10:
```suggestion
 * be able to connect to Postgres database using knex
```

Issue #1095, opened 2020-04-10
Date or DateTime type


lastmjs:
Is there no native date or datetime type? I would really like to have these kinds of types. They should handle date inputs and store themselves in a universal format, like an ISO UTC string or perhaps milliseconds since the unix epoch. Prisma's date (or DateTime?) type stores itself as an ISO UTC string, which has worked very well for me. You can input either a JavaScript date type or a date string, and it all works. Something like that would be nice to have, and seems essential actually.

wtrocki:
Great feedback! We were talking with the team about this. 
Generally, we decided that Scalars like Date etc. should be the choice of the template and our current templates do not have any loaded. Graphback itself works as great addition to existing GraphQL projects that might have scalars. 
If we start adding scalars in the core we might get conflicts with already defined one. 

Ideally we should at least have Date, JSONObject and couple others from graphql-scalars in the default template to show that they work quite nicely and test them. In terms of having then out of the box in model I will need to rely on @craicoverflow opinion as this might affect the way we migrate database.

I think this could be a bug to see how database migration will work with scalars and also feature request to have scalars in core or templates

craicoverflow:
As @wtrocki said - idealy we would like to leave the choice of scalars to the template. 

I just tested migrations with [graphql-iso-date](https://github.com/excitement-engineer/graphql-iso-date) and the field will only get generated when it has a `@db.type: 'timestamp` annotation.

Obviously we do not want our users to need to specify an annotation for a `DateTime` field in their data model for migrations. I think this could be pretty easily resolved by adding custom scalar mapping in the `dbmigrations` config. Currently the default scalar mappings are embedded into the code ([see here](https://github.com/aerogear/graphback/blob/bd328b581202a15b53083f39b783c63bec7dc9ff/packages/graphql-migrations/src/abstract/getColumnTypeFromScalar.ts#L49)).

We could make it simpler again by embedding a default mapping between DateTime and timestamp. This would not require any additional dependencies, but it would just mean that the user does not have to add any additional config.

wtrocki:
Related to #1114

wtrocki:
It is important to actually support his out of the box so resolver plugin should add date and time resolvers (with ability to suppress it)

Problem is that people will need to do 
```
npm install --save graphql-scalars
```

Alternative will be to have this embedded into runtime package with ability to get all resolvers. Adding this issue to the most critical we need to solve for the next release

Artimunor:
[...]

PR #1236, opened 2020-05-06
Deprecate db and resolvers generator

wtrocki:
This PR started as simple change but I have been continuing work by picking nondestructive and conflicting changes 

1) Remove cmd DB migrations (mostly docs and cmd - all removals so get)
We need to do this to have a unified way to solve migrations and address numerous issues from the community: #1095 etc.

2) Remove graphql-codegen-resolvers integration from source code, while keeping resolvers plugin still alive for a while to not trigger conflicts with other PR's etc.

3) Cleanup templates and give the ability to compose them together.
We have added the ability to clean up templates so we can use multiple layers of them when doing init. This enabled us to have multiple variations runtime templates with little effort. I still kept the 

4) Remove migration config and improvements 

The purpose of this PR is to bring the changes that will minimize the amount of the work required for the CRUD services while still being mostly deleting and reorganizing things. 

## Verification

1. cd packages/graphback-cli
2. yarn build
3. yarn cmdinit 
4. Pick first template (it composes from 2 templates)
5. All should be able to compose 
6. Template should be runnable


docs/releases.md
```diff
@@ -10,11 +10,18 @@ Please follow individual releases for more information.
 
 # 0.14.0
```
machi1990, line 11: :-)

craicoverflow:
These are being removed not deprecated

.vscode/launch.json
```diff
@@ -28,37 +28,11 @@
         "DB_HOST": "localhost",
         "DB_PORT": "55432"
       },
-      "args": [
```
craicoverflow, line 31: You still need to have an "args" config to point to the file to run the app.

wtrocki:
Resolver plugin is actually staying there it is simply deprecated by not mentioning it in the future docs. Previous docs will still get it and we going to release new version of the resolver plugin to people despite not having any templates there. 

As for CLI I was looking for an effective way to deprecate command and I found that deprecating it will make no impact. If we remove codegen template there will be no way to use it anyway so having it here - even as deprecated could cause potential harm (especially that some of the migration stuff will need to be refactored to cather new CRUD). 

I think we reached the point where we need to 
I'm going to create new document to show how things could play together and draft the architecture of the datasync in the diagram that we all can relate too.

.vscode/launch.json
craicoverflow: Add this:

```
"args": [
   "${workspaceRoot}/templates/ts-apollo-postgres-backend/src/index.ts"
],
```

wtrocki:
@craicoverflow I will help with rebasing your changes. Fixing tests

PR #1240, opened 2020-05-07
Knex CRUD improvements

craicoverflow:
## Description

This PR brings some general refactoring of the core CRUD model as well as fixes some bugs and adds the ability to order items by a field in ascending or descending order

- Added separate input types for create/update and delete mutations
- Add order by types and implementation for Postgres
- Add some tests for ordering.

## Verification

```
query findNotes {
  findNotes(limit:5, orderBy:{field:"id"}) {
    items {
      id
      title
    }
  }
}
```


craicoverflow:
@wtrocki what are we going to do with [](https://github.com/aerogear/graphback/blob/7ecea7d39d7d96dd01edc19f441d74fd96aac212/packages/graphback-runtime/src/data/RESTDataProvider.ts)?

Is there even any point in having this, since people's REST API will likely not follow this format exactly, and advanced filtering is not possible in this way.

wtrocki:
RestDataProvider will be removed. I did not wanted to remove it in my PR to not cause conflicts - I'm avoiding runtime changes now. 

Target is to merge both PR's at some point an then start with further refactoring of runtime

PR #1243, opened 2020-05-08
Implement new crud for mongo

ssd71:
This PR implements the new CRUD spec for MongoDB

### DIFF
--- a/packages/graphback-runtime-mongodb/src/queryBuilder.ts
+++ b/packages/graphback-runtime-mongodb/src/queryBuilder.ts
@@ -54,17 +54,17 @@ const operatorTransform: {
     },
     contains: (value: string): OperatorTransform => {
         return [
-            [operatorMap.eq, new RegExp(value, 'g')]
+            ['$regex', new RegExp(value, 'g')]
         ]
     },
-    startswith: (value: string): OperatorTransform => {
+    startsWith: (value: string): OperatorTransform => {
         return [
-            [operatorMap.eq, new RegExp(`^${value}`, 'g')]
+            ['$regex', new RegExp(`^${value}`, 'g')]
         ]
     },
-    endswith: (value: string): OperatorTransform => {
+    endsWith: (value: string): OperatorTransform => {
         return [
-            [operatorMap.eq, new RegExp(`${value}$`, 'g')]
+            ['$regex', new RegExp(`${value}$`, 'g')]
         ]
     },
 }
@@ -78,7 +78,7 @@ function isPrimitive(test: any): boolean {
 function traverse(filter: any): any {
 
     Object.keys(filter).forEach((key: string) => {
-        if (filter.hasOwnProperty(key)) {
+
             // Transform the operators to mongodb operators
 
             // Check if it can be directly substituted
@@ -102,12 +102,10 @@ function traverse(filter: any): any {
                     filter[operator] = value;
                 });
             }
-        }
     });
 
     // If there is nesting, recursively transform all operators
     Object.keys(filter).forEach((key: string) => {
-        if (filter.hasOwnProperty(key)) {
             if (['$and', '$or', '$nor'].includes(key)) {
 
                 // If AND, OR, NOT do not have an array as their contents
@@ -122,7 +120,6 @@ function traverse(filter: any): any {
             if (!isPrimitive(filter[key])) {
                 filter[key] = traverse(filter[key]);
             }
-        }
     });
 
     return filter
@@ -132,6 +129,5 @@ export function buildQuery(filter: AdvancedFilter) {
     let query = {};
     if (filter)
         {query = traverse(filter);}
-
     return query;
 }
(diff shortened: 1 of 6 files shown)


wtrocki:
I think that having single test executed against mongo and pg. Adding this filtering to mongo dB should be quick now and we can test it end to end

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -0,0 +1,132 @@
...
+            [operatorMap.ge, values[0]],
+            [operatorMap.le, values[1]]
+        ]
+    },
+    nbetween: (values: [any, any]): OperatorTransform => {
+
+        values.sort();
+
+        return [
+            [
```
wtrocki, line 42: Can use operator above

wtrocki:
This looks amazing! Really good job. Tried it out with and query only. Can you connect it to mongodb data provider? Adding couple unit tests will be good as well

ssd71:
Connected to Mongo Data Provider :100: 

Writing tests now

packages/graphback-runtime-mongodb/src/queryBuilder.ts
ssd71: This is actually wrong implementation of this operator, fixing it now

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const AND_FIELD = 'and';
+const OR_FIELD = 'or';
+const NOT_FIELD = 'not';
+
+// Operators that work with simple substitution :P
+const operatorMap = {
+    [AND_FIELD]: '$and',
+    [OR_FIELD]: '$or',
+    [NOT_FIELD]: '$nor',
```
craicoverflow, line 11: Should this be `$not`?

craicoverflow, line 11: $nor means NOT OR right? Shouldn't NOT_FIELD map to $not?

I am not too familiar with MongoDB syntax

ssd71: [$not](https://docs.mongodb.com/manual/reference/operator/query/not/) works slightly differently in Mongo as compared to SQL, I cannot give multiple fields or even multiple objects to $not, it only works for single field like so
```
{
  description: {
    $not: {
      $eq: "MongoDB"
    }
  },
  title: {
    $not: {
      $eq: "MongoDB"
    }
  }
}
```
but with $nor, we can do:
```
{
  $nor: [
    {
      title: { $eq: "MongoDB" }
      description: { $eq: "MongoDB" }
    }
  ]
}
```
It all works out because of the way boolean nor operator works :)

ssd71:
Written some basic tests for advanced filtering, mostly for functionality that I had to implement

wtrocki:
Some small linting issues breaking build

wtrocki:
Build is failing on something unrelated

@craicoverflow any comments before we merge that back to feature branch for testing?

ssd71:
@wtrocki  Should I try and implement orderBy and resolve the merge conflict?

wtrocki:
No rush. Merge conflict should be resolved. You can add more features in upcoming days on chillout mode

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (filter.hasOwnProperty(key)) {
+            // Transform the operators to mongodb operators
+
+            // Check if it can be directly substituted
+            if (operatorMap[key]) {
+
+                // Substitute it with the mongodb operator
+                filter[operatorMap[key]] = filter[key];
+                /*eslint-disable-next-line*/
+                delete filter[key];
```
craicoverflow, line 90: https://palantir.github.io/tslint/rules/no-dynamic-delete/

> Deleting dynamically computed keys is dangerous and not well optimized.

It looks like this forEach is modifying the original `filter`, so maybe a map could be used instead?

ssd71:
Fixed all issues the issues @craicoverflow encountered, trying to use Maps instead of raw objects now

wtrocki:
Trying that now

ssd71:
Encountered errors with incorrect implementation of operators 'contains, endswith, startswith' , also fixed now

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
```diff
@@ -1,154 +1,369 @@
...
+        (
+          (post.likes >= 25) && (post.likes <= 75)
+        )
+      );
+
+      expect(
+        (post.text === 'post')
+        ||
+        (post.text === 'post2')
+      );
```
ssd71, line 329: Is there some alternative to using matchers like this? I looked but I couldn't find anything

wtrocki:
@ssd71 Checking now

PR #1244, opened 2020-05-08
fix: set default value as undefined instead of null

craicoverflow:
graphql-migrations generates a slightly different abstract database when comparing the existing database to the schema. One sets defaultValue on columns to undefined, and the other uses null. Because of this it is repeating the same migration operations every time.

## Verification

Run graphql-migrations on an already up to date database. The number of operations should be 0.


wtrocki:
Looking into it

PR #1246, opened 2020-05-11
test: replace the usage of @db.model with @model

machi1990:


machi1990:
Thanks for the quick review.

wtrocki:
It was automatically verified by CI so it was a matter to check code :D

PR #1255, opened 2020-05-12
Auth package

darahayes:


packages/graphback-keycloak-auth/README.md
```diff
@@ -0,0 +1,16 @@
...
+  <img width="400" src="https://github.com/aerogear/graphback/raw/master/website/static/img/graphback.png">
+  <br/>
+  Auto generate database structure, <br/>
+  GraphQL Resolvers and Queries from GraphQL types 🚀
+</p>
+
+**Documentation**: https://graphback.dev
+**Repository**: https://github.com/aerogear/graphback/
+
+## Graphback-Codegen-Offix
```
wtrocki, line 12: Rename

wtrocki:
The code seems well refined and put it has a systematic problem.
I have problems to see how this solves auth issues and will integrate with other capabilities and since it is indirectly referenced to implementation of the schema plugin (which is not guaranteed) it can cause some issues (sec vulnerabilities that users will experience.

Package tries hard to avoid integrating with our stack by using some hardcoded method names that are not guaranteed to be there. This indirect format could be really easy to get some unexpected results and generally not something I will be comfortable to use in my own project or verify - too much magic related to wrapping with no control over the results (and possible bugs)

It is really sad for me to come with this feedback - let's try to have faster feedback cycles to avoid this happening in the future.

Apart from systematic problems, I would avoid putting owner implementation initially - owner implementation that assumes owner field can be created on creating which might not be the case. I see it is that if you have use case for owner you should have it in the model itself rather than adding owner as a hardcoded feature. This is hard for us to document and propagate - we basically hiding some query capabilities from users that will need to inspect our code to understand it. 

I can put some comments on the individual code snippets if that will be more helpful but I personally prefer if we use something like CRUDService. 

If we extend CRUDService or change some of the args in the future we will be able to track it here. This indirect integration + auth + resolver wrapping seems like a more risky approach - while code might still be fully functional for some of the cases.

The best way I think we can proceed here is to still leave most of your work:

- Have a simple CRUD service that gives RBAC only (Copy paste spike from the starter)
- Have set of predicates that need to be applied by developers to specific resolver method - just owner function that can be applied on service.
- Reuse current implementation later to deliver some model-based definitions as you have here but - due to change based on if we actually generate schema or not (currently we are not) - I see this as least priority considering the amount of work we need to pull out, but code can stay there 

We do not have much time to spend on this feature so I will prefer something more refined and production-ready with fewer features rather than tons of fancy stuff that could possibly cause problems to the community and invalidate the entire feature for people who will need it. If we can get RBAC solid and tested by the end of the sprint that will be an amazing outcome here.

This way we do not throw away the general ideas of this PR but still keep a conservative approach to auth and ability to connect this with graphback ecosystem.

darahayes:
Cool, really appreciate the in depth feedback. I think you're right. I think I fell into the classic pit of trying to do too much in one go and it just led to some underlying problems. What I will do is close this PR and leave the branch hanging around, and we can go with the crud service for now, covering the standard RBAC cases to begin with. Let's get that solid and from there we can iterate on it and gather feedback from wider usage. Thanks 🙂

PR #1257, opened 2020-05-12
feat: keycloak-authz package

darahayes:
This PR adds the a `KeycloakCrudService` which extends the default `CrudService` from `graphback-runtime`.

So far, this crud service adds role based access control to the standard crud methods. RBAC on subscriptions and individual fields are not yet supported.

You can see this code being used in the following PR: https://github.com/aerogear/datasync-starter/pull/412

Still todo:

- [x] Add tests
- [x] Documentation


wtrocki:
>  RBAC on subscriptions and individual fields are not yet supported.

Individual fields are stretch that we might not need for POC but thinking about subscriptions - is there challenge in that or it is just the amount of time needed to land this change in this sprint?

packages/graphback-keycloak-authz/README.md
```diff
@@ -0,0 +1,16 @@
...
+  Auto generate database structure, <br/>
+  GraphQL Resolvers and Queries from GraphQL types 🚀
+</p>
+
+**Documentation**: https://graphback.dev
+**Repository**: https://github.com/aerogear/graphback/
+
+## Graphback Keycloak Authz
+
+Add Keycloak authorization to Graphback runtime applications
```
wtrocki, line 14: Let's add some info how this can be consumed. 
Taking look and playing with datasync starter to see how we can do this.

packages/graphback-keycloak-authz/src/createCrudServiceAuthConfigs.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Convert global auth config supplied by the user into a set of configs that can be applied
+ * to each model's CRUD Service
+ * 
+ * Done in this way so that the user facing config can support 
+ * additional features such as owner based auth in the future without severe breaking changes
+ * 
+ * @param models The model configs that contain auth rules for each model
+ */
+export function createCrudServiceAuthConfigs(models: ModelAuthConfigs): CrudServiceAuthConfig {
```
wtrocki, line 14: Really clever mapping. However should this actually come from user? Not strong opinion but would developers define this rules per model/service based approach - this kinda boils back to what runtime API will look like, but want to understand why we need that transformation

wtrocki:
I might have more questions once finished verification. In the mean time we can refine this to pass tests add doc to readme. 

Stretch goals:
- Relationships ( Sub Queries/Batch) 
- Subscriptions

packages/graphback-keycloak-authz/jest.config.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable */
+const baseConfig = require("../../jest.config");
+const packageName = 'graphback-keycloak-auth';
```
wtrocki, line 3:
```suggestion
const packageName = 'graphback-keycloak-authz';
```

packages/graphback-keycloak-authz/src/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  for (const operation of supportedCrudOperations) {
+    serviceConfig[operation] = { roles: [] };
+  }
+
+  return serviceConfig;
+}
+
+export function isAuthorizedByRole(roles: string[], context?: any) {
+  for (const role of roles) {
+    if (context.kauth.hasRole(role)) {
```
wtrocki, line 30: So this package will require our keycloak library as an indirect dependency? Is there easy win to document this/add keycloak-connect-graphql to dependencies?

This is about: How to connect that to the `context.kauth` with indirect dependency

packages/graphback-keycloak-authz/src/createCrudServiceAuthConfigs.ts
wtrocki: I would need to know what was your intention on this one.
Is this just helper and if we want to generate other queries/mutations there are no problems to extend it?

darahayes: It's really just a transform function that allows a developer to define auth rules in a similar way to how other projects do the same thing. That particular format could be extended to support owner based auth and so on. However, maybe we don't really need at as a user could define the configs on a service level. Something like this:

```js
const authConfigs = {
  Task: {
    create: { roles: ['developer']},
    read: { roles: ['developer']},
    update: { roles: ['admin']},
    delete: { roles: ['admin']},
  },
  Note: {
    ...
  }
}
```

Yeah this will probably be easier for a user to use.

packages/graphback-keycloak-authz/src/utils.ts
darahayes: Yeah understood. I'll have a think about it. Perhaps this package could take the keycloak-connect-graphql lib as a dependency and rexport it and then we have tests to verify the integration. I will think of other ideas.

wtrocki: No strong opinion on this - if you manage to document it right (as this will be added manually by users so they need some clear docs for this

darahayes: Yeah totally agreed!

darahayes:
Okay so based on discussion in the PR. The next things to do:

* Figure out if there's a way to have a clearer more direct dependency on keycloak-connect-graphql
* Remove the createCrudServiceConfig thing
* Add clear documentation on how to use the library
* Integration tests

packages/graphback-keycloak-authz/src/createCrudServiceAuthConfigs.ts
wtrocki: Transformation might have bugs or people might expect to do some magic. 
Having more implicit way to define stuff it ok:

```
const authConfigs = {
  Task: {
    create: { roles: ['developer']},
    read: { roles: ['developer']},
    update: { roles: ['admin']},
    // Roles are optional
    //delete: { roles: ['admin']},
  },
  Note: {
    ...
  }
}
```

This format is also good to work with later for relations:

```
const authConfigs = {
  Task: {
    create: { roles: ['developer']},
    comment: { roles: ['admin']},
  }
}
```

wtrocki: Made edit.

wtrocki:
Ok to have run?

packages/graphback-keycloak-authz/README.md
```diff
@@ -0,0 +1,285 @@
...
+Then follow the [Getting started instructions for `keycloak-connect-graphql`](https://github.com/aerogear/keycloak-connect-graphql#getting-started)
+
+Once the getting started instructions are covered, you must create a configuration that defines the authorization rules for each model within your Graphback application.
+
+Here is an example configuration.
+
+```ts
+const authConfig = {
+  Task: {
+    create: {},
```
wtrocki, line 37: Question. Can create be omitted?

darahayes: Yep, these can be omitted.

wtrocki: Sorry. I typed to fast. I meant to actually call out that in some note in the docs that we can omitt those

packages/graphback-keycloak-authz/README.md
```diff
@@ -0,0 +1,285 @@
...
+  dataProvider: KnexDBDataProvider,
+  crudService: CustomCrudService
+})
+```
+
+## Full Example showing Keycloak integration in apollo-server-express
+
+The following example is a full that shows how to set up a Graphback runtime application that uses Keycloak authentication and authorization.
+
+```ts
```
wtrocki, line 156: I would put that into file for better maintenance :)
Example will be anyway in:

- DataSync repo
- DataSync starter
- OVP 

so this could be great place for reference. We will want to add that to OVP after release and push some short blog post about auth approach. 

This docs works perfectly. We will need 2-5 sentences introduction and we can push that as new blog

wtrocki:
Verifying

PR #1276, opened 2020-05-15
Custom indexing for mongodb

ssd71:
This PR is meant for discussion about different approaches for implementation of custom indexes in mongodb, as such, mostly monkey patches here and there, will add tests after finalizing approach

With the example schema [here](https://gist.github.com/ssd71/3858d67bc58df2610c944872bbda8439)
We get indexes on:
- noteId field for comment collection
- title field for note collection


packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -20,6 +21,33 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
```
ssd71, line 22: I am very curious about this field and what it is being used for besides idField

ssd71:
I see that we are not adding relationship metadata to modelTableMap:
https://github.com/aerogear/graphback/blob/442cc70dfa2951e7a70589ed24a0bebeb36cb48a/packages/graphback-core/src/db/buildModelTableMap.ts#L55-L63
I wonder if we can implement that and perhaps use it here somehow

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
craicoverflow: This is being used for id field mapping and the table name => GraphQLObjectType name mapping.

The eventual aim was to also do a GraphQL field ==> column name mapping and this is what `fieldMap` is for. This is partially implemented but we never actually used it in the database providers, because it is a lower priority for us right now.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -20,6 +21,33 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
...
+          if (e) {
+            console.error(e);
+          } else {
+            console.log(JSON.stringify(r,null,4))
+          }
+        });
+      }
+
+      const otherAnnotations = parseAnnotations('db', baseType.getFields()[k].description) as any;
+      if (otherAnnotations?.index) {
```
craicoverflow, line 39: If someone changes the indexed column - there should probably also be a scan to remove old index?

I don't know a lot about Mongo, so forgive if I am way off :D

ssd71:
Also wondering, is this a good idea to do it in the constructor?

Can we see somebody creating and possibly deleting indexes on a production database just by creating a Mongo Data Provider?

wtrocki:
> Also wondering, is this a good idea to do it in the constructor?

Constructor would be just initial place to put it. Generally we might need to have mongocounterpart for the db migrations package, but first I want to see how much we will need to have there to justify having separate package etc.

wtrocki:
> Can we see somebody creating and possibly deleting indexes on a production database just by creating a Mongo Data Provider?

We should not deal with deleting indexed data. This is very bad advice based on now SQL databases work. For Mongo there is no way to rename field. It requires map reduce on entire collection that is typically something that you would be very careful with.

ssd71:
Renaming a field in MongoDB is equivalent to doing both $set/$unset in single operation, essentially the field is removed from the document, meaning that the any existing index on that field will need to be rebuilt: https://docs.mongodb.com/manual/reference/operator/update/rename/#behavior, https://stackoverflow.com/a/10345169

ssd71:
Indexing is also a very intensive operation, can be particularly troublesome if it had to remove and entirely recreate an index on it's own

What do you guys think?

wtrocki:
Yep.. It should not be added lightly. So it sounsd like if you rename you will have @db.index in renamed field as well so it will be rebuild.

ssd71:
> So it sounsd like if you rename you will have @db.index in renamed field as well so it will be rebuild.
Yes! Hence the concerns

wtrocki:
I generally like to point out that we mixing use cases here. 
WE ARE NOT DOING ANY PRODUCTION OPEARATIONS. 

What we want to do is to setup developer level stuff.
Which means that we might be actually having limited exposure to the indexing itself unless we can port them as some statement for production for developer level collections.

craicoverflow:
> Indexing is also a very intensive operation, can be particularly troublesome if it had to remove and entirely recreate an index on it's own

If this just to happen at development level and shouldn't happen in production, maybe this should not matter too much?

ssd71:
Sounds good to me, so in this context, should we still consider removing/recreating indexes?

wtrocki:
I do think that we have all we wanted.

craicoverflow:
> Sounds good to me, so in this context, should we still consider removing/recreating indexes?

You could argue that in this context it doesn't matter about removing old indexes in development either, since it would not affect production code.

However I do think it should keep the db structure matching how it would be in production, as this can also affect the ordering of how data is retrieved which might influence the design of the app that consumes the data.

wtrocki:
Challenge here is that indexes have value in production. In SQL land indexes will be ported when doing schema analysis. So we applying indexes here on dev for purpose of performance tests etc. but giving people free will to do it in prod

wtrocki:
@ssd71 This requires rebase now to drop all irrelevant commits since it was based on the feature branch. I would like to review that again

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -20,6 +21,33 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
+    Object.keys(baseType.getFields()).forEach((k: string) => {
+      const relationshipData = parseRelationshipAnnotation(baseType.getFields()[k].description)
+      if (relationshipData?.kind && ['manyToOne', 'manyToMany'].includes(relationshipData.kind)) {
+        this.db.collection(this.collectionName).createIndex({
```
wtrocki, line 27: You can have ensureIndex that checks if index exist

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -20,6 +21,33 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
+    Object.keys(baseType.getFields()).forEach((k: string) => {
```
wtrocki, line 24: For each is not good way to iterate. use for of etc.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
wtrocki: Without support for renames changes for columns in mongo are not allowed

wtrocki:
Good job. I do think that we will need separate package for this at some point. Let's hold of after we rename runtime

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
ssd71: [ensureIndex](https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#ensureIndex) is actually deprecated, createIndex officially replaces ensureIndex, and if index already exists, createIndex does not touch anything else

ssd71:
I believe that this is back on table now

wtrocki:
Yes

ssd71:
Have rebased now and added index for datasync, and I have some thoughts about this
This is how I add index in datasync:
https://github.com/aerogear/graphback/blob/fc52edb02d56383e8717c471bd86408407f47d01/packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts#L14-L22

and then this is how I add index in mongo provider:
https://github.com/aerogear/graphback/blob/fc52edb02d56383e8717c471bd86408407f47d01/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L27-L53

I feel like there is something similar about these but problem is that we do not actually add _deleted field to baseType so I cannot do something like add `@index` to _deleted field and datasync will just reuse the code in runtime mongo to add index. wdyt

Also, what do you think about creating new package now? I think this much functionality can just be extracted out to separate file instead of creating separate package altogether, though if we do decide to make separate package, we need more functionality than just adding indexes, just my opinion :smile:

wtrocki:
No strong opinion to create new package.

wtrocki:
I like how simple this is. great work

ssd71:
Of course there is the question that how we will execute it only in dev environment if we keep it in Mongo provider

wtrocki:
So.. in production you would have already this index so there is no cost right? And if you do not then it means that we probably thinking about new collection and not large one. More like doc problem than actuall code issue. This does the job

ssd71:
Made some changes, ready for review, also we need tests for this

wtrocki:
I like the changes. I'm off laptop but generally this is getting better. I would verify it on Monday and look for @craicoverflow opinion

ssd71:
Ran into #1436 with schema:
```graphql
""" @model """
type Comment {
  id: ID!
  
  """@db(index: 1)"""
  text: String
  description: String
}
```
but otherwise it works.

As an aside: does `index: 1` make sense as somebody using it? I am currently only using it to check if index is required, is `index:true` better/supported?

craicoverflow:
> As an aside: does `index: 1` make sense as somebody using it? I am currently only using it to check if index is required, is `index:true` better/supported?

`@index` alone parses as `index:true` so you could just use that :)

wtrocki:
Yes. Let's use @index - directive here makes no sense.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -165,6 +178,60 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
...
+  }
+
+  private getCustomIndex(field: GraphQLField<any, any>): IndexSpecification {
+    const otherAnnotations = parseMetadata('db', field.description) as any;
+      if (otherAnnotations?.index) {
+        return {
+          key: {
+            [field.name]: 1
+          },
+          name: `graphback_custom_index_${field.name}_on_${this.collectionName}`
```
machi1990, line 214: It'd be awesome if we could be able to provide index name too. 

e.g via a `db.index: 'index-name-here'`

/cc @craicoverflow @wtrocki

machi1990:
> Ran into #1436 with schema:
> 
> ```graphql
> """ @model """
> type Comment {
>   id: ID!
>   
>   """@db(index: 1)"""
>   text: String
>   description: String
> }
> ```
> 
> but otherwise it works.
> 

I am having this error/warning when testing this locally:

```
Can't parse annotation index:1  ): Unexpected token ')'

```

My model is:
```
""" @model """
type Note {
  id: ID!
  """@db(index:1)"""
  title: String!
  description: String
  """
  @oneToMany field: 'note'
  """
  comments: [Comment]!
}

""" @model """
type Comment {
  id: ID!
  text: String
  description: String
}

```

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -16,16 +17,28 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
...
+
+    if (this.indexes.length > 0) {
+      // Apply indexes
+      this.db.collection(this.collectionName).createIndexes(this.indexes).catch((err: any) => {
+        throw Error(err)
+      });
+    }
   }
   
   public setBaseType(baseType: GraphQLObjectType<any, any, { [key: string]: any; }>): void {
```
craicoverflow, line 39: This function can be taken out now

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -165,6 +178,60 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
...
+        res.push(relationIndex);
+      }
+
+    })
+    
+    return res;
+  }
+
+  private getCustomIndex(field: GraphQLField<any, any>): IndexSpecification {
+    const otherAnnotations = parseMetadata('db', field.description) as any;
```
craicoverflow, line 208:
```suggestion
    const otherAnnotations: any = parseMetadata('db', field.description);
```

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
craicoverflow: +1 on that @machi1990 

It could go one step further and parse the index annotation directly as an `IndexSpecification` interface, falling back on defaults otherwise?

```
"""
@index(
  name: 'my_index',
  unique: true,
  expiresAfterSeconds: 3600,
  collationDocument: {
    locale: 'en_IE'
  }
)
"""
```

ssd71: Sounds cool! How about:
```
@index(name: 'index123')
```
With this, we could even use `@index` and it would use a default index name

wtrocki:
@machi1990 World class review! Thanks for double checking :D

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
machi1990: +1 on all your  suggestions. Good stuffs.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
craicoverflow, line 208: Let's not use the `db` namespace, see https://github.com/aerogear/graphback/issues/1436#issuecomment-644045719

ssd71:
What about compound indexes? Does postgres support those? I was wondering if we could implement that here as well?

craicoverflow:
`Can't parse annotation index:1  ): Unexpected token ')'` happens because `buildModelTableMap` parses all `db` namespaced annotations. If we use an `@index` namespace this issue will be resolved. See https://github.com/aerogear/graphback/issues/1436#issuecomment-644045719

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
craicoverflow: > With this, we could even use @index and it would use a default index name

Exactly. Only `key` is required in `IndexSpecification`.

>  If unspecified, MongoDB generates an index name by concatenating the names of the indexed fields and the sort order.

wtrocki:
Conflict with master

ssd71:
@craicoverflow I have moved the testing setup function in mongo to a different file, I did check it out, but would love for you point out if I left something out while rebasing :smile: :
https://github.com/aerogear/graphback/blob/7490b389e18eed8a8515a47b44a047d0c34d4884/packages/graphback-runtime-mongodb/tests/__util__.ts#L13

ssd71:
> Conflict with master

Rebased!

ssd71:
> I can see db index being created, but we are missing tests I'd suggest to add those.

@machi1990 Tests added as well! feel free to take a look

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -165,6 +173,61 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     return document;
   }
 
+  protected getIndexFields(baseType: GraphQLObjectType): IndexSpecification[] {
+    const res: IndexSpecification[] = [];
+    const fields = baseType.getFields();
+    Object.keys(fields).forEach((k: string) => {
+      const field = fields[k];
+
+      // Add custom Index if found e.g. @db(index: 1)
```
machi1990, line 182: This comment is not valid anymore, 

I believe it is `@index(name: ....)` now right?

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -165,6 +173,61 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
...
+      if (relationIndex !== undefined) {
+        res.push(relationIndex);
+      }
+
+    })
+    
+    return res;
+  }
+
+  private getCustomIndex(field: GraphQLField<any, any>): IndexSpecification {
```
craicoverflow, line 202: t

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
```diff
@@ -279,3 +222,122 @@ describe('MongoDBDataProvider Basic CRUD', () => {
     expect(next.createdAt).toEqual(createDate.getTime())
   })
 });
+
+describe('MongoDB indexing', () => {
```
craicoverflow, line 226: When/if we add a file for indexing, the tests can be moved too.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -107,4 +107,20 @@ export class DataSyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
+      ...res,
+      {
+        key: {
+          // Add Index on deleted field
+          _deleted: 1
+        },
+        name: 'graphback_datasync_softdeletion_index'
+      }
+    ]
+    // TODO: Add TTL after DateTime scalars added
```
wtrocki, line 124: Every TODO = GitHub issue

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -165,6 +173,61 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     return document;
   }
 
+  protected getIndexFields(baseType: GraphQLObjectType): IndexSpecification[] {
```
wtrocki, line 176: This should be done outside this methods. Like helper class

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
wtrocki: Docs docs docs

ssd71:
Something to remind myself:
> Starting in version 4.2, the createIndexes command and the mongo shell helpers db.collection.createIndex() and db.collection.createIndexes() report an error if you create an index with one name, and then try to create the same index again but with another name.

taken from [here](https://docs.mongodb.com/manual/reference/command/createIndexes/#index-names)

Since this error varies through mongo versions, I think I might have to cover this in docs

ssd71:
I am in the process of adding docs for metadata in #1422 , would it be okay to do that there?

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
ssd71: What do you think about letting these tests stay as a sort of integration test, if yes, should I do some unit tests for this as well?

wtrocki:
@ssd71 Perfect. you have many PR's so you can consolidate them as well if needed or merge them separately. I requested review from the team. This one seems to be the most complex change so far so let's see if we have all issues addressed.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -22,12 +24,12 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
     this.fieldTransformMap = getFieldTransformations(baseType);
+
+    findAndCreateIndexes(baseType, this.db.collection(this.collectionName)).catch((e: Error) => {
```
machi1990, line 28: Seems like the `applyIndexes` function already catches the error and does not re-thrown: https://github.com/aerogear/graphback/pull/1276/files#diff-0310a0356ee20ce16b7b5d3f6d8d6420R16, maybe this catch  clause here is redudant.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -11,6 +11,15 @@ export class DataSyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
 
   public constructor(baseType: GraphQLObjectType, client: any) {
     super(baseType, client);
+    applyIndexes([
+      {
+        key: {
+          _deleted: 1
+        }
+      }
+    ],this.db.collection(this.collectionName)).catch((e: any) => {
```
machi1990, line 20: Seems like the `applyIndexes` function already catches the error and does not re-thrown: https://github.com/aerogear/graphback/pull/1276/files#diff-0310a0356ee20ce16b7b5d3f6d8d6420R16, maybe this catch  clause here is redudant.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
ssd71: I agree, but there's [this](https://palantir.github.io//tslint/rules/no-floating-promises/) rule, I will add a tslintignore comment in sometime, wdyt

machi1990: Okay, we can look at that later on. It was a minor nitpick.

PR #1277, opened 2020-05-15
Schema fixes

wtrocki:
Restoring some old code to get relationships and adjusting models to new format:
https://gist.github.com/wtrocki/f5ca13f32baac323a53be0c383ab916c~

WIP

- [x] Schema updates
- [x] Adjust client side
- [x] Regenerate client/backend
- [ ] Fix integration tests


packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -157,18 +150,35 @@ export const buildFilterInputType = (modelType: GraphQLObjectType) => {
   });
 }
 
-export const buildCreateMutationInputType = (modelType: GraphQLObjectType) => {
+// Fetching all fields required for input type including relationships
+const getAllModelFields = (modelType: GraphQLObjectType<any, any, { [key: string]: any; }>, model: ModelDefinition) => {
```
craicoverflow, line 154: Don't need to pass both the graphql type and model definition, since the model definition has both.

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -52,15 +53,16 @@ export const getFieldName = (typeName: string, action: GraphbackOperationType):
...
-      return `Create${typeName}Input`
-    case GraphbackOperationType.UPDATE:
-      return `Update${typeName}Input`
-    case GraphbackOperationType.DELETE:
-      return `Delete${typeName}Input`
+      return `${finalName}CreationInput`
+    case GraphbackOperationType.UPDATE || GraphbackOperationType.DELETE:
+      return `${finalName}MutationInput`
+    case GraphbackOperationType.SUBSCRIPTION_CREATE || GraphbackOperationType.SUBSCRIPTION_UPDATE || GraphbackOperationType.SUBSCRIPTION_DELETE:
+      return `${finalName}SubscriptionFilter`
```
craicoverflow, line 65: I think you can do:

```ts
    case GraphbackOperationType.SUBSCRIPTION_CREATE:
    case GraphbackOperationType.SUBSCRIPTION_UPDATE:
    case GraphbackOperationType.SUBSCRIPTION_DELETE:
      return `${finalName}SubscriptionFilter`
```

Which is a bit neater

wtrocki: Switch case is evil :)

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
wtrocki: Not sure what you mean. I'm adding idField here :D

craicoverflow, line 154: I think you can just have one function argument:

```suggestion
const getAllModelFields = (model: ModelDefinition) => {
```

modelType is a child of model.

wtrocki: Yep. Autogenerated. I'm adding aggregate for pagination now

wtrocki: Yes. This was autogenerated by IDE

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -157,18 +150,35 @@ export const buildFilterInputType = (modelType: GraphQLObjectType) => {
...
+  const modelFields = Object.values(modelType.getFields());
+  const scalarFields = modelFields.filter((f: GraphQLField<any, any>) => isScalarType(getNamedType(f.type)));
+  const relationshipFields = model.relationships.map((relationship: FieldRelationshipMetadata) => {
+    return {
+      name: getInputFieldName(relationship.ownerField),
```
craicoverflow, line 159: This is throwing an 'Not inputtable error' because you are trying to map a 1:N field to an input, but this is invalid (inputs should only accept scalar with reference to foreign key field). Basically this should be filtered out before this mapping is applied.

craicoverflow: Since this functions gets all model fields and not necessarily input fields, there should be no need to map the fields to input fields at this point.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -186,16 +196,43 @@ export const buildCreateMutationInputType = (modelType: GraphQLObjectType) => {
   });
 }
 
-export const buildUpdateMutationInputType = (modelType: GraphQLObjectType) => {
+export const buildSubscriptionFilterType = (model: ModelDefinition) => {
+  const modelType = model.graphqlType;
+  const inputTypeName = getInputTypeName(modelType.name, GraphbackOperationType.CREATE);
+  const modelFields = Object.values(modelType.getFields());
+  const scalarFields = modelFields.filter((f: GraphQLField<any, any>) => isScalarType(getNamedType(f.type)));
```
craicoverflow, line 203: Should this also get generated relationship fields?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -120,6 +123,16 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       queryTypes = this.createQueries(model, queryTypes);
       mutationTypes = this.createMutations(model, mutationTypes);
       subscriptionTypes = this.createSubscriptions(model, subscriptionTypes);
+
+      const modifiedType = schemaComposer.getOTC(model.graphqlType.name);
+      const modelRelationshipFilterFields = buildRelationshipFilterField(model);
+
+      // TODO: Fix error "Error: Type with name "CommentFilter" does not exists"
```
craicoverflow, line 130: @wtrocki i think once this is resolved it should be working.

wtrocki:
I have adjusted schema to support pagination containers and fixed problems with the client side plugins. Build fails for me on relationships. 

I will adjust integration tests (despite they might be failing now) leaving you to deal with relationships implementation - we can pair program on this

craicoverflow:
> I have adjusted schema to support pagination containers and fixed problems with the client side plugin. Build fails for me on relationships.
> 
> I will adjust integration tests (despite they might be failing now) leaving you to deal with relationships implementation - we can pair program on this

https://github.com/aerogear/graphback/pull/1277/commits/bb43c2d14edeb7618589e3c3985dd9fff9cf8847 should resolve this. We were still mixing non graphql-compose and graphql-compose to build our schema. We should always use graphql-compose in plugin to avoid chicken and egg problem.

Need to resolve this next:

Screenshot from 2020-05-19 13-12-09

Returning nested data from mutation is failing

craicoverflow:
I am wondering whether we should go further and embed the GraphQL Compose Object Type Composer into the model definition. Since this metadata is for generation only it would make it much easier to work with schema information.

packages/graphback-codegen-schema/tests/__snapshots__/GraphQLSchemaCreatorTest.ts.snap
```diff
@@ -109,13 +114,27 @@ type Comment {
   noteComment: Note!
 }
 
-input CommentFilterInput {
+input CommentCreationInput {
+  id: ID
+  title: String!
+  description: String!
+  note_id: ID!
```
wtrocki, line 121: @craicoverflow This is example of the problem you mentioned on the call. Automatic generation of the opposite side not always make sense. For example here we getting capability to move comment to different note which in real life makes no sense

craicoverflow: I don't fully understand - without looking into this too much `note_id` appears to be correct?

wtrocki: yes. it behaves correctly but it makes no sense for end-user. 
Why would I need ability to move my comment to different Note etc.
That is why more explicit relationships make sense.

wtrocki:
I have hit some challenge around framents for pagination vs without pagination.
I will need to implement pagination container (using Unions) and then made 
some conditional queries or always assume pagination is there (which then might break the client-side query generation as a decision about return format needs to be made there. Will research more into Unions to see how we can still get what we want. When querying I need to assume that there is pagination container etc.


Exact problem I'm getting
```
GraphQLDocumentError: Fragment "CommentExpandedFields" cannot be spread here as objects of type "CommentResultList" can never be of type "Comment".
```

craicoverflow:
Here is an example of the inputs we talked about creating:

```
""" Represents Input for the exact object in create"""
input CreateCommentInput {
  someid: ID
  title: String!
  description: String!
}

""" Represents Input for the update/delete """
input CommentMutateInput {
  """ This is dynamic"""
  someid: ID!
  title: String
  description: String
}

""" Represents object filter used for subscriptions """
input CommentSubscriptionFilter {
  someid: ID
  title: String
  description: String
}

""" Used for find """
input CommentFilter {
  someid: IDInput
  text: StringInput
  description: StringInput
  title: StringInput
  and: [CommentFilterInput]
  or: [CommentFilterInput]
  not: CommentFilterInput
}
```

However we are naming them `CommentCreationInput`, `CommentMutationInput`. Any reason for this change?

I prefer `CreateCommentInput` and `MutateCommentInput`. This is closer to how you would say it using natural human language, so it is much easier to recognise the input meaning from just the name.

packages/graphback-codegen-schema/tests/__snapshots__/GraphQLSchemaCreatorTest.ts.snap
craicoverflow: Yeah explicit relationships do make more sense, but we put explicit because of those rare cases where a model references another in two fields, for example:

```gql
type Message {
  """
  @manyToOne field: 'sentMessages'
  """
  from: User
  """
  @manyToOne field: 'receivedMessages'
  """
  to: User
}
```

We are aware of the need to enable explicit relationships alright.

craicoverflow:
> I will need to implement pagination container (using Unions) and then made
some conditional queries or always assume pagination is there (which then might break the client-side query generation as a decision about return format needs to be made there.

Could you generate fragments for both cases and let the user load the correct one (pagination versus none)?

wtrocki:
> `prefer CreateCommentInput and MutateCommentInput`

Yes. It is actually easy to make this change now as both client and schema use the same helpers.

wtrocki:
> Could you generate fragments for both cases and let the user load the correct one (pagination versus none)?

That still causing issues when someone do not provide page. 
Generally unions are not something that regular folks will use.

I have tried to generate query with pagination and query without pagination and that would work nicely but it is too verbose.

For the moment I think it is easier for us to stick with the pagination container as default and mitigate other issues

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -69,13 +69,14 @@ export class LayeredRuntimeResolverCreator {
...
         const findOneField = getFieldName(modelName, GraphbackOperationType.FIND_ONE);
+        const primaryKeyLabel = getPrimaryKey(resolverElement.graphqlType).name;
         //tslint:disable-next-line: no-any
         resolvers.Query[findOneField] = (parent: any, args: any, context: any) => {
           if (!this.services[modelName]) {
             throw new Error(`Missing service for ${modelName}`);
           }
 
-          return this.services[modelName].findOne(args, context)
+          return this.services[modelName].findOne({ [primaryKeyLabel]: args.id }, context)
```
craicoverflow, line 79: There needs to be an annotation provided to map the static `id` to the true primary key, for cases where the primary key is something different, for example `email`.

craicoverflow: Never mind I was misunderstanding it.

craicoverflow: My biggest issue with only performing ID mapping at resolver level is that this is not obvious from the schema. The user may not make a connection between the static id and their primary key.

```gql
type User {
  id: ID!
  """
  @db.primary
  """
  email: String
}
```

In this example, `findOne` will remap id to email, but from the user perspective they may expect it to fetch by their explicitly defined ID field.

We should additionally have an annotation to let the user add this mapping.

wtrocki: Happy for you to take over with something different. 
I never actually liked get in the first place :D

PR #1288, opened 2020-05-21
Keycloak crud api

darahayes:


wtrocki:
@darahayes Would you mind doing release for this as fav :)
Thank you!

PR #1295, opened 2020-05-22
fix: escape regex characters in string searches(mongo)

ssd71:
Fixes #1266 for mongo

# Verification

```
mutation create {
  createNote(input: { title: "23yeet39", description: "1" }) {
    title,
    id,
    description
  }
}

query find {
  findNotes(filter: { title: { contains: "y..t" } }) {
    items {
      title
    }
  }
}
```

Issue #630, opened 2020-01-14
Hook into existing database without migrations


lastmjs:
I'm not sure if this is possible in graphback, but I would love it to be. One essential thing that Prisma 1 does is allow users to hook into an existing database, and allow other tools to handle migrations. This allowed me to generate all of my GraphQL infrastructure using Prisma, but for an already-existing Postgres DB whose migrations are handled with Rails' Active Record. Because we're still in the process of migrating away from our Rails app, it is convenient to allow the Rails app to handle database migrations.

Does graphback support this type of workflow? Could it? As Prisma 2 seems to be going in a direction that I don't agree with, I'm assessing the viability of graphback as a replacement

lastmjs:
It would also be nice if graphback could be used entirely without Docker, and just hook into an existing Postgres database through a config file that points to the database URL, username, password, etc

wtrocki:
Graphback is designed as a set of tools rather than closed ecosystem. It consists:

- A number of code generators for schema, resolvers etc.
Each of those can be swapped for developer needs
- Runtime layer (similar to Prisma) resolves the challenge of making performant queries to the database without generating a lots of complex codebase.
- DB migrations package that makes changes in db

Each of those can be used separately or together. I will add more info into the docs to make this clear.  

As for the "bring your own database" - technically this is really simple to achieve however, graphback at the moment is not providing db to graphql mapping package, which means that data from the database needs to conform with the graphql fields and types. This limitation can be resolved by:

1) simply providing your own runtime layer that will understand how your database is structured 
https://graphback.dev/docs/runtime

2) writing some generator that will create some ORM types etc. 

In both cases there will be some extra work needed that might depend on the structure of your database etc.
Our team can check if we can build some example that does that kind of mapping to understand if runtime will allow easily to support those cases. We have written custom solutions for these cases in the past (with a strong focus on performance). 
I'm going to pin this issue to see if we can get some opinions about this and possible approaches.

To support that we can have a community runtime layer that works with Bookshelf. See: https://github.com/knex/knex/issues/498 

When we think about it we are closer to AWS App Sync rather than Prisma2 only because Schema really drives the data.

**In TL;DR Graphback will give much more flexibility and freedom (and migration patterns) but it will mean much more work required on developer side to figure out the best pattern**

> It would also be nice if graphback could be used entirely without Docker, and just hook into an existing Postgres database through a config file that points to the database URL, username, password, etc

[...]

PR #1301, opened 2020-05-23
feat: add support for createdAt,updatedAt directives for Offix Mongo

ssd71:
This PR will (eventually) add support for auto-managing of fields marked with @createdAt, @updatedAt directives for Offix Mongo
Work Is In Progress.

# Current Status

```
directive @createdAt on FIELD_DEFINITION
directive @updatedAt on FIELD_DEFINITION

"""@model"""
type Vote {
  id: ID!
  text: String
  created: String @createdAt
  lastModified: String @updatedAt
}
```

## TO DO

- [x] Add functionality for @createdAt @updatedAt for OffixMongo DataProvider
- [x] Add tests
- [x] Cleanup debug output (where applicable)
- [x] ~Add functionality for automatically adding typedefs for directives to schema (is this something that we should do?)~
- [x] Add filtering features
- [x] Add filtering tests


wtrocki:
Thanks for creating PR. I'm worried we did not had discussion on how we going to mark existence of this fields. Generally we do not use directives anywhere in GB. any reason to use directives over annotations.

ssd71:
Agreed @wtrocki, I was having some trouble with directives anyway :smiley: 
In with annotations, out with directives then?

wtrocki:
I actually like this approach. Initially I was thinking that data layer will add those and users will need to explicitly call out those in schema. This gives ultimate flexibility

ssd71:
Fair enough :thinking: So we go on this?

wtrocki:
Put +1 on comment, but making sure you have seen it

ssd71:
@wtrocki I have made some changes to Mongo provider tests in #1295 , would like for you to review and consider merging it(or not) before writing more tests (So I don't have to solve merge conflicts :D)

wtrocki:
Sorry. Done that in wrong PR. Other PR is merged now

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
+  protected updatedAtField: string;
+  protected createdAtField: string;
 
   public constructor(baseType: GraphQLObjectType, client: any) {
     super(baseType, client);
+    this.createdAtField = undefined;
+    this.updatedAtField = undefined;
+    Object.keys(baseType.getFields()).forEach((k: string) => {
+      if (baseType.getFields()[k]?.astNode?.directives) {
+        baseType.getFields()[k].astNode.directives.forEach((directive: any) => {
```
ssd71, line 22: I have changed up how we look for directives, is this okay?

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
+      res = await super.update({
+        ...res,
+        id: res.id,
+        [this.updatedAtField]: (new ObjectId(res.id)).getTimestamp()
+      });
+    }
 
-    return super.create(data);
+    return this.mapFields(res);
   }
```
machi1990, line 57: Can we issue one query to mongo here? 

I am thinking something like 
```javascript
const creationDate = new Date(); // or Date.now()
return super.create({
...data,
[this.createdAtField]: this.createdAtField ? creationDate : undefined,
[this.updatedAtField]: this.updatedAtField ? creationDate : undefined
}) 
```

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -42,13 +75,25 @@ export class OffixMongoDBDataProvider<Type = any, GraphbackContext = any> extend
...
     }
 
     throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
+
+  protected mapFields(document: any): any {
+    document = super.mapFields(document);
+    if (this.createdAtField) {
+      document[this.createdAtField] = new ObjectId(document.id).getTimestamp();
+    }
```
machi1990, line 95: We should probably do the same for `createdAtField`?

packages/graphback-runtime-mongodb/tests/OffixMongoDBDataProviderTest.ts
```diff
@@ -0,0 +1,97 @@
...
+            text: 'note'
+        },
+        {
+            text: 'note2'
+        }
+    ]
+
+    afterEach(async () => {
+        await context.server.stop();
+    });
```
machi1990, line 72:
```suggestion
    afterEach(() => context.server.stop());
```

no need for async/await, you could just return the func and jest runner will handle for you.

packages/graphback-runtime-mongodb/tests/OffixMongoDBDataProviderTest.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    it.only('can maintain createdAt timestamp', async () => {
```
machi1990, line 74: We should remove `.only` before merging this one ;-)

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: Currently, there is no actual createdAt field in the actual documents, it is a pseudo-field?(help me) 
I am getting creation timestamp from the ObjectId of the document see [this](https://docs.mongodb.com/manual/reference/method/ObjectId/#ObjectId)
https://github.com/aerogear/graphback/blob/d50fcdbec9b2d6e7b36250c13c301a156b3d5666/packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts#L91
So at this point we could very well store both fields, I am not against that approach if it aligns more with what we are doing/ :)

wtrocki: Out of curiosity what and why is this bad idea? Using ObjectId as creation timestamp is the only way to ensure that this time is accurate. I will check what mongose folks are doing but I guess they would do something similar

wtrocki: I was wrong - they use `new Date()`
https://mongoosejs.com/docs/guide.html#timestamps

ssd71: I think if somebody uses graphback as their primary crud tool, they will be surprised if they didn't find any created at field in the actual document. 

Also if we go with creating an actual field, there is more flexibility: if somebody doesn't want to use extra field they could implement this objectID-to-timestamp approach  (in client side maybe) and just omit the @createdAt directive from schema (only possible for mongo and not postgres). If someone is working with mongo extensively, they would normally know this and use this as well.

We also have option of creating new field using the createdAt directive (I am mulling over what advantages this has besides simplifying the the [filtering code](https://kchodorow.com/2011/12/20/querying-for-timestamps-using-objectids/) a **lot** and perhaps making the data more portable? I'm not sure), but I don't think this has any significant disadvantages at least (besides using more space)

I would also assume the new field option would be less prone to bugs (since less code actually involved), plus easier to test as well.

I am kinda impartial to both so would love some opinions!

ssd71:
Added some more tasks (filtering w/ created/updatedAt) to TODO, will take some more time for this

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
wtrocki: Ok. We have some prototype and investigation. Let's step back to the issue and explain how this might work (for both DBs)

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
  */
 export class OffixMongoDBDataProvider<Type = any, GraphbackContext = any> extends MongoDBDataProvider<Type, GraphbackContext> {
+  protected updatedAtField: string;
+  protected createdAtField: string;
 
   public constructor(baseType: GraphQLObjectType, client: any) {
     super(baseType, client);
+    this.createdAtField = undefined;
+    this.updatedAtField = undefined;
+    Object.keys(baseType.getFields()).forEach((k: string) => {
```
craicoverflow, line 20: This can be implemented in MongoDBDataProvider instead so that it can be use without having to install Offix.

You could also abstract the directive parsing out of the data provider so it can be reused across different providers (since there is nothing specific about Mongo here yet).

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
+      if (baseType.getFields()[k]?.astNode?.directives) {
+        baseType.getFields()[k].astNode.directives.forEach((directive: any) => {
+          if (directive?.name?.value === "createdAt") {
+            if (this.createdAtField === undefined) {
+              this.createdAtField = baseType.getFields()[k]?.name;
+            } else {
+              throw Error("Cannot have more than one field with createdAt directive");
```
craicoverflow, line 27: This validation should be handled in SchemaCRUDPlugin. This way it is caught really early before the server starts and the user won't get any surprise when the API is deployed somewhere.

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
   public constructor(baseType: GraphQLObjectType, client: any) {
     super(baseType, client);
+    this.createdAtField = undefined;
+    this.updatedAtField = undefined;
+    Object.keys(baseType.getFields()).forEach((k: string) => {
+      if (baseType.getFields()[k]?.astNode?.directives) {
+        baseType.getFields()[k].astNode.directives.forEach((directive: any) => {
+          if (directive?.name?.value === "createdAt") {
+            if (this.createdAtField === undefined) {
+              this.createdAtField = baseType.getFields()[k]?.name;
```
craicoverflow, line 25: This is an idea to make more generic directives, it would mean they would need to be implemented per database type, but means logic could be reused in any CRUD method

```ts
this.directives: DirectiveObj = {
  updatedAt: {
    fieldName: baseType.getFields()[k]?.name,
    resolve: (id) => {
      return (new ObjectId(id)).getTimestamp()
    }
  }
}
```

Then in your CRUD function:

```ts
const { updatedAt } = this.directives;

res = await super.update({
  ...res,
  id: res.id,
  [updatedAt.fieldName]: updatedAt.resolve(res.id)
});
```

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -42,13 +75,25 @@ export class OffixMongoDBDataProvider<Type = any, GraphbackContext = any> extend
...
+        return this.mapFields(data);
       }
     }
 
     throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
+
+  protected mapFields(document: any): any {
+    document = super.mapFields(document);
+    if (this.createdAtField) {
```
craicoverflow, line 93: This will make the createdAt field get overwritten each time, correct? This should only happen in `create`

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: Sorry! I am not sure I understand :(

What I was thinking was we would store both of these things in the database itself so that would eliminate the need to doing all the hacking that is present in this PR currently

ssd71: What you are saying is a good idea, I was also thinking it would be better to unify these 'things'(directives/metadata?), but I am not sure I completely understand your example

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: Actually I am using this function to map fields on docs that the data provider returns. It is very ugly but I will try to explain anyhow. The provider does not actually store a createdAt field in the database. I am using this method to add that field to output documents by getting the timestamp from the document's ObjectId.

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
craicoverflow: > What I was thinking was we would store both of these things in the database itself

Store what exactly in the database itself?

> Sorry! I am not sure I understand :(

No problem :)

Basically my example is very similar to yours, except instead of implementing the logic in the provider method, it could be be built beforehand, Ive made a little gist prototype:

https://gist.github.com/craicoverflow/a165dbb42688e26d794f6f0a18bd2ebc#file-offixmongodbdataproviderwithdirectives-ts-L21

https://gist.github.com/craicoverflow/a165dbb42688e26d794f6f0a18bd2ebc#file-offixmongodbdataproviderwithdirectives-ts-L86

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -10,17 +10,50 @@ import { MongoDBDataProvider } from './MongoDBDataProvider';
...
   }
 
   public async create(data: any): Promise<Type> {
     if (!data.version) {
       data.version = 1;
     }
+    const o = await super.create(data);
+    let res = o as any;
+    if (this.updatedAtField) {
+      res = await super.update({
```
craicoverflow, line 49: Why is there a second mutation in `create`? Can't this all be done in a single create?

machi1990: Same remark as the one I added here https://github.com/aerogear/graphback/pull/1301#discussion_r430042997

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: Oh I see! That is indeed useful!

> Store what exactly in the database itself?

I was actually thinking something else, so that statement isn't actually relevant to the discussion here

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: This is a hack for setting the updatedAt field of a newly created document to be the same as it's createdAt

ssd71:
I am going all-metadata from here on out, just saying

craicoverflow:
> I am going all-annotations no-directives from here on out, just saying

I'm on board with that :wink:

ssd71:
I have:
- extracted annotation parsing logic
- implemented timestamps using annotation in mongo provider
- added some tests and removed irrelevant tests

I am thinking about removing Offix tests, unless any functionality is to be added there

wtrocki:
I will check it first thing in the morning

ssd71:
It seems that I cannot use advanced filtering with timestamps as Numbers, because GraphQL does not support 64bit Ints

craicoverflow:
> It seems that I cannot use advanced filtering with timestamps as GraphQL `Int`, because GraphQL does not support 64bit Ints
> 
> Although GraphQL `Float` seems to work

We intend to add a filter inputs specifically for dealing with dates, after adding DateTime scalars (probably using https://github.com/excitement-engineer/graphql-iso-date)

Will it be possible to use DateTime scalar at the GraphQL level and this would still be stored as a number in MongoDB?

wtrocki:
> Will it be possible to use DateTime scalar at the GraphQL level and this would still be stored as a number in MongoDB?

This is really bad advice. Why not use BigDecimal scalar?

ssd71:
How about something like:
```gql
"""
@model
@db.createdAt
@db.updatedAt: 'lastUpdated'
"""
type Vote {
  id: ID!
  text: String
}
```
producing runtime schema like:
```gql
type Vote {
  id: ID!
  text: String
  createdAt: String
  lastUpdated: String
}
```
I am thinking until we have DateTime scalars, we could keep this as a string in runtime schema, and the database details would be managed by data provider(s) and graphql migrations (for postgres), unless, if I should hold off on this until we get the scalars from #1315 ?

ssd71:
I am not sure if any of my points are valid for a discussion, so if you think something is trivial feel free to point it out if I am overthinking/over-optimising

> Will it be possible to use DateTime scalar at the GraphQL level and this would still be stored as a number in MongoDB

We can. If we do that:
- We have to use `344555632.543` instead of `344555632543`, which is still technically a `Number` in javascript but Databases would respectfully disagree that it is not Int, so we have to store a Float in the database. 
- Or we could just floor it and call it a day :) e.g use `344555632` instead of `344555632543`. The important distinction here is that Unix timestamps are in **seconds** since Epoch but in JavaScript `Date.now()` gives **milliseconds** since Epoch. I didn't know that. So this approach actually gives us Unix timestamps and not millisecond timestamps. I feel that this is definitely more portable e.g if people try to use `graphback` for backend and another stack for the frontend or wherever they make queries. Since we can parse Date object from datestring anyway, I think this approach is perfectly valid. Feel free to ignore first point :D

> Why not use BigDecimal scalar?

If we plan to have people actually use the `createdAt`, then wouldn't it look better to have a DateTime scalar? I mean, sure, it is the same because we have to parse the query result anyway (perf diff between iso date string vs unix timestamp ?), but using DateTime sounds a bit more sane to me. Of course, if we just use BigDecimal, code will be easier for me so I won't complain :D

craicoverflow:
> How about something like:
> 
> ```graphql
> """
> @model
> @db.createdAt
> @db.updatedAt: 'lastUpdated'
> """
> type Vote {
>   id: ID!
>   text: String
> }
> ```
> 
> producing runtime schema like:
> 
> ```graphql
> type Vote {
>   id: ID!
>   text: String
>   _createdAt: String
>   lastUpdated: String
> }
> ```

That looks confusing to me. You essentially have to write the field information anyway but in an annotation.

Here is what I think it should work (don't implement this yet, @wtrocki might disagree from a DataSync perspective).

Model:

```gql
"""
@model
@deltasync
"""
type Vote {
  id: ID!
  text: String
}
```

Generates:

```
"""
@model
"""
type Vote {
  id: ID!
  text: String
  """
  @createdAt
  """
  createdAt: String
  """
  @updatedAt
  """
  lastUpdated: String
}
```

This way, users can still have control over their timestamps if they want but can also let Graphback fully handle them.

1. Include deltasync annotation and let Graphback generate fields.
2. Or don't include deltasync and add custom timestamp fields.
3. Graphback will always look for the field annotations when setting timestamps in providers.

> 
> I am thinking until we have DateTime scalars, we could keep this as a string in runtime schema

Agreed.

> 
> I think this would work with postgres out of the box, if we have correct column data type for them?

For migrations to translate a string field to a timestamp the field can have a `@db.type: 'timestamp' annotation`, so you could generate this field annotation too and we can remove it when we add scalars.

wtrocki:
Never dissagree :) Good idea. Minimize complexity and give some control.

wtrocki:
actually deltasync or versioned can be used. Also generated one should still.have deltasync annotations (probably)

ssd71:
Some questions:

- Where should we do this SchemaCRUD plugin or Offix plugin?

- If we do it in Schema Plugin, will we run the risk of colliding with Offix plugin in the future?

wtrocki:
As we talked before..

`@versioned` is generic and datasync will based on versioned. 
When using `@delta` for datasync we will imply versioned.
No collisions possible (2 different annotations)

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -115,7 +115,12 @@ function traverse(filter: any): any {
                     filter[key] = [filter[key]];
                 }
             }
-
+            if (["createdAt", "updatedAt"].includes(key)) {
+                const entries = Object.entries(filter[key]).map((entry: [string, string])=> {
+                    return [entry[0], parseInt(entry[1], 10)]
+                })
+                filter[key] = Object.assign({}, ...Array.from(entries, ([k, v]: [string, any]) => ({[k]: v}) ));
```
machi1990, line 122: I assume here we want to create the object from entries, if so here, we can achieve this with this

```suggestion
                filter[key] = Object.fromEntries(entries);
```

ssd71: I tried that too but I got `Property 'fromEntries' does not exist on type 'ObjectConstructor'`

ssd71:
Schema plugin now adds createdAt and updatedAt fields when it sees a `@versioned` or `@delta` on the type

wtrocki:
Requires rebase

packages/graphback-runtime-mongodb/src/queryBuilder.ts
machi1990: Okay, you must have at least node 12.0.0.

machi1990: Anyway, it was a minor suggestion. You can ignore it for now and focus on the big stuffs ;-)

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
```diff
@@ -213,4 +213,59 @@ describe('MongoDBDataProvider Basic CRUD', () => {
...
+  })
+
+  test('updatedAt', async () => {
+    context = await createTestingContext(`
+    """@model"""
+    type Note {
+      id: ID!
+      text: String
+      """
+      @createdAt
```
wtrocki, line 247: I thought we agreed with different format ( `@versioned` or something like that

ssd71: Haven't updated tests yet

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -30,11 +33,16 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
       delete data[idField.name];
     }
 
+    this.fieldTransforms
```
wtrocki, line 36: Did we just implemented a generic field transformation engine?

ssd71: It was discussed here: https://github.com/aerogear/graphback/pull/1301#discussion_r430277272

But I have messed up naming the types and hence the result.

wtrocki: This is amazing. So technically this can be used for different reasons - for example we can hook field name into it and do CAMEL_CASE transformation per type or each individual field. You have implemented base for https://github.com/aerogear/graphback/issues/630

image

ssd71: Happy that I helped 😄

ssd71:
Is this final schema okay?
```gql
"""
@model

"""
type Vote {
  id: ID!
  text: String

  """
  @createdAt
  @db.type: 'timestamp'
  """
  createdAt: String

  """
  @updatedAt
  @db.type: 'timestamp'
  """
  updatedAt: String
}
```

wtrocki:
I would still add @versioned marker (keep it as it is in the model right?)

ssd71:
> I would still add @versioned marker (keep it as it is in the model right?)

The `@delta` marker persists though, do we still need versioned to persist to final schema as well 🤔

wtrocki:
I'm ok with that. I was thinking that you have been removing it. if yes do not bother not should stay

packages/graphback-runtime-mongodb/src/queryBuilder.ts
ssd71: I think we need to add [es2019.object](https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2019.object.d.ts) to tsconfig.lib to make it work

Perhaps here would be fine: https://github.com/aerogear/graphback/blob/db8d973108a973665488416dd761a2f68a4b8c41/packages/graphback-runtime-mongodb/tsconfig.json#L10

ssd71:
I am only removing the `@versioned` marker here:
https://github.com/aerogear/graphback/blob/db8d973108a973665488416dd761a2f68a4b8c41/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L293

Is this safe to do?

wtrocki:
But what is the purpose of that. There is no need to remove it. I will be confusing

ssd71:
Yep. Sorry, I seem to have misunderstood.

wtrocki:
Looks good to me but requires verification. @machi1990 @craicoverflow

machi1990:
> Looks good to me but requires verification. @machi1990 @craicoverflow

I'll need to sit around this and and take a detailed look. But, I must say this is a wonderful job.

packages/graphback-runtime-mongodb/tests/queryBuilderTest.ts
```diff
@@ -266,4 +267,45 @@ describe('MongoDBDataProvider Advanced Filtering', () => {
...
+    }
+    `)
+    const startTime = 1590679886048;
+    advanceTo(startTime);
+
+    // Create some posts
+    await Promise.all(
+
+      ["hi guys", "not yet", "bye guys"]
+        .map(async (postTitle: string) => {
```
machi1990, line 295:
```suggestion
        .map((postTitle: string) => {
```

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
```diff
@@ -42,6 +43,13 @@ export class OffixMongoDBDataProvider<Type = any, GraphbackContext = any> extend
...
+
+      this.fieldTransforms
+      .filter((t: FieldTransform) => t.transformType === TransformType.UPDATE)
+      .forEach((f: FieldTransform) => {
+        data[f.fieldName] = f.transform(f.fieldName);
+      });
+
       // TODO use findOneAndUpdate to check consistency afterwards
       const result = await this.db.collection(this.collectionName).updateOne({ _id: new ObjectId(idField.value) }, { $set: data });
       if (result.result?.ok) {
```
machi1990, line 55: Is it possible to re-use the `update` method we have in `MongoDBDataProvider.ts`?

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -46,13 +54,18 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
       throw new NoDataError(`Cannot update ${this.collectionName} - missing ID field`)
     }
 
+    this.fieldTransforms
+      .filter((t: FieldTransform) => t.transformType === TransformType.UPDATE)
+      .forEach((f: FieldTransform) => {
+        data[f.fieldName] = f.transform(f.fieldName);
+      });
+
```
machi1990, line 62: We should probably cache these transforms instead of doing this at each update.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -30,11 +33,16 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
       delete data[idField.name];
     }
 
+    this.fieldTransforms
+      .filter((t: FieldTransform) => t.transformType === TransformType.CREATE)
```
machi1990, line 37: I think we should cache the `create` transformers too.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -68,8 +68,9 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     const schemaComposer = new SchemaComposer(schema)
 
     this.buildSchemaModelRelationships(schemaComposer, models);
-
+    
     this.buildSchemaForModels(schemaComposer, models);
+    this.addMetaFields(schemaComposer, models);
```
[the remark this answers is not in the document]
craicoverflow: Ah but this would mean these fields will automatically get added to the mutation input types, which we don't want.

I think it is fine link this for now, but in future schemaDefinitions should detect and filter out a predefined set of fields that should not be included in mutation inputs

https://github.com/aerogear/graphback/blob/c6f7fba82b107fd2c09075362e8fc7ee9582898d/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts#L99

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
```diff
@@ -213,4 +231,49 @@ describe('MongoDBDataProvider Basic CRUD', () => {
...
+    type Note {
+      id: ID!
+      text: String
+    }
+    `);
+    const cDate = new Date(2020, 5, 26, 18, 29, 23);
+    advanceTo(cDate);
+
+    const res = await context.providers.Note.create({ text: 'asdf' });
+    expect(res.createdAt).toEqual(cDate.getTime());
```
machi1990, line 250: We should test for updatedAt right? 

It is normal practice to have the `updatedAt` field set to `createdAt` field during creation.

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
```diff
@@ -213,4 +231,49 @@ describe('MongoDBDataProvider Basic CRUD', () => {
...
+    type Note {
+      id: ID!
+      text: String
+    }
+    `);
+    let cDate = new Date(2020, 5, 26, 18, 29, 23);
+    advanceTo(cDate);
+
+    const res = await context.providers.Note.create({ text: 'asdf' });
+    expect(res.updatedAt).toEqual(cDate.getTime());
```
machi1990, line 268: Probably worth to add the check that `createdAt` value has not changed

wtrocki:
Docs, Release notes and... This is mongo only right?

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
ssd71: Ah, yes, thanks!

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
ssd71: I think this functionality is tested in the test imediately after this

packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts
ssd71: I believe so, yes, thanks!

packages/graphback-runtime-mongodb/tests/MongoDataProviderTest.ts
machi1990: Yes, we have the `updateAt test` but what I am asking is, should we also check that the `updateAt` is set immediately after creating an entity? I'd see that test here, and the `note.updatedAt` should be equal to `node.createdAt`.

ssd71:
> Added a few minor comments.

Noted, thanks for reviewing :)

> One question though, does this work with PostgreSQL? Or it is the next step?

It does not, only implemented for Mongo as of yet

ssd71:
> I have done verification and it indeed works! Great job +1
> 
> Some comments to address first.

Noted! Thanks! Have resolved all, feel free to check :)

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -115,7 +115,12 @@ function traverse(filter: any): any {
                     filter[key] = [filter[key]];
                 }
             }
-
+            if (["createdAt", "updatedAt"].includes(key)) {
```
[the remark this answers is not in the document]
wtrocki: Having a hack embedded in function is bad. Let's simply move that to separate functions we can remove. We also need an issue to reference

wtrocki:
This is going on for long time. Please make final touches and request review if possible. We are still missing docs updates etc.

packages/graphback-runtime-mongodb/src/queryBuilder.ts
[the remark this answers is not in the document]
ssd71: Done that now

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -75,6 +75,24 @@ function isPrimitive(test: any): boolean {
 };
 
 
+function dangerousHack(filter: any, key: string): any {
```
machi1990, line 78: What is this hack doing? Maybe we can name it that way and have a comment about the hack and what's needed to get rid of the hack in the future - e.g a link to an issue / PR etc.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -281,6 +283,43 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     schemaComposer.Query.addFields(queryFields)
   }
 
+  protected addMetaFields(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
+    models.forEach((model: ModelDefinition, index: number)=> {
+      const name = model.graphqlType.name; 
+      const modelTC = schemaComposer.getOTC(name);
+      const desc = model.graphqlType.description;
+      if (parseMarker("versioned", desc) || parseMarker("delta", desc)) {
```
machi1990, line 291: should `versioned`, `delta`, `createdAt` etc be extracted somewhere e.g in a `data-sync.ts` file?

packages/graphback-runtime-mongodb/src/queryBuilder.ts
ssd71: Thanks! Please see #1349 I believe it covers this, plus I have added some comments in the same file:
https://github.com/aerogear/graphback/blob/4fd1c80b7f0126e366924ab8b7c1ddde7e8bcbad/packages/graphback-runtime-mongodb/src/queryBuilder.ts#L77-L84

machi1990:
> ... We'll need documentation and a line or two in release notes.

As for the docs, I am seeing something in line to what requested in this issue https://github.com/aerogear/graphback/issues/1323, it could be done in separate PR. Thoughts? @wtrocki 

/cc @ssd71 would you like to take a look at the issue?

packages/graphback-runtime-mongodb/src/queryBuilder.ts
machi1990: This is great. Nice work.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
ssd71: I am not sure if I understand, could you elaborate :)

machi1990:
> > One question though, does this work with PostgreSQL? Or it is the next step?
> 
> It does not, only implemented for Mongo as of yet

Created https://github.com/aerogear/graphback/issues/1355 for this.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: WDYT @ssd71 ?

ssd71:
> /cc @ssd71 would you like to take a look at the issue?

Sure! Once this stuff is merged, would love to try and spike!

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
ssd71: By extracting them do you mean parsing them and processing them separately? Just making sure we are on same page, do you mean that we should have single function `parseDataSync` that would be called in `addMetadataFields` ?

machi1990: Sorry, bad wording. I simply meant instead of having hard coded constants, we could refactor the code a bit by extracting the constants into meaningful variable e.g `const createdAt = "createdAt"` etc and use it throughtout the codebase, same for the other constants. Just a nitpick, nothing major. If that sounds okay with you?

machi1990:
> > /cc @ssd71 would you like to take a look at the issue?
> 
> Sure! Once this stuff is merged, would love to try and spike!

Awesome :-)

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
ssd71: Cool! Sounds great!

ssd71: How does this look:
https://github.com/aerogear/graphback/blob/da8e46fc7b61fe76f974df4b2eec9f0ef5441e94/packages/graphback-codegen-schema/src/metadataAnnotations.ts#L1-L11

wtrocki: delta will be not in the core. We will rename `offix` plugin to `datasync` and use delta there exclusively.

ssd71: Cool, so we remove it from here as well?
https://github.com/aerogear/graphback/blob/da8e46fc7b61fe76f974df4b2eec9f0ef5441e94/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L292

ssd71: https://github.com/aerogear/graphback/blob/3eb2e6591edf0738a470702dbb22aea116e13bef/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L292-L293
Now this only works for `@versioned`
WDYT @wtrocki

wtrocki:
@machi1990 Kinda having some other changes now. Do you have time to verify this changes?

packages/graphback-runtime/src/utils/fieldTransformHelpers.ts
```diff
@@ -0,0 +1,54 @@
...
+    UPDATE = "onUpdateFieldTransform",
+    CREATE = "onCreateFieldTransform"
+}
+
+export type FieldTransformMap = {
+    [TransformType.CREATE]: FieldTransform[],
+    [TransformType.UPDATE]: FieldTransform[]
+};
+
+export function getFieldTransformations(baseType: GraphQLObjectType): FieldTransformMap {
```
machi1990, line 22: This function should probably handle `@versioned` marker too.

wtrocki: I think this works differently. Versioned is just used at generation phase and it is adding 2 new annotations to target schema. If we were about to add versioned here we could remove other markers like createdAt etc.

ssd71: Would love to help, are you using schema plugin? We are handling `@versioned` in the schema plugin:
https://github.com/aerogear/graphback/blob/06ea54ebd4650586a6d20830b467fdf687ddd676/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L287-L322

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -15,11 +15,13 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
   protected db: Db;
   protected collectionName: string;
   protected tableMap: ModelTableMap;
+  protected fieldTransformMap: FieldTransformMap;
 
   public constructor(baseType: GraphQLObjectType, db: any) {
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
+    this.fieldTransformMap = getFieldTransformations(baseType);
```
machi1990, line 24: Because of how the runtime is setup:

https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/src/graphql.ts#L29-L34

We might actually miss the `createdAt` and `updatedAt` fields which are automatically added when we encounter `@versioned` marker, since this (transformer retrival) happens before the `SchemaPlugin`, specifically this line https://github.com/aerogear/graphback/pull/1301/files#diff-bb76265ef553643b897bd0eb1ab72edfR292 is called.

One of the problem we could be having is actually missing the `createdAt` and `updatedAt` fields when an entity is created or updated since we do not have the appropriate transformers. 

See https://github.com/aerogear/OpenVolunteerPlatform/pull/93#discussion_r434476757 for details. 

/cc @ssd71 @wtrocki @craicoverflow

packages/graphback-runtime/src/utils/fieldTransformHelpers.ts
machi1990: Yes, OVP is using the schema plugin https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/.graphqlrc.yml#L17

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
wtrocki: Generally, this PR went to the very complex form of us utilizing 3 different markers etc. I would personally prefer only a single marker (versioned). It is hard to understand the implications and transformation phase it redundant - we can actually follow @machi1990 original comment and use only versioned annotation everywhere. 
Especially now when we do not have physical schema file so all transformations that will be added will not be visible to users unless they inspect schema in playground.

packages/graphback-runtime/src/utils/fieldTransformHelpers.ts
machi1990: > I think this works differently. Versioned is just used at generation phase and it is adding 2 new annotations to target schema. If we were about to add versioned here we could remove other markers like createdAt etc.

Yes, I expected to have the `createdAt` and `updatedAt` fields to be generated on the fly (and persisted, which is the current bug I am seeing), with the usage of `@versioned`, at least from this is what I see from the code shared in https://github.com/aerogear/graphback/pull/1301#discussion_r434479544

machi1990: I wanted to bring the behaviour tested here https://github.com/aerogear/graphback/pull/1301/files#diff-5e6f4bf330b4e97ce59c1869385ee32bR235

PR #1315, opened 2020-05-27
fix: Initial version for generic scalar building

wtrocki:
Inititial version

### DIFF
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -2,10 +2,10 @@
 import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
 import { getFieldName, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap } from '@graphback/core'
-import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt } from 'graphql';
+import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt, GraphQLFloat, GraphQLScalarType, isScalarType, isSpecifiedScalarType } from 'graphql';
 import { SchemaComposer } from 'graphql-compose';
 import { gqlSchemaFormatter, jsSchemaFormatter, tsSchemaFormatter } from './writer/schemaFormatters';
-import { buildFilterInputType, createModelListResultType, StringScalarInputType, IntScalarInputType, FloatScalarInputType, BooleanScalarInputType, SortDirectionEnum, buildCreateMutationInputType, buildFindOneFieldMap, buildMutationInputType, OrderByInputType, buildSubscriptionFilterType, IDScalarInputType, PageRequest } from './definitions/schemaDefinitions';
+import { buildFilterInputType, createModelListResultType, StringScalarInputType, BooleanScalarInputType, SortDirectionEnum, buildCreateMutationInputType, buildFindOneFieldMap, buildMutationInputType, OrderByInputType, buildSubscriptionFilterType, IDScalarInputType, PageRequest, createInputTypeForCustomScalar } from './definitions/schemaDefinitions';
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -57,6 +57,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
 
   public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
     const schema = metadata.getSchema();
+
     const models = metadata.getModelDefinitions();
     if (models.length === 0) {
       this.logWarning("Provided schema has no models. Returning original schema without any changes.")
@@ -68,7 +69,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
 
     this.buildSchemaModelRelationships(schemaComposer, models);
 
-    this.buildSchemaForModels(schemaComposer, models);
+    this.buildSchemaForModels(schemaComposer, models, schema);
 
     return schemaComposer.buildSchema()
   }
@@ -109,8 +110,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     return SCHEMA_CRUD_PLUGIN_NAME;
   }
 
-  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
-    this.createStandardTypes(schemaComposer)
+  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[], schema: GraphQLSchema) {
+    this.createStandardInputTypes(schemaComposer, schema);
 
     for (const model of Object.values(models)) {
       this.createQueries(model, schemaComposer);
@@ -125,7 +126,6 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       modifiedType.addFields(modelRelationshipFilterFields)
     }
   }
-
 
   protected createSubscriptions(model: ModelDefinition, schemaComposer: SchemaComposer<any>) {
     const name = model.graphqlType.name
@@ -281,14 +281,21 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     schemaComposer.Query.addFields(queryFields)
   }
 
-  protected createStandardTypes(schemaComposer: SchemaComposer<any>) {
-    schemaComposer.add(IDScalarInputType)
-    schemaComposer.add(StringScalarInputType)
-    schemaComposer.add(IntScalarInputType)
-    schemaComposer.add(FloatScalarInputType)
-    schemaComposer.add(BooleanScalarInputType)
-    schemaComposer.add(SortDirectionEnum)
-    schemaComposer.add(PageRequest)
+  private createStandardInputTypes(schemaComposer: SchemaComposer<any>, schema: GraphQLSchema) {
+    schemaComposer.add(PageRequest);
+    schemaComposer.add(IDScalarInputType);
+    schemaComposer.add(SortDirectionEnum);
+    schemaComposer.add(StringScalarInputType);
+    schemaComposer.add(BooleanScalarInputType);
+    schemaComposer.add(createInputTypeForCustomScalar(GraphQLInt));
+    schemaComposer.add(createInputTypeForCustomScalar(GraphQLFloat));
+    
+    const schemaTypes = Object.values(schema.getTypeMap());
+    for (const schemaType of schemaTypes) {
+      if (isScalarType(schemaType) && !isSpecifiedScalarType(schemaType)) {
+        schemaComposer.add(createInputTypeForCustomScalar(schemaType));
+      }
+    }
   }
 
   /**
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -3,9 +3,9 @@ import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
 import { getFieldName, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap } from '@graphback/core'
 import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt, GraphQLFloat, GraphQLScalarType, isScalarType, isSpecifiedScalarType } from 'graphql';
-import { SchemaComposer } from 'graphql-compose';
+import { SchemaComposer, NamedTypeComposer } from 'graphql-compose';
 import { gqlSchemaFormatter, jsSchemaFormatter, tsSchemaFormatter } from './writer/schemaFormatters';
-import { buildFilterInputType, createModelListResultType, StringScalarInputType, BooleanScalarInputType, SortDirectionEnum, buildCreateMutationInputType, buildFindOneFieldMap, buildMutationInputType, OrderByInputType, buildSubscriptionFilterType, IDScalarInputType, PageRequest, createInputTypeForCustomScalar } from './definitions/schemaDefinitions';
+import { buildFilterInputType, createModelListResultType, StringScalarInputType, BooleanScalarInputType, SortDirectionEnum, buildCreateMutationInputType, buildFindOneFieldMap, buildMutationInputType, OrderByInputType, buildSubscriptionFilterType, IDScalarInputType, PageRequest, createInputTypeForScalar } from './definitions/schemaDefinitions';
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -69,7 +69,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
 
     this.buildSchemaModelRelationships(schemaComposer, models);
 
-    this.buildSchemaForModels(schemaComposer, models, schema);
+    this.buildSchemaForModels(schemaComposer, models);
 
     return schemaComposer.buildSchema()
   }
@@ -110,8 +110,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     return SCHEMA_CRUD_PLUGIN_NAME;
   }
 
-  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[], schema: GraphQLSchema) {
-    this.createStandardInputTypes(schemaComposer, schema);
+  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
+    this.createSchemaCRUDTypes(schemaComposer);
 
     for (const model of Object.values(models)) {
       this.createQueries(model, schemaComposer);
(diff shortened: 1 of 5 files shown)


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -281,14 +281,21 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+  private createStandardInputTypes(schemaComposer: SchemaComposer<any>, schema: GraphQLSchema) {
+    schemaComposer.add(PageRequest);
+    schemaComposer.add(IDScalarInputType);
+    schemaComposer.add(SortDirectionEnum);
+    schemaComposer.add(StringScalarInputType);
+    schemaComposer.add(BooleanScalarInputType);
+    schemaComposer.add(createInputTypeForCustomScalar(GraphQLInt));
+    schemaComposer.add(createInputTypeForCustomScalar(GraphQLFloat));
+    
+    const schemaTypes = Object.values(schema.getTypeMap());
```
wtrocki, line 293: Kinda liked previous helper as it could become an core thing :D

packages/graphback-codegen-schema/tests/mock.graphql
```diff
@@ -48,10 +48,13 @@ See issue https://github.com/aerogear/graphback/issues/1299
 type  Issue {
   id: ID!
   category: IssueCategory
+  createdAt: DateTime
```
wtrocki, line 51: CC @ssd71 We going to enable scalar support in this PR so your hacks might be swapped later. However we will still need to supply set of the `graphback supported` scalars that we will require for both core and datasync in separate PR so please bear with us until we get this sorted.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: I thought I could remove it for now since it meant iterating twice on.

machi1990:
Seeing a failure in CI https://app.circleci.com/pipelines/github/aerogear/graphback/4310/workflows/6e15ccf8-4604-4797-b996-c41de4737635/jobs/4387 that I am not getting when I ran integrations test locally:

```log
yarn test:integration
yarn run v1.22.4
$ cd integration && yarn && yarn integration
[1/4] 🔍  Resolving packages...
success Already up-to-date.
$ jest --runInBand
ts-jest[versions] (WARN) Version 26.0.1 of jest installed has not been tested with ts-jest. If you're experiencing issues, consider using a supported version (>=25.0.0 <26.0.0). Please do not report issues in ts-jest if you are using unsupported versions.
  console.log
    querying object Note with filter undefined

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    querying object Note with filter undefined

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    querying object Note with filter undefined

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    querying object Comment with filter undefined

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object CommentMetadata with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object Note with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object CommentMetadata with args {"id":2}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object Note with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object Note with args {"id":"1"}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    querying object Comment with filter {"noteId":{"eq":"1"}}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object CommentMetadata with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object Note with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    querying object Comment with filter {"noteId":{"eq":"1"}}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object CommentMetadata with args {"id":2}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    fetching single object Note with args {"id":1}

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    Updating object Note

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    Creating object Note

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

  console.log
    deleting object Note

      at Object.log (../packages/graphback-runtime/src/utils/Logger.ts:21:17)

 PASS  tests/runtime-workflow.ts (9.888 s)
  ✓ Find all notes (65 ms)
  ✓ Find all notes except the first (20 ms)
  ✓ Find at most one note (22 ms)
  ✓ Find all comments (89 ms)
  ✓ Note 1 should be defined (14 ms)
  ✓ Find at most one comment on Note 1 (28 ms)
  ✓ Find comments on Note 1 except first (41 ms)
  ✓ Should update Note 1 title (33 ms)
  ✓ Should create a new Note (21 ms)
  ✓ Delete Note 1 (25 ms)

Test Suites: 1 passed, 1 total
Tests:       10 passed, 10 total
Snapshots:   1 passed, 1 total
Time:        10.854 s
Ran all test suites.
```

Have we seen it before?

wtrocki:
@machi1990 Time for Pong?

wtrocki:
We have changed names of the InputTypes which is going to affect client side queries. I think it will be better to revert that and introduce this change separately after @Eunovo will finish his refactor for client side.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -1,45 +1,55 @@
...
 
 const PageRequestTypeName = 'PageRequest';
 const SortDirectionEnumName = 'SortDirectionEnum';
 const OrderByInputTypeName = 'OrderByInput';
 
 const getScalarInputName = (type: GraphQLNamedType) => {
-  if (isSpecifiedScalarType(type)) {
-    return `${type.name}Input`;
+  if (isEnumType(type)) {
+    return `StringFilterInput`
```
wtrocki, line 10: I think we mighth revert this.

wtrocki:
As for failure. Make sure that you have all packages build locally when executing the test.

machi1990:
> @machi1990 Time for Pong?
> 
> EDIT:
> 
> Let's not rebase this branch as it will make it hard to collaborate

Yeap, you can take over or I'll continue tomorrow. As you prefer :-)

wtrocki:
This looks like it is going to work. So it needs little bit of polish and we can get that over with client side changes and other PR's into next dev release

machi1990:
> As for failure. Make sure that you have all packages build locally when executing the test.

Yeap, I did ran `yarn build` before the `yarn test:integration` cmd. 

Would need further investigation but hard to do so if a CI error is not reproducible locally. Let me know if you have the error too.

wtrocki:
It fails locally. Sounds like yarn.lock problem

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: You can getTypeMap on schemaComposer, so you don't need to pass schemaComposer and schema.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -109,8 +110,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     return SCHEMA_CRUD_PLUGIN_NAME;
   }
 
-  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
-    this.createStandardTypes(schemaComposer)
+  protected buildSchemaForModels(schemaComposer: SchemaComposer<any>, models: ModelDefinition[], schema: GraphQLSchema) {
+    this.createStandardInputTypes(schemaComposer, schema);
```
craicoverflow, line 114: This method can probably be renamed, since we are also generating dynamic input types based on unknown scalars.

machi1990: Good idea. Do you have any naming suggestion for inspiration ? ;-)

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -281,14 +281,21 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
-    schemaComposer.add(BooleanScalarInputType)
-    schemaComposer.add(SortDirectionEnum)
-    schemaComposer.add(PageRequest)
+  private createStandardInputTypes(schemaComposer: SchemaComposer<any>, schema: GraphQLSchema) {
+    schemaComposer.add(PageRequest);
+    schemaComposer.add(IDScalarInputType);
+    schemaComposer.add(SortDirectionEnum);
+    schemaComposer.add(StringScalarInputType);
+    schemaComposer.add(BooleanScalarInputType);
+    schemaComposer.add(createInputTypeForCustomScalar(GraphQLInt));
```
craicoverflow, line 290: Are we using `createInputTypeForCustomScalar` for Float and Int because the values are the same? If so, maybe `createInputTypeForCustomScalar` is a misleading name.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -1,45 +1,55 @@
...
-    le: { type: GraphQLFloat },
-    lt: { type: GraphQLFloat },
-    ge: { type: GraphQLFloat },
-    gt: { type: GraphQLFloat },
-    in: { type: GraphQLList(GraphQLFloat) },
-    between: { type: GraphQLList(GraphQLFloat) }
-  }
-})
+export const createInputTypeForEnum = (scalarType: GraphQLScalarType) => {
+  const newInput = new GraphQLInputObjectType({
```
craicoverflow, line 37: The input object fields looks to be the same as that returned in `createInputTypeForCustomScalar`

craicoverflow:
> As for failure. Make sure that you have all packages build locally when executing the test.

And run `yarn clean` first.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Hmm. `createSchemaCRUDTypes`? Since it creates the various types for CRUD like pagination, input types, enums and so on.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: I am not sure if this is even used?  I think we should treat enumInputs like a StringInput, WDYT? @wtrocki

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: Yeap, that's the initial idea we had with @wtrocki but could not find a simple way to get them from schemaComposer. Suggestions welcomed.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: How about `createInputTypeForScalar`

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: I tried this out and it seems to work:

```ts
private createStandardInputTypes(schemaComposer: SchemaComposer<any>, schema: GraphQLSchema) {
    schemaComposer.add(PageRequest);
    schemaComposer.add(IDScalarInputType);
    schemaComposer.add(SortDirectionEnum);
    schemaComposer.add(StringScalarInputType);
    schemaComposer.add(BooleanScalarInputType);
    schemaComposer.add(createInputTypeForCustomScalar(GraphQLInt));
    schemaComposer.add(createInputTypeForCustomScalar(GraphQLFloat));

    schemaComposer.forEach((tc: NamedTypeComposer<any>) => {
      const namedType = tc.getType()
      if (isScalarType(namedType) && !isSpecifiedScalarType(namedType)) {
        schemaComposer.add(createInputTypeForCustomScalar(namedType));
      }
    })
}
```

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
wtrocki: Yep. I pushed that by mistake. We should use strings

machi1990: Okay, will remove this.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: Okay, let's use that version and use schemaComposer as single source of truth.

craicoverflow:
If we are going to support any scalar without knowledge of the scalar, graphql-migrations should also map the scalar to a string by default, because right now graphql-migrations fails.

```
🚀  Server ready at http://localhost:4000/graphql
index.ts:28
Unsupported type Booooolean on field Note.test.
generateAbstractDatabase.ts:223
Migrated database
```

machi1990:
> If we are going to support any scalar without knowledge of the scalar, graphql-migrations should also map the scalar to a string by default, because right now graphql-migrations fails.
> 
> ```
> 🚀  Server ready at http://localhost:4000/graphql
> index.ts:28
> Unsupported type Booooolean on field Note.test.
> generateAbstractDatabase.ts:223
> Migrated database
> ```

Where are you seeing this error?

craicoverflow:
> Where are you seeing this error?

If you add a custom scalar to your data model
Run runtime example
It should appear in the output log of the template

machi1990:
> > Where are you seeing this error?
> 
> If you add a custom scalar to your data model
> Run runtime example
> It should appear in the output log of the template
> It is caused by graphql-migrations

We think we'd have the same issue in current master. I think this is the manifestation of what https://github.com/aerogear/graphback/issues/1095 is about - supporting custom scalars out of the box, which this PR does not address. What this PR is trying to address is just create the appropriate FilterType for custom scalars with the assumption that the user will have the necessary bits to handle the scalar type in question. For the OVP side, we have this [graphql-scalars](https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/package.json#L35) doing the work for us. So the idea is just to have the appropriate input filter type to be make the filter work.

craicoverflow:
> > > Where are you seeing this error?
> > 
> > 
> > If you add a custom scalar to your data model
> > Run runtime example
> > It should appear in the output log of the template
> > It is caused by graphql-migrations
> 
> We think we'd have the same issue in current master. I think this is the manifestation of what #1095 is about - supporting custom scalars out of the box, which this PR does not address. What this PR is trying to address is just create the appropriate FilterType for custom scalars with the assumption that the user will have the necessary bits to handle the scalar type in question. For the OVP side, we have this [graphql-scalars](https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/package.json#L35) [](/aerogear/OpenVolunteerPlatform/blob/HEAD@{2020-05-28T10:31:00Z}/platform/server/package.json#L35) doing the work for us. So the idea is just to have the appropriate input filter type to be make the filter work.

Yep, good points!

wtrocki:
Also Adding Date scalar to datasync-starter - dates are very common and almost always needed.

wtrocki:
Rebase needed.

machi1990:
Ci failure is not because of this issue, looks like a global one - 

> Rebase needed.

Done.

craicoverflow:
The dynamic scalar input type has fields that would work great when filtering, but is missing ones that would be useful for non-numbers - mainly String.

Should it have more fields? startsWith, contains, endsWith..

```gql
input TestInput {
  ne: Test
  eq: Test
  le: Test
  lt: Test
  ge: Test
  gt: Test
  in: [Test]
  between: [Test]
}
```

wtrocki:
Yes. That is tradeoff here. I see it this way. We will have this unbound, but we will need to bring multiple scalars into graphback core at some point. 

That is why we put some basics things now. 
startsWith for Date makes no sense etc.

machi1990:
> Yes. That is tradeoff here. I see it this way. We will have this unbound, but we will need to bring multiple scalars into graphback core at some point.
> 
> That is why we put some basics things now.
> startsWith for Date makes no sense etc.

+1

craicoverflow:
If someone has 10 custom scalars, we generate 10 identical filter inputs for them. Why not just use 1 generic `ScalarInput` for this?

wtrocki:
There is no generics in GraphQL. 
See: https://github.com/graphql/graphql-spec/issues/697

craicoverflow:
We do not have support for generics in GraphQL
See: `https://github.com/graphql/graphql-spec/issues/697`

I feel like this should work just fine:

```gql
type Game {
  highScore: BigInt
  name: Text
}

type GameFilter {
  highScore: ScalarFilterInput
  name:      ScalarFilterInput
}
```

machi1990:
> If someone has 10 custom scalars, we generate 10 identical filter inputs for them. Why not just use 1 generic `ScalarInput` for this?

I feel like we prior to this patch we had the `StringInput` as the default filter which worked great where a field is actually stored as string in db e.g an `enum` type, and it was not working in some cases e.g Date. 

> 
> ```graphql
> type Game {
>   highScore: BigInt
>   name: Text
> }
> 
> type GameFilter {
>   highScore: ScalarFilterInput
>   name:      ScalarFilterInput
> }
> ```

I am curious to see how would the `ScalarFilterInput` look like.

wtrocki:
See my comment above - this is not possible without reinventing Scalars on the server or generic scalars ( see issue )

```graphql

Input ScallarTypeInput {
  field: GenericScalar
}
```
Then when using this input we will need to supply type info everytine which is uber bad

wtrocki:
Verification wise -  we can link this to OVP or release new Dev version

craicoverflow:
> I am curious to see how would the `ScalarFilterInput` look like.

It would look like this:

```gql
input ScalarFilterInput {
  ne: Test
  eq: Test
  le: Test
  lt: Test
  ge: Test
  gt: Test
  in: [Test]
  between: [Test]
}
```

What we are doing right now is generating a new input for each custom scalar, but the only thing that differs between them is the input name and under the hood they would be used the exact same way.

PR #1321, opened 2020-05-28
test: fix failing runtime-workflow integration tests

machi1990:
Fix the issue by making sure that there is no shared context between each test.

Previous after the first test is executed, subsequent tests were failing with the following message:
```
TypeError: Cannot redefine property: Symbol(apolloServerExecutionDispatcherWillResolveField)

        at Function.defineProperty (<anonymous>)

      at Object.<anonymous> (../node_modules/apollo-server-core/src/requestPipeline.ts:431:14)

      at fulfilled (../node_modules/apollo-server-core/dist/requestPipeline.js:5:58)
 ```
While looking at the code, it seemed to me like the context we passed to apollo server was a global one and hence between multiple tests which resulted in a failure as we attempted to redifine a property that already was defined in the previous ran.

### DIFF
--- a/integration/tests/runtime-workflow.ts
+++ b/integration/tests/runtime-workflow.ts
@@ -11,13 +11,18 @@ import { loadDocuments } from '@graphql-toolkit/core';
 import { GraphQLFileLoader } from '@graphql-toolkit/graphql-file-loader';
 import * as Knex from 'knex';
 import { GraphbackRuntime, GraphbackGenerator } from "graphback/src";
-import { buildSchema, printSchema } from 'graphql';
+import { buildSchema, printSchema, GraphQLSchema } from 'graphql';
 import { migrateDB } from '../../packages/graphql-migrations/src';
 import { createKnexPGCRUDRuntimeServices } from "../../packages/graphback-runtime-knex/src"
 
 /** global config */
-let client: ApolloServerTestClient;
 let db: Knex;
+let server: ApolloServer; 
+let modelSchema: GraphQLSchema;
+let client: ApolloServerTestClient;
+let runtimeEngine: GraphbackRuntime;
+
+const documentsCache = {};
 
 beforeAll(async () => {
   try {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by machi1990** [comment]
Opening as draft to see what CI will say and we can chat about the approach.

File: integration/tests/runtime-workflow.ts [hunk 1]
```diff
@@ -45,28 +50,31 @@ beforeAll(async () => {
...
     console.log(e);
     throw e;
   }
 })
 
+beforeEach(() => {
+  const services = createKnexPGCRUDRuntimeServices(runtimeEngine.getDataSourceModels(), modelSchema, db, new PubSub());
+  const runtime = runtimeEngine.buildRuntime(services);
+  server = new ApolloServer({
+    typeDefs: printSchema(runtime.schema),
```
craicoverflow, line 68: `printSchema` actually removes directives from the schema, so if our tests have them in future this won't validate them.

You could replace this with

```suggestion
schema: runtime.schema,
```

I haven't tested the implications of this, but it should work. If we still need to use typeDefs you can use this `printSchemaWithDirectives`:
 https://github.com/aerogear/graphback/blob/2c72ddb03058a6b5c478fe28067325aebd172d5e/packages/graphback-core/src/utils/printSchemaWithDirectives.ts#L4

integration/tests/runtime-workflow.ts
```diff
@@ -130,13 +138,18 @@ const getConfig = async () => {
...
+  }
+
+  const loadedDocuments = await loadDocuments(path.resolve(`./output/client/**/${name}.graphql`), {
     loaders: [
       new GraphQLFileLoader()
     ]
   });
 
-  return documents[0];
+  documentsCache[name] = loadedDocuments[0];
```
craicoverflow, line 151: :+1:

wtrocki:
Amazing work @machi1990 +1 from me!

integration/tests/runtime-workflow.ts
machi1990: I did not change the way schema are passed around, but it is a solid suggestion which I have applied, let's see what CI will say :-)

Issue #1296, opened 2020-05-22
An error is thrown if `plugins` is not defined in .graphqlrc.yml file


machi1990:
With `0.14.0-dev2`, when plugins key is not defined in the configuration file, I am seeing an error 
```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
```
I believe it is coming from here https://github.com/aerogear/graphback/blob/8fa8dcdebab081313fa9daf2fa21ea52d3311efe/packages/graphback/src/loadPlugins.ts#L6 since the `pluginConfig` is `undefined`.

craicoverflow:
https://github.com/aerogear/graphback/issues/1068

Yeah I have seen this before. To be completely honest I don't remember the exact reason that I closed it.

Why do you not have any plugin define? Isn't SchemaCRUDPlugin needed?

machi1990:
> #1068
> 
> Yeah I have seen this before. To be completely honest I don't remember the exact reason that I closed it.

Oh the issue is closed that's why it did not show up when I did a quick search.

So for me was when I did not have plugins setup that I was seeing this error. 
 
> Why do you not have any plugin define? Isn't SchemaCRUDPlugin needed?

I thought with stuffs being done at runtime, I could try to remove some plugins `graphback-schema` being one of them hence the error. Probably a user mistake and if so, we should at least have a better error message.

craicoverflow:
Runtime still uses plugins to transform schema:

https://github.com/aerogear/graphback/blob/83def270c729aa2e003996961640439742280495/packages/graphback/src/GraphbackRuntime.ts#L22

> Probably a user mistake and if so, we should at least have a better error message.

Absolutely, we are going to move to a code-based plugin approach anyway (see https://github.com/aerogear/graphback/issues/1283) so this can be resolved there :)

wtrocki:
So templates in graphback have examples. We still need plugins there. We just do not need to use cli to generate code anymore (unless client side queries)

machi1990:
Just ran into this issue.

craicoverflow:
Where are you seeing this?

I can see that the only places `GraphbackGenerator` is being used is in integration tests and the `generate` command.

I think client generation will be moved to Offix at some point, then we can remove `GraphbackGenerator` and this problem should go away.

But since it is probably a quick fix, if you are in that area feel free to action it :D

wtrocki:
Yep. Same here. Quick fix needed.

[...]

Issue #1283, opened 2020-05-20
Simplify schema plugin API


craicoverflow:
The barrier to custom plugin creation is too high and it is too complex.

1. Create a plugin module and publish it to NPMJS
2. Instal plugin as a dependency
2. Reference the plugin name in the config file. This would not work for namespaced plugins like `@myorg/graphback-plugin`
3. The plugin engine loads the plugin dynamically from the name.

Why not something like this?

```ts
const myPlugin: GraphbackPlugin = {
  transformSchema(...) { ... }
}

const plugins: GraphbackPlugin[] = [
  myPlugin
]

const { resolvers, typeDefs } = graphback(schema, {
 ...
 plugins: plugins
})
```

wtrocki:
Worht to add - 
Plugins can move from declarative to code-based (as runtime). The only need for declarative plugins is an client side generation now.

Issue #1157, opened 2020-04-21
Runtime API should have option to return `typeDefs`


craicoverflow:
I have a usecase where I want to use `typeDefs` instead of a GraphQLSchema object in my GraphQL app.

The API returns `schema` and `resolvers`:

```ts
const { typeDefs, resolvers } = createRuntime();
```

If I want to use `makeExecutableSchema` from apollo-server I need `typeDefs`:

```ts
const { typeDefs, resolvers } = createRuntime();

const schema = makeExecutableSchema({
  typeDefs,
  resolvers
})
```

Given that the schema is completely in-memory it is important to offer this flexibility for various applications.

wtrocki:
Even better. I do not see any use case where returning schema makes sense as it is not executable schema (it is missing directives and resolvers anyway)

craicoverflow:
`createKnexPGCRUDRuntimeServices`

wtrocki:
`although currently in runtime this is not attached to context it could be.`

What you mean by this?

craicoverflow:
A user may want to use the services created by `createKnexPGCRUDRuntimeServices` in their own custom resolvers. For this they would want to attach the services to the resolver context so they would require a schema object to do this.

wtrocki:
I think we have incorporated that. Reassigned that to yourself and closing but feel free to reopen if needed.

craicoverflow:
Im going to keep this open until I get to it, I don't believe this is implemented.

Issue #1072, opened 2020-04-08
Schema and Resolver plugins should not require `outputPath` and `format` for runtime API


craicoverflow:
It is confusing to have to specify an `outputPath` and `format` to the plugin configuration when using the runtime API.

An error is thrown in the constructor of these plugins. If we move the check into the plugin `createResources` plugin functions then we can throw an error here instead, as runtime does not execute this function. This simplifies configuration.

PR #1324, opened 2020-05-28
New runtime api

craicoverflow:
## Motivation

Closes:

- https://github.com/aerogear/graphback/issues/1157
- https://github.com/aerogear/graphback/issues/1071
- https://github.com/aerogear/graphback/issues/1283
- https://github.com/aerogear/graphback/issues/1296
- https://github.com/aerogear/graphback/issues/1027
- https://github.com/aerogear/graphback/issues/1072
- https://github.com/aerogear/graphback/issues/1064

This PR changes significantly alters the runtime API to reduce overall complexity of configuring and using Graphback.

## Implementation

Essentially we have cut out a multi-step approach to configuring a Graphback server and made it into a single function API to create the services, resolvers, datasources and schema. Here is how simple it has become:

```ts
const { typeDefs, resolvers, services } = buildGraphbackAPI(modelDefs, {
  dataProviderCreator: createKnexDbProvider(dbConfig),
  plugins: [
    new ClientCRUDPlugin({
      format: 'graphql',
      outputFile: path.resolve('./client/src/graphql/graphback.graphql')
    })
  ]
});

  typeDefs,
  resolvers,
  context: (context) => ({
    ...context,
    services
  })
})
```

This approach uses the Factory pattern to handle creation of the service and data sources after the final schema is created.

For now this is done for PostgreSQL only, and has been updated in `ts-apollo-postgres-backend`, but once agreed on Postgres approach we can do the others.

## Tasks

- [x] Implement new API for Postgres 
- [x] Implement new API for Mongo
- [x] Update graphql-serve to use new API
- [x] Add tests
- [x] Documentation

## Verification

1. Run `ts-apollo-postgres-backend` template and test API
2. Change some config values in the API to see that they work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,8 @@ jobs:
     docker:
       # Node 8 LTS
       - image: circleci/node:lts-browsers
+        environment:
+          NODE_OPTIONS: --max_old_space_size=4096
       - image: circleci/postgres:latest
         environment:
           POSTGRES_PASSWORD: postgres


packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/tslint/config
+class ServiceBuilder implements GraphbackServiceBuilder {
+  private datasourceConfig: ModelDataSourceConfig;
+  private options: ServiceBuilderConfig;
+  public constructor(builderConfig: ServiceBuilderConfig) {
+    this.options = {
+      ...builderConfig,
+      pubSub: new PubSub()
```
craicoverflow, line 31: Should come first

packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const service = new CRUDService(this.options.pubSub)
+    service.setDataProvider(this.options.defaultDataProvider)
+
+    this.datasourceConfig = {
+      default: {
```
craicoverflow, line 38: Sets a default data provider that will be used for each model, unless it is overridden.

packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+        service
+      }
+    };
+  }
+
+  public setDatasource(modelName: string, provider: GraphbackDataProvider): void {
```
craicoverflow, line 44: This allows the user to change the dataprovider for a model from the template.

packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    this.datasourceConfig[modelName] = {
+      service: this.datasourceConfig.default.service
+    }
+
+    this.datasourceConfig[modelName].service.setDataProvider(provider)
+  }
+
+  public setService(modelName: string, service: GraphbackCRUDService): void {
```
craicoverflow, line 59: Allows changing the CRUD service for a model

packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      return;
+    }
+
+    this.datasourceConfig[modelName] = {
+      service
+    }
+  }
+
+  public build(models: ModelDefinition[]): { [modelName: string]: GraphbackCRUDService } {
```
craicoverflow, line 72: For each model definition, set a CRUD service. and return

packages/graphback/src/buildServices.ts
```diff
@@ -0,0 +1,108 @@
...
+      const modelDsConfig = this.datasourceConfig[model.graphqlType.name];
+      const pubSubConfig = this.getModelPubSubConfig(model)
+
+      if (modelDsConfig) {
+        modelServiceConfig[model.graphqlType.name] = modelDsConfig.service;
+      } else {
+        modelServiceConfig[model.graphqlType.name] = this.datasourceConfig.default.service
+      }
+
+      modelServiceConfig[model.graphqlType.name].setModelConfig(model.graphqlType, pubSubConfig)
```
craicoverflow, line 85: Setting model config and data provider outside of the constructors because it is way too complicated for a user to set their own CRUD service if they have to pass the GraphQLObject type in the constructor.

wtrocki:
Nice! 🔥
Checking this now

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -16,16 +16,17 @@ import { buildQuery } from './knexQueryMapper';
...
   protected baseType: GraphQLObjectType;
   protected tableName: string;
   protected tableMap: ModelTableMap;
 
-  public constructor(baseType: GraphQLObjectType, db: Knex) {
-    this.db = db;
-    this.baseType = baseType;
-    //TODO build and use mapping here
+  public constructor(dbConfig: Knex.Config) {
+    this.db = Knex(dbConfig);
```
craicoverflow, line 25: Trimmed this down to only require the database config. If we think about the user adding their own custom default crud service, it would not work because the constructor values would be different for every model.

See API:
```
const services = buildGraphbackServices({
  defaultDataProvider: new PgKnexDBDataProvider(dbConfig),
  pubSub
})
```

If the model was required here, this simply would not work.

packages/graphback-runtime/src/service/GraphbackCRUDService.ts
```diff
@@ -18,83 +21,86 @@ export interface ResultList<T = any> {
  */
 //tslint:disable-next-line: no-any
 export interface GraphbackCRUDService<Type = any, GraphbackContext = any> {
+  setModelConfig(modelType: GraphQLObjectType, pubSubConfig: GraphbackPubSubModel): void
```
craicoverflow, line 24: Not entirely sure if it is a good idea to have these setters in the CRUD service interface.

wtrocki:
```
const services = buildGraphbackServices({
  defaultDataProvider: new PgKnexDBDataProvider(dbConfig),
  pubSub
})
```

`pubSub` is specific to the service. Would we be able to handle more arguments in the future?

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
wtrocki: Yes. This makes PgKnexDBDataProvider stateless so service itself will hold some state for types. I really like how you figured that out. 

What service this is going to use? 
For example for upstream you are using CRUDService, For datasync it will use DataSyncService (as compositve of Auth, Metrics and CRUD service). 

Are we simplifying things only for the upstream and going to have some hacks for the other use cases? (Which is fine tradeoff)

packages/graphback-runtime-knex/src/createKnexServices.ts
```diff
@@ -6,38 +6,18 @@ import { PgKnexDBDataProvider } from './PgKnexDBDataProvider';
...
-    const objectDB = new PgKnexDBDataProvider(modelType, db)
-
-    services[model.name] = new CRUDService(modelType, objectDB, {
-      pubSub,
-      ...model.pubSub
-    })
-
-    return services;
-  }, {})
+  throw new Error("Deprecated")
```
wtrocki, line 20: Like that we actually giving this backwards compatibility. We can even return link to docs :D

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -21,6 +21,10 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
   }
+  
+  public setBaseType(baseType: GraphQLObjectType<any, any, { [key: string]: any; }>): void {
```
wtrocki, line 25: I do not think that setter with work in this scenario. 

What graphback is now is just list of the services where each service contains other service (thru inheritance or composition) and then contains data provider.

wtrocki:
What graphback is now is just a list of the services where each service contains other services (through inheritance or composition) and then contains data providers. 

This structure needs to be created somehow for a specific number of objects in the schema. 

```
user -> AuthService -> CRUDService -> DataProvider 
```

Each of those can be mixed together in any way and that makes the API more complex.

Developers need to manually create those per each type if they want this to be custom and compose them manually with the database provider. 

Going back to the reference approach requirements for the API could be:

1) Being able to create runtime by a single line of code that works with database x and default Service. 

2) Being able to create various levels of services and data layers where services can accept data layers and even use different layers for specific crud methods.

That is mostly for DataSync starter 

2.1) Provide alternative API for DataSync starter.

## Approaches 

I think we can do something very simple hereby a focus on one database and service as a helper (means single method per database type) that will use default service.

What was initially build by @darahayes was actually very similar to what SQLMancer has? So you can actually provide a reference to the implementation that will be used. 

### Creator approach

The creator approach requires to decouple data layer from service layer using context. The main challenge of the API we have now is that service is actually containing an instance of the data provider. 

If data provider will be available thru context it will actually give much more freedom to users to use data and services thru context using name (same as prisma).

Then for customization developers will be able to use two factory methods. 
I tried builder today and I think it is not resolving this problem but factory is.

Having two factories (that are optional) could give flexibility, but I'm not sure how you will feel about this

wtrocki:
let me know what do you think.

wtrocki:
Let me know what you think

wtrocki:
Actualy this will invalidate the req to have different services by default. it will require new helper per service per db

wtrocki:
Made edits. ignore emails and check github

templates/ts-apollo-postgres-backend/src/index.ts
```diff
@@ -1,21 +1,51 @@
...
+
+const pubSub = new PubSub();
+
+const dbConfig = getConnectionDetails()
+
+const services = buildGraphbackServices({
+  defaultDataProvider: new PgKnexDBDataProvider(dbConfig),
+  pubSub
+})
+
```
machi1990, line 35: I'll be good to be able to create runtime setups with a single line or few lines (less than this) of code. I have seen this in [OVP](https://github.com/aerogear/OpenVolunteerPlatform/pull/75#discussion_r430341699). I am wondering if the user has to care about how to load graphback config and passing it around some services to finally create `schema` and `resolvers` - what I'd really hope to see is the ability to be able to get the needed bits from graphaback i.e schema, typeDefs, resolvers, context, that could be passed directly to ApolloServer constructor. 

``` 
  const {typeDefs, schema, resolvers, graphbackContext, whetever} = graphbackApi(...)  
```

/!\PS: Ignore naming in the pseudo code above.

craicoverflow: Graphback config will be gone from the templates soon :)

Technically this works without it, you can simply pass your own model string directly to `buildGraphback` if you want, I just didn't want to change the template too much while this spike is still in place as it would distract from what is really changed.

> what I'd really hope to see is the ability to be able to get the needed bits from graphaback i.e schema, typeDefs, resolvers, context, that could be passed directly to ApolloServer constructor

Yep, this is what I am aiming for.

https://github.com/aerogear/graphback/pull/1324/files#diff-9d04c14f3c4e1969e7db314752020f1bR36-R39

The complicated thing about having this all in a one liner is our data providers and CRUD services all rely on having the GraphQL type in the ctor to set up table mapping information etc. 

So it is difficult for a user to provide their own custom CRUD service because to create it they need to have the GraphQLObjectType and create a new one for that model and supply it.

```ts
const myCustomNoteService = new CustomCRUDService(Note);

const { typeDefs, schema, graphbackContext } = graphbackApi(...)
```

This is why I added getter/setters to the CRUD service interface, so that the constructor only requires non-unique information to the contructor, like the dbConfig to the DataProvider, and PubSub to the CRUDService, and then the Graphback service builder would later add model information to each.

Another option is to make the service builder independent, similar to how it is now. But this would still not work since data providers require the model in the constructor.

```ts
const { typeDefs, schema } = graphbackApi(...)

const serviceBuilder = new GraphbackServiceBuilder(typeDefs, new CRUDService(...))
```

craicoverflow:
> ```
> const services = buildGraphbackServices({
>   defaultDataProvider: new PgKnexDBDataProvider(dbConfig),
>   pubSub
> })
> ```
> 
> `pubSub` is specific to the service. Would we be able to handle more arguments in the future?

This is not the model specific pub sub:

```ts
const pubSub = new PubSub();

const services = buildGraphbackServices({
  defaultDataProvider: new PgKnexDBDataProvider(dbConfig),
  pubSub
})
```

Good point though, this won't work.

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
craicoverflow: The constructor can still accept whatever generic information it needs for configuration, but this approach will fail if someone wants to add some model specific info to the CRUD service in the constructor.

But they can't really do this now anyway, since they don't have that information?

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
craicoverflow: Base type is the only fundamental variable that is needed in every data provider.

What other arguments, model-specific auth etc?

craicoverflow: Shouldn't thinks like metrics, auth happen in CRUD service anyway?

craicoverflow:
```ts
const serviceOverrides = { user: new Service(whatever)};
const dataOverrides = { user: new DataLayer(whatever) }
const API = new MongoGraphbackAPI(documents, createGlobalService, serviceOverrides, dataOverrides)

API.resolvers
API.schema
```

This looks great but at what stage does the services receive model information?

`new Service(whatever).create(...)`

This is why I added a setter so that this information can be added by the builder when it has model information, but in this creator approach I do not see where this would be added.

I will need to read up/experiment with this approach.

craicoverflow:
> What was initially build by @darahayes was actually very similar to what SQLMancer has? So you can actually provide a reference to the implementation that will be used.

From what I could see from this approach, this would work when you know the exact  arguments of the CRUDService, which we do know and could defer construction to happen in Graphback, but I opted against this for the purpose of letting users set their own default CRUD service, which could have any custom arguments and therefore we could not construct in Graphback.

wtrocki:
> This looks great but at what stage does the services receive model information?

Global creator will get it nicely. but.. you are right - service overrides will be hard to build as they need their type.

craicoverflow:
> > This looks great but at what stage does the services receive model information?
> 
> Global creator will get it nicely as argument. but.. you are right - service overrides will be hard to build as they need their type.
> Dam this stuff is hard :)

Yep, this is hard for sure. Only reason I added setters (I didn't like adding it :P)

wtrocki:
WDYT?

wtrocki:
I will need to check how setters.work so it will be safe to use them across different services

templates/ts-apollo-postgres-backend/src/index.ts
machi1990: > This is why I added getter/setters to the CRUD service interface, so that the constructor only requires non-unique information to the contructor, like the dbConfig to the DataProvider, and PubSub to the CRUDService, and then the Graphback service builder would later add model information to each.

I generally vouch for immutability as much as I can and avoid `getter/setter`.  Are they intended to be called by the end user? Or just the internal contract we'd have so that the Graphback service builder will be able to set some values to create the `schema`, `typeDefs` and resolvers? 

> Another option is to make the service builder independent, similar to how it is now. But this would still not work since data providers require the model in the constructor.

```ts
const { typeDefs, schema } = graphbackApi(...)

const serviceBuilder = new GraphbackServiceBuilder(typeDefs, new CRUDService(...))
```

From example above, will the problem be solved if we pass `CRUDServiceBuilder` instead of `new CRUDService()`?

wtrocki: Yep. We need to pass specification of the Service that will be able to build service itself. This will add another abstraction

machi1990: > Yep. We need to pass specification of the Service that will be able to build service itself.

Gotcha!

> This will add another abstraction

Yeap, let's start with the setter/getter approach. 

The biggest problem I had with the runtime api is the easy of getting started - a lot of setups need to be done before and this PR is already addressing that, that's awesome.

craicoverflow: > Yep. We need to pass specification of the Service that will be able to build service itself. This will add another abstraction

The only problem with this that I can see is that this is an abstraction which complicates the user API. 

User wants to create a CustomNoteService, CustomCommentService (which both differ) - so they also have to create a builder class for each.

craicoverflow: Architecturally this is much cleaner and desired - however at a compromise of adding boilerplate to the API.

wtrocki:
Agreement is to have API as follows:

```js
const API = new GraphbackAPI(documents, {
dataSourceCreator: createGlobalMongoDS,
serviceCreator: createGlobalService,
plugins)
```

packages/graphback-runtime/src/service/ServiceCreator.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphbackDataProvider } from '../data/GraphbackDataProvider';
+import { PubSubConfig } from '../GraphbackPubSubModel';
+import { GraphbackCRUDService } from './GraphbackCRUDService';
+
+/**
+ * Abstract class for creating a new CRUD Service
+ */
+export abstract class ServiceCreator {
+  public abstract createService(modelName: string, dataProvider: GraphbackDataProvider, pubSubConfig: PubSubConfig): GraphbackCRUDService;
```
craicoverflow, line 9: Help wanted! Actually I do not see how this would work. This `createService` method assumes a very strict signature.

Am I implementing this wrong @wtrocki ?

packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -0,0 +1,117 @@
...
+ * @param {GraphQLSchema|string} model - Data model as a string or GraphQL schema. Used to generate the Graphback API resolvers, services and database
+ * @param {GraphbackAPIConfig} config
+ * @param {ServiceCreator} [config.serviceCreator] - Creator class specifying which default CRUD service should be created for each model.
+ * @param {DataProviderCreator} config.providerCreator - Creator class specifying which default database provider should be created for each model.
+ * @param {GraphbackCRUDGeneratorConfig} [config.crud] - Global CRUD configuration for the Graphback API.
+ * @param {GraphbackPlugin[]} [config.plugins] - Schema plugins to perform automatic changes to the generated schema
+ *
+ * @returns {GraphbackAPI} Generated schema, CRUD resolvers and services
+ */
+export function buildGraphbackAPI(model: string | GraphQLSchema, config: GraphbackAPIConfig): GraphbackAPI {
```
craicoverflow, line 83: This is the API entrypoint.

templates/ts-apollo-postgres-backend/src/index.ts
```diff
@@ -4,19 +4,43 @@ import { ApolloServer } from "apollo-server-express"
...
-const { schema, resolvers } = createRuntime();
+const modelDefs = loadSchemaSync(path.join(__dirname, '../model/*.graphql'), {
+  loaders: [
+    new GraphQLFileLoader()
+  ]
+})
+
+const dbConfig = loadDBConfig()
+
+const { typeDefs, resolvers, services } = buildGraphbackAPI(modelDefs, {
```
craicoverflow, line 27: New way to create Graphback API

packages/graphback-runtime/src/service/ServiceCreator.ts
craicoverflow: We have agreed on the creator pattern, however this is a bit ambiguous as there seems to be multiple creator pattern types (factory, builder being two).

I've implemented factory pattern, because I believe we discussed this. However I can't make it generic enough to work for all cases.

craicoverflow: I think Builder pattern makes more sense, and is more flexible.

packages/graphback-runtime/src/service/CRUDServiceCreator.ts
```diff
@@ -0,0 +1,34 @@
...
+import { CRUDServiceOptions, CRUDService } from './CRUDService';
+
+export interface CRUDServiceCreatorConfig {
+  pubSub: PubSubEngine
+  logger?: GraphbackMessageLogger
+}
+/**
+ * Creator class to create a new CRUDService
+ */
+export class CRUDServiceCreator extends ServiceCreator {
```
wtrocki, line 15: Any reason why this needs to be class vs function?

packages/graphback-runtime/src/service/ServiceCreator.ts
wtrocki: I did not clearly explain that so happy to jump to some call to avoid misunderstanding. 

WHat I mean that both `ServiceCreator` and `createService` abstraction is redundant. 
You can see this as callback and the only granted values that will be generic in the callback would be `model`?

packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export interface GraphbackAPIConfig {
+  /**
+   * Creator class which creates a default CRUD Service for every data model
+   */
+  serviceCreator?: ServiceCreator
+  /**
+   * Creator class which creates a default data provicer for every data model
+   */
+  providerCreator: DataProviderCreator
```
wtrocki, line 15: dataProviderCreator?

packages/graphback/src/buildGraphbackAPI.ts
wtrocki: Yep. Very clean and simple!

packages/graphback-runtime/src/service/ServiceCreator.ts
craicoverflow: Think I get you now. Will get back to this tomorrow!

machi1990: +1 for a function.

packages/graphback-runtime/src/service/CRUDServiceCreator.ts
machi1990: +1

packages/graphback/src/buildGraphbackAPI.ts
machi1990: Like it.

packages/graphback-runtime/src/service/ServiceCreator.ts
craicoverflow: >  the only granted values that will be generic in the callback would be model?

I don't think this is guaranteed to always be the case.

Look at KeycloakCrudService. The constructor signature is different to the standard CRUD service and it has other model config specific to this service which we cannot account for at runtime.
 
https://github.com/aerogear/graphback/blob/0fdc6fb7babe4b70831e9140efe6bb4d2057aae5/packages/graphback-keycloak-authz/src/KeycloakCrudService.ts#L35

craicoverflow: See that authConfig is custom to that model. 

Graphback has to know the structure of the service when creating it:

https://github.com/aerogear/graphback/blob/57fb74e082f213cbe4c9f109ed8387b0d81a1f60/packages/graphback/src/buildGraphbackAPI.ts#L64

But this won't work for KeycloakCRUDService.

I kind of feel like it the original helpers like `createKnexCRUD...RuntimeServices` were better simply because they have the final schema and it is easy to create the services and data providers. 

Yes they don't allow overriding of the services, and it involves having to create one if you want to use a custom services and/or provider.

I will need to think about this.

craicoverflow: Actually I might be wrong about that.

wtrocki: >  Look at KeycloakCrudService. The constructor signature is different to the standard CRUD service and it has other model config specific to this service which we cannot account for at runtime.

I think we failing to understand each other here. Call badly needed.

packages/graphback-keycloak-authz/src/createKeycloakCRUDService.ts
```diff
@@ -0,0 +1,30 @@
+import { ModelDefinition } from '@graphback/core';
+import { PubSubEngine, PubSub } from 'graphql-subscriptions';
+import { GraphbackCRUDService, GraphbackDataProvider } from '@graphback/runtime'
+import { KeycloakCrudServiceOptions, KeycloakCrudService } from './KeycloakCrudService';
+import { CrudServiceAuthConfig } from './definitions';
+
+export interface CreateKeycloakCRUDServiceOptions {
```
wtrocki, line 7: This is amazing!

packages/graphback-runtime-knex/src/OffixDataProvider.ts
```diff
@@ -9,34 +10,34 @@ import { PgKnexDBDataProvider } from './PgKnexDBDataProvider';
  * Note: this provider is still experimental not fully functional
  */
 export class OffixDataProvider extends PgKnexDBDataProvider {
-    public constructor(baseType: GraphQLObjectType, db: any) {
-        super(baseType, db);
-    }
+  public constructor(db: Knex.Config, modelType: GraphQLObjectType) {
```
wtrocki, line 13: Awesome! Very breaking so needs release notes.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -37,33 +34,33 @@ export class LayeredRuntimeResolverCreator {
         const resolverCreateField = getFieldName(modelName, GraphbackOperationType.CREATE);
         //tslint:disable-next-line: no-any
         resolvers.Mutation[resolverCreateField] = (parent: any, args: any, context: any) => {
-          if (!this.services[modelName]) {
+          if (!context.services[modelName]) {
             throw new Error(`Missing service for ${modelName}`);
           }
 
-          return this.services[modelName].create(args.input, context)
+          return context.services[modelName].create(args.input, context)
```
wtrocki, line 41: Awesome! This addresed problem with ability to use services in your own resolver if needed.

packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -0,0 +1,106 @@
...
+import { GraphbackPlugin, GraphbackPluginEngine, GraphbackCRUDGeneratorConfig, printSchemaWithDirectives, ModelDefinition } from '@graphback/core';
+import { SchemaCRUDPlugin } from '@graphback/codegen-schema';
+import { LayeredRuntimeResolverCreator, GraphbackCRUDService, DataProviderCreator, createCRUDService } from '@graphback/runtime';
+import { PubSub } from 'graphql-subscriptions';
+
+export interface GraphbackAPIConfig {
+  /**
+   * Function which creates a default CRUD Service for every data model
+   */
+  serviceCreator?: Function
```
wtrocki, line 11: Hmm.. the only drawback of this is to not being able to define format of the return type from function. I think we can use typescript argument array format to build top level function but then return strongly typed creator one

packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -1,14 +1,17 @@
...
-import { GraphQLObjectType } from 'graphql';
 import { PubSubEngine } from 'graphql-subscriptions';
 import { GraphbackDataProvider } from "../data/GraphbackDataProvider";
 import { defaultLogger, GraphbackMessageLogger } from '../utils/Logger';
 import { GraphbackPage } from "../GraphbackPage"
 import { GraphbackOrderBy } from '../GraphbackOrderBy';
 import { GraphbackCRUDService, ResultList } from "./GraphbackCRUDService";
-import { GraphbackPubSub } from "./GraphbackPubSub"
 
+export interface CRUDServiceConfig {
```
wtrocki, line 10: After seeing complains in issues I think we need to document PubSub usage. any comment will work here

packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Schema plugins to perform automatic changes to the schema
+   */
+  plugins?: GraphbackPlugin[]
+}
+
+interface GraphbackServices {
+  [modelName: string]: GraphbackCRUDService
+}
+
+export interface GraphbackAPI {
```
wtrocki, line 30: Comment missing

templates/ts-apollo-mongodb-backend/src/index.ts
```diff
@@ -1,23 +1,40 @@
...
 
 async function start() {
   const app = express()
 
   app.use(cors())
 
-  // connect to db
-  const { schema, resolvers } = await createRuntime();
+  const modelDefs = loadSchemaSync(path.resolve('./model/*.graphql'), {
+    loaders: [
```
wtrocki, line 20: I think file loader is default one. need to check if we can avoid specifying this

templates/ts-apollo-postgres-backend/src/index.ts
```diff
@@ -4,19 +4,50 @@ import { ApolloServer } from "apollo-server-express"
...
+  dataProviderCreator: createKnexDbProvider(dbConfig),
+  plugins: [
+    new ClientCRUDPlugin({
```
wtrocki, line 31: This makes no sense in this context. we not going to generate client queries at runtime.

Config need to stay to use cli to generate client side. we will think how to transform that outside this PR

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
wtrocki: Docs needed

wtrocki:
Would you see me as ashole to ask to made some docs updates in this PR. Happy to have separate if needed.

templates/ts-apollo-mongodb-backend/src/index.ts
craicoverflow: The options object was at least required, I did not check if `loadSchemaSync(dir, {})` would work but without second arg it didnt work.

templates/ts-apollo-postgres-backend/src/index.ts
craicoverflow: Cool

packages/graphback/src/buildGraphbackAPI.ts
craicoverflow: > Would you see me as ashole to ask to made some docs updates in this PR. Happy to have separate if needed.
> 
> Sorry for pointing some obvious docs stuff. Just making sure that we have the same vision of docs

I wanted to get the code for review and I can work on docs while ye review (quicker turnaround)

templates/ts-apollo-mongodb-backend/src/index.ts
craicoverflow: Screenshot from 2020-06-05 14-10-54

Needs to specify it (although their docs say differently).

craicoverflow:
Will have to do service overrides in a new PR, this one is too big.

packages/graphback-keycloak-authz/package.json
```diff
@@ -35,6 +35,7 @@
     "typescript": "3.9.5"
   },
   "dependencies": {
+    "@graphback/core": "^0.12.1",
```
machi1990, line 38: Should this be `0.14.0-aplha2`?

craicoverflow: yep, spotted the same issue in a different, thanks for seeing this one.

templates/ts-apollo-mongodb-backend/package.json
```diff
@@ -13,6 +13,8 @@
   "license": "Apache 2.0",
   "dependencies": {
     "@graphback/runtime-mongo": "0.14.0-alpha2",
+    "@graphql-tools/graphql-file-loader": "^6.0.7",
+    "@graphql-tools/load": "^6.0.7",
```
machi1990, line 17: I think we should fix versions here.

packages/graphback-keycloak-authz/package.json
machi1990: Cool.

packages/graphback-runtime-knex/src/SQLiteKnexDBDataProvider.ts
```diff
@@ -0,0 +1,32 @@
...
+import { GraphQLObjectType } from 'graphql';
+import { KnexDBDataProvider } from './KnexDBDataProvider';
+
+/**
+ * Knex.js database data provider exposing basic CRUD operations.
+ *
+ * NOTE: This class implements SQLite specific implementaion
+ */
+//tslint:disable-next-line: no-any
+export class SQLiteKnexDBDataProvider<Type = any, GraphbackContext = any> extends KnexDBDataProvider<Type, GraphbackContext>{
```
craicoverflow, line 13: I have added this class mainly for some tests which still use SQLite

packages/graphback-runtime-knex/src/PgKnexDBDataProvider.ts
```diff
@@ -13,18 +13,19 @@ import { KnexDBDataProvider } from './KnexDBDataProvider';
 //tslint:disable-next-line: no-any
 export class PgKnexDBDataProvider<Type = any, GraphbackContext = any> extends KnexDBDataProvider<Type, GraphbackContext>{
 
-    public constructor(baseType: GraphQLObjectType, db: Knex) {
-        super(baseType, db);
-    }
+  public constructor(baseType: GraphQLObjectType, dbConfig: Knex.Config) {
+    console.warn('Deprecated! Please use KnexDBDataProvider')
```
craicoverflow, line 17: Deprecating this because we can use KnexDBDataProvider as the default.

docs/intro/serverless.md
```diff
@@ -10,46 +10,78 @@ Schema and resolvers are created in-code and are passed to an Apollo GraphQL or
 
 ## Adding runtime layer to your application
 
-To create GraphQL Layer at runtime developers need to initialize `GraphbackRuntime` instance as follows:
+The `buildGraphbackAPI` method creates a schema, resolvers and CRUD services at runtime using your data model and some minimal configuration. The example below shows how you can use Graphback with Apollo and a Postgres database.
```
craicoverflow, line 13: Added documentation on how to use the new API. In general the website docs are outdated as they still refer to multiple ways of getting started.

wtrocki:
That is good stuff! I noticed conflict. Would like to play with this over the next 2 days and come back to you but it looks like solution for all our needs is there.

wtrocki:
Happy to accept changes if manyanda is ok with this as well

machi1990:
> Happy to accept changes if manyanda is ok with this as well

This is really good stuff already. Hoping w get the integration tests fixed and we should be good to go.

machi1990:
```
FAIL  packages/graphback-runtime-knex/tests/service/DefaultCRUDServiceTest.ts
 ● Test suite failed to run

    packages/graphback-runtime-knex/tests/service/DefaultCRUDServiceTest.ts:55:67 - error TS2345: Argument of type 'Config<any>' is not assignable to parameter of type 'Knex<any, unknown[]>'.
      Type 'Config<any>' is missing the following properties from type 'Knex<any, unknown[]>': VERSION, __knex__, raw, transactionProvider, and 122 more.

    55     const modelProvider = new SQLiteKnexDBDataProvider(modelType, dbConfig);
``` 


seems like the signature change is not one throughout.

craicoverflow:
@wtrocki @machi1990 just going to add a test for buildGraphbackAPI but everything else is ready

wtrocki:
all good. LGTM

packages/graphback/tests/buildGraphbackAPI.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      connection: {
+        filename: ':memory:'
+      },
+      useNullAsDefault: true
+    })
+
+    return { db }
+  }
+
+  test('Creates a Graphback API using the default configuration with Knex', () => {
```
craicoverflow, line 22: Tests now added :)

machi1990: Nice. Should the name of the test file be `buildGraphbackAPITest`?

packages/graphback-keycloak-authz/tests/KeycloakCrudService.test.ts
```diff
@@ -205,11 +210,11 @@ test('multiple roles can be applied to each operation', async () => {
 
   const context = { kauth: new KeycloakContextBase(Token) }
 
-  const val = {test: 'value'}
+  const val = { test: 'value' }
```
machi1990, line 213: Not related, but I think we should automate this to avoid these linting noises in a PR :-)

packages/graphback/tests/buildGraphbackAPI.test.ts
craicoverflow: I will change it for consistency, I prefer `buildGraphbackAPI.test.ts` though because it is much easier to see at a glance that it is a test. I think it is a more common convention in JS/TS to do `.spec.ts` or `.test.ts`. 

Maybe we will change them all sometime.

image

machi1990:
@craicoverflow I do not see the runtime api changes. The only change that got merged is the memory increase for ci.

machi1990:
> @craicoverflow I do not see the runtime api changes. The only change that got merged is the memory increase for ci.

Fixed by https://github.com/aerogear/graphback/pull/1403

PR #1335, opened 2020-05-29
fix(template): fix incorrect template paths

ssd71:


ssd71:
Wrong branch, apologies.

wtrocki:
Ok.. Tried to fix it. we need that on master

PR #1365, opened 2020-06-03
Deprecate offix into graphback-datasync package

ssd71:
Work is in progress


wtrocki:
@ssd71 Let's avoid huge PR's like previous one. Ideally we should have max 1-2 days PR's. If we split work to only move and rename this PR could work nicely. 

THere is no documentation for offix and this is not going to affect anything

packages/graphback-datasync/src/providers/DatasyncKnexDataProvider.ts
```diff
@@ -0,0 +1,42 @@
+import { GraphQLObjectType } from 'graphql';
```
wtrocki, line 1: I will keep it mongo only for the moment. 
We will focus on Mongo and bring knex at later point

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -0,0 +1,53 @@
...
+import { getDatabaseArguments } from '@graphback/core';
+import { ObjectId } from 'mongodb';
+import { MongoDBDataProvider } from '@graphback/runtime-mongo';
+
+/**
+ * Mongo provider that contains special handlers for offix conflict resolution format:
+ *
+ * https://offix.dev/docs/conflict-server#structure-of-the-conflict-error
+ */
+export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> extends MongoDBDataProvider<Type, GraphbackContext> {
```
wtrocki, line 12: DataSync

wtrocki:
@ssd71 Let's avoid huge PR's like previous one. Ideally we should have max 1-2 days PR's. If we split work to only move and rename this PR could work nicely. 

THere is no documentation for offix and this is not going to affect anything

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: I am a bit puzzled as to what to do here, do you mean to change docstring or name of class

ssd71:
> @ssd71 Let's avoid huge PR's like previous one. Ideally we should have max 1-2 days PR's. If we split work to only move and rename this PR could work nicely.
> 
> THere is no documentation for offix and this is not going to affect anything

Agreed

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
wtrocki: I ment rename. You can rename to camelcase. We can do it afterwards.

PR #1370, opened 2020-06-03
Production ready client generation

Eunovo:
This PR modifies the client codegen to use specified `outputPath` as file path for output. This PR also contains Doc fixes

### DIFF
--- a/docs/plugins/graphback-client.md
+++ b/docs/plugins/graphback-client.md
@@ -37,9 +37,8 @@ Plugin allows to create files in many formats to suit the needs of your applicat
    *
    * - ts - typescript file output (backwards compatibility)
    * - graphql - .graphql file
-   * - gqlwithfragment - complete graphql queries containing fragments for redundancy
    */
-  format: 'ts' | 'graphql' | 'gqlwithfragment'
+  format: 'ts' | 'graphql'
 
   /**
    * Generate only fragments and skip query, mutation and subscription elements
@@ -49,9 +48,9 @@ Plugin allows to create files in many formats to suit the needs of your applicat
   fragmentOnly?: boolean
 
   /**
-   * RelativePath for the output files created by generator
+   * RelativePath for the output file created by generator
    */
-  outputPath: string
+  outputFile: string
 ```
 
 Example plugin configuration in YAML: 
@@ -59,7 +58,7 @@ Example plugin configuration in YAML:
  plugins:
   graphback-client:
     format: 'graphql'
-    outputPath: ./client/src/graphql
+    outputFile: ./client/src/graphql/graphback.graphql
 ```
 
 > Note: Plugin require `crud` configuration specified as part of the `graphback` extension 
--- a/docs/plugins/graphback-client.md
+++ b/docs/plugins/graphback-client.md
@@ -27,6 +27,11 @@ apolloClient.mutate(createUserQuery)
 ```
 
 Plugin allows to create files in many formats to suit the needs of your application.
+The client documents for graphql format are generated in a single file and this should be used by generators, like;
+[GraphQL Code Generator](https://graphql-code-generator.com/)
+[Apollo Generator](https://github.com/apollographql/apollo-tooling#code-generation)
+
+You can also generate the client documents in typescript for direct use if you are not use a generator
 
 ## Plugin Config
 
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -20,6 +20,11 @@ Check configuration of the resolvers plugin to see where they are located.
 2. Remove @graphback/codegen-resolvers from your package and the config.
 3. Use Graphback runtime Getting Started Guide 
 4. Change your client output format from `gqlwithfragment` to `graphql`
+5. Replace `outputPath` with `outputFile` in your client config 
+5. The `outputFile` specifies the output path of the generated file
+6. Change your documents path in graphback config to match new `outputFile`
+7. Delete previously generated client documents and rerun generate command
+8. Be sure to correct your documents path in your codegen config if you use GraphQL-Code-Generator
 
 ## Breaking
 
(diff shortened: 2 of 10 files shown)


docs/plugins/graphback-client.md
```diff
@@ -49,7 +48,7 @@ Plugin allows to create files in many formats to suit the needs of your applicat
   fragmentOnly?: boolean
 
   /**
-   * RelativePath for the output files created by generator
+   * RelativePath for the output file created by generator
    */
   outputPath: string
```
wtrocki, line 53: Output file then?

docs/plugins/graphback-client.md
```diff
@@ -59,7 +58,7 @@ Example plugin configuration in YAML:
  plugins:
   graphback-client:
     format: 'graphql'
-    outputPath: ./client/src/graphql
+    outputPath: ./client/src/graphql/graphback.graphql
```
wtrocki, line 61: generated.graphql?

docs/releases.md
```diff
@@ -20,6 +20,10 @@ Check configuration of the resolvers plugin to see where they are located.
 2. Remove @graphback/codegen-resolvers from your package and the config.
 3. Use Graphback runtime Getting Started Guide 
 4. Change your client output format from `gqlwithfragment` to `graphql`
+5. Change your client output path to be the output path of generated file
+6. Change your documents path in graphback config to match new `outputPath`
+7. Delete previously generated client documents and rerun generate command
+8. Be sure to correct your documents path in your codegen config if you use `graphql-codegen`
```
wtrocki, line 26: GraphQL-Code-Generator - codegen is internal slang we use as shortcut

packages/graphback-codegen-client/src/helpers/writeDocuments.ts
```diff
@@ -2,15 +2,16 @@ import { existsSync, mkdirSync, writeFileSync } from 'fs';
...
-    }
-
     const implementations: string[] = [];
     Object.keys(generatedClient).forEach((folder: string) => {
         //tslint:disable-next-line: no-any
         generatedClient[folder].forEach((c: any) => implementations.push(c.implementation));
     })
 
-    writeFileSync(`${clientPath}/graphback.${extension}`, implementations.join('\n\n'));
+    const isExtensionSpecifiedRegExp = new RegExp(/\.([a-zA-Z])+$/);
```
wtrocki, line 11: This is redundant. It should be GraphQL but some people might call this .query etc.

machi1990:
@Eunovo thanks for taking care of this. 

This will need a rebase.

docs/plugins/graphback-client.md
Eunovo: Does it matter? The name can be changed and `graphback.graphql` doesn't seem bad

docs/plugins/graphback-client.md
Eunovo: Will change `outputPath` to `outputFile`

docs/releases.md
Eunovo: All right

packages/graphback-codegen-client/src/helpers/writeDocuments.ts
Eunovo: > This might br redundant. Filem send with GraphQL but some people might call this .query etc.
>

I don't understand what you mean here

wtrocki: Sorry was typing on the phone. I do not think this is needed. We should allow people to specify whatever extension they want. Also implementation using regexp is quite inefficient

Eunovo: I have removed the regexp. I'm using path now.

People can specify whatever extension they want. I only add an extension when none is specified. So, you want me to allow people specify no extension, right?

wtrocki: I will leave it. Generally I tend to do less of stuff like this as it is more code to maintain etc.

docs/plugins/graphback-client.md
```diff
@@ -37,9 +37,8 @@ Plugin allows to create files in many formats to suit the needs of your applicat
    *
```
wtrocki, line 37: We should mention that we generate graphql as single file and this should be used by generators. We should list apollo generator and graphql-code-gen.

Then for non generator approach, we generating typescript to be used directly

packages/graphback-codegen-client/src/helpers/writeDocuments.ts
Eunovo: @wtrocki

wtrocki:
Docs changes still required

docs/releases.md
```diff
@@ -20,6 +20,11 @@ Check configuration of the resolvers plugin to see where they are located.
 2. Remove @graphback/codegen-resolvers from your package and the config.
 3. Use Graphback runtime Getting Started Guide 
 4. Change your client output format from `gqlwithfragment` to `graphql`
+5. Replace `outputPath` with `outputFile` in your client config 
+5. The `outputFile` specifies the output path of the generated file
```
machi1990, line 24: Should it be `6.` ?

docs/plugins/graphback-client.md
```diff
@@ -27,6 +27,11 @@ apolloClient.mutate(createUserQuery)
 ```
 
 Plugin allows to create files in many formats to suit the needs of your application.
+The client documents for graphql format are generated in a single file and this should be used by generators, like;
+[GraphQL Code Generator](https://graphql-code-generator.com/)
+[Apollo Generator](https://github.com/apollographql/apollo-tooling#code-generation)
+
+You can also generate the client documents in typescript for direct use if you are not use a generator
```
machi1990, line 34: `You can also generate the client documents in typescript for direct use if you are not using a the above mentioned generators`.

Eunovo: @wtrocki didn't limit `graphql` format to only the mentioned generators. What if more are created?

docs/releases.md
Eunovo: I'll fix that

docs/plugins/graphback-client.md
wtrocki: This is to show people examples. There are couple other generators already but we mentioning 2 most popular ones.

Eunovo: No problem.

PR #1376, opened 2020-06-04
Add support for delta queries

ssd71:
This PR adds experimental support for delta queries using the graphback-datasync package, approach is still under investigation, feel free to review


packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -87,6 +88,20 @@ export class LayeredRuntimeResolverCreator {
...
+        resolvers.Query[diffQuery] = (parent: any, args: any, context: any) => {
+          const filter: any = {};
+          if (args?.lastSync) {
+            filter.updatedAt = { gt: args.lastSync};
+          }
+          
+          return this.services[modelName].findBy(filter, undefined, undefined, context);
+        }
+      }
+
```
ssd71, line 104: Is there some other way to add the delta resolvers? This is kinda uncomfortable to do 😄

packages/graphback-datasync/src/DatasyncPlugin.ts
```diff
@@ -76,13 +77,13 @@ export class DatasyncPlugin extends GraphbackPlugin {
...
+            if (parseMarker('delta', model.graphqlType.description)) {
+                const diffQuery = `get${model.graphqlType.name}Delta`
                 schemaComposer.Query.addFields({
-                    [diffQuery]: `[${model.graphqlType.name}]!`
-                })
+                    [diffQuery]: `[${model.graphqlType.name}]`
+                });
                 schemaComposer.Query.addFieldArgs(diffQuery, {
-                    lastSync: 'String!'
+                    lastSync: 'String'
```
ssd71, line 86: This is where it adds delta queries to the schema, I am going to replace the resolver Type with something more involved, `ResultList` type is a candidate for this but I think it wouldn't suffice as it does not have a `lastSync`

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -87,6 +88,20 @@ export class LayeredRuntimeResolverCreator {
         }
       }
 
+      // If delta marker is encountered, add resolver for `delta` query
+      if (parseMarker('delta', resolverElement.graphqlType.description)) {
```
craicoverflow, line 92: To save having to parse the delta marker every time like this, you could add this configuration value to the ModelDefinition [here](https://github.com/aerogear/graphback/blob/f553235391fd3b6554c0acb751e159cdbe3fe4c8/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts#L75).

So it could be refactored to `if (resolverElement.deltaSync) {`

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -87,6 +88,20 @@ export class LayeredRuntimeResolverCreator {
...
+        const diffQuery = `get${resolverElement.graphqlType.name}Delta`
```
craicoverflow, line 93: This is likely going to be used elsewhere, so we should centralise the naming in `mappingHelpers`, similar to [getSubscript](https://github.com/aerogear/graphback/blob/cf09a8d914a274c2344d7a8de2e66fc5e597b37e/packages/graphback-core/src/crud/mappingHelpers.ts#L77)

craicoverflow: `get<Model>Delta` implies this should only return a single item, since `get<Model>` is used to fetch a single record by its ID.

This should be named something different, like `sync<Model>s`.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -87,6 +88,20 @@ export class LayeredRuntimeResolverCreator {
...
+
+        resolvers.Query[diffQuery] = (parent: any, args: any, context: any) => {
+          const filter: any = {};
+          if (args?.lastSync) {
+            filter.updatedAt = { gt: args.lastSync};
```
craicoverflow, line 98: This should merge the user filter from args too.

craicoverflow: Ignore that statement ^ I see there are no filter arguments in the resolver method.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -87,6 +88,20 @@ export class LayeredRuntimeResolverCreator {
         }
       }
 
+      // If delta marker is encountered, add resolver for `delta` query
+      if (parseMarker('delta', resolverElement.graphqlType.description)) {
+        const diffQuery = `get${resolverElement.graphqlType.name}Delta`
+
+        resolvers.Query[diffQuery] = (parent: any, args: any, context: any) => {
+          const filter: any = {};
+          if (args?.lastSync) {
```
craicoverflow, line 97:
```suggestion
```

Since `lastSync` is a required argument, it is safe to assume `args` and `lastSync` are defined, as GraphQL will validate this client side.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
craicoverflow: This seems like a good approach to me, what makes it uncomfortable?

packages/graphback-datasync/src/DatasyncPlugin.ts
```diff
@@ -76,13 +77,13 @@ export class DatasyncPlugin extends GraphbackPlugin {
 
 
             // Diff queries
-            if (this.pluginConfig.generateDeltaQueries) {
-                const diffQuery = `${model.graphqlType.name}Delta`
+            if (parseMarker('delta', model.graphqlType.description)) {
```
wtrocki, line 80: @craicoverflow I thought we changed api in https://github.com/aerogear/graphql-metadata

craicoverflow: Yes, you can now do:

`parseMarker('delta', model.graphqlType)` or preferably `parseMetadata('delta', model.graphqlType)`.

`parseMarker` accepts to make it backwards compatible.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: I was a bit uncomfortable adding datasync stuff to the runtime itself, but nevermind that 😄

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: Do you think we should have that though? Might be useful

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: I actually have made it optional here:
https://github.com/aerogear/graphback/blob/72cc030a1fb4a190d1295b67def411289b324fcd/packages/graphback-datasync/src/DatasyncPlugin.ts#L85-L87

I was thinking if it could be used for getting data initially?

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
craicoverflow: @wtrocki would be best to answer this one. I can see it being useful but it may cause some side effects I am unaware of.

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
wtrocki: It should be required for deltaQueries

packages/graphback-datasync/src/DatasyncPlugin.ts
ssd71: This is done

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: Resolving this, feel free to open if I am mistaken

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: This is also done: https://github.com/aerogear/graphback/blob/bf46cbf020736b3edfe55b2632a369a736f40270/packages/graphback-core/src/crud/mappingHelpers.ts#L178

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: Please check:
https://github.com/aerogear/graphback/blob/bf46cbf020736b3edfe55b2632a369a736f40270/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts#L80-L84

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: This is done now:
https://github.com/aerogear/graphback/blob/c5783e234d8c30f6e6bed1284223b14d44c95fc5/packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts#L94-L99
https://github.com/aerogear/graphback/blob/c5783e234d8c30f6e6bed1284223b14d44c95fc5/packages/graphback-datasync/src/DatasyncPlugin.ts#L78

packages/graphback-datasync/src/DatasyncPlugin.ts
```diff
@@ -56,32 +57,24 @@ export class DatasyncPlugin extends GraphbackPlugin {
...
-                if (inputType) {
-                    inputType.addFields({
-                        version: 'Int'
-                    });
+            const SyncResultList = new GraphQLObjectType({
+                name: `${model.graphqlType.name}DeltaList`,
+                fields: {
+                    items: { type: GraphQLList(model.graphqlType)},
+                    lastSync: { type: GraphQLString}
                 }
```
machi1990, line 65: I am wondering if we should deal / or there are plans to deal with pagination too?

ssd71: Presently this setup returns all changed documents, but I think paginating it could be worthwhile, @wtrocki  what do you think?

craicoverflow: Offset-based pagination won't work well with data synchronisation I believe, since the result set is dynamic and can be very different after each request. 

We would first need to implement cursor-based pagination, which is probably outside the scope of this PR.

https://www.sitepoint.com/paginating-real-time-data-cursor-based-pagination/

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -170,3 +170,11 @@ export function getInputFieldType(field: GraphQLField<any, any>): GraphQLInputTy
 
   return undefined;
 }
+
+/**
+ * Get the name of delta query for a model
+ * @param typeName Name of the model
+ */
+export function getDeltaQuery(typeName: string) {
+  return `sync${pluralize(typeName)}`;
```
machi1990, line 179: Why do we pluralise? I have seen it all over in the file so this is a general remark not specific for this PR.

I am wondering if the lib used takes into account complex pluralisation rules from other languages? E.g in [French](https://grammar.collinsdictionary.com/french-easy-learning/forming-plurals#:~:text=Most%20French%20nouns%20form%20their,plural%2C%20with%20a%20few%20exceptions.) the rules are so different that suffix an `s` to a business model `Hopital` (Hospital in French) will be wrong, as it should be `Hopitaux`.  I'd generally tend to stay off from doing this kind of stuff and just use the model name.

packages/graphback-datasync/src/DatasyncPlugin.ts
machi1990: +1 with what @craicoverflow said. I am fine with the answers here.

packages/graphback-core/src/crud/mappingHelpers.ts
craicoverflow: But then you lead to misleading names in the schema like:

`findNote` instead of `findNotes`.

One implies fetching a single record, the other implies multiple.

I just checked and it seems it only supports English as a language

craicoverflow: I'd still like to use it, as it will work for the majority of cases.

But as you said there are cases where it simply won't.

A suggestion is to have a model annotation which can override the default resolver names:

```gql
"""
@model
@crud.find: 'trouverDesHôpitaux'
"""
```

This is something we can deal with if a request comes in, but for now sticking with pluralize will do.

machi1990: Let's keep it this way considering the naming constraints we have. I'll keep in my head and come back to this once I have a better proposition.

ssd71:
> Would love to see some tests and documentations.

@machi1990 Would love to add docs in separate PR :smile:

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -28,12 +29,12 @@ export interface DatasyncPluginConfig {
 export const SCHEMA_CRUD_PLUGIN_NAME = "DatasyncPlugin";
 
 /**
- * Ofix plugin
+ * DataSync plugin
  *
- * Plugin is enabled by """ @datasync """ annotation
- * It will add version field to the schema and also generate diffQueries
+ * Plugin is enabled by """ @ddelta """ annotation
```
machi1990, line 34:
```suggestion
 * Plugin is enabled by """ @delta """ annotation
```

machi1990:
> > Would love to see some tests and documentations.
> 
> @machi1990 Would love to add docs in separate PR 😄

Sure. 

Also, about test, I was thinking of integration level tests where we could actually send a gql query. Like the one we have here - https://github.com/aerogear/graphback/tree/master/integration. Let me know what you think.

ssd71:
@machi1990 Sure! Sounds good!

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
wtrocki: Yes. Delta should be filtered the same way as the regular find method.

Issue #1379, opened 2020-06-04
graphback-runtime-mongodb/tests/queryBuilderTest.ts is unstable


machi1990:
I have seen this test failing randomly a couple of times already. Below is an extract of [the latest failure](https://circleci.com/gh/aerogear/graphback/4578?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

```log
Summary of all failing tests
 FAIL  packages/graphback-runtime-mongodb/tests/queryBuilderTest.ts (5.173 s)
  ● queryBuilder scalar filtering › can filter @versioned metadata fields

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
    -   "hi guys",
        "not yet",
        "bye guys",
    +   "hi guys",
      ]

      302 |     const posts = await context.providers.Post.findBy({ createdAt: { gt: startTime } });
      303 |     expect(posts.length).toEqual(3);
    > 304 |     expect(posts.map((post: any) => post.text)).toEqual(["hi guys", "not yet", "bye guys"]);
          |                                                 ^
      305 | 
      306 |     // Get all posts created after the first post
      307 |     const newPosts = await context.providers.Post.findBy({ createdAt: { gt: posts[0].createdAt } });

      at packages/graphback-runtime-mongodb/tests/queryBuilderTest.ts:304:49
      at fulfilled (node_modules/tslib/tslib.js:112:62)
          at runMicrotasks (<anonymous>)
```

PR #1380, opened 2020-06-04
test: fix random mongo queryBuilder test failure

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1379


machi1990:
> Thanks! 

Yw. 

> .. I also encountered this just now! 😄

Was hit by it a couple of time in CI and locally. :-)

PR #1391, opened 2020-06-05
Do not overwrite existing docker-compose file

### LINKED ISSUE #1387: Graphback overwrote my docker-compose.yml
I've been incorporating graphback into a project that is already using docker (Prisma to be specific). I'm still using Prisma as I slowly refactor it away. I somehow missed it, but I believe that during the graphback initialization process, my docker-compose.yml was replaced with this:

```yaml
version: '3'

services:
  postgres:
    image: postgres:9.6
    ports:
      - "55432:5432"
    environment:
      POSTGRES_PASSWORD: postgres
      POSTGRES_USER: postgresql
      POSTGRES_DB: users
```
I was not expecting this to happen, and I was given no warning. It's only a week or two later that I've found this problem, as my normal docker containers weren't showing up.

I think a warning should be given when there is already a docker-compose file present, or perhaps rethink the docker thing entirely.

machi1990:
Fixes #1387


packages/graphback-cli/src/templates/configTemplates.ts
```diff
@@ -1,48 +1,47 @@
...
  * Different database choices
  */
 const getConfig = (database: string) => {
-  if (database === 'pg') {
-    return [readFileSync(`${configFilesPath}/postgres.json`, 'utf8'), readFileSync(`${dockerFilesPath}/postgres.yml`, 'utf8')]
-  } else if (database === 'MongoDB') {
-    return ["", readFileSync(`${dockerFilesPath}/mongodb.yml`, 'utf8')]
-  } else if (database === 'sqlite3') {
-    return [readFileSync(`${configFilesPath}/sqlite3.json`, 'utf8'), undefined]
+  if (database === postgreSQL) {
```
wtrocki, line 19: this name was used for knex to identify database. If we use it here we will need to map it in the knex config

wtrocki: Ah.. ignore me - we no longer actually put any db info into config! All good

machi1990: Yeap, I had to double check before doing it. 

Actually, this function could only return the docker-compose part as the first element of the array is never used.

wtrocki:
Tried to run this with mongo version on top of existing template.
I could not find any occurrence in the docs that will be affected.

All works. @machi1990 absolute pleasure to work with you!
💯

machi1990:
> Tried to run this with mongo version on top of existing template.
> I could not find any occurrence in the docs that will be affected.
> 
> All works. @machi1990 absolute pleasure to work with you!
> 💯

It is the mutual feeling. Thanks for the swift review.

PR #1394, opened 2020-06-07
Add soft deletes for Mongo

ssd71:
This PR adds soft deletion functionality to DataSyncMongoDBProvider

### DIFF
--- a/packages/graphback-core/src/crud/mappingHelpers.ts
+++ b/packages/graphback-core/src/crud/mappingHelpers.ts
@@ -178,11 +178,3 @@ export function getInputFieldType(field: GraphQLField<any, any>): GraphQLInputTy
 export function getDeltaQuery(typeName: string) {
   return `sync${pluralize(typeName)}`;
 }
-
-export function getDeltaListType(typeName: string): string {
-  return `${typeName}DeltaList`;
-}
-
-export function getDeltaType(typeName: string) {
-  return `${typeName}Delta`;
-}
--- a/packages/graphback-core/src/crud/mappingHelpers.ts
+++ b/packages/graphback-core/src/crud/mappingHelpers.ts
@@ -178,3 +178,11 @@ export function getInputFieldType(field: GraphQLField<any, any>): GraphQLInputTy
 export function getDeltaQuery(typeName: string) {
   return `sync${pluralize(typeName)}`;
 }
+
+export function getDeltaListType(typeName: string): string {
+  return `${typeName}DeltaList`;
+}
+
+export function getDeltaType(typeName: string) {
+  return `${typeName}Delta`;
+}
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -1,8 +1,9 @@
 
-import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, getDeltaQuery, getDeltaType, getDeltaListType, parseRelationshipAnnotation } from '@graphback/core'
+import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, parseRelationshipAnnotation, getDeltaQuery } from '@graphback/core'
 import { GraphQLObjectType, GraphQLInt, GraphQLNonNull, GraphQLList, GraphQLSchema, buildSchema, GraphQLString, GraphQLBoolean } from 'graphql';
 import { parseMetadata } from "graphql-metadata";
 import { SchemaComposer, ObjectTypeComposerFieldConfig } from 'graphql-compose';
+import { getDeltaType, getDeltaListType } from "./deltaMappingHelper";
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -60,6 +61,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
             const modelName = model.graphqlType.name;
             const modifiedType = schemaComposer.getOTC(modelName);
             const entries = Object.entries(modifiedType.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
+                // Remove relationship fields from delta Types
                 return parseRelationshipAnnotation(e[1].description) === undefined
             })
 
@@ -67,7 +69,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
             schemaComposer.createObjectTC(getDeltaType(modelName)).addFields({
                 ...fields,
-                deleted: 'Boolean',
+                _deleted: 'Boolean',
             })            
 
             schemaComposer.createObjectTC({
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -1,8 +1,8 @@
 
-import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, getDeltaQuery } from '@graphback/core'
-import { GraphQLObjectType, GraphQLInt, GraphQLNonNull, GraphQLList, GraphQLSchema, buildSchema, GraphQLString } from 'graphql';
+import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, getDeltaQuery, getDeltaType, getDeltaListType, parseRelationshipAnnotation } from '@graphback/core'
+import { GraphQLObjectType, GraphQLInt, GraphQLNonNull, GraphQLList, GraphQLSchema, buildSchema, GraphQLString, GraphQLBoolean } from 'graphql';
 import { parseMetadata } from "graphql-metadata";
-import { SchemaComposer } from 'graphql-compose';
+import { SchemaComposer, ObjectTypeComposerFieldConfig } from 'graphql-compose';
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -55,24 +55,34 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
             return schema;
         }
-
         models.forEach((model: ModelDefinition) => {
-            const SyncResultList = new GraphQLObjectType({
-                name: `${model.graphqlType.name}DeltaList`,
-                fields: {
-                    items: { type: GraphQLList(model.graphqlType)},
-                    lastSync: { type: GraphQLString}
-                }
+            
+            const modelName = model.graphqlType.name;
+            const modifiedType = schemaComposer.getOTC(modelName);
+            const entries = Object.entries(modifiedType.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
+                return parseRelationshipAnnotation(e[1].description) === undefined
             })
 
+            const fields = Object.assign({}, ...Array.from(entries, ([k, v]: [string, any]) => ({ [k]: v })));
+
+            schemaComposer.createObjectTC(getDeltaType(modelName)).addFields({
+                ...fields,
+                deleted: 'Boolean',
+            })            
+
+            schemaComposer.createObjectTC({
+                name: getDeltaListType(modelName),
+                fields: {
+                    items: `[${getDeltaType(modelName)}]!`,
+                    lastSync: `String`
+                }
+            })
             // Diff queries
             if (parseMetadata('delta', model.graphqlType)) {
                 
                 const deltaQuery = getDeltaQuery(model.graphqlType.name)
                 schemaComposer.Query.addFields({
-                    [deltaQuery]: {
-                        type: GraphQLNonNull(SyncResultList)
-                    }
+                    [deltaQuery]: `${getDeltaListType(modelName)}!`
                 });
                 schemaComposer.Query.addFieldArgs(deltaQuery, {
                     lastSync: 'String!'
--- a/packages/graphback-datasync/src/deltaMappingHelper.ts
+++ b/packages/graphback-datasync/src/deltaMappingHelper.ts
@@ -1,0 +1,8 @@
+import * as pluralize from 'pluralize'
+
+export function getDeltaListType(typeName: string): string {
+  return `${typeName}DeltaList`;
+}
+export function getDeltaType(typeName: string) {
+  return `${typeName}Delta`;
+}
--- a/packages/graphback-datasync/src/deltaMappingHelper.ts
+++ b/packages/graphback-datasync/src/deltaMappingHelper.ts
@@ -1,5 +1,3 @@
-import * as pluralize from 'pluralize'
-
 export function getDeltaListType(typeName: string): string {
   return `${typeName}DeltaList`;
 }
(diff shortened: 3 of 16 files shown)


packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -178,3 +178,11 @@ export function getInputFieldType(field: GraphQLField<any, any>): GraphQLInputTy
 export function getDeltaQuery(typeName: string) {
   return `sync${pluralize(typeName)}`;
 }
+
+export function getDeltaListType(typeName: string): string {
```
wtrocki, line 182: -1 for core. Delta will be very specific to datasync

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -55,24 +55,34 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+            const modifiedType = schemaComposer.getOTC(modelName);
+            const entries = Object.entries(modifiedType.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
+                return parseRelationshipAnnotation(e[1].description) === undefined
+            })
+
+            const fields = Object.assign({}, ...Array.from(entries, ([k, v]: [string, any]) => ({ [k]: v })));
+
+            schemaComposer.createObjectTC(getDeltaType(modelName)).addFields({
+                ...fields,
+                deleted: 'Boolean',
```
wtrocki, line 70: _deleted could be better to not clash with existing names

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -55,24 +55,34 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+            })            
+
+            schemaComposer.createObjectTC({
+                name: getDeltaListType(modelName),
                 fields: {
-                    items: { type: GraphQLList(model.graphqlType)},
-                    lastSync: { type: GraphQLString}
+                    items: `[${getDeltaType(modelName)}]!`,
+                    lastSync: `String`
```
wtrocki, line 77: I will add description to that field if compose allows us to do it

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -1,29 +1,25 @@
...
-  }
-
-  public async create(data: any): Promise<Type> {
-    if (!data.version) {
-      data.version = 1;
-    }
 
-    return super.create(data);
+    // Mark any created items with 'deleted' flag set to false
+    this.fieldTransformMap[TransformType.CREATE].push({
```
wtrocki, line 16: I think we overusing this. Since we will have provider for every strategy we might directly out version,_deleted etc. into create method. 
No strong opinion but at the first glance looks like overkill without any functional reason. is there functional benefit to have it this way?

packages/graphback-core/src/crud/mappingHelpers.ts
ssd71: I agree, should we put all getDelta* in datasync?

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: > I think we overusing this. Since we will have provider for every strategy we might directly out version,_deleted etc. into create method.

I have removed version field, do we need it?

> No strong opinion but at the first glance looks like overkill without any functional reason. is there functional benefit to have it this way?

Not really, was just trying to reuse that code

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Done

packages/graphback-core/src/crud/mappingHelpers.ts
ssd71: I have moved stuff as non destructively as I could, please check

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: I am not sure I understand :smile: could you explain?

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: Made some changes, please check

packages/graphback-datasync/src/DataSyncPlugin.ts
wtrocki: Ignore this. Every field in GraohQL can have description that users can read..This is small improvement I can do later on master

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -92,7 +92,7 @@ export class LayeredRuntimeResolverCreator {
         const deltaQuery = getDeltaQuery(resolverElement.graphqlType.name)
 
         resolvers.Query[deltaQuery] = async (parent: any, args: any, context: any) => {
-          const res = await this.services[modelName].findBy({ updatedAt:{ gt: args.lastSync }}, undefined, undefined, context);
+          const res = await this.services[modelName].findBy({ updatedAt:{ gt: args.lastSync }, _deleted: { in: [true, false]}}, undefined, undefined, context);
```
wtrocki, line 95: This is going to affect all datasources sadly we cannot do that because it will break non DataSync graphback

wtrocki: You probably mean datasource or DataSync service

ssd71: Yes about this, I cannot find an optimal way to add the delta resolver without doing it in layered runtime resolver

wtrocki: Can we just extend this class and add deltas there?

wtrocki: Also I meant that deleted should be specified on the datasource level or service

machi1990: > Also I meant that _deleted should be specified on the datasource level or service

+1 with this.

packages/graphback-datasync/src/deltaMappingHelper.ts
```diff
@@ -0,0 +1,8 @@
+import * as pluralize from 'pluralize'
```
machi1990, line 1: Unused import?

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const { Post }  = context.providers;
+    const startTime = 1590679886048;
+    const deleteTime = 1590679886048;
+    advanceTo(startTime);
+    const { id } = await Post.create({ text: 'TestPost' });
+    advanceTo(deleteTime);
+    const deletedPost = await Post.delete({ id });
+    expect(deletedPost._deleted).toEqual(true);
+    expect(deletedPost.updatedAt).toEqual(deleteTime);
```
machi1990, line 106: I am wondering whether we should have a `deletedAt` field too?

machi1990, line 106: Can we call `Post.findBy` and make sure that we do not have the deleted item?

ssd71: I think this has been brought up up before, with conclusion that updatedAt can store this timestamp, @wtrocki what do you think?

wtrocki: deletedAt = deleted=true and updatedAt..

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -32,16 +27,14 @@ export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
     }
 
     // TODO Can be improved by conditional updates
-    const queryResult = await this.db.collection(this.collectionName).find({ _id: new ObjectId(idField.value) }).toArray();
+    const queryResult = await this.db.collection(this.collectionName).find({ _id: new ObjectId(idField.value), _deleted: false }).toArray();
     if (queryResult && queryResult[0]) {
       queryResult[0][idField.name] = queryResult[0]._id;
```
machi1990, line 32: We could use `this.findOne()` here.

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
```diff
@@ -0,0 +1,108 @@
...
+import { advanceTo, advanceBy } from "jest-date-mock";
+import { SchemaCRUDPlugin } from "@graphback/codegen-schema";
+import { DataSyncMongoDBDataProvider, DataSyncPlugin, createDataSyncMongoCRUDRuntimeContext } from '../src';
+
+export interface Context {
+    providers: { [modelname: string]: DataSyncMongoDBDataProvider };
+    server: MongoMemoryServer
+  }
+  
+  export async function createTestingContext(schemaStr: string, config?: { seedData: { [collection: string]: any[] } }): Promise<Context> {
```
machi1990, line 14: I remember seeing this function somewhere, it'll probably be useful to extract this somewhere

machi1990: Not to be done in this PR of course.

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
machi1990: > I think this has been brought up up before, with conclusion that updatedAt can store this timestamp, @wtrocki what do you think?

Ah sorry, I might have missed the conversation.

ssd71:
I have made some rather significant changes, we now have DataSyncCRUDService and some other minor fixes. I still had to use a hack to add the resolvers:
https://github.com/aerogear/graphback/blob/f8fbd3cfcf55b0cb2f4492d2768e6a016b4f13a2/packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts#L90-L103

wtrocki:
I would not call it a hack - it should be this way.

@craicoverflow Any blockers to get that merged?

wtrocki:
Works from general DataSync standpoint.
I will look for approval of the lead before merge

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
ssd71: At this point there is no way to `update` a soft deleted object that I know of,  There was a bug where it could be soft deleted multiple times but I have fixed that now :) Feel free to point out if you see any bugs 🐞

wtrocki: Hm.. deletion of deleted item should throw error. edits are imposible

machi1990: Yeap. Edits and I think deletion was possible from the previous patch (before latest fixes) hence my comment.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -50,4 +43,68 @@ export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
+    data = {};
+
+    this.fieldTransformMap[TransformType.UPDATE]
+      .forEach((f: FieldTransform) => {
+        data[f.fieldName] = f.transform();
+      });
+
+    data._deleted = true;
+
+    const result = await this.db.collection(this.collectionName).updateOne({ _id: new ObjectId(idField.value), _deleted: false }, { $set: data });
```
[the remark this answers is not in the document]
ssd71: the `update` method needs an `updatedAt` in the data argument for conflict detection, for this we would need to call findOne to get updatedAt and then pass it to the update method

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -50,4 +43,68 @@ export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
+    if (result.result.nModified === 1) {
+      const queryResult = await this.db.collection(this.collectionName).find({ _id: new ObjectId(idField.value) }).toArray();
```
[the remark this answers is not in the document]
ssd71: the `findOne` method will miss any soft deleted items, since we are fetching the item that we just soft deleted, I don't think we can use findOne here

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -50,4 +43,68 @@ export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
+    }
+    throw new NoDataError(`Cannot find a result for ${this.collectionName} with filter: ${JSON.stringify(filter)}`);
+  }
+
+  public async findBy(filter?: any, orderBy?: GraphbackOrderBy, page?: GraphbackPage): Promise<Type[]> {
+    if (filter === undefined) {
+      filter = {};
+    }
+
+    filter._deleted = false
```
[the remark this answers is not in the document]
ssd71:
```typescript
filter._deleted = false
```
 is just shorthand for that 😄

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -50,4 +43,68 @@ export class DatasyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
+  }
+
+  public sync(lastSync: string): Promise<Type[]> {
+    const filter = {
+      updatedAt: {
+        gt: lastSync
+      },
+    };
+
+    return super.findBy(filter, undefined, undefined);
```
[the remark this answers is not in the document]
ssd71: Ah yes! Thanks! I will put that down

craicoverflow:
Oh by the way don't worry about the plugin not loading - with https://github.com/aerogear/graphback/pull/1324 getting merged we can do code-based not config based plugins so it will work then :)

wtrocki:
@craicoverflow I'm really sorry for merging. It was a misunderstanding. Somehow got mixed with @machi1990 approval.

craicoverflow:
> @craicoverflow I'm really sorry for merging. It was a misunderstanding. Somehow got mixed with @machi1990 approval.

No problem :smile:

ssd71:
> Oh by the way don't worry about the plugin not loading - with #1324 getting merged we can do code-based not config based plugins so it will work then :)

I was actually doing '@graphback/datasync' to include it for testing, and `private: true` is so we don't end up releasing the package accidentally (rather cynical of me) , but I am not sure if we should remove `private: true` in current state 😄

craicoverflow:
> but I am not sure if we should remove `private: true` in current state

Ah fair enough :)

Issue #1322, opened 2020-05-28
Rename db.primary to db.id for better visibility.


wtrocki:

craicoverflow:
+1

Or just """@id"""

machi1990:
Should we keep `db.primary` around and deprecate it to avoid breaking change?

craicoverflow:
Yeah I guess so :)

wtrocki:
Trick is to do as many braking changes as.possible until 1.0 so after that it will.be not needed and code will be easy to maintain

machi1990:
So we can safely rename and add a breaking change note in the release notes?

wtrocki:
Yep. It is couple min fix on user side anyway

wtrocki:
0.14.0 is uber breaking so we can break as much we want

machi1990:
Sure, I'll make it `@db.id`. So that it is under the `@db` namespace like the others.

PR #1396, opened 2020-06-08
feat(core): rename `@db.primary` to `@id` for better visibility

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1322

### DIFF
--- a/docs/intro/datamodel.md
+++ b/docs/intro/datamodel.md
@@ -36,7 +36,7 @@ type Note {
 }
 ```
 
-You can set a custom primary key field using the `@db.primary` field annotation.
+You can set a custom primary key field using the `@id` field annotation.
 
 ```graphql
 """
@@ -45,7 +45,7 @@ You can set a custom primary key field using the `@db.primary` field annotation.
 type Note {
   id: ID!
   """
-  @db.primary
+  @id
   """
   email: String
 }
--- a/packages/graphback-core/src/db/getPrimaryKey.ts
+++ b/packages/graphback-core/src/db/getPrimaryKey.ts
@@ -1,9 +1,9 @@
 import { GraphQLField, GraphQLObjectType, getNamedType } from "graphql";
-import { parseDbAnnotations } from '../annotations/parser';
+import { parseMetadata } from 'graphql-metadata';
 
 /**
  * Returns the primary key field of a GraphQL object.
- * First looks for the existence of a `@db.primary` field annotation,
+ * First looks for the existence of a `@id` field annotation,
  * otherwise tries to find an `id: ID` field.
  * 
  * @param graphqlType 
@@ -14,10 +14,10 @@ export function getPrimaryKey(graphqlType: GraphQLObjectType): GraphQLField<any,
   let primaryKey: GraphQLField<any, any>;
   let primariesCount = 0;
   for (const field of fields) {
-    const dbConfig: any = parseDbAnnotations(field);
+    const hasIdMarker = parseMetadata("id", field);
     const baseType = getNamedType(field.type);
 
-    if (dbConfig.primary) {
+    if (hasIdMarker) {
       primaryKey = field;
       primariesCount += 1;
     } else if (field.name === 'id' && baseType.name === 'ID') {
@@ -26,7 +26,7 @@ export function getPrimaryKey(graphqlType: GraphQLObjectType): GraphQLField<any,
   }
   
   if (primariesCount > 1) {
-    throw new Error(`${graphqlType.name} type should not have multiple '@db.primary' annotations.`)
+    throw new Error(`${graphqlType.name} type should not have multiple '@id' annotations.`)
   }
 
   if (!primaryKey) {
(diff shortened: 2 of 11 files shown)


craicoverflow:
How does this affect graphql-migrations?

docs/intro/datamodel.md
```diff
@@ -45,7 +45,7 @@ You can set a custom primary key field using the `@db.primary` field annotation.
 type Note {
   id: ID!
   """
-  @db.primary
+  @db.id
```
craicoverflow, line 48: Thoughts on just `@id`?

I'd rather:

- `@unique` over `@db.unique`
- `@oneToMany` over `@db.oneToMany`

etc. where possible because `@db` does not add much information to above examples.

It comes from graphql-migrations and we could find more approiate namespaces like:

`@col(type: 'varchar', length: 200)` instead of `@db(type: 'varchar', length: 200)`

machi1990: I honestly thought that everything was under `@db` hence I went for consistency. Though I agree that `@db` does not add much value, there is probably a historical technical reason on top of it coming from graphql migrations which I would like to know :-)

machi1990: > How does this affect graphql-migrations?

Yeap, migrations was not handled. Good spot, I can add it too but let's decide ourselves first if we want `@db.id` or `@id`.

craicoverflow: graphql-migrations is a fork of [graphql-migrate](https://github.com/Akryum/graphql-migrate), which used `@db` everywhere.

My guess is because it was a tool for one thing - db migrations.

Graphback ecosystem uses these annotations in several context - for migrations, at runtime in CRUD services and data providers, and for schema generation.

machi1990: Thanks.

machi1990: So should we stick with `@db.id` or start the work of removing the `@db` prefix here? - meaning keeping this PR as simple as possible and do for `id` and followup PR for the rest of the remaining annotations.

craicoverflow: We can just do `@id` here, and do the other annotations in a separate sprint (we will need to migrate to a new format anyway, meaning a breaking change for these but not for `@id` as format wont change.

In fact you can use `parseMetadata` for `@id` now, it should accept it.

craicoverflow:
Also I would see this as more of a `feat` over `refactor` since it is a breaking change.

machi1990:
> Also I would see this as more of a `feat` over `refactor` since it is a breaking change.

Cool. Thanks for spotting that one, I was not sure how to qualify this one really.

packages/graphql-migrations/tests/specs/abstract-database.ts
```diff
@@ -155,6 +155,29 @@ test('default primary index', async () => {
   expect(id.columns).toEqual(['id'])
 })
 
+test('primary key with the @id annotation', async () => {
+  const schema = buildSchema(`
+      """
+      @model
+      """
+      type User {
+        id: ID!
```
machi1990, line 164: This is ugly probably because of https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts#L149, I think we should be able to just do 
```
"""
@model
"""
type ModelWithCustomPrimaryKey {
        """
        @id
        This will be the primary key
        """
        email: String!
}
```

machi1990: Thoughts @craicoverflow

machi1990: I think we should move the check after all the fields have been parsed. But wanted to get your thoughts first.

packages/graphback-core/src/db/getPrimaryKey.ts
```diff
@@ -14,10 +14,10 @@ export function getPrimaryKey(graphqlType: GraphQLObjectType): GraphQLField<any,
   let primaryKey: GraphQLField<any, any>;
   let primariesCount = 0;
   for (const field of fields) {
-    const dbConfig: any = parseDbAnnotations(field);
+    const hasIdMarker = parseMarker("id",String(field.description));
```
craicoverflow, line 17: This should work too:

```suggestion
    const hasIdMarker = parseMetadata("id", field);
```

packages/graphql-migrations/tests/specs/abstract-database.ts
craicoverflow: There is no reason to force an `id` field anymore because we map these at the resolver level.

Graphback works without an `id` field (if you are not using migrations), so there is no reason for graphql-migrations to restrict this anymore

This can be removed (whether you want to do it now or in another PR, doesn't matter :)
https://github.com/aerogear/graphback/blob/875f24585a534cedc7cf805d671c38592ea8fb53/packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts#L149

machi1990: Cool, let me put that one is separate commit.

machi1990:
@craicoverflow we should be good now.

machi1990:
> I checked the migrated database and it does not set the corresponding ID column.
> However I also tested this on master and it is an existing issue. So it deserves a separate issue and fix.
> 
> image

Sure, let's create another isssue and I'll have a look.

PR #1403, opened 2020-06-09
feat: new runtime API

craicoverflow:
BREAKING


wtrocki:
Tests are printing a looooot now. Do we print snapshots or something?

wtrocki:
Tests are printing a looooot now. Do we print snapshots or something?

wtrocki:
Ah.. Requires metadata lib bump :D

machi1990:
I am seeing test failure fr changes brought by https://github.com/aerogear/graphback/pull/1394 and related to constructor signature changes of the `CRUDService`.

craicoverflow:
> I am seeing test failure fr changes brought by #1394 and related to constructor signature changes of the `CRUDService`.

Yeah working on it, Shibani's PR did not have the new runtime API so how you create services is a lot different now.

packages/graphback-datasync/src/services/createDataSyncService.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const serviceConfig: CRUDServiceConfig = {
+      ...config,
+      crudOptions: model.crudOptions
+    }
+
+    if (isDataSyncModel(model)) {
+      return new DataSyncCRUDService(model.graphqlType.name, dataProvider, serviceConfig)
+    } else {
+      return new CRUDService(model.graphqlType.name, dataProvider, serviceConfig)
```
craicoverflow, line 30: DataSyncProvider CRUD methods differ quite a lot from standard MongoDBDataProvider, but there is no way to determine if this is Mongo/Postgres.

So we will still need to add serviceOverrides to the runtime API.

craicoverflow: cc @wtrocki @machi1990

wtrocki: This is bad.. DataSync that should be evaluating on top of graphback is leaking to graphback core. This will never have happened if we would work on separate repository.

wtrocki: I will work on this with @ssd71 to find best approach

PR #1408, opened 2020-06-09
fix: wrong path to service property

craicoverflow:
The path to the Datasync service was wrong (my bad).

this is not easily verifiable unless we add a data sync template.


machi1990:
> The path to the Datasync service was wrong (my bad).
> 
> this is not easily verifiable unless we add a data sync template.

Yeap. Or have some sort of integration test for it.

ssd71:
+1 for template, I believe integration test is also needed, though separate issues

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -89,7 +89,7 @@ export class LayeredRuntimeResolverCreator {
         const deltaQuery = getDeltaQuery(resolverElement.graphqlType.name)
 
         resolvers.Query[deltaQuery] = async (parent: any, args: any, context: any) => {
```
machi1990, line 91: I am thinking, would having a "full type context object" helped here?.

craicoverflow: probably yep, I will add it :)

PR #1419, opened 2020-06-10
Remove sqlite from package dependencies

craicoverflow:
In some cases, sqlite3 was not needed at all. In other cases, it is fine to keep in devDependencies (for automaated tests)


craicoverflow:
This is still a draft, need to do some verifications.

Issue #1415, opened 2020-06-10
Heap out of memory error when running tests in my local machine


machi1990:
When running the test locally,  I am getting a heap out of memory error

```

 RUNS   graphql-serve  packages/graphql-serve/tests/GraphbackServer.test.ts

<--- Last few GCs --->

[100208:0x4b85480]    55571 ms: Mark-sweep 2031.0 (2064.1) -> 2025.9 (2052.6) MB, 463.3 / 0.1 ms  (+ 128.7 ms in 28 steps since start of marking, biggest step 5.2 ms, walltime since start of marking 606 ms) (average mu = 0.191, current mu = 0.096) allocat[100208:0x4b85480]    56193 ms: Mark-sweep 2030.6 (2052.6) -> 2027.4 (2051.8) MB, 610.8 / 0.1 ms  (average mu = 0.112, current mu = 0.018) allocation failure scavenge might not succeed


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x13cb519]
Security context: 0x08039c9008d1 <JSObject>
    1: new Map(aka Map) [0x8039c90a1f9](this=0x3deebcc00591 <the_hole>)
    2: ConstructFrame [pc: 0x134c7fa]
    3: createSymbolTable [0x18b36a0c4409] [/home/manyanda/Documents/Projects/aerogear/graphback/node_modules/typescript/lib/typescript.js:~12970] [pc=0x3082f17874e4](this=0x07d449cfcf79 <Object map = 0x3cf48f738d9>,0x3deebcc004b1 <undefined>)
    4: decla...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xa07f90 node::Abort() [/usr/local/bin/node]
 2: 0xa0839c node::OnFatalError(char const*, char const*) [/usr/local/bin/node]
 3: 0xb80d9e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 4: 0xb81119 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: 0xd2d875  [/usr/local/bin/node]
 6: 0xd2df06 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [/usr/local/bin/node]
[...]

PR #1420, opened 2020-06-10
fix: heap out of memory error

craicoverflow:
See #1415 

This PR was to fix an out of memory error in CI. The initial fix in https://github.com/aerogear/graphback/commit/4a16de7ea20accb8daf0b4922d95e862ab999770 increased the memory available to the CI but we wanted to see if we could fix this locally too without having to increase memory.

I am now seeing errors in the Mongo tests - it is related to thew tests timing out waiting for MongoMemoryServer to download.

I have always seen this error locally, but it was never an issue on CI until now. @wtrocki @ssd71 did not experience this as far as I am aware.

I tried resolving it by ensuring the MongoMemoryServer only gets downloaded once per test (see https://github.com/aerogear/graphback/pull/1420/commits/6f8556cc211fc103b1c6db109d91c4b4bcac798e), because it was being re-downloaded in every test.

Help/ideas appreciated @ssd71 @machi1990 @wtrocki


machi1990:
@craicoverflow amazing job.I'lll have to try this in my local machine. 

Look like there are some conflicts, can you do a rebase?

craicoverflow:
> @craicoverflow amazing job.I'lll have to try this in my local machine.
> 
> Look like there are some conflicts, can you do a rebase?

Done.

packages/graphback-cli/package.json
```diff
@@ -35,7 +35,7 @@
     "lint-staged": "10.2.9",
     "ncp": "2.0.0",
     "rimraf": "3.0.2",
-    "ts-jest": "25.5.1",
+    "ts-jest": "26.0.1",
```
machi1990, line 38: Looks like `26.0.1` is not there, should be `26.1.0`. 

I had to manually pick the version locally after yarn proposed it to me. 

Also CI is failing for similar reason 
```
Error: Couldn't find any versions for "ts-jest" that matches "26.0.1"
 
    at MessageError.ExtendableBuiltin (/opt/yarn-v1.22.4/lib/cli.js:721:66)
 
    at new MessageError (/opt/yarn-v1.22.4/lib/cli.js:750:123)
 
    at Function.<anonymous> (/opt/yarn-v1.22.4/lib/cli.js:50545:13)
 
    at step (/opt/yarn-v1.22.4/lib/cli.js:310:30)
 
    at /opt/yarn-v1.22.4/lib/cli.js:321:13
 
    at runMicrotasks (<anonymous>)
 
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
 
``

machi1990:
Verified this locally a couple of times and did not get the error.

ssd71:
> I am now seeing errors in the Mongo tests - it is related to thew tests timing out waiting for MongoMemoryServer to download.
> 
> I have always seen this error locally, but it was never an issue on CI until now. @wtrocki @ssd71 did not experience this as far as I am aware.
> 
> I tried resolving it by ensuring the MongoMemoryServer only gets downloaded once per test (see [6f8556c](https://github.com/aerogear/graphback/commit/6f8556cc211fc103b1c6db109d91c4b4bcac798e)), because it was being re-downloaded in every test.

I actually did run into this in the past week, but wasn't sure if it was repoducible or not, just assumed it was bandwidth problem. Since then I have been using the jest extension for vscode so haven't run into this after that.

Issue #1397, opened 2020-06-08
graphql-migrations does not set primary fields


machi1990:
- https://github.com/aerogear/graphback/pull/1396#pullrequestreview-426343623 for context. 

/cc @craicoverflow opening an issue to drive the investigation as suggested on the linked PR.

machi1990:
I have ran the migration we have in the postgres template https://github.com/aerogear/graphback/blob/master/templates/ts-apollo-postgres-backend/src/index.ts#L32 

I could see that the primary key index is there
```log
users-# \d
                List of relations
 Schema |      Name      |   Type   |   Owner    
--------+----------------+----------+------------
 public | comment        | table    | postgresql
 public | comment_id_seq | sequence | postgresql
 public | note           | table    | postgresql
 public | note_id_seq    | sequence | postgresql
(4 rows)

users-# \d comment
                                    Table "public.comment"
    Column     |          Type          |                      Modifiers                       
---------------+------------------------+------------------------------------------------------
 text          | character varying(255) | 
 description   | character varying(255) | not null
 title         | character varying(255) | not null
 noteCommentId | integer                | 
 id            | integer                | not null default nextval('comment_id_seq'::regclass)
Indexes:
    "comment_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "comment_notecommentid_foreign" FOREIGN KEY ("noteCommentId") REFERENCES note(id)

users-# \d note
                                   Table "public.note"
   Column    |          Type          |                     Modifiers                     
-------------+------------------------+---------------------------------------------------
 title       | character varying(255) | not null
 description | character varying(255) | not null
 id          | integer                | not null default nextval('note_id_seq'::regclass)
Indexes:
    "note_pkey" PRIMARY KEY, btree (id)
Referenced by:
    TABLE "comment" CONSTRAINT "comment_notecommentid_foreign" FOREIGN KEY ("noteCommentId") REFERENCES note(id)
```

How did you reproduce it? @craicoverflow

machi1990:
Closing this for now as it was a false issue.

craicoverflow:
I should have been more specific in my original comment - it will not work if you try to set a custom primary key with the `@id` annotation, but works on the standard id field.

For this this model

```gql
"""
@model
"""
type User {
  """
  @id
  """
  email: String
}
```

Produces:

```sh
users=# \d user
             Table "public.user"
 Column |          Type          | Modifiers 
--------+------------------------+-----------
[...]

Issue #936, opened 2020-03-18
Implicit relationships


lastmjs:
Would it be possible to implicitly identify relationships, without the extra `one-to-many` directive? It is a lot of boilerplate and confusing having to write that in.

Instead of this:

```graphql
"""
@model
"""
type Note {
  id: ID!
  title: String!
  """
  @oneToMany field: 'note'
  """
  comments: [Comment]
}
```

Why not this?

```graphql
"""
@model
"""
type Note {
  id: ID!
  title: String!
  comments: [Comment]
}

type Comment {
  id: ID!
  note: Note
}
```
Prisma is able to identify this in most cases if I'm not mistaken.

craicoverflow:
Hey @lastmjs !

Yes this would definitely be possible and it is something we want to implement soon.

We released this relationship implementation with one path/one way of doing it initially so that we could receive feedback on the general approach before refining it.

It is definitely on our backlog.

lastmjs:
Excellent news! Keep up the good work

lastmjs:
Is this still on the backlog? I would love to simplify and remove the relationship annotations wherever possible. The most confusing thing right now is the many-to-many relationships, and requiring an explicit join table in the schema. Prisma did this very well, essentially allowing you to create a many-to-many relationship with two array types, one on each object type. Dealing with the join table has led to unnecessary confusion on my team

wtrocki:
Added to milestone.

PR #1421, opened 2020-06-10
fix(graphql-migrations): set primary keys when a column has `@id`

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1397

Opening as draft to get early feedback but from what I am seeing the `@id` or its predecessor `@db.primary` never really worked before. 

Here is what  I mean, apart from the primary field not being set, which should be fixed by this patch, another issue is when you have a relation like the one we have in https://github.com/aerogear/graphback/blob/24894c385f1ad846312c9382dae5468f98dc1f73/templates/ts-apollo-postgres-backend/model/datamodel.graphql where the migration will fail with the below error:
```
Foreign field id on type Note not found on field note.
Foreign field id on type Note not found on field note.
(node:136301) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'name' of undefined
    at /home/manyanda/Documents/Projects/aerogear/graphback/packages/graphql-migrations/dist/abstract/generateAbstractDatabase.js:389:80
    at Array.find (<anonymous>)
    at AbstractDatabaseBuilder.hasColumn (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphql-migrations/dist/abstract/generateAbstractDatabase.js:389:38)
    at AbstractDatabaseBuilder.createOneToManyRelationship (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphql-migrations/dist/abstract/generateAbstractDatabase.js:368:19)
    at AbstractDatabaseBuilder.build (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphql-migrations/dist/abstract/generateAbstractDatabase.js:71:18)
    at Object.<anonymous> (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphql-migrations/dist/abstract/generateAbstractDatabase.js:40:24)

```

as it still expects the `id` to be a primary field. I'd consider that a separate issue and I can investigate further once we are okay with this patch. 

/cc @craicoverflow


machi1990:
> ok. so you have it figured out and fix seems to addressing main challenges. Lets have great evening and relax. I will verify this for you tommorow

Yes, this is what I have when I ran the migration with for the model below:

## model
```
""" @model """
type Note {
  """
  @id
  """
  title: String!
  """
  @db.foreign: 'title'
  @oneToMany field: 'note'
  """
  comments: [Comment]!
}

""" @model """
type Comment {
  """
  @id
  """
  text: String
  description: String
}
```

## database
```
users=# \d note
             Table "public.note"
 Column |          Type          | Modifiers 
--------+------------------------+-----------
 title  | character varying(255) | not null
Indexes:
    "note_pkey" PRIMARY KEY, btree (title)
Referenced by:
    TABLE "comment" CONSTRAINT "comment_noteid_foreign" FOREIGN KEY ("noteId") REFERENCES note(title)

users=# \d comment
              Table "public.comment"
   Column    |          Type          | Modifiers 
-------------+------------------------+-----------
 text        | character varying(255) | not null
 description | character varying(255) | 
 noteId      | character varying(255) | 
Indexes:
    "comment_pkey" PRIMARY KEY, btree (text)
Foreign-key constraints:
    "comment_noteid_foreign" FOREIGN KEY ("noteId") REFERENCES note(title)
```

Now, going back to the PR description:

> Here is what I mean, apart from the primary field not being set, which should be fixed by this patch, another issue is when you have a relation like the one we have in https://github.com/aerogear/graphback/blob/24894c385f1ad846312c9382dae5468f98dc1f73/templates/ts-apollo-postgres-backend/model/datamodel.graphql where the migration will fail with the below error:
> 
> ```
> Foreign field id on type Note not found on field note.
> Foreign field id on type Note not found on field note.

The error described here is that we cannot automatically infer a primary key because we always seem to be defaulting to `id` - https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts#L245

So have to explicitly specify foreign key using the `@db.foreign` as can be seen from the model above, which is fine but I was thinking maybe we should go further (in a followup PR) and infer this stuff since we have all the info to do so.

packages/graphql-migrations/src/diff/computeDiff.ts
```diff
@@ -194,7 +194,7 @@ class Differ {
 
     // Columns
     for (const column of table.columns) {
-      if (column.name === 'id') {
+      if (column.isPrimaryKey && column.autoIncrementable) {
```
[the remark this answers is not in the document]
craicoverflow: https://github.com/machi1990/graphback/blob/845d890472666972da743a17686d2046ea4521a7/packages/graphql-migrations/src/connector/read.ts#L101

Apologies, I see it now: `ID` scalar maps to increments Knex type automatically. I just forgot how this was being handled.

craicoverflow:
I've done some quick verification checks. 

Graphback allows you to set `id: ID` and set a different custom ID field with `@id`. However graphql-migrations tries to set multiple PKs:

```gql
"""
@model
"""
type User {
  id: ID!
  """
  @id
  """
  email: String
}
```

The model above (which would be acceptable in Graphback, throws the following error in migrations:

```
UnhandledPromiseRejectionWarning: error: alter table "public"."user" add constraint "user_pkey" primary key ("email2") - multiple primary keys for table "user" are not allowed
```

Perhaps `graphql-migrations` should read `getPrimaryKey` from `@graphback/core`, it is already a dependency.

There appears to be a problem when trying to add new columns after first migration.

```gql
"""
@model
"""
type User {
  """
  @id
  """
  email2: String
  email3: String
}
```

The above model performs a migration as expected.

```
users=# \d user
             Table "public.user"
 Column |          Type          | Modifiers 
--------+------------------------+-----------
 email2 | character varying(255) | not null
 email3 | character varying(255) | 
Indexes:
    "user_pkey" PRIMARY KEY, btree (email2)
```

However an error occurs when I added a new regular field `email3: String` and tried to run another migration.

```
UnhandledPromiseRejectionWarning: error: alter table "public"."user" alter column "email2" drop not null - column "email2" is in a primary key
```

For some reason it tries to drop email2, even though I have not made any change to it.

machi1990:
> I've done some quick verification checks.
> 
> Graphback allows you to set `id: ID` and set a different custom ID field with `@id`. However graphql-migrations tries to set multiple PKs:
> 
> ```graphql
> """
> @model
> """
> type User {
>   id: ID!
>   """
>   @id
>   """
>   email: String
> }
> ```
> 
> The model above (which would be acceptable in Graphback, throws the following error in migrations:
> 
> ```
> UnhandledPromiseRejectionWarning: error: alter table "public"."user" add constraint "user_pkey" primary key ("email2") - multiple primary keys for table "user" are not allowed
> ```

> Perhaps `graphql-migrations` should read `getPrimaryKey` from `@graphback/core`, it is already a dependency.

Very good point. 

> There appears to be a problem when trying to add new columns after first migration.
> 
> ```graphql
> """
> @model
> """
> type User {
>   """
>   @id
>   """
>   email2: String
>   email3: String
> }
> ```
> 
> The above model performs a migration as expected.
> 
> ```
> users=# \d user
>              Table "public.user"
>  Column |          Type          | Modifiers 
> --------+------------------------+-----------
>  email2 | character varying(255) | not null
>  email3 | character varying(255) | 
> Indexes:
>     "user_pkey" PRIMARY KEY, btree (email2)
> ```
> 
> However an error occurs when I added a new regular field `email3: String` and tried to run another migration.
> 
> ```
> UnhandledPromiseRejectionWarning: error: alter table "public"."user" alter column "email2" drop not null - column "email2" is in a primary key
> ```
> 
> For some reason it tries to drop email2, even though I have not made any change to it.

Yeap, I have encountered the issue too. I suppose the change I have added in the computeDiff are not enough, I am looking at it now.

craicoverflow:
> So have to explicitly specify foreign key using the @db.foreign as can be seen from the model above, which is fine but I was thinking maybe we should go further (in a followup PR) and infer this stuff since we have all the info to do so.

Fully agreed. [Feedback](https://github.com/aerogear/graphback/issues/936) suggests fewer annotations needed the better (and rightly so)

machi1990:
> > So have to explicitly specify foreign key using the @db.foreign as can be seen from the model above, which is fine but I was thinking maybe we should go further (in a followup PR) and infer this stuff since we have all the info to do so.
> 
> Fully agreed. [Feedback](https://github.com/aerogear/graphback/issues/936) suggests fewer annotations needed the better (and rightly so)

Thanks for the linked issue for this. Totally agree.

packages/graphql-migrations/src/connector/read.ts
```diff
@@ -83,20 +82,28 @@ class Reader {
...
+          const info = columnInfo[key];
+          const {args, type} = getTypeAlias(info.type, info.maxLength);
+          const foreign = foreignKeys.find((k: any) => k.column === key);
+          const isPrimaryKey  = primaries.some((primary:  {column: string, indexName: string }) => primary.column === columnName);
           const column: TableColumn = {
             name: columnName,
+            isPrimaryKey,
+            autoIncrementable: type === "increments", // TODO
             comment: this.getComment(columnComments, key),
             annotations: {},
```
craicoverflow, line 103: So is `args and `type` moved out of annotations to a top-level property in `TableColumn` now?

machi1990: They were always top level property because of the destructing that happened in this line https://github.com/aerogear/graphback/pull/1421/files#diff-2038add530cf853b61f15d1ee2036174L98

I just extracted the info caused I needed it in line 101.

craicoverflow:
Checking this out locally now.

craicoverflow:
It appears that changing the field annotated with """@id""" after the first migration does not update the database primary key. Do you see this too @machi1990?

machi1990:
> It appears that changing the field annotated with """@id""" after the first migration does not update the database primary key. Do you see this too @machi1990?
> 
> ```
> error: alter table "public"."user" alter column "email" drop not null - column "email" is in a primary key
> ```

Ah strange as this should have been addressed by https://github.com/aerogear/graphback/pull/1421/files#diff-4972a24a7bb3dda74ca34232ed6243aeR304
 
> I don't know if this is fixable, as long as the error is thrown though it is probably fine. Perhaps we should add this as note to website documentation?

From the test I did, looks like the code I shared above fixed it Might have been a false positive. Let me keep on looking and if not fixable I can add a note in the documentation as suggested.

machi1990:
@craicoverflow have pushed new changes hoping to fix the `drop no null` issue on a previous primary key column. Can you re-review please?

packages/graphql-migrations/src/connector/read.ts
```diff
@@ -83,20 +82,28 @@ class Reader {
...
-          const info = columnInfo[key]
-          const foreign = foreignKeys.find((k: any) => k.column === key)
+          const info = columnInfo[key];
+          const {args, type} = getTypeAlias(info.type, info.maxLength);
+          const foreign = foreignKeys.find((k: any) => k.column === key);
+          const isPrimaryKey  = primaries.some((primary:  {column: string, indexName: string }) => primary.column === columnName);
           const column: TableColumn = {
             name: columnName,
+            isPrimaryKey,
+            autoIncrementable: type === "increments", // TODO
```
craicoverflow, line 101: Knex has bigIncrements for bigint, which could be mapped by the user in `migrateDB` to override the default mapping increments.

`` `suggestion
            autoIncrementable: type === "increments" || type === "bigIncrements", // TODO
```

machi1990: Good catch.

Issue #1076, opened 2020-04-08
Postgres 12 has issues (Initial migration only created the table, no fields)


lastmjs:
Here's my model:

```graphql
""" @model """
type User {
  id: ID!
  
  """ @db.unique """
  address: String!
  
  gender: GENDER!
  age: Int!
  
  """ @db.default: 0 """
  numFlags: Int!

  # pastSpeedDates: [User!]!
}

enum GENDER {
  MALE
  FEMALE
}
```

Here's my `.graphqlrc.yml`:

```yml
## GraphQL Config Generated by Graphback
## Please review configuration and adjust it for your own project
schema: ./backend/schema/*.graphql
extensions:
  graphback:
    model: ./backend # TODO I am not sure why this is a directory, shouldn't it point to a file?
    crud: # TODO I wonder if these should all just be enabled by default?
      create: true
      update: true
      findAll: true
      find: true
      delete: false
      subCreate: false
      subUpdate: false
      subDelete: false
    plugins:
      graphback-schema:
        format: graphql
        outputPath: ./backend/schema
      graphback-resolvers:
        format: ts
        outputPath: ./backend/resolvers
  dbmigrations:
    client: pg
    connection:
      host: ${SS_DB_HOST}
      port: ${SS_DB_PORT}
      database: ${SS_DB_NAME}
      user: ${SS_DB_USERNAME}
      password: ${SS_DB_PASSWORD}
```

I ran `graphback db`, and no errors were indicated. I looked in my database and the `user` table is there, along with the `user_id_seq`, but nothing else.

lastmjs:
Okay, so it looks like the fields were created, but I couldn't see them with my version of psql for some reason. I'm using Postgres from within a Docker container, but I was using psql on my local machine. I used psql from within the docker container and I can see the columns now.

But there is still a problem. I simplified my model to look like this:

```graphql
""" @model """
type User {
  id: ID!
  
  # """ @db.unique """
  # address: String!
  
  # gender: GENDER!
  # age: Int!
  
  # """ @db.default: 0 """
  # numFlags: Int!

  # pastSpeedDates: [User!]!
}

# enum GENDER {
#   MALE
#   FEMALE
# }
```

After deploying to the database, I can see that column. I then uncommented everything but `pastSpeedDates`:

```graphql
""" @model """
type User {
  id: ID!
  
  """ @db.unique """
  address: String!
  
  gender: GENDER!
  age: Int!
  
  """ @db.default: 0 """
[...]

PR #1425, opened 2020-06-11
fix: postgres 12 issue when running migrations multiple times

machi1990:
When checking constraints in postgres 12, the sql query belowed failed
with `column c.consrc does not exist`

```
select
    c.conname as "indexName",
    array_to_string(array_agg(a.attname), ',') as "columnNames",
    c.consrc as "expression"
  from
    pg_class t,
    pg_constraint c,
    pg_namespace n,
    pg_attribute a
  where
    c.conrelid = t.oid
    and n.oid = c.connamespace
    and a.attrelid = t.oid
    and a.attnum = ANY(c.conkey)
    and c.contype = 'c'
    and t.relname = ?
    and n.nspname = ?
  group by
    t.relname,
    c.conname,
    c.consrc;
```

This was because the `consrc` column was removed from the `pg_constraint` table starting from pg 12.

Starting the version 11, https://www.postgresql.org/docs/11/catalog-pg-constraint.html, it is recommended to use pg_get_constraintdef()
to extract the definition of a check constraint.

Fixes https://github.com/aerogear/graphback/issues/1076

### DIFF
--- a/packages/graphql-migrations/src/util/getCheckConstraints.ts
+++ b/packages/graphql-migrations/src/util/getCheckConstraints.ts
@@ -5,7 +5,7 @@ const queries: any = {
     sql: `select
     c.conname as "indexName",
     array_to_string(array_agg(a.attname), ',') as "columnNames",
-    c.consrc as "expression"
+    pg_get_constraintdef(c.oid) as "expression"
   from
     pg_class t,
     pg_constraint c,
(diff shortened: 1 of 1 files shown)


machi1990:
Opening as draft since I need to check if nothing else if affected with the upgrade.

machi1990:
> Awesome! I've verified this works with 12.3 and 9.6

Thanks for checking and review. Let's see what @wtrocki will say.

wtrocki:
Trusting enda on this. No time to dig deeper

Issue #1432, opened 2020-06-11
Can't add custom field with custom type to model type


lastmjs:
I am trying to add a custom field to one of my models. The custom field maps to a custom type, which is not a model. I'm going to implement the resolver myself, which won't have anything to do with my database. I keep getting this error when trying to do so:

```Generation failed: TypeError: Cannot read property 'name' of undefined```

For example:

```graphql
""" @model """
type Invoices {
  id: ID!
  stripe_invoice: StripeInvoice
}

type StripeInvoice {
  totalPaid: Int!
  totalCharged: Int!
}
```

The `Invoices` type is a model and has fields in the database, but the `stripe_invoice` field will have a custom resolver that actually hits the Stripe API. I can't seem to do this right now because of the error during generation.

wtrocki:
Hi

We are aware of the issue and going to resolve it as soon as possible in the next alpha.

machi1990:
I'll have a look.

wtrocki:
This has been addressed

wtrocki:
Going to land in alpha4

PR #1433, opened 2020-06-12
fix(codegen-schema): can't add field with custom type to model type

machi1990:
Adding a field with custom type was breaking since the custom type is neither a scalar , enum or a
model type so we cannot really generate the necessary crud bits for it. Let's ignore the field with
undefined type as it is breaking the CRUD bits generation part. `undefined` is returned by this
method
https://github.com/aerogear/graphback/blob/ae5c6d5148877fb7ce6fa622036dcd31ad6b9946/packages/graphback-core/src/crud/mappingHelpers.ts#L159-L173

Fixes https://github.com/aerogear/graphback/issues/1432

### DIFF
--- a/docs/intro/datamodel.md
+++ b/docs/intro/datamodel.md
@@ -51,6 +51,34 @@ type Note {
 }
 ```
 
+Your data models can also contain custom types that are not Graphback models as shown below. 
+However, it is expected that you provide the custom resolvers for your custom type as Graphback will not generate one for you.
+```graphql
+"""
+@model
+"""
+type Note {
+  id: ID!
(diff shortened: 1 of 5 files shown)


wtrocki:
Amazing work. Verifying

wtrocki:
Amazing work. Generally you can still add that to database but it is not breaking badly for people. We might doc this carefully for 0.14

machi1990:
Pushed a commit to add docs. Let me know what you think.

wtrocki:
Amazing. So many fixes for the new alpha release

PR #1441, opened 2020-06-15
Add ability to filter sync queries

ssd71:
This PR adds a filter argument to the sync query in the datasync package

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -61,7 +61,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
                 });
                 const filterTypeName = getInputTypeName(modelName, GraphbackOperationType.FIND);
                 schemaComposer.Query.addFieldArgs(deltaQuery, {
-                    lastSync: 'String!',
+                    lastSync: GraphQLNonNull(GraphQLString),
                     filter: filterTypeName
                 })
             }
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -60,7 +60,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
                     [deltaQuery]: `${getDeltaListType(modelName)}!`
                 });
                 schemaComposer.Query.addFieldArgs(deltaQuery, {
-                    lastSync: 'String!'
+                    lastSync: 'String!',
+                    filter: `${getInputTypeName(modelName, GraphbackOperationType.FIND)}`
                 })
             }
         })
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -59,9 +59,10 @@ export class DataSyncPlugin extends GraphbackPlugin {
                 schemaComposer.Query.addFields({
                     [deltaQuery]: `${getDeltaListType(modelName)}!`
                 });
+                const filterTypeName = getInputTypeName(modelName, GraphbackOperationType.FIND);
                 schemaComposer.Query.addFieldArgs(deltaQuery, {
                     lastSync: 'String!',
-                    filter: `${getInputTypeName(modelName, GraphbackOperationType.FIND)}`
+                    filter: filterTypeName
                 })
             }
         })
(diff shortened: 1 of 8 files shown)


wtrocki:
Checking

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -60,7 +60,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
                     [deltaQuery]: `${getDeltaListType(modelName)}!`
                 });
                 schemaComposer.Query.addFieldArgs(deltaQuery, {
-                    lastSync: 'String!'
+                    lastSync: 'String!',
+                    filter: `${getInputTypeName(modelName, GraphbackOperationType.FIND)}`
```
wtrocki, line 64: I will extract this to var

packages/graphback-datasync/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@graphback/core": "0.14.0-alpha3",
+    "@graphback/codegen-schema": "0.14.0-alpha3",
     "@graphback/runtime": "0.14.0-alpha3",
     "@graphback/runtime-knex": "0.14.0-alpha3",
```
wtrocki, line 43: Why knex and mongo here. I will see only mongo

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
```diff
@@ -108,4 +108,39 @@ describe('Soft deletion test', () => {
...
+    }
+
+    // Sync query
+    const deltaPosts = await Post.sync("0", {
+      text: {
+        startsWith: 'post'
+      }
+    });
+
+    expect(deltaPosts.length).toBeGreaterThanOrEqual(1);
```
machi1990, line 139: Should we change this assertion to:

```suggestion
    expect(deltaPosts.length).toEqual(2));
```

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
```diff
@@ -108,4 +108,39 @@ describe('Soft deletion test', () => {
...
+
+    const { Post }  = context.providers;
+
+    // Create some posts
+    for (const postTitle of ["post", "post2", "trains"]) {
+      await Post.create({ text: postTitle });
+    }
+
+    // Sync query
+    const deltaPosts = await Post.sync("0", {
```
wtrocki, line 133: I would recommend to do second sync with return timestamp and check if result is zero etc. (maybe separate tests)

packages/graphback-datasync/tests/DataSyncPluginTest.ts
```diff
@@ -1,6 +1,7 @@
 import { readFileSync } from 'fs';
 import { buildSchema, printSchema } from 'graphql';
 import { GraphbackCoreMetadata } from '@graphback/core';
+import { SchemaCRUDPlugin } from "@graphback/codegen-schema";
```
craicoverflow, line 4: If this is just used in tests (looks that way) this should be a devDependency

packages/graphback-datasync/package.json
```diff
@@ -38,8 +38,8 @@
   },
   "dependencies": {
     "@graphback/core": "0.14.0-alpha3",
+    "@graphback/codegen-schema": "0.14.0-alpha3",
```
craicoverflow, line 41: If this is just used in tests (looks that way) this should be a devDependency

packages/graphback-datasync/tests/DataSyncPluginTest.ts
ssd71: I think we actually need the schema plugin to do `@versioned` when using datasync

wtrocki: Yep. We can think of aggregating all 3 annotations into one as consolidation. WDYT? I can create an optional issue to deal with this.

wtrocki:
Looking for @ssd71 to address some comments

packages/graphback-datasync/tests/DataSyncPluginTest.ts
ssd71: I am onboard :+1:  with this!

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
ssd71: On it

packages/graphback-datasync/tests/DataSyncPluginTest.ts
craicoverflow: > I think we actually need the schema plugin to do @versioned when using datasync

Yes, but this happens independently of the datasync package in the plugin chain. The datasync package only uses it in test files, so it should be a devDependency

image

wtrocki:
Ah really sad that did not make it to alpha

Issue #637, opened 2020-01-16
Custom Database Column => GraphQL Field mapping solution


craicoverflow:
## Overview

Graphback gives capability to customise database column name during database generation with the `@db.name` annotation. See [Changing column names in graphback](https://graphback.dev/docs/database-schema-migrations#changing-column-names-in-graphback).

Graphback does not offer a simple way to map between a custom database column name and its GraphQL type at the resolver level, i.e. when performing queries or mutations. Currently the only way to do this is to manually map fields in each resolver like this:

```ts
// In your data provider
data.title = data['note_title'];

return data;
```

Which would be overwritten in any of the generated resolver functions during the next `graphql generate`. This also means that column customisation is completely unsupported in runtime applications.

We need a way to map between customised tables and the GraphQL schema.

## Benefits

- Users would be able to use an existing database with Graphback by manually adding the annotations on their data model fields.
- Greater customisation support.

### Option 1

Generate a mapping JSON object which can be consumed by the application at startup (by attaching to context) which can map between fields.

## Option 2

Use annotations from generated schema to build mapping and attach time runtime context.

Eg:

```ts
const context = createKnexRuntimeContext(db, pubSub, generatedSchema)

export const createKnexRuntimeContext = (db: Knex, pubSub: PubSubEngine): GraphbackRuntimeContext => {
  const crudDb = new PgKnexDBDataProvider(db);
  const crudService = new CRUDService(crudDb, pubSub);

 const fieldMapping = {
   "user": {
     "id" => "id",
     "name" => "user_name",
     "age" => "age"
  }

  return {
    crudService,
    crudDb,
    pubSub,
    fieldMapping
  };
}
```

This would then be passed to resolver which can transform the mapping dynamically to perform queries and mutations.

The benefit of this approach is that the generated schema is used as single source of truth. Having a mapping file could become outdate or the user would be able to modify it, breaking the mappings.

wtrocki:
From my point option nr 2 will be no brainer in the new architecture we have proposed. Generating json schema will bring some additional challenges and complexity.

lastmjs:
Just to make sure, will a custom schema directives be able to be used still? I want to be able to mark up my schema, even manually, to describe custom columns or table names or potentially even column types

wtrocki:
Yes. All will be possible by annotations only on schema. 
More info here: https://graphback.dev/docs/database-schema-migrations

craicoverflow:
## CRUD Level

[...]

PR #1450, opened 2020-06-15
feat: rename `@db.default` to `@default`

machi1990:


wtrocki:
-1 for picking random bugs. Still cool to have it done, but I would not consider it something to be done even for the 0.14.0 in the meant time there are dozens of issues that we need to address for the 0.14.0

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -1,3 +1,4 @@
+import {parseMetadata} from "graphql-metadata";
```
craicoverflow, line 1:
```suggestion
import { parseMetadata } from "graphql-metadata";
```

packages/graphql-migrations/src/abstract/generateAbstractDatabase.ts
```diff
@@ -48,6 +48,12 @@ const INDEX_UNIQUE_TYPES = [
   }
 ];
 
+const ANNOTATIONS = {
```
craicoverflow, line 51: :clap: good thinking.

docs/db/migrations.md
```diff
@@ -191,7 +191,7 @@ type Note {
   id: ID!
   title: String!
   """
-  @db.default: false
```
craicoverflow, line 194: Migrations are only for Postgres - since default values work for any data source there should also be a section added to [Data Model](https://graphback.dev/docs/intro/datamodel)

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -173,31 +174,40 @@ export const buildFilterInputType = (modelType: GraphQLObjectType) => {
...
-  const idField = getPrimaryKey(modelType)
+function buildMutationInputFields(modelType: GraphQLObjectType, operation: GraphbackOperationType) {
+  const inputTypeName = getInputTypeName(modelType.name, operation);
+  const idField = getPrimaryKey(modelType);
   const allModelFields = getModelInputFields(modelType);
+  const fields = () => allModelFields
+  .reduce((fieldObj: any, field: GraphQLInputField) => {
+    const { name, type } = field
+    let fieldType: GraphQLNamedType;
+    let defaultValue: any;
```
craicoverflow, line 192: We could start creating interfaces from here on for new annotations and eventually add ones for old annotations at some point. What you think?

```suggestion
    let defaultValue: DefaultMetadata;
```

Or...Is it overkill to create a class instead which extends a BaseAnnotation.

```
const default = new DefaultAnnotation(field);
default.getValue();
```

Maybe a separate less pressing task, but this could make our annotations really maintainable in future. All could be in `graphback-core/src/annotations`

docs/releases.md
```diff
@@ -67,6 +67,20 @@ For more advanced usage, refer to the [Runtime API](./intro/serverless) document
 
 We have replaced the `@db.primary` annotation with `@id`.
 
+#### `@db.default` is removed, use `@default` instead
+
+We have replaced the `@db.default` annotation with `@default`.
+Now you can provide default values like this:
+"""
```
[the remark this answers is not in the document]
machi1990: Good catch. Let me review the docs.

docs/db/migrations.md
machi1990: Thanks for the pointers.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: > We could start creating interfaces from here on for new annotations and eventually add ones for old annotations at some point. What you think?

Good idea.

> Or...Is it overkill to create a class instead which extends a BaseAnnotation.
> 
> ```
> const default = new DefaultAnnotation(field);
> default.getValue();
> ```
> 
> Maybe a separate less pressing task, but this could make our annotations really maintainable in future. All could be in `graphback-core/src/annotations`

+1 on making this a separate task. I'll add the `DefaultMetadata` first.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -173,31 +174,40 @@ export const buildFilterInputType = (modelType: GraphQLObjectType) => {
...
+export const buildMutationInputType = (modelType: GraphQLObjectType) => {
+  return buildMutationInputFields(modelType, GraphbackOperationType.UPDATE);
+}
 
-  const idField = getPrimaryKey(modelType)
+function buildMutationInputFields(modelType: GraphQLObjectType, operation: GraphbackOperationType) {
+  const inputTypeName = getInputTypeName(modelType.name, operation);
+  const idField = getPrimaryKey(modelType);
   const allModelFields = getModelInputFields(modelType);
+  const fields = () => allModelFields
```
wtrocki, line 188: Very complex. Begs for extraction to separate method

wtrocki:
How this will work for MongoDB? Would we do some map reduce to add default value to existing collections or try to inspect array of the results after query?

I truly believe that this belongs to the mapping feature that we have on roadmap for a while but decided to stage off. It is good change, but it might be worth revisiting this when we going to focus on mapping?

machi1990:
> How this will work for MongoDB? Would we do some map reduce to add default value to existing collections or try to inspect array of the results after query?

This is really independent of any database. What we do it not set the database on the db level but create a mutation input type with fields having a default value.
```
input CreateNoteInput {
  id: ID
  title: String = "text default value"
  description: String = "description default value"
}
```
 
Nothing needs to be done or existing collections.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: I can create a higher order function taking `allModelFields` as parameter and return a function   that returns `fields` (this function).

craicoverflow:
> I truly believe that this belongs to the mapping feature that we have on roadmap for a while but decided to stage off. It is good change, but it might be worth revisiting this when we going to focus on mapping?

Isn't it better on the GraphQL schema level? Default values are visible in the input types, types are automatically casted. Adding this to mapping introduces more complexity/validation/things that could go wrong.

@machi1990 I've approved, but let's wait a bit to see @wtrocki's thoughts on why we should not do this in schema.

wtrocki:
This requires discussion as it is really going to break for people.

Imagine scenario - I have stuff in db.
Adding new field with default and expect it to be returned from server all the time.

But old data has no value. Adding this what it looks to nice feature will break on production as Graphql will not allow to return data that is null for non null cases.

wtrocki:
Rule is simple - if something has possibility to blow out to our faces it will blow to our faces for sure. I see it now when trying to put simple demo. 

Adding default is very complex issue that involves possible data loss, unexpected behaviour and issues. There is no any functional limitations for this as people might add this in:

- Their own services
- Leaf resolvers when data is returned
- Datatabase

So unless there is some way to avoid issues mentioned above let's not merge this change. If we make it only PG specific (default on DB level) we do not have mentioned problems - so this will make sense and it will be easy to explain.

wtrocki:
I also think that we should stop adding new annotations that work only under some undocumented conditions until we have proper docs for them. `lastUpdated` am looking at you :D

machi1990:
Thanks for the discussion. 

We'll only change the `@db.default` annotation here to `@default(value)` and add the new annotation in the [metadata](https://github.com/aerogear/graphback/blob/1d5b7d02d81e1402160176e4461c158fd8b9ac74/docs/metadata.md) guide.

Discussions on https://github.com/aerogear/graphback/issues/637 will drive how we do default values independent of any database instead of having them written on schema.

craicoverflow:
Reverified :+1:

packages/graphback-core/src/annotations/DefaultMetadata.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * An interface that represents the parsed metadata from the `@default(value: <metadata-value>)` annotation.
+ *
+ */
+export interface DefaultMetadata {
```
wtrocki, line 5:
```suggestion
export interface DefaultMarkerMetadata {
```

docs/db/migrations.md
```diff
@@ -186,12 +186,14 @@ type User {
 
 #### Default field value
 
+You can specify a default value using the `@default` field annotation as shown below.
```
wtrocki, line 189:
```suggestion
You can specify a default value using the `@default` field marker as shown below.
```

docs/metadata.md
```diff
@@ -84,6 +84,27 @@ type Comment {
 }
 ```
 
+### `@default`
+
+The `@default` annotation is used to specify a default value. This is currently only supported for PostgreSQL database through the [graphql-migrations](db/migrations.md) package.
```
wtrocki, line 89:
```suggestion
The `@default` marker is used to specify a default value. This is currently only supported for PostgreSQL database through the [graphql-migrations](db/migrations.md) package.
```

packages/graphback-core/src/annotations/DefaultMetadata.ts
craicoverflow: > DefaultMarkerMetadata

Why?

docs/db/migrations.md
craicoverflow: We had a discussion yesterday about annotations versus markers and decided to use annotations.

packages/graphback-core/src/annotations/DefaultMetadata.ts
wtrocki: `DefaultMetadata` got me confused as `This is default metadata` instead of `DefaultMarkerMetadata`

wtrocki: Also we have `annotations` here
We use annotations, markers and directives in many places. 
We need to stick to single name :D

machi1990: How about `DefaultAnnotationMetadata` ?

craicoverflow: > DefaultValueAnnoation
> DefaultValueMetadata

docs/db/migrations.md
wtrocki: ah.. this is confusing as we deprecated `annotations` in the graphql-`metadata`

packages/graphback-core/src/annotations/DefaultMetadata.ts
craicoverflow: I prefer `DefaultValueAnnotation` - if we are going to use this term in docs we should do the same in code.

wtrocki: Anything will work - current name is too short and confusing

machi1990: `DefaultValueAnnotation` is much better.

docs/db/migrations.md
craicoverflow: We deprecated `parseAnnotations` - doesnt mean we should retire the use of "annotations" :)

packages/graphback-core/src/annotations/DefaultMetadata.ts
machi1990: done.

PR #1451, opened 2020-06-15
chore(lint): disable use no-use-before-define rule

machi1990:


.eslintrc.json
```diff
@@ -62,7 +62,7 @@
         "@typescript-eslint/no-unnecessary-qualifier": "error",
         "@typescript-eslint/no-unnecessary-type-arguments": "error",
         "@typescript-eslint/no-unnecessary-type-assertion": "error",
-        "@typescript-eslint/no-use-before-define": "error",
+        "@typescript-eslint/no-use-before-define": "off",
```
craicoverflow, line 65: Would it also work if you just remove the entry completely?

PR #1453, opened 2020-06-15
fix: add ability to build dynamic context api

wtrocki:
Allow us to configure how elements are exposed to context and add additional elements in future. Api will use function instead of the manually attaching changes.


packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -89,7 +94,9 @@ export function buildGraphbackAPI(model: string | GraphQLSchema, config: Graphba
   const serviceCreator = config.serviceCreator || createCRUDService({ pubSub: new PubSub() })
 
   const services = createServices(models, serviceCreator, config.dataProviderCreator)
-
+  const contextCreator = () => {
```
craicoverflow, line 97: Where is this getting overriden with the custom one from `GraphbackAPIConfig`?

wtrocki: Done!

templates/ts-apollo-mongodb-backend/src/index.ts
```diff
@@ -24,17 +24,14 @@ async function start() {
 
   const db = await connectDB()
 
-  const { typeDefs, resolvers, services } = buildGraphbackAPI(modelDefs, {
+  const { typeDefs, resolvers, contextCreator } = buildGraphbackAPI(modelDefs, {
```
craicoverflow, line 27: Looks good! Tested. Needs docs

wtrocki: Done

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -145,11 +145,11 @@ export class LayeredRuntimeResolverCreator {
       resolvers.Subscription[operation] = {
         //tslint:disable-next-line: no-any
         subscribe: (_: any, __: any, context: GraphbackContext) => {
-          if (!context.services || !context.services[modelName]) {
+          if (!context.graphback || !context.graphback[modelName]) {
             throw new Error(`Missing service for ${modelName}`);
           }
 
-          return context.services[modelName].subscribeToUpdate({}, context);
+          return context.graphback[modelName].subscribeToUpdate({}, context);
```
wtrocki, line 152: exposing as graphback to avoid conflicts + make it easy to find all related items.

Issue #1418, opened 2020-06-10
Move resolver creation to plugins


craicoverflow:
Original discussion is in https://github.com/aerogear/graphback/issues/1411

`LayeredRuntimeResolverCreator` creates the standars CRUD resolvers but is in no way scalable or extendable.

It makes sense to move resolver generation to the existing plugin system as it is easy for developers to extend Graphback in one place.

craicoverflow:
We can extend [`GraphbackPlugin`](https://github.com/aerogear/graphback/blob/9df2aa574299db464b7f77293bb493f1f799c522/packages/graphback-core/src/plugin/GraphbackPlugin.ts#L15) easily without making major changes to the plugin architecture.

The resolvers array can be a new property of `GraphbackCoreMetadata`.

```ts
export abstract class GraphbackPlugin {
    public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
        return metadata.getSchema();
    }

    public createResolvers(metadata: GraphbackCoreMetadata): GraphbackResolvers {
      const resolvers: GraphbackResolvers = {
        Query: {
          hello: () => {
            return 'Hello World!'
          }
        }
      }

      return resolvers
    }
}
```

I think we should be opinionated with resolver format in plugins (choosing Apollo format) to prevent different plugins creating resolvers in multiple incompatible* formats.

I don't know if they are incompatible, what happens if you try to merge x and y resolvers, x being the Apollo format and y being in the GraphQL.js format?

wtrocki:
hmm.. we should use graphql-tools everywhere - GraphQL-tools for schema creation will resolve format problem and document that to create schema you need to use graphql-tools. 

Also - we should pass resolvers array to the method rather than create it itself. 
The first plugin will get empty array.

wtrocki:
I remember having discussion on this already. Anyway.. This is very pressing matter and worth to do it (considering it will be trivial copy paste change)

craicoverflow:
> Also - we should pass resolvers array to the method rather than create it itself.
The first plugin will get empty array.

I thought about this, but above approach would mean the plugin would have to do minimal amount (create its resolvers). Plugin engine would then add it to array.

machi1990:
> Also - we should pass resolvers array to the method rather than create it itself.
> The first plugin will get empty array.

[...]

Issue #1448, opened 2020-06-15
`printSchemaHandler` still uses GraphbackRuntime


craicoverflow:
[printSchemaHandler](https://github.com/aerogear/graphback/blob/18b80864290803c7d86736c36e8626de60cbddbd/packages/graphql-serve/src/components/printSchemaHandler.ts#L12) from graphql-serve needs to be updated to use the new runtime API, then we can remove GraphbackRuntime.

When GraphbackRuntime is removed, `LayeredRuntimeResolverCreator` and any related unused files should be removed too.

craicoverflow:
graphql-serve can dynamically load Graphback plugins from the `.graphqlrc` file, whereas our new API uses code-loaded plugins.

These approaches are still compatible, however I know that `graphback-datasync` plugin will not work loaded from a config file because Graphback rewrites `graphback` prefixed plugins to `graphback-codegen-x`.

All we need to do is remove `codegen` from the renaming in [loadPlugins#L10](https://github.com/aerogear/graphback/blob/21d8cbf02efd6b72cee51d10565534f605de168e/packages/graphback/src/loadPlugins.ts#L10) and it should work.

wtrocki:
Nice!

machi1990:
> graphql-serve can dynamically load Graphback plugins from the `.graphqlrc` file, whereas our new API uses code-loaded plugins.
> 
> These approaches are still compatible, however I know that `graphback-datasync` plugin will not work loaded from a config file because Graphback rewrites `graphback` prefixed plugins to `graphback-codegen-x`.
> 
> All we need to do is remove `codegen` from the renaming in [loadPlugins#L10](https://github.com/aerogear/graphback/blob/21d8cbf02efd6b72cee51d10565534f605de168e/packages/graphback/src/loadPlugins.ts#L10) and it should work.

If we drop `codegen` from the renaming, it will mean having to change configurations like this https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/.graphqlrc.yml#L17 to add the codegen?

craicoverflow:
> If we drop `codegen` from the renaming, it will mean having to change configurations like this [aerogear/OpenVolunteerPlatform:platform/server/.graphqlrc.yml@`master`#L17](https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/.graphqlrc.yml?rgh-link-date=2020-06-18T12%3A57%3A08Z#L17) to add the codegen?

[...]

PR #1455, opened 2020-06-15
Create resolvers in SchemaCRUDPlugin

craicoverflow:
Closes https://github.com/aerogear/graphback/issues/1418

This PR does two things:

1. Makes some changes to the plugin engine to allow creation of resolvers (separate from the schema) https://github.com/aerogear/graphback/commit/89ea898d13261f1ac3752188c4edde3562766de4
2. Add resolvers from `@graphback/codegen-schema` plugin

## Tasks

- [x] Add ability to add resolvers in plugins https://github.com/aerogear/graphback/commit/89ea898d13261f1ac3752188c4edde3562766de4
- [x] Add a CRUD resolver creation to plugin
- [x] Add tests (minimal, ported from old resolver creator package)
- [x] Documentation


packages/graphback-resolvers/README.md
```diff
@@ -0,0 +1,20 @@
+## Graphback Codegen Schema
```
craicoverflow, line 1: @craicoverflow TODO: Change README

packages/graphback-resolvers/src/ResolversCRUDPlugin.ts
```diff
@@ -0,0 +1,199 @@
...
+ *    Query: {...}
+ *    Mutation: {...}
+ *    Subscription: {...}
+ * }
+ * ```
+ */
+export class ResolversCRUDPlugin extends GraphbackPlugin {
+
+  public addResolvers(metadata: GraphbackCoreMetadata): IResolvers {
+    const resolvers: IResolvers = {
```
craicoverflow, line 23: This is simply a port of [LayeredRuntimeResolverCreator](https://github.com/aerogear/graphback/blob/61b24b7b9b83bf361bdacb754d5a32efeea4833f/packages/graphback-runtime/tests/LayeredRuntimeResolverCreatorTest.ts#L25)

packages/graphback/src/buildGraphbackAPI.ts
```diff
@@ -73,6 +75,7 @@ function createServices(models: ModelDefinition[], createService: Function, crea
 export function buildGraphbackAPI(model: string | GraphQLSchema, config: GraphbackAPIConfig): GraphbackAPI {
   const schemaPlugins: GraphbackPlugin[] = [
     new SchemaCRUDPlugin,
+    new ResolversCRUDPlugin,
```
craicoverflow, line 78: Actually don't think this is a good idea, because it means there is no way to NOT use Graphback resolvers, but I like the idea of the user not having to import the resolver plugin

wtrocki:
hmm.. I would see this as just part of the schema plugin. 

It is strange to have 2 plugins that will be relying on each other?
WDYT?

wtrocki:
I do not have actual comment on implementation as it is something we widely agreed and it will really make difference, but feeling that separate package is widely redundant and gives not much benefits overall having this code in schema pluggin. If we make change in schema plugin we will always need to change resolve r plugin

craicoverflow:
> hmm.. I would see this as just part of the schema plugin.

I did not want to two be coupled together. SchemaCRUDPlugin gives a solid foundation for the rest of the plugins to use as an API reference. 

Someone could come along and add their own resolver plugin that is implemented slightly differently but complies with the schema. If we add resolvers in SchemaCRUDPlugin resolvers would essentially be generated twice (okay, the latter created resolvers would override the originals, but it would still be a startup performance issue).

That is the main reason I went with separate plugin.

>  If we make a change in schema plugin we will always need to change resolver plugin

The same amount of things would be changed, just in a different plugin file.

wtrocki:
> I did not want to two be coupled together. SchemaCRUDPlugin gives a solid foundation for the rest of the plugins to use as an API reference.

Yep. Ultimate flexibility :) Sorry I forgot we still doing it :)
To be serious I think this is a valid point. Let's do it.

wtrocki:
Would this plugin be default as well or require to be passed explicitly. 
If it is default it is not much value to have it as you will need to override it in your own plugin which you will be able to do it anyway. 

Not convinced separate package is worth it, but also no strong opinion and as lead you have final voice.

craicoverflow:
> Would this plugin be default as well or require to be passed explicitly.
> If it is default it is not much value to have it as you will need to override it in your own plugin which you will be able to do it anyway.
> 
> Not convinced separate package is worth it, but also no strong opinion and as lead you have final voice.

Actually you are right - at present it is a default plugin, so it is not able to be swapped anyway. If we ever do need to separate them again in the future it won't break anything, so I think we can keep it in SchemaCRUDPlugin.

craicoverflow:
After some more thoughts, I will keep this in a ResolverPlugin. I will separate each rsolver creation into its own method so anyone can extend the plugin to easily override a single resolver if they need to.

wtrocki:
Resolvers are functions. Anyone can extend them anyway. Do we provide ability to extend existing resolvers by additional function?

craicoverflow:
> Resolvers are functions. Anyone can extend them anyway.

Yep, but having a separate plugin means people can extend it easily to customise their resolvers without having to create a new plugin from scratch. If we conceal resolver creation in one or two methods it will be reeally hard to extend because you would have to rewrite so much extra code. So we should separate individual resolvers create methods.

This is the one of the biggest benefits of having plugins and it would be as shame if they are hard to extend/use.

wtrocki:
Yes, but what we doing here is already possible (users get resolvers object that can extend in their apps and datasync starter will have this. They do not need to create plugins. PLugins are created by us for datasync etc. and they will still be created no matter what. We also have service abstraction that give people ability to control the entity level behaviour. Bringing another thing is making it hard to document and not really valuable IMHO considering range of options available for developers that are much natural to work with

wtrocki:
To be clear there is actually one problem that we do not resolve - strong typing of the context in people's resolvers, but I consider that something we can address by actually code generation that were left there later rather than plugin itself.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -78,6 +79,36 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     return schemaComposer.buildSchema()
   }
 
+  /**
+   * Creates CRUD resolvers
+   *
+   * @param {GraphbackCoreMetadata} metadata - Core metatata containing all model information
+   */
+  public addResolvers(metadata: GraphbackCoreMetadata): IResolvers {
```
machi1990, line 87: The method name is misleading. When I read `addResolvers(metadata)`, it appears like we are adding resolvers to metadata but in fact what we are doing here is creating resolvers from metadata. Should this be renamed to `createResolvers` ?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
           });
         }
       }
-
     });
   }
 
+  /**
+   * Create resolvers function that
```
machi1990, line 338: incomplete sentence

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+   *
+   * @param inputContext
+   * @param options
```
machi1990, line 341: Seems like this comment is not accurate

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    throw Error("Invalid format specified. `options.format` supports only `ts`, `js` and `graphql` flags");
+  }
+
+  /**
+   * Create Query resolver fields
+   *
+   * @param {ModelDefinition} model - The model definition with CRUD config and GraphQL typr
+   * @param {IFieldResolver} queryObj - Query resolver object
+   */
+  protected addQueryResolvers(model: ModelDefinition, queryObj: any) {
```
machi1990, line 365:
```suggestion
  protected addQueryResolvers(model: ModelDefinition, queryObj: IFieldResolver) {
```

Not sure how accurate the typing is just copied it from method comments. So it will need verification

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    }
+  }
+
+  /**
+   * Create Mutation resolver fields
+   *
+   * @param {ModelDefinition} model - The model definition with CRUD config and GraphQL typr
+   * @param {IFieldResolver} mutationObj - Mutation resolver object
+   */
+  protected addMutationResolvers(model: ModelDefinition, mutationObj: any) {
```
machi1990, line 382:
```suggestion
  protected addMutationResolvers(model: ModelDefinition, mutationObj: IFieldResolver) {
```

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    }
+  }
+
+  /**
+   * Create Subscription resolver fields
+   *
+   * @param {ModelDefinition} model - The model definition with CRUD config and GraphQL typr
+   * @param {IFieldResolver} subscriptionObj - Subscription resolver object
+   */
+  protected addSubscriptionResolvers(model: ModelDefinition, subscriptionObj: any) {
```
machi1990, line 402:
```suggestion
  protected addSubscriptionResolvers(model: ModelDefinition, subscriptionObj: IFieldResolver) {
```

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+  }
+
+  /**
+   * Create relationship resolver fields
+   *
+   * @param {FieldRelationshipMetadata[]} relationships - Model relationship metadata
+   * @param {IResolvers} resolverObj - Resolvers object
+   */
+  protected addRelationshipResolvers(relationships: FieldRelationshipMetadata[], resolverObj: IResolvers) {
+    if (!relationships.length) { return undefined }
```
machi1990, line 423: Not needed

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    }
+  }
+
+  /**
+   * Creates a Create mutation resolver field
+   *
+   * @param {GraphQLObjectType} modelType - Model GraphQL object type
+   * @param {IFieldResolver} mutationObj - Mutation resolver object
+   */
+  protected addCreateMutationResolver(modelType: GraphQLObjectType, mutationObj: any) {
```
machi1990, line 440:
```suggestion
  protected addCreateMutationResolver(modelType: GraphQLObjectType, mutationObj: IFieldResolver) {
```

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: Also, it has to be applied throughout

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    const modelName = modelType.name;
+    const resolverCreateField = getFieldName(modelName, GraphbackOperationType.CREATE);
+
+    mutationObj[resolverCreateField] = (_: any, args: any, context: any) => {
+      if (!context.services || !context.services[modelName]) {
```
machi1990, line 445: Do we still have `services` in context? IIRC we change that in https://github.com/aerogear/graphback/pull/1453/files#diff-1dac687b85bfc24533f0b1a094ad644cR42.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -320,10 +331,318 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+   * Creates a Create mutation resolver field
+   *
+   * @param {GraphQLObjectType} modelType - Model GraphQL object type
+   * @param {IFieldResolver} mutationObj - Mutation resolver object
+   */
+  protected addCreateMutationResolver(modelType: GraphQLObjectType, mutationObj: any) {
+    const modelName = modelType.name;
+    const resolverCreateField = getFieldName(modelName, GraphbackOperationType.CREATE);
+
+    mutationObj[resolverCreateField] = (_: any, args: any, context: any) => {
```
machi1990, line 444:
```suggestion
    mutationObj[resolverCreateField] = (_: any, args: any, context: GraphbackContext) => {
```

machi1990, line 444: I think the context should be properly typed to avoid problems like the one solved by this https://github.com/aerogear/graphback/pull/1408#discussion_r437495973

```suggestion
    mutationObj[resolverCreateField] = (_: any, args: any, context: any) => {
```

templates/ts-apollo-postgres-backend/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "graphback-templates-runtime-mongo",
+  "name": "graphback-templates-runtime-postgres",
```
machi1990, line 2: Good catch

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Good point. I will change that

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Yep wanted to come back to this but wanted to push before my power went (planned power outage here today) - it is back now.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: This was before rebase from master, should be in next push.

craicoverflow:
> To be clear there is actually one problem that we do not resolve on current master - strong typing of the context in people's resolvers, but I consider that something we can address by actually code generation that were left there later rather than plugin itself.

GraphbackContext is coupled to graphback-runtime package. We could 

a) be happy with importing graphback-runtime in codegen-schema. Seems like a code smell.

b) Put GraphbackCRUDService, GraphbackDataProvider, GraphbackContext interfaces into graphback-core (seems logical to me)

c) Other suggestions @machi1990 @wtrocki ?

machi1990:
> > To be clear there is actually one problem that we do not resolve on current master - strong typing of the context in people's resolvers, but I consider that something we can address by actually code generation that were left there later rather than plugin itself.
> 
> GraphbackContext is coupled to graphback-runtime package. We could
> 
> a) be happy with importing graphback-runtime in codegen-schema. 

I think this is okay, I am failing to see how the codegen-schema plugin will work without the runtime.  

> b) Put GraphbackCRUDService, GraphbackDataProvider, GraphbackContext interfaces into graphback-core (seems logical to me)

I think this is already the case?

craicoverflow:
> > GraphbackContext is coupled to graphback-runtime package. We could
> > a) be happy with importing graphback-runtime in codegen-schema.
> 
> I think this is okay, I am failing to see how the codegen-schema plugin will work without the runtime.

codegen-schema is working right now without any dependency on graphback-runtime. The only reason it would be need to be used in codegen-schema is to strongly type the context using `GraphbackContext`

> 
> > b) Put GraphbackCRUDService, GraphbackDataProvider, GraphbackContext interfaces into graphback-core (seems logical to me)
> 
> I think this is already the case?

These are all defined in `@graphback/runtime`:

[GraphbackContext](https://github.com/aerogear/graphback/blob/53eefefadd55a28a66c34d1327f1f04a7372ebe5/packages/graphback-runtime/src/service/GraphbackContext.ts), [GraphbackCRUDService](https://github.com/aerogear/graphback/blob/61b24b7b9b83bf361bdacb754d5a32efeea4833f/packages/graphback-runtime/src/service/GraphbackCRUDService.ts), [GraphbackDataProvider](https://github.com/aerogear/graphback/blob/61b24b7b9b83bf361bdacb754d5a32efeea4833f/packages/graphback-runtime/src/data/GraphbackDataProvider.ts)

craicoverflow:
Working and ready for review - let's not let the context issue block this, that can happen in a different PR if need be.

machi1990:
> > > GraphbackContext is coupled to graphback-runtime package. We could
> > > a) be happy with importing graphback-runtime in codegen-schema.
> > 
> > 
> > I think this is okay, I am failing to see how the codegen-schema plugin will work without the runtime.
> 
> codegen-schema is working right now without any dependency on graphback-runtime. The only reason it would be need to be used in codegen-schema is to strongly type the context using `GraphbackContext`

Right. Thanks. 

> > > b) Put GraphbackCRUDService, GraphbackDataProvider, GraphbackContext interfaces into graphback-core (seems logical to me)
> > 
> > 
> > I think this is already the case?
> 
> These are all defined in `@graphback/runtime`:
> 
> [GraphbackContext](https://github.com/aerogear/graphback/blob/53eefefadd55a28a66c34d1327f1f04a7372ebe5/packages/graphback-runtime/src/service/GraphbackContext.ts), [GraphbackCRUDService](https://github.com/aerogear/graphback/blob/61b24b7b9b83bf361bdacb754d5a32efeea4833f/packages/graphback-runtime/src/service/GraphbackCRUDService.ts), [GraphbackDataProvider](https://github.com/aerogear/graphback/blob/61b24b7b9b83bf361bdacb754d5a32efeea4833f/packages/graphback-runtime/src/data/GraphbackDataProvider.ts)

Did a quick check in my IDE and it deceived me and thought I had seem them in the core.

In that case, yes I'll vote for solution (b) as you proposed. But let's wait for others opinion.

docs/plugins/graphback-schema.md
```diff
@@ -35,16 +41,23 @@ npm install @graphback/codegen-schema
...
+
+### Usage
+
+```ts
+  const schemaPlugin = new SchemaCRUDPlugin({
+    format: 'graphql',
+    outputPath: './src/schema'
+  });
+
+  buildGraphbackAPI(modelDefs, {
```
machi1990, line 53:
```suggestion
 const {schema, typeDefs, resolvers, contextCreator} = buildGraphbackAPI(modelDefs, {
```

To show the results of the above call?

craicoverflow:
> In that case, yes I'll vote for solution (b) as you proposed. But let's wait for others opinion.

I think our agreement is enough, I will do (b). Thanks for input!

wtrocki:
> b) Put GraphbackCRUDService, GraphbackDataProvider, GraphbackContext interfaces into graphback-core (seems logical to me)

That is cool. Then what will be left in runtime then? We moving runtime resolvers generator responsibilities out as well.

craicoverflow:
runtime will be minimal but it will have CRUD service creator function, `NoDataError`, `Logger` etc.

We can still export `GraphbackCRUDService`, `GraphbackContext` from runtime package so people can still use it as normal without having to install core in their template.

wtrocki:
+100!

craicoverflow:
I could not remove completely `LayeredRuntimeResolverCreator` as it is still used in graphql-serve - it can be fixed in https://github.com/aerogear/graphback/issues/1448

docs/plugins/graphback-schema.md
craicoverflow: I can add it in, but did not bother as assumed they have this knowledge already. Good to be explicit though!

wtrocki:
Why not resolve that issue now and make things clean? We also need graphql serve for testing and performance stuff.

wtrocki:
Not in this PR, but overall have this stuff completed rather than having tons of halfbacked stuff and relying on issues to remind us about some unfinished work.

craicoverflow:
> Why not resolve that issue now and make things clean? We also need graphql serve for testing and performance stuff. Should not take much time and it will make entire area clean.

It is a separate issue that will take some time. This PR is already 28 files and I do not want to increase that.

packages/graphback-core/src/runtime/GraphbackContext.ts
```diff
@@ -1,4 +1,4 @@
-import { GraphbackCRUDService } from './GraphbackCRUDService';
+import { GraphbackCRUDService } from '@graphback/core/src/runtime/GraphbackCRUDService';
```
machi1990, line 1:
```suggestion
import { GraphbackCRUDService } from '@graphback/core/GraphbackCRUDService';
```

Since have them exported here https://github.com/aerogear/graphback/pull/1455/files#diff-621d05b95d70a70d7e4794643ed2d0e9R2

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -1,5 +1,5 @@
 import { getFieldName, getSubscriptionName, GraphbackOperationType, ModelDefinition, getPrimaryKey, FieldRelationshipMetadata, getDeltaQuery } from '@graphback/core';
-import { GraphbackContext } from '../service/GraphbackContext';
+import { GraphbackContext } from '@graphback/core/src/runtime/GraphbackContext';
```
machi1990, line 2:
```suggestion
import { GraphbackContext } from '@graphback/core/GraphbackContext';
```

They are exposed here https://github.com/aerogear/graphback/pull/1455/files#diff-621d05b95d70a70d7e4794643ed2d0e9R2

packages/graphback-core/src/runtime/GraphbackContext.ts
craicoverflow: Thanks! I should not allow VS Code to change import paths!

Issue #1452, opened 2020-06-15
Generating resolvers fails with `XXXFilter not found` if `@crud.find` is false and filtering is activated for the model in question through a `@oneToMany` relationship


machi1990:
This happens at https://github.com/aerogear/graphback/blob/00b7366d20284269666080995bbc3bee3c2d30d8/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L277

Typical example is here https://github.com/aerogear/datasync-starter/blob/6b10346983c9394a9d5850fedca61c32b46d6bdb/server/model/task.graphql

machi1990:
I'll create a draft PR to drive the discussion.

PR #1457, opened 2020-06-15
fix(codegen-schema): filter on relationship when `@model.find` is false

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1452


machi1990:
> Looks good, but needs to be updated since the `@crud` namespace is removed now.

Good catch. I have update the mock.graphql schema.

wtrocki:
All good for me! Also worth to see if we can plan things more so we will have less clashes conflicts etc.

wtrocki:
This was soo simple. Thanks for great fix!

Issue #1465, opened 2020-06-16
[DataSync] Broken delete operation 


wtrocki:
It looks like deletion and edit input types are not properly created.
They not accepting the updatedAt field which is critical for the conflict resolution.

Relates to #1460 

Error
```
message: "Variable "$input" got invalid value { id: "5ee7a844cfd59c84cd672d25", title: "123", description: "123", version: null, status: "OPEN", type: null, priority: null, public: null, startDate: "2020-06-15T16:56:29.258Z", payload: null, createdAt: "1592240196780", updatedAt: "1592240196780" }; Field "createdAt" is not defined by type MutateTaskInput."
1: {,…}
```

wtrocki:
FYI @ssd71

ssd71:
@wtrocki I think we need another issue for conflict or can we discuss here?

wtrocki:
see linked issue above. Requires update

wtrocki:
@ssd71 Any work needed here?

ssd71:
It does accept the updatedAt field now, I think not

wtrocki:
Closing then. Good job!

Issue #1460, opened 2020-06-15
Remove existing conflict resolution from datasync offix


wtrocki:
We need to remove or temporary disable version based conflict resolution.
Alternatively we can add early conflict by requiring update and delete function to specify lastUpdated.

@ssd71 Whatever way you prefer. We will need to restore this later anyway.

ssd71:
I think this issue has been addressed, we can research conflict resolution in #1445 wdyt

wtrocki:
Would we use timestamp now instead of version?

wtrocki:
The other issue is for delta store.
We haven't made this change and it is kinda making entire implementation unusable. See: 
https://github.com/aerogear/graphback/blob/ed6a766f15fa7dbacd2f1f0070799ec90dcab5aa/packages/graphback-runtime-mongodb/src/OffixMongoDBDataProvider.ts

PR #1467, opened 2020-06-16
Add updatedAt to mutate inputs in datasync and remove existing conflict resolution

ssd71:
Addresses #1465 #1460


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -294,6 +293,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       const name = model.graphqlType.name;
       const modelTC = schemaComposer.getOTC(name);
       const desc = model.graphqlType.description;
+      const {markers, fieldNames} = metadataMap;
```
wtrocki, line 296:
```suggestion
      const { markers, fieldNames } = metadataMap;
```

wtrocki:
So we removing it, but if we add that now to update requirements we no longer need to remove it. In fact swapping towards token will make conflict work (assuming that we add the same for the delete operations.

WDYT? Sorry for misleading information. I think it is easier to just swap version based stuff to the timestamp ones.

ssd71:
> So we removing it, but if we add that now to update requirements we no longer need to remove it. In fact swapping towards token will make conflict work (assuming that we add the same for the delete operations.
> 
> WDYT? Sorry for misleading information. I think it is easier to just swap version based stuff to the timestamp ones.

Sure! sounds cool

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -64,6 +64,20 @@ export class DataSyncPlugin extends GraphbackPlugin {
                     lastSync: GraphQLNonNull(GraphQLString),
                     filter: filterTypeName
                 })
+
+                // Add updatedAt arg to update and delete input types
+                // for conflict resolution
+                const {fieldNames} = metadataMap;
+                [GraphbackOperationType.DELETE, GraphbackOperationType.UPDATE].map((operation: GraphbackOperationType) => {
```
machi1990, line 71: we do not need an array here since this call https://github.com/aerogear/graphback/pull/1467/files#diff-37ea145aded7cc3fa4e8f167fac3af2bR72 will return the same inputType because of https://github.com/aerogear/graphback/blob/eec1d8bfc6395311c8c61959ade0c9dddf86cb34/packages/graphback-core/src/crud/mappingHelpers.ts#L62-L64

ssd71: Good point, thanks!

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -64,6 +64,19 @@ export class DataSyncPlugin extends GraphbackPlugin {
                     lastSync: GraphQLNonNull(GraphQLString),
                     filter: filterTypeName
                 })
+
+                // Add updatedAt arg to update and delete input types
```
wtrocki, line 68: Instead of inline comments function will be preferred.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -64,6 +64,19 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+                // for conflict resolution
+                const {fieldNames} = metadataMap;
+
+                const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
+                if (inputType) {
+                    inputType.addFields({
+                        [fieldNames.updatedAt]: {
+                            type: GraphQLNonNull(GraphQLString)
```
wtrocki, line 76: Forgot to mention! Nice that you put not null here!

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Good point, I will add in this and then restore conflict resolution and push again

wtrocki:
Reviewed it yesterday. I think we can adjust conflict in separate PR to make it use timestamps. No problem with that for the moment

PR #1447, opened 2020-06-15
fix: independent model loading

wtrocki:
Load model independent of working dir.

PR #1469, opened 2020-06-16
Revert "fix: independent model loading"

wtrocki:
Reverts aerogear/graphback#1447


wtrocki:
This will require templates branch update as well :(

machi1990:
> This will require templates branch update as well :(
> Revert is not solving the issue of the PWD not being guaranteed.
> 
> Maybe we should be using typescript graphql file loader instead and include model file into compilation instead of using FS? New issue?

Yeap, let's create an issue to investigate that. 
While we are at it, we can maybe add some kind of smoke test in templates.

PR #1476, opened 2020-06-17
test(codegen-schema): update snapshot to make test pass

machi1990:
Noticed that master build is broken cause of https://github.com/aerogear/graphback/pull/1457 but the PR did not have latest changes from master hence the snapshot issue.


craicoverflow:
To prevent this happening by accident, I've enabled **Require branches to be up to date before merging**.

If we spot issues with this when using it we can change back.

image

Issue #1459, opened 2020-06-15
[DataSync] Validation of the lastsync key


wtrocki:
Putting any invalid value into sync key results in error

```graphql
query syncTasks{
  syncTasks(lastSync:""){
    items{
      id
      title
    }
    lastSync
  }
}
```

Error:
```
 "message": "Cannot read property 'sync' of undefined",
      "locations": [
```

PR #1422, opened 2020-06-10
Add datasync docs

ssd71:
This is very much a work in progress, although feel free to provide feedback if you feel it is lacking in some regard. I am also planning to add docs for metadata, although I think that would suffice with a link to graphql-migrations and some graphback only metadata like `versioned`

Issue #1409, opened 2020-06-09
Template(s) for Data Synchronization


ssd71:
We need templates that show how to use the datasync package

wtrocki:
We have one:
https://github.com/aerogear/datasync-starter

ssd71:
Yep, my bad, I almost forgot about that

craicoverflow:
Closing as it looks like this one was solved.

PR #1477, opened 2020-06-17
Move datasync resolvers to datasync plugin and other minor fixes

### LINKED ISSUE #1475: Create DataSync resolvers in DataSyncPlugin
With https://github.com/aerogear/graphback/issues/1418 complete, DataSync resolvers can be added in the DataSyncPlugin

ssd71:
Fixes #1475 #1459

### DIFF
--- a/packages/graphback-datasync/package.json
+++ b/packages/graphback-datasync/package.json
@@ -43,12 +43,12 @@
     "@graphback/runtime-mongo": "0.14.0-alpha4",
     "@graphql-toolkit/common": "0.10.7",
     "@graphql-toolkit/core": "0.10.7",
-    "@graphql-tools/utils": "^6.0.10",
+    "@graphql-tools/utils": "6.0.10",
     "graphql-compose": "7.18.1",
     "graphql-metadata": "0.6.3",
     "graphql-subscriptions": "1.1.0",
     "mongodb": "3.5.9",
-    "pluralize": "^8.0.0"
+    "pluralize": "8.0.0"
   },
   "peerDependencies": {
     "graphql": "^0.12.0 || ^0.13.0 || ^14.0.0"
--- a/packages/graphback-datasync/package.json
+++ b/packages/graphback-datasync/package.json
@@ -43,11 +43,12 @@
     "@graphback/runtime-mongo": "0.14.0-alpha4",
     "@graphql-toolkit/common": "0.10.7",
     "@graphql-toolkit/core": "0.10.7",
+    "@graphql-tools/utils": "^6.0.10",
     "graphql-compose": "7.18.1",
     "graphql-metadata": "0.6.3",
     "graphql-subscriptions": "1.1.0",
     "mongodb": "3.5.9",
-    "pluralize": "8.0.0"
+    "pluralize": "^8.0.0"
   },
   "peerDependencies": {
     "graphql": "^0.12.0 || ^0.13.0 || ^14.0.0"
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -5,6 +5,7 @@ import { parseMetadata } from "graphql-metadata";
 import { SchemaComposer, ObjectTypeComposerFieldConfig } from 'graphql-compose';
 import { IResolvers, IFieldResolver } from '@graphql-tools/utils'
 import { getDeltaType, getDeltaListType, getDeltaQuery } from "./deltaMappingHelper";
+import { getDataSyncService } from "./util";
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -124,11 +125,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
     const deltaQuery = getDeltaQuery(modelType.name)
 
     queryObj[deltaQuery] = async (_: any, args: any, context: GraphbackContext) => {
-      const dataSyncService: any = context.graphback[modelName];
-
-      if (dataSyncService.sync === undefined) {
-        throw Error("Please use DataSync provider for delta queries");
-      }
+      const dataSyncService = getDataSyncService(context.graphback[modelName]);
 
       return dataSyncService.sync(args.lastSync);
     }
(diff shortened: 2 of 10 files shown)


wtrocki:
Awesome work!

packages/graphback-datasync/package.json
```diff
@@ -43,11 +43,12 @@
     "@graphback/runtime-mongo": "0.14.0-alpha4",
     "@graphql-toolkit/common": "0.10.7",
     "@graphql-toolkit/core": "0.10.7",
+    "@graphql-tools/utils": "^6.0.10",
     "graphql-compose": "7.18.1",
     "graphql-metadata": "0.6.3",
     "graphql-subscriptions": "1.1.0",
     "mongodb": "3.5.9",
-    "pluralize": "8.0.0"
+    "pluralize": "^8.0.0"
```
machi1990, line 51:
```suggestion
    "pluralize": "8.0.0"
```

ssd71: Ah, sorry about that, I miss that all the time :D

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
```diff
@@ -84,21 +84,7 @@ export class LayeredRuntimeResolverCreator {
...
-      }
-
-      // If delta marker is encountered, add resolver for `delta` query
-      if (resolverElement.config.deltaSync) {
-        const deltaQuery = getDeltaQuery(resolverElement.graphqlType.name)
-
-        resolvers.Query[deltaQuery] = async (parent: any, args: any, context: GraphbackContext) => {
-          const dataSyncService: any = context.graphback[modelName];
-
-          if (dataSyncService.sync === undefined) {
```
craicoverflow, line 96: This would technically work with some other class that happens to have the `.sync` property or function. You should use `instanceof` to ensure the class and give full typing

```ts
function isDataSyncService(service: any): service is DataSyncCRUDService {
  return service instanceof DataSyncCRUDService
}
```

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -88,10 +88,19 @@ function stringTimestampsToInt(filter: any, key: string): any {
     // operators on these fields to Integers
     if (["createdAt", "updatedAt"].includes(key)) {
         if (isPrimitive(filter[key])) {
-            filter[key] = parseInt(filter[key], 10);
+            const n = parseInt(filter[key], 10);
+            if (isNaN(n)) {
+                throw Error("Please provide a proper timestamp")
```
craicoverflow, line 93:
```suggestion
                throw Error("Please provide a valid timestamp")
```

packages/graphback-runtime-mongodb/src/queryBuilder.ts
```diff
@@ -88,10 +88,19 @@ function stringTimestampsToInt(filter: any, key: string): any {
...
+            }
+            filter[key] = n;
         } else {
             const entries = Object.entries(filter[key]).map((entry: [string, string]) => {
-                return [entry[0], parseInt(entry[1], 10)];
+                const n = parseInt(entry[1], 10);
+                if (isNaN(n)) {
+                    throw Error("Please provide a proper timestamp")
```
craicoverflow, line 100:
```suggestion
                    throw Error("Please provide a valid timestamp")
```

packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts
ssd71: Fair point, Thanks!

wtrocki:
`datasync-is-free` ROFL

wtrocki:
@ssd71 Amazing change. Tried that with datasync-starter branch.

How do you test these features yourself?
I think we will need to add some of the elements into release notes for data sync as it is important to collect it while we develop rather than try to remember what we have done. We can do it in #1422

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -94,4 +120,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
     return SCHEMA_CRUD_PLUGIN_NAME;
```
craicoverflow, line 120: I'd suggest changing the variable name as it is not accurate.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -94,4 +120,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
   }
 
+  protected addDeltaSyncResolver(modelType: GraphQLObjectType, queryObj: IFieldResolver<any, any>) {
+    const modelName = modelType.name;
+    const deltaQuery = getDeltaQuery(modelType.name)
```
craicoverflow, line 125: You already have `modelName` on line 124

packages/graphback-datasync/src/util.ts
```diff
@@ -1,6 +1,15 @@
...
 export function isDataSyncModel(model: ModelDefinition): boolean {
   return parseMetadata("delta", model.graphqlType) && parseMetadata("versioned", model.graphqlType)
 }
+
+export function getDataSyncService(service: GraphbackCRUDService): DataSyncCRUDService {
+  if (service instanceof DataSyncCRUDService) {
+    return service;
+  }
+
+  throw Error("Service is not a DataSyncCRUDService. Please use DataSyncCRUDService and DataSync-compliant DataProvider with DataSync Plugin to get Delta Queries.")
```
craicoverflow, line 14: This error is really only useful in the `DataSyncPlugin`, but if someone wants to fetch the DataSyncCRUDService or undefined (quite common) they then have to catch an error. I'd throw the error in the DataSyncPlugin instead and change this to `isDataSyncService`

wtrocki:
Ouch.. Sorry for premature merge @craicoverflow 

@ssd71 Would you mind applying this in separate PR.

ssd71:
> Ouch.. Sorry for premature merge @craicoverflow
> 
> @ssd71 Would you mind applying this in separate PR.

Sure! Will do that first thing tomorrow, also thanks for feedback @craicoverflow :D

ssd71:
> How do you test these features yourself?

I had almost created a separate mongodb-datasycn template thanks to #1409 , so using that for testing hehe

> I think we will need to add some of the elements into release notes for data sync as it is important to collect it while we develop rather than try to remember what we have done. We can do it in #1422

Will do!

wtrocki:
@ssd71 If @craicoverflow allows maybe we can drop one template into Graphback

craicoverflow:
> @ssd71 If @craicoverflow allows maybe we can drop one template into Graphback.
> It will be kinda duplicating a lot but might be better to have it tested in the repo

I think that is fine, I have thought about this as I was finding it difficult to verify data sync functionality. It could be a minimal template (they are quite small already).

PR #1489, opened 2020-06-18
refactor(datasync): minor refactoring

ssd71:
Following up with feedback from #1477

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -12,7 +12,7 @@ import { isDataSyncService } from "./util";
  */
 
 
-export const DATASYNC_PLUGIN_NAME = "DatasyncPlugin";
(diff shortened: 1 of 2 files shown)


packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -5,14 +5,14 @@ import { parseMetadata } from "graphql-metadata";
...
-import { getDataSyncService } from "./util";
+import { isDataSyncService } from "./util";
 
 /**
  * Configuration for Schema generator CRUD plugin
  */
 
 
-export const SCHEMA_CRUD_PLUGIN_NAME = "DatasyncPlugin";
+export const DATASYNC_PLUGIN_NAME = "DatasyncPlugin";
```
machi1990, line 15:
```suggestion
export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
```
WDYT?

PR #1492, opened 2020-06-18
feat: restore basic conflicts in datasync

ssd71:
This PR restore the old conflict functionality in datasync.

Docs needed.


packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -30,16 +31,19 @@ export class DataSyncMongoDBDataProvider<Type = any, GraphbackContext = any> ext
...
 
     // TODO Can be improved by conditional updates
-    const queryResult = await this.db.collection(this.collectionName).find({ _id: new ObjectId(idField.value), _deleted: false }).toArray();
-    if (queryResult && queryResult[0]) {
-      queryResult[0][idField.name] = queryResult[0]._id;
-
+    const queryResult = await this.db.collection(this.collectionName).findOne({ _id: new ObjectId(idField.value), _deleted: false });
+    if (queryResult) {
+      queryResult[idField.name] = queryResult._id;
+      if (parseInt(data[fieldNames.updatedAt], 10) !== queryResult[fieldNames.updatedAt]) {
```
wtrocki, line 44: I would prefer to compare them as strings

wtrocki: So lastSync should be the label not number on its own. Then people can decide what field they will use for comparison - out of the box this will be lastUpdatedAt but do not need to be actually that field itself

Issue #1482, opened 2020-06-18
Add unique index for `@id` annotation for mongodb


ssd71:
I was just thinking that since this works for Knex Data source, would be good to do it here as well

machi1990:
Very good idea @ssd71. I had this on my mind after https://github.com/aerogear/graphback/pull/1276 landed.

PR #1498, opened 2020-06-19
Mongo id index and datasync conflict fix

ssd71:
This PR addresses: #1482 and addresses feedback from #1492 hence also addresses #1460


packages/graphback-runtime-mongodb/src/utils/createIndexes.ts
```diff
@@ -85,4 +95,20 @@ export function getRelationIndex(field: GraphQLField<any, any>): IndexSpecificat
   } else {
     return undefined;
   }
+}
+
+export function isCustomIdField(field: GraphQLField<any, any>): IndexSpecification {
```
machi1990, line 100: This is more like like `getIndexForPrimaryKey` since it is returning the index?

ssd71: Yes, I'll fix that, thanks!

packages/graphback-runtime-mongodb/src/utils/createIndexes.ts
```diff
@@ -85,4 +95,20 @@ export function getRelationIndex(field: GraphQLField<any, any>): IndexSpecificat
...
+  const indexMetadata: any = parseMetadata('id', field.description);
+  if (indexMetadata === true) {
+    const indexSpec: IndexSpecification = {
+      key: {
+        [field.name]: 1
```
machi1990, line 105: How will this work with 
```
""" 
@model
"""
type SomeType{
 """ 
 @id
 @db.name: 'primary_key_colulm'
 """ 
  key: String!
}
```

ssd71: We don't have renaming fields in Mongo yet, but I think we probably need `id:ID!`

machi1990: Oh sure, then this comment can be safely ignored.

ssd71: I do think a custom primary key would be worthwhile looking into instead of just indexing it, so I will try and open an issue for this after a bit of research, Thanks for pointing it out :smile:

machi1990: Amazing thanks for the pro-activity.

wtrocki: Hmm.. reminded me that we need to decide what we do with _id.

Issue #1234, opened 2020-05-06
Proper logging format using pino/winston TBD


machi1990:
## Feature Request

Current when querying the server logs the something like 

```
Creating object Recipient
Creating object PrefferedProduct
querying object Product with filter {"id":"5eb16d8d54bbea1125a1fa33"}
querying object Recipient with filter {"id":"5eb290a7bd50c03b80b7261e"}
querying object Recipient
querying object Product with filter {"id":"5eb2745fda8343363625fd2d"}
```

It would be good if these logs would use a proper logger so that logs are properly formatted. 

These seems like an issue for graphback repo but opening here first for discussion.

/cc @wtrocki

PR #1499, opened 2020-06-19
Remove logger from the graphback services

wtrocki:
Motivation is to remove pointless logger that is not easy to remove from the service.
We have a separate issue to improve general logging but logging should be disabled by default.


packages/graphback-runtime/src/service/CRUDService.ts
```diff
@@ -11,10 +10,7 @@ export interface CRUDServiceConfig {
    * PubSub implementation for creating subscriptions
    */
   pubSub?: PubSubEngine
-  /**
-   * Optional logging service
-   */
-  logger?: GraphbackMessageLogger
```
craicoverflow, line 17: Hmm - I could see it as valuable to have but disabled by default, so the logger would be set but it would have an `enabled` flag passed to the constructor

wtrocki: hmm.. general we need to have platform level logger in core
having separate logger for the each component line service, data provider, plugin etc. will be really really bad API. We have issue to improve logging as whole (thx @machi1990) and we will action it there. This was more like hotfix as extensive logging in node.js is problem (especially console.log)

machi1990: Yes we have this issue https://github.com/aerogear/graphback/issues/1234 to address that. And I could see the logger being configured by the user (to make it custom) so that we do not bring our own logger lib.

Issue #793, opened 2020-02-26
Integrate graphql-query-mapper into graphback queries


wtrocki:
When making queries we always select all data which is no efficient.
We can use graphql-query-mapper library to see what kind of data was requested and map it directly in our API. 

## Targets

- Eliminate select * from all queries.
- Add a way to integrate query-mapper - selectionSet into context so CRUD layer can detect it and act on it.

## Stretch Goals

- Knex doesn't seem to be generic to be able to implement all crud methods and swap databases. There are differences which means that we will need to implement each db directly. Instead of using knex we can rely on actual db drivers which will allow us to get better insights into performance and consistency handlers

wtrocki:
Target here is to avoid * when doing selects. We can use some parts of the query mapper to get actual fields.

machi1990:
This is something I'll be willing to give a try at some point. 

WDYT? @wtrocki @craicoverflow

wtrocki:
Yes. This can be perfect issue to grab

craicoverflow:
Related https://github.com/aerogear/graphback/issues/1281

craicoverflow:
> This is something I'll be willing to give a try at some point.

Yep, this would be great to have.

wtrocki:
We need that for 0.14.0 (specifically we want to execute total count only when a client requests it.

wtrocki:
Bumping this. WE need that for 0.14.0

PR #1509, opened 2020-06-19
feat: retrieve query-able fields for selection in database

machi1990:
Fixes https://github.com/aerogear/graphback/issues/793

- [x] Make sure that foreign key fields coming from relationhips are retrieved
- [x] ~~Use the info to build pagination (could be done in separate PR)~~ Issue https://github.com/aerogear/graphback/issues/1517 opened, I'll work on it on separate PR after this one is merged.


wtrocki:
Amazing work @machi1990.
Considering that query mapper requires more work. Woud you see bring this logic into graphback or use separate package?

wtrocki:
Amazing work @machi1990.
Considering that query mapper requires more work. Woud you see bring this logic into graphback or use separate package?

wtrocki:
@machi1990 I'm happy with the changes. I think that graphql-query-mapper never took of and as you seen there are other libraries out there that do this work well (I believe graphql-tools 5.0 have this as well) so query mapper has no sense to be used. 

I'm looking for PR to be marked as ready to do final review.

packages/graphback-core/src/runtime/GraphbackCRUDService.ts
```diff
@@ -23,44 +23,48 @@ export interface GraphbackCRUDService<Type = any, GraphbackContext = any> {
...
 
   /**
    * Fetch a single record by its unique attribute(s)
    *
    * @param filter - the unique attributes to fetch the record with
+   * @param fields fields to select from datasource
    * @param context context object from GraphQL/REST layer
    */
-  findOne(filter: Type, context?: GraphbackContext): Promise<Type>;
+  findOne(filter: Type, fields: string[], ccontext?: GraphbackContext): Promise<Type>;
```
wtrocki, line 56: `ccontext` :)


Also -1 for adding ` fields: string[]` it brings negative pattern to break API and increases number of arguments.

There are two options here:

1) Introduce options object (we had that before for subscriptions flags but since then we use context instead
2) use `context.graphbackOptions` or something similar and make context non optional

No strong opinion but utilizing context will setup good pattern for possible extensions/requirements in future. Current implementation will encourage constant breaking when we need yet another thing to add

machi1990: Also my plan to to get early feedback on the api design. This is really useful feedback that I'll look to bring on later and mark the PR ready for review.

packages/graphback-core/src/runtime/GraphbackContext.ts
```diff
@@ -11,5 +11,6 @@ export interface GraphbackServiceConfigMap {
  * GraphQL context interface according to Graphback runtime layer format
  */
 export interface GraphbackContext {
-  graphback: GraphbackServiceConfigMap
+  graphback: GraphbackServiceConfigMap,
+  selectedFields: string[]
```
machi1990, line 15: Is this okay?

craicoverflow:
Just curious - what was the reason(s) to not use graphql-query-mapper? I assume it is not up to scratch in some way.

Where does that leave us with the library though? Are we going to leave it there as is or look to maintain it, are there people using it?

cc @wtrocki

machi1990:
> Just curious - what was the reason(s) to not use graphql-query-mapper? I assume it is not up to scratch in some way.

I ran into this issue - https://github.com/aerogear/graphql-query-mapper/issues/19 when trying to integrate it.

packages/graphback-core/src/runtime/GraphbackContext.ts
craicoverflow: I think @wtrocki mentioned putting into an options object on the context

```ts
export interface GraphbackContext {
  graphback: GraphbackServiceConfigMap
  graphbackOptions: {
    selectedFields: string[]
  }
}
```

I can see why this is a good idea - to keep the top level context small.

wtrocki: I would be looking to have typescript definitions of our context so it will be strongly typed and easy to work with. We have generics already just need to start with some base:

context.graphback - services
context.graphbackOptions - request level arguments
context.graphbackConfig - configuration (we are passing this around in services but we can have it mounted in context as well)

machi1990: Isn't confusing to have `graphback` and `graphbackOptions` on top level object?

wtrocki: We are free to propose different namespaces. 
```
graphback.services.Tastk
graphback.config
graphback.optons
```

machi1990: +1 on having
 ```
context.graphback.options.selectedFields
context.graphback.services.Entity
context.graphback.config.crud.find
```

Comment (wtrocki) [reply packages/graphback-core/src/runtime/GraphbackContext.ts:15]: @machi1990 would you make this small changes for us?

File: packages/graphback-runtime-knex/src/KnexDBDataProvider.ts [hunk 3]
```diff
@@ -28,25 +28,24 @@ export class KnexDBDataProvider<Type = any, GraphbackContext = any> implements G
     this.baseType = baseType;
   }
 
-  public async create(data: Type): Promise<Type> {
+  public async create(data: Type, context: any): Promise<Type> {
```
craicoverflow, line 31: Should context be strongly typed to GraphbackContext?

packages/graphback-core/src/runtime/GraphbackContext.ts
machi1990: > +1 on having
> 
> ```
> context.graphback.options.selectedFields
> context.graphback.services.Entity
> context.graphback.config.crud.find
> ```

This should be the case now for `services` and `options` from my recent push. I did not want to bring in the `config` change yet.If we can open an issue for that, it can be looked post this PR.

wtrocki: Config is redundant - it was used as example. 
All good with options and services. 
@craicoverflow WDYT?

craicoverflow: Agreed that there is no need for both options and config.

packages/graphback-core/src/runtime/GraphbackDataProvider.ts
```diff
@@ -22,7 +22,7 @@ export interface GraphbackDataProvider<Type = any, GraphbackContext = any> {
    * @param data input data
    * @param context context object passed from graphql or rest layer
    */
-  create(data: Type, context?: GraphbackContext): Promise<Type>;
+  create(data: Type, context: GraphbackContext): Promise<Type>;
```
craicoverflow, line 25: I have a concern with passing the entire context object down to the data provider with every request. All that is used is the selectedFields.

Context can have:

- A CRUD service for every model
- A data provider for every model
- 0...N data loaders
- Request/response data
- Selected fields
- Other unknown user context data

This seems too much to pass down to the provider for the purpose of using only a tiny fraction of this.

packages/graphback-core/src/runtime/GraphbackContext.ts
craicoverflow: I think there is a need to have The `options` key as a strongly typed interface though, so it can be passed down to the data providers instead of the full context.

At present the entire context is getting passed which I  think is overkill.

See https://github.com/aerogear/graphback/pull/1509/files#r444043195

packages/graphback-core/src/runtime/GraphbackDataProvider.ts
wtrocki: Objects are passed by reference and context is a general-purpose function that is widely used across requests. We have been silently using it for many purposes so it will make sense to standardize it and use it even more :D

craicoverflow: Good point.

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
machi1990: The interfaces:
1. https://github.com/aerogear/graphback/blob/623426121c3573d28730c903d7949e7bbac5f3ea/packages/graphback-core/src/runtime/GraphbackDataProvider.ts
2. https://github.com/aerogear/graphback/blob/623426121c3573d28730c903d7949e7bbac5f3ea/packages/graphback-core/src/runtime/GraphbackCRUDService.ts 

are typed already.

craicoverflow: Yep in the interfaces, but they are not typed in the class, which is where they are used. If you were to make a typo you would not know because it is `any`.

machi1990: Yes, not sure how that will play with the type defs we have in the interface. 

Just a note, adding a type here will make the tests hard to maintain: This file for example, if we add something e.g `config` to the contex but not using it in `DataProvider`, we'll have to properly add it in all tests files so that the compiler is happy https://github.com/aerogear/graphback/blob/83ca6c904d80b151b97fae1cc309ee172afe886c/packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts 

Just to get historical reasons, the context is `any` to allow user custom things to be added in context right?

craicoverflow: Ah, you can do `context: Partial<GraphbackContext>`, making all fields optional.

I think the context is any because at the time we were not using it and did not have any use for it, though I am not sure, @wtrocki ?

That raises a good point though. Maybe the interface should have `any`, and the KnexDBDataProvider should have the typed interface. 

Not every data provider will need GraphbackContext (a custom one for example), but the interface makes it required. So I think it should be the other way around?

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -167,6 +172,16 @@ export class MongoDBDataProvider<Type = any, GraphbackContext = any> implements
     return document;
   }
 
+  protected buildProjectionOption(context: any) {
```
craicoverflow, line 175: Could just pass `options` to this method?

packages/graphback-core/src/plugin/getSelectedFieldsFromResolverInfo.ts
```diff
@@ -0,0 +1,33 @@
+import { GraphQLResolveInfo } from 'graphql';
```
wtrocki, line 1: So we will keep this in core? Nice:)

wtrocki: Would that execute for every query now?

wtrocki:
Minor rebase needed and we can merge. (we do not squash commits so not worth to do it manually

wtrocki:
Awesome work!

Issue #1444, opened 2020-06-15
Single annotation to rule them all


wtrocki:
Instead of using `model, versioned and delta` we can use `datasync` to imply all 3 of them.

craicoverflow:
Misleading title, I thought you meant one namespace for ALL metadata.

I still think `@model` should be separate, using `@datasync` will cause issues in migrations and everywhere that depends on `@model` to build metadata for plugins etc.

wtrocki:
Agreed!

PR #1513, opened 2020-06-20
Fix datasync using multiple annotations

ssd71:
Addresses #1444 
:D

### DIFF
--- a/docs/datasync/datasync-soft-delete.md
+++ b/docs/datasync/datasync-soft-delete.md
@@ -194,13 +194,13 @@ Notice that there are is no relationship field in the `CommentDelta` type.
 
 ### `@delta` annotation
 
-The `@datasync` annotation provided by `@graphback/datasync` is a shorthand for two separate annotations: `@delta` and `@versioned`. Vanilla Graphback supports `@versioned` out of the box, see [this](../metadata.md) page for it's thorough documentation.
+The `@datasync` annotation provided by `@graphback/datasync` is a shorthand for two separate annotations: `@delta` and `@versioned`. Vanilla Graphback supports `@versioned` out of the box, see [this](../metadata.md) page.
 
 The `@delta` annotation is only provided by the `@graphback/datasync` package. However, we do not support using this annotation on it's own without `@versioned`.
 
 #### Example
 
-Replacing `@datasync` in any the above examples yields the exact same results, for example:
+Replacing `@datasync` in any the above examples with `@versioned` and `@delta` yields the exact same results, for example:
 ```graphql
 """ 
 @model
@@ -218,7 +218,8 @@ This model exactly yields all of the type-definitions outlined above:
 ```graphql
 """ 
 @model
-@datasync 
+@versioned
+@delta 
 """
 type Comment {
   id: ID!
@@ -229,4 +230,4 @@ type Comment {
 }
 ```
 
-Similarly, this also yields a `Delta` type, a `DeltaList` type and a `sync` query.
+Similarly, this will also yield a `Delta` type, a `DeltaList` type and a `sync` query.
--- a/docs/datasync/datasync-soft-delete.md
+++ b/docs/datasync/datasync-soft-delete.md
@@ -82,7 +82,7 @@ The `DeltaList` is a container for `Delta` type, which also returns a `lastSync`
 
 ### Modify the template to support Data Synchronization
 
-In the [`src/index.ts`](https://github.com/aerogear/graphback/blob/templates-0.14.0/templates/ts-apollo-mongodb-backend/src/index.ts) file of the template, use  `DataSyncPlugin` and compliant data sources in `buildGraphbackAPI`:
+In the [`src/index.ts`](https://github.com/aerogear/graphback/blob/master/templates/ts-apollo-mongodb-backend/src/index.ts) file of the template, use  `DataSyncPlugin` and compliant data sources in `buildGraphbackAPI`:
 
 ```typescript
 import { createDataSyncMongoDbProvider, createDataSyncCRUDService, DataSyncPlugin } from '@graphback/datasync'
--- a/docs/datasync/datasync-soft-delete.md
+++ b/docs/datasync/datasync-soft-delete.md
@@ -12,13 +12,12 @@ Start off with the official Graphback template for MongoDB [*here*](https://GitH
 
 ### Annotate the required models
 
-Add the `@versioned` and `@delta` annotations to your model(s) in your GraphQL SDL found in the model folder:
+Add the `@datasync` annotation to your model(s) in your GraphQL SDL found in the `model` folder:
 
 ```graphql
 """ 
 @model
-@versioned
-@delta
+@datasync
 """
 type Comment {
   id: ID!
@@ -27,19 +26,16 @@ type Comment {
 }
 ```
 
-The `@versioned` annotation ensures consistency of your data and the `@delta` annotation gives you delta queries.
+The `@datasync` annotation ensures consistency of your data and gives you delta queries.
 
-The `@graphback/datasync` package also provides a `@datasync` annotation, which is a shorthand for both `@versioned` and `@delta`.
+> **NOTE**: The `@datasync` annotation is a shorthand for two annotations `@versioned` and `@delta`. Refer to the last section for details.
 
-> **NOTE**: While the `@versioned` annotation can be used independent of `@graphback/datasync`, both `@versioned` and `@delta` are needed on a model for adding delta queries
-
-`@versioned` transforms your model by adding two fields: `updatedAt` and `createdAt`
+`@datasync` transforms your model by adding two fields: `updatedAt` and `createdAt`
 
 ```graphql
 """ 
 @model
-@versioned
-@delta 
+@datasync 
 """
 type Comment {
   id: ID!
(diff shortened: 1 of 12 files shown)


wtrocki:
Sorry for late review. I needed to think about this more. I talked with couple guys about their datasync use cases.

To have DataSync successful as feature we need ability to pick any field from the type to check for consistency.

Plugin with @delta implementation exposes delta schema queries
@versioned adds lastUpdatedAt and createdAt as generic feature
@datasync ties both delta and versioned into shorthand format so you have this out of the box

I'm thinking - should our DataProvider accept field name or function that will return value that will be used for comparison. This requirement is already implemented in offix (offix conflict server package) where we have pluggable conflict resolution.

Would that be confusing or overkill?
Is this making sense?
While this PR makes sense I think that we still need delta on your own 

T

wtrocki:
What I mean here is that what we have done in this PR is 100% correct but we might actually still need delta annotation. Would love to get your feedback first

ssd71:
Delta annotations is actually still there, I have just removed relevant docs, will restore that part today

Also about pluggable conflicts: sounds really cool! What do you think about using DataProvider taking a function that will override default conflicts?

ssd71:
Do you mean that we should change the approach of how it (datasync being shorthand) is done in code?

wtrocki:
What I wanted to have here is to not remove delta docs completely as it will be useful

> What do you think about using DataProvider taking a function that would be used as custom logic for conflict checking?(is this what you were talking about?)

Exactly, we can do it later. Just if you can keep docs of both for the moment - like mention this annotation as shorthand

packages/graphback-datasync/src/util.ts
```diff
@@ -3,7 +3,8 @@ import { parseMetadata } from 'graphql-metadata';
 import { DataSyncCRUDService } from "./services";
 
 export function isDataSyncModel(model: ModelDefinition): boolean {
-  return parseMetadata("delta", model.graphqlType) && parseMetadata("versioned", model.graphqlType)
+  // (Both delta and versioned) or (just datasync)
+  return (parseMetadata("delta", model.graphqlType) && parseMetadata("versioned", model.graphqlType))||(parseMetadata('datasync', model.graphqlType))
```
wtrocki, line 7: Logish enough to have it split into more lines

wtrocki:
> Do you mean that we should change the approach of how it (datasync being shorthand) is done in code?

Code is good. approved already

ssd71:
Made some changes in docs as well :D, wdyt

docs/datasync/datasync-soft-delete.md
```diff
@@ -46,8 +48,19 @@ type Comment {
...
+```
+
+The `@delta` annotation adds a `sync` query or a delta query:
+```graphql
+type Query {
+  syncComments(lastSync: String!, filter: CommentFilter): CommentDeltaList!
+}
+```
 
+This allows you to get all the changed(updates and deletes) documents in a collection since the `lastSync` timestamp. Internally this uses the `updatedAt` database field to check if any documents in the database have been modified, by comparing client provided `lastSync` timestamp value. The documents to be returned in a `sync` query may also be filtered, so you can, for example, only `sync` comments on a specific post. last stuff
```
machi1990, line 60: I was lost when reading this sentence:  `The documents to be returned in a `sync` query may also be filtered, so you can, for example, only `sync` comments on a specific post. last stuff ` 

What did you want to say here?

wtrocki: Good point. This can be simplified

docs/datasync/datasync-soft-delete.md
```diff
@@ -61,22 +74,17 @@ type CommentDeltaList {
...
 
-The `@delta` annotation also adds a `_deleted` field to the `delta` type that tracks if a document has been deleted. Thus `delete` mutations only mark documents with `_deleted: true` instead of actually removing them.
+The `DeltaList` is a container for `Delta` type, which also returns a `lastSync` timestamp, that can be used in a subsequent `sync` query.
 
 > **NOTE**: `Soft Deletes` strategy can only get you the latest version of changed documents ignoring any in-between states that may have transpired between `lastSync` and now.
 
 ### Modify the template to support Data Synchronization
 
-In the [`src/index.ts`]() file of the template, use  `DataSyncPlugin` and compliant data sources in `buildGraphbackAPI`:
+In the [`src/index.ts`](https://github.com/aerogear/graphback/blob/templates-0.14.0/templates/ts-apollo-mongodb-backend/src/index.ts) file of the template, use  `DataSyncPlugin` and compliant data sources in `buildGraphbackAPI`:
```
machi1990, line 87: Is there some way to use a link of the latest templates releases and avoid the versioned link e.g `-0.14.0`? doing so we probably wouldn't have to care about revising the fine grained details in docs when doing the upgrade.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -301,29 +301,15 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       const desc = model.graphqlType.description;
       const { markers, fieldNames } = metadataMap;
       if (parseMarker(markers.versioned, desc)) {
+        const metadataFields = createMetadataFields();
```
machi1990, line 304: `MetadataFields` is broad. Can we be specific here about the fields we are adding? I am thinking of `createDataSyncDatesFields`, open for other suggestions.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -133,4 +136,25 @@ export class DataSyncPlugin extends GraphbackPlugin {
       return dataSyncService.sync(args.lastSync);
     }
   }
+
+  protected addMetaFields(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
```
machi1990, line 140: The function name could be `addDataSyncMetaDataFields` considering my suggestions above

docs/datasync/datasync-soft-delete.md
wtrocki: I think we might need minimal datasync template anyway :)

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -133,4 +136,25 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+    models.forEach((model: ModelDefinition, index: number) => {
+      const name = model.graphqlType.name;
+      const modelTC = schemaComposer.getOTC(name);
+      const desc = model.graphqlType.description;
+      const { markers } = metadataMap;
```
machi1990, line 145: This is not used

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -133,4 +136,25 @@ export class DataSyncPlugin extends GraphbackPlugin {
       return dataSyncService.sync(args.lastSync);
     }
   }
+
+  protected addMetaFields(schemaComposer: SchemaComposer<any>, models: ModelDefinition[]) {
+    models.forEach((model: ModelDefinition, index: number) => {
```
machi1990, line 141: index parameter is not used.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -31,6 +32,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
       return schema;
     }
+
+    this.addMetaFields(schemaComposer, models);
```
machi1990, line 36: With this call, we are introducing a double loop. 

Do you think we can perform the actions in https://github.com/aerogear/graphback/pull/1513/files#diff-37ea145aded7cc3fa4e8f167fac3af2bR140 in the this forEach loop? https://github.com/aerogear/graphback/pull/1513/files#diff-37ea145aded7cc3fa4e8f167fac3af2bR37 

What I have in mind is to the `addMetaFields` (which is to be named) method could instead take the model and not the list of models.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
ssd71: What do you think about `createVersionFields` or `createVersionedFields`, a bit reluctant to put datasync stuff in schema plugin, plus this functionality is independent of datasync so might cause confusion

machi1990: Okay, works for me. 

But we have the same re-occurence of the method name in the `DataSyncPlugin.ts` file, you can bring in the suggestions contextual.

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: I see, Thanks! :D

docs/datasync/datasync-soft-delete.md
ssd71: I have changed it to point to master for the meantime, I realise that this is not optimal, but I don't see any other workarounds here except actually pushing a datasync template and linking to it, unless I am missing something :D

We could always remove this link, it's not exactly being that much more useful as it is

docs/datasync/datasync-soft-delete.md
```diff
@@ -182,4 +187,47 @@ type CommentDelta {
...
+Notice that there are is no relationship field in the `CommentDelta` type.
+
+
+## Advanced topics
+
+### `@delta` annotation
+
+The `@datasync` annotation provided by `@graphback/datasync` is a shorthand for two separate annotations: `@delta` and `@versioned`. Vanilla Graphback supports `@versioned` out of the box, see [this](../metadata.md) page.
+
+The `@delta` annotation is only provided by the `@graphback/datasync` package. However, we do not support using this annotation on it's own without `@versioned`.
```
wtrocki, line 199: This paragraph is good but.. Let's create issue to supply custom function to select field name that will be used for deltas. WDYT?

ssd71: Definitely!

Issue #1472, opened 2020-06-17
[DataSync] Add TTL index for soft deletion


ssd71:

wtrocki:
Wondering what TTL has to do to the GraphQL schema? Do you think users should configure it using GraphQL mutations?

I thought I will be set internally by some config and managed internally by some Cron job etc.

ssd71:
My bad, I should provide some context.

>Wondering what TTL has to do to the GraphQL schema? Do you think users should configure it using GraphQL mutations?

Not really, I was thinking about setting it internally.

> I thought I will be set internally by some config and managed internally by some Cron job etc. DateTime is supported on mongo so we have all enabled

I have some thoughts about this:
MongoDB has a feature called TTL Indexes which only works on Date values [<sup>1</sup>](https://stackoverflow.com/questions/53766763/how-to-create-ttl-index-on-long-timestamp-in-mongodb), if we were to use this feature, we would need to store the timestamps as Date values in the database. This wouldn't normally be a problem except when a query returns a Date value, it will show up as an ISO Datestring in the response of the query and not as a timestamp.

Of course, this feature is not very extensible because we will obviously need a completely different approach for Knex data sources, so I am open to using cron or some equivalent approach.

wtrocki:
Persistent Cron jobs would be the key. TTL is internal - no need to return it

ssd71:
Cool! I'll update the issue!

ssd71:
I have been ruminating quite a bit about this(quite honestly a bit lost here :D) and I'd love it if you could just give some feedback about this:

1.  A place I could see cron jobs happening is putting a script in templates or something similar so there is a separate 'entity' that only manages the database, wdyt? Does this sound wasteful?

> TTL is internal

2. Do you mean that we should have a TTL field in documents that says when to expire documents?

wtrocki:
`Soft deletes` do not delete data from the collection. For some use cases that is ok, but in cases where data is deleted frequently, it can be a problem. If we keep deleted data indefinitely it can lead to overgrown database and increased memory usage. To prevent from that we can enable configurable job (for example once a week will delete week old `soft deletes`). There are numerous npm libraries that support this use case (it is very common to prune some data on timely basis). I do not want to give you solution on silver platter but this should be relatively simple to do with some library

ssd71:
Yep, I was a bit concerned about concurrency troubles you see :D, I will spike and ping you on it later today, thanks!

wtrocki:
[...]

PR #1515, opened 2020-06-22
Add ttl to datasync mongo

ssd71:
This PR adds TTL for datasync mongo, It is a work in progress, related to #1472

### DIFF
--- a/docs/datasync/datasync-soft-delete.md
+++ b/docs/datasync/datasync-soft-delete.md
@@ -12,13 +12,12 @@ Start off with the official Graphback template for MongoDB [*here*](https://GitH
 
 ### Annotate the required models
 
-Add the `versioned` and `delta` annotations to your model(s) in your GraphQL SDL found in the `model` folder:
+Add the `@datasync` annotation to your model(s) in your GraphQL SDL found in the `model` folder:
 
 ```graphql
 """ 
 @model
-@versioned
-@delta 
+@datasync
 """
 type Comment {
   id: ID!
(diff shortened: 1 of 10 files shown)


ssd71:
For implementing this, I am using [TTL Indexes](https://docs.mongodb.com/manual/core/index-ttl/) from MongoDB, I think while this is not available for Postgres, It is the most optimal approach for MongoDB.

I was confused if we should use cron for this: If there are multiple instances of Graphback running, we might run into issues with scaling, although to be fair, if we might use a filter to delete documents(maybe `_ttl < Date.now()`?), I don't see many issues with this

WDYT

wtrocki:
Now I get `concurency`. The way I seen it is that we have one job starting once some time and it it checks TTL value on deleted fields and prune them. TTL indexes are more sophisticated mechanism. Any benefits of that?

wtrocki:
So we need job that does something - and we will get other kinds of jobs in future that TTL indexes will not support

ssd71:
I see what you mean, thanks :D

docs/datasync/datasync-soft-delete.md
```diff
@@ -61,22 +68,17 @@ type CommentDeltaList {
...
   lastSync: String
 }
-
-type Query {
-  ...
-  syncComments(lastSync: String!, filter: CommentFilter): CommentDeltaList!
-}
 ```
 
- The `@delta` annotation adds a `sync` query or a delta query as shown above. This allows you to get all the changed(updates and deletes) documents in a collection since the `lastSync` timestamp. Internally this uses the `updatedAt` database field to check if any documents in the database have been modified, by comparing client provided `lastSync` timestamp value. The documents to be returned in a `sync` query may also be filtered, so you can, for example, only `sync` comments on a specific post.
```
wtrocki, line 71: I will have 

# Advanced  

chapter and then mention delta there.

ssd71:
Closing this for now as new approach needed, will open new PR with forementioned approach

wtrocki:
@ssd71 I would rather keep the code and PR for context but it is ok for you to manage things you want to work.

PR #1518, opened 2020-06-22
fix(graphql-serve): TypeError: get_port_1.default is not a function

craicoverflow:
This fixes a bug in `graphql-serve@0.14.0-alpha6`

```
❯ gqls serve .                         

No port number specified.
Starting server on random available port...

(node:307354) UnhandledPromiseRejectionWarning: TypeError: get_port_1.default is not a function
    at GraphbackServer.<anonymous> (/home/ephelan/code/aerogear/graphback/packages/graphql-serve/dist/GraphbackServer.js:31:52)
```


craicoverflow:
@wtrocki we might need alpha7

PR #1519, opened 2020-06-22
feat: allow specifying of file name in outputPath option

### LINKED ISSUE #1512: graphback-schema plugin outputPath specify filename
As far as I can tell, the `outputPath` for the `graphback-schema` plugin only allows me to specify a path, and not a filename. I would like the option to specify an exact filename, and not have it necessitate creating a directory. Right now I have to have the generated schema in its own directory. Ideally for me, I want my model and generated schema files in the same directory as siblings

craicoverflow:
Fixes #1512 

## Description

This change allows the specifying of the schema in a single config option, `outputPath`:

```ts
const schemaPlugin = new SchemaCRUDPlugin({
  outputPath: './path/to/my/schema.js'
});
```

You can still let graphback create a default file type with:

```ts
const schemaPlugin = new SchemaCRUDPlugin({
  outputPath: './path/to/schema'
});
```

This creates `/path/to/schema/schema.graphql`

This is a breaking change because it removed the `format` and `outputFileName` options.

### DIFF
--- a/docs/plugins/graphback-schema.md
+++ b/docs/plugins/graphback-schema.md
@@ -21,24 +21,14 @@ npm install @graphback/codegen-schema
 
 ### Configuration
 
-`SchemaCRUDPlugin` can be configured to write your generated schema to a file:
+By default `SchemaCRUDPlugin` will not generate any schema file, instead creating an in-memory schema. You might sometimes need the schema file for compatibility with other libraries or frameworks, so Graphback allows you to automate the persistence of the schema to a file in your project:
 
 
 ```ts
   /**
-   * Output format for schema
-   */
-  format: 'ts' | 'js' | 'graphql',
-
-  /**
    * RelativePath for the output files created by generator
    */
   outputPath: string
-
-  /**
-   * Name of the output file (by default `schema`)
-   */
-  outputFileName?: string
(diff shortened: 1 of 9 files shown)


integration/tests/runtime-workflow.ts
```diff
@@ -40,23 +72,22 @@ beforeAll(async () => {
         password: "postgres",
         database: "users",
         host: "localhost",
-        port: 5432
+        port: process.env.PORT || 5432
```
craicoverflow, line 75: Now you can run integration tests locally without going in to code :)

docs/plugins/graphback-schema.md
```diff
@@ -21,24 +21,14 @@ npm install @graphback/codegen-schema
...
 ```ts
-  /**
-   * Output format for schema
-   */
-  format: 'ts' | 'js' | 'graphql',
-
   /**
    * RelativePath for the output files created by generator
    */
   outputPath: string
```
wtrocki, line 31: outputFile?

craicoverflow: But you can just specify the folder also, which is not a file. Path is still technically correct for both scenarios.

PR #1525, opened 2020-06-23
Add TTL for datasync

ssd71:
This PR is meant for discussion of and eventually implementation of an approach to add TTLs to DataSync and possibly to add feasible infrastructure for running jobs in vanilla Graphback

### DIFF
--- a/packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
+++ b/packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
@@ -1,8 +1,8 @@
 import { Db } from 'mongodb';
 import { ModelDefinition, defaultTableNameTransform } from '@graphback/core';
 import { MongoDBDataProvider, GraphbackDataProvider } from "@graphback/runtime-mongo"
-import * as Agenda from "agenda";
 import { isDataSyncModel } from '../util';
+import { PruneScheduler, PruneConfiguration, TTLMap } from "../scheduler";
 import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
 
 /**
@@ -10,44 +10,17 @@ import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
  *
  * @param {Db} db - MongoDb connection
  */
-export function createDataSyncMongoDbProvider(db: Db, TTL:number = -1, jobConfig: Agenda.AgendaConfiguration = {}): (...args: any[]) => GraphbackDataProvider {
-  if (jobConfig?.db === undefined && jobConfig?.mongo === undefined) {
-    jobConfig.mongo = db
-  }
-  const agenda = new Agenda(jobConfig);
-  agenda.start().catch((e: any) => {
-    console.error(`Error occurred while starting job service: ${e}`)
-  })
+export function createDataSyncMongoDbProvider(db: Db, ttlMap?:TTLMap, pruneConfig: PruneConfiguration = {}): (...args: any[]) => GraphbackDataProvider {
+  const scheduler = new PruneScheduler(pruneConfig, db);
   
   return (model: ModelDefinition): GraphbackDataProvider => {
     if (isDataSyncModel(model)) {
-      if (TTL >= 0) {
-        const modelName = model.graphqlType.name;
+
+      const modelName = model.graphqlType.name;
+      const modelTTL = ttlMap[modelName];
+      if (modelTTL >= 0) {
         const collectionName = defaultTableNameTransform(modelName, 'to-db');
-        const pruneJobName = `prune:${collectionName}`;
-
-        // Define prune job processor for a collection
-        agenda.define(pruneJobName, async (_: any, done: (e: any) => void) => {
-          db.collection(collectionName).deleteMany({
-            _deleted: true,
-            updatedAt: {
-              $lte: Date.now() - (TTL * 1000)
-            }
-          })
-            .then(() => {
-              done(undefined);
-            })
-            .catch((e: any) => {
-              console.error('prune job failed: ', JSON.stringify(e, undefined, 4));
-              done(e);
-            })
-        });
-
-        // Run prune job every 10 seconds
-        agenda.every('10 seconds', pruneJobName).catch((e:any) => {
-          console.error(`Error: prune job for ${modelName}`)
-        })
-
+        scheduler.defineAndStartPruneJob(collectionName, modelTTL);
       }
 
       return new DataSyncMongoDBDataProvider(model.graphqlType, db)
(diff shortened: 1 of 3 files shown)


ssd71:
A workaround to test it out:
- replace `index.ts` in mongo template with [this](https://gist.github.com/ssd71/f9bf3a3865e95a958d95845a0d857269)
Note that the second argument to createDataSyncMongoDbProvider is the TTL offset in seconds:
https://gist.github.com/ssd71/f9bf3a3865e95a958d95845a0d857269#file-index-ts-L31
- add `@datasync` annotation to your model

The prune job is being run every 10 seconds:
https://github.com/aerogear/graphback/blob/f56a5adf2335b35dab6cee62a3d254aa5727bb39/packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts#L47

Would be nice if we could also agree on what to change this interval to :D

wtrocki:
This looks like good prototype of what we should do. 
Challenge I see now is how to make this easy to configure. 
TTL per entity type or global? Separate config for the prune job trigger time etc.

ssd71:
Yes, I am working on it, if you have any pointers feel free to tell me :smile: 
- TTL is global for now, I am thinking it would be better to have it per entity, what do you think? That way we would have to specify it for every model
- Yes, we really might need configurable trigger time

ssd71:
- TTL is per entity now
- prune Interval is configurable now

See [here](https://github.com/aerogear/graphback/pull/1525#issuecomment-647926810) for testing, will keep updating this gist from here on out

packages/graphback-datasync/src/scheduler.ts
```diff
@@ -0,0 +1,76 @@
+import * as Agenda from "agenda";
+import { Db, MongoClient } from "mongodb";
+
+export interface PruneConfiguration {
+  agendaConfig?: Agenda.AgendaConfiguration
+  pruneInterval?: string | number
+}
+
+export interface TTLMap {
```
machi1990, line 9: How is this constructed? 

Via something like a `@ttl` annotation in the model? Or an argument for `@datansync` annotation?

ssd71: It is constructed in code, passed as argument to createDataSyncMongoDbProvider
But what you said sounds really cool! I think it would be nice if it was an argument to datasync :smiley: 
wdyt @wtrocki

packages/graphback-datasync/src/scheduler.ts
```diff
@@ -0,0 +1,76 @@
...
+export class PruneScheduler {
+  public objectDb: Db;
+  public pruneInterval: string | number;
+  public agenda: Agenda;
+
+  public constructor(config: PruneConfiguration, db: Db) {
+    this.agenda = new Agenda(config.agendaConfig);
+    if (config.agendaConfig?.db === undefined && config.agendaConfig?.mongo === undefined) {
+      this.agenda.mongo(db);
+    }
```
machi1990, line 25: This will create a collection in the db to track scheduled jobs, which is okay since we want the jobs to be distributed jobs and be run once even if we have multiple instances. Does the lib take care of wiping out old scheduled jobs?

packages/graphback-datasync/src/scheduler.ts
machi1990: > It is constructed in code, passed as argument to createDataSyncMongoDbProvider

I saw the part where it is passed as argument in `createDataSyncMongoDbProvider` but not the part where the ma is constructed.

packages/graphback-datasync/src/scheduler.ts
ssd71: Nope, we have to do that ourselves using [purge](https://www.npmjs.com/package/agenda#purge)

ssd71: I haven't added that functionality yet, will try something in next push regarding this

packages/graphback-datasync/src/scheduler.ts
ssd71: Sorry, I see what you mean, we are not constructing it,  I am thinking about constructing it from something like: `@datasync(TTL: 3600)` should be good, what do you think?

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
```diff
@@ -1,20 +1,31 @@
...
+import { PruneScheduler, PruneConfiguration, TTLMap } from "../scheduler";
 import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
 
 /**
  * Creates a new Data synchronisation data provider for MongoDb
  *
  * @param {Db} db - MongoDb connection
  */
-export function createDataSyncMongoDbProvider(db: Db): (...args: any[]) => GraphbackDataProvider {
+export function createDataSyncMongoDbProvider(db: Db, ttlMap?:TTLMap, pruneConfig: PruneConfiguration = {}): (...args: any[]) => GraphbackDataProvider {
```
wtrocki, line 13: Instead of map I will carry this information in model.
Benefit is that we can expose this later in schema or something or have global TTL for the moment

packages/graphback-datasync/src/scheduler.ts
wtrocki: Yep. see my comment above. That will be exactly the target, however TTL will be specific to some strategies so we might need to figure that out later.

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
```diff
@@ -1,20 +1,31 @@
...
+  const scheduler = new PruneScheduler(pruneConfig, db);
+  
   return (model: ModelDefinition): GraphbackDataProvider => {
     if (isDataSyncModel(model)) {
+
+      const modelName = model.graphqlType.name;
+      const modelTTL = ttlMap[modelName];
+      if (modelTTL >= 0) {
+        const collectionName = defaultTableNameTransform(modelName, 'to-db');
```
craicoverflow, line 22:
```suggestion
        const collectionName = getTableName(model.graphqlType);
```

https://github.com/aerogear/graphback/blob/9df2aa574299db464b7f77293bb493f1f799c522/packages/graphback-core/src/db/buildModelTableMap.ts#L27

packages/graphback-datasync/src/scheduler.ts
craicoverflow: +1 on `@datasync(ttl: 3600)`. It looks like this won't work without the `datasync` annotation anyway, so it should not be as a separate annotation.

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
ssd71: Yup! Thanks!

wtrocki:
@ssd71 Let's revisit this again

wtrocki:
Seening this PR I'm glad we did not merged it. Closing this PR

PR #1529, opened 2020-06-23
Remove runtime

wtrocki:
TL;DR deprecate runtime and remove all related classes from it

We moving every part into core.
Our entire codebase have been reimporting
core/runtime form different packages making it really hard
to get what package offers what class and
also not being able to do proper code jumps in the VSCode.

This change:
- Moves all runtime Services into core
- Removes runtime from all packages
- Deprecates runtime itself

## Progress 

- [x] Make required code changes
- [ ] fix unit tests and integration tests

I'm not going to update docs in this PR as our docs require a more comprehensive approach


wtrocki:
Note that I split this with the keycloak PR so we can later ignore this changes when reviewing keycloak changes, which will be relatively small. All changes here are very cosmetic and mostly related to the import types and exposing additional types.

Tests will fail due to missing keycloak implementation.

machi1990:
I noticed test failures but you have them on your TODO. Otherwise, looks good and I'll prefer we get this in first before #1533

packages/graphback-keycloak-authz/src/KeycloakCrudService.ts
```diff
@@ -32,11 +27,11 @@ export type KeycloakCrudServiceOptions = {
 export class KeycloakCrudService<T = any> implements GraphbackCRUDService<T> {
 
   private authConfig: CrudServiceAuthConfig;
-  private wrappingService: GraphbackCRUDService;
+  private protectedService: GraphbackCRUDService;
```
machi1990, line 30: `securedService` maybe, so that to not confuse it with `protected` access level.

packages/graphback-runtime-knex/tests/service/DefaultCRUDServiceTest.ts
```diff
@@ -117,7 +116,7 @@ test('find Todo by text', async () => {
 
   const todoResults = await services.Todo.findBy({
     text: { eq: 'the second todo' },
-  }, {graphback: {services: {}, options: { selectedFields: ["id"]}}});
+  }, { graphback: { services: {}, options: { selectedFields: ["id"] } } });
```
machi1990, line 119: Sad that when I ran `yarn lint-fix` the linter does not pick these.

PR #1533, opened 2020-06-23
feat: return pagination info if requested

### LINKED ISSUE #1517: Find and return pagination info only when requested
Follows up https://github.com/aerogear/graphback/issues/793#issuecomment-631384247 

Opening this issue to leave the PR https://github.com/aerogear/graphback/pull/1509 only about adding the means to select fields in database only if they are requested. I'll work on this issue after #1509 is merged so that we properly close the ideas we had for #793.

machi1990:
Fixes #1517

### DIFF
--- a/packages/graphback-core/src/runtime/CRUDService.ts
+++ b/packages/graphback-core/src/runtime/CRUDService.ts
@@ -83,7 +83,16 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
 
   //tslint:disable-next-line: no-any
   public async findBy(filter: any, context: GraphbackContext, orderBy?: GraphbackOrderBy, page?: GraphbackPage): Promise<ResultList<T>> {
-    const items = await this.db.findBy(filter, context, orderBy, page);
+    let count: number;
+    if (context.graphback.options.aggregations?.count) {
+      count = await this.db.count(filter);
+    }
+
+    let items: T[];
+
+    if (context.graphback.options.selectedFields.length) {
+      items = await this.db.findBy(filter, context, orderBy, page);
+    }
 
     // set page values for returned object
     const resultPageInfo = {
@@ -93,6 +102,7 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
 
     return {
       items,
+      count,
       offset: 0,
       ...resultPageInfo
     }
--- a/packages/graphback-core/src/runtime/interfaces.ts
+++ b/packages/graphback-core/src/runtime/interfaces.ts
@@ -12,7 +12,10 @@ export interface GraphbackServiceConfigMap {
  * Contains resolver options
  */
 export interface GraphbackResolverOptions {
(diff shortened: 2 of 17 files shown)


machi1990:
Opened as draft to see if this does not conflict with #1529

wtrocki:
Finished reviewing and verifying now

packages/graphback-core/src/runtime/interfaces.ts
```diff
@@ -12,7 +12,10 @@ export interface GraphbackServiceConfigMap {
  * Contains resolver options
  */
 export interface GraphbackResolverOptions {
-  selectedFields: string[]
+  selectedFields: string[],
+  aggregations?: {
```
wtrocki, line 16: I will add a comment to this field as it is not obvious what it does.

+100 to have it wrapped.

machi1990: The idea is that this will be a container for aggregations operations.

machi1990: Something along those lines will be okay?

wtrocki:
How count is being executed. I see this added into dataprovider but I do not see it being called?
Any example query?

packages/graphback-core/src/runtime/interfaces.ts
wtrocki: Perfect!

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -83,7 +83,16 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
 
   //tslint:disable-next-line: no-any
   public async findBy(filter: any, context: GraphbackContext, orderBy?: GraphbackOrderBy, page?: GraphbackPage): Promise<ResultList<T>> {
-    const items = await this.db.findBy(filter, context, orderBy, page);
+    let count: number;
+    if (context.graphback.options.aggregations?.count) {
+      count = await this.db.count(filter);
+    }
```
machi1990, line 89: @wtrocki this is where it will be executed

machi1990: And this will be executed if the gql asks for it. 
e.g 
```
findNotes {
 items {
  id
}
count
}
```
Will execute the count query in db but 
```
findNotes {
 items {
  id
}
}
```

won't

machi1990: Something similar will happen for `items`, if it is not asked in the incoming gql, we won't send the query to db.

wtrocki:
Sorry - Been on meetings.

wtrocki:
Amazing work! this is not only very very flexible but also it is done in such a way it is easy to move to different mechanisms. Really really good job!

machi1990:
Thanks for the swift review. @wtrocki @craicoverflow

Issue #1526, opened 2020-06-23
Documentation refactor


craicoverflow:
There have been a number of major changes between between 0.12.x and 0.14.x.

The documentation and website requires major restructuring/refactoring to adapt to the significant changes in Graphback.

This is a chance to simplify and focus the documentation to make it the best experience possible for users. This should be easier to achieve, since we now have reduced file generation to client documents and schema (optionals).

To get an idea of what has changed between these releases check https://github.com/aerogear/graphback/blob/master/docs/releases.md#0140

## Sitemap Progress

- [x] Getting Started
    - [x] Why use Graphback? #1636
    - [x] Creating a new application #1636
    - [x] Adding Graphback to your project #1636
- [x] Data Model
    - [x] Common #1651 
    - [x] For Postgres #1651 
    - [x] For MongoDB #1651 
- [x] CRUD #1655
    - [x] Introduction #165
    - [x] Queries #1655
        - [x] findMany
        - [x] getOne #1663 
    - [x] Mutations #1633
        - [x] createOne #1633
        - [x] updateOne #1633
        - [x] deleteOne #1633
    - [x] Subscriptions
        - [x] onCreate
        - [x] onUpdate
        - [x] onDelete
- [x] Databases.
    - [x] Introduction
    - [x] MongoDB
        - [x] Installation
        - [x] Connecting to the database via mongodb package
        - [x] Provider usage
        - [x] ~~Mongo specific annotation references?~~ Covered in the **Data Model** page.
    - [x] PostgreSQL
        - [x] Installation 
        - [x] Connecting to the database via Knex. Show example of `db` pooling etc
        - [x] Provider Usage
        - [x] ~~Postgres annotation references?~~ Covered in the **Data model** page
        - [x] ~~Migrations~~ Covered in the**GraphQL Migrations** page, We'll add a link to this page
        - [x] ~~Custom scalar mapping~~ Covered in the **GraphQL Migrations** page
    - [x] SQLLite
         - [x] Installation
         - [x] Provider Usage 
- [x] Subscriptions
    - [x] Live Updates
- [x] Authentication
    - [x] Introduction
    - [x] Keycloak
- [x] Data Synchronization https://github.com/aerogear/graphback/pull/1665
    - [x] Introduction
    - [x] Soft deletes
- [x] Plugins
    - [x] SchemaCRUDPlugin
    - [x] ClientCRUDPlugin?
    - [x] Adding your own plugin https://github.com/aerogear/graphback/pull/1706
[...]

PR #1536, opened 2020-06-23
feat: se `parseMetadata` syntax for annotations

### LINKED ISSUE #1449: Migrate to use `parseMetadata` in annotations
`parseAnnotations` format of graphql-metadata is deprecated and will be removed in the future. `parseMetadata` introduces a new syntax, which we should migrate annotations to. This should happen before 0.14.0 stable as it is a breaking change.

Docs should also be updated to reference this format.

craicoverflow:
Resolves #1449 

This PR updates all references to the old annotation syntax: `@db.type: 'timestamp'` with the new format: `@db(type: 'timestamp')`.

Key benefits:

- More concise (can be done in one line instead of a new line for each annotation)
- Familiar syntax. This copies the GraphQL directive syntax, making it easy for people to understand

## Tasks

- [x] Remove to `parseAnnotations` and `parseMarker`
- [x] Update tests

## Verification

Functionality remains the exact same, the only updates to the tests were the schema snapshots which have the new annotation syntax.

## Other

Not doing anything with docs since this will be done in https://github.com/aerogear/graphback/issues/1526


docs/metadata.md
```diff
@@ -151,7 +151,7 @@ type Post {
 
 | Argument | Description | Example |
 |-|-|-|
-|`key`| Optionally specifies the name of foreign key field on the foreign object. Accepts a string value. Defaults to `<typeName>Id` | `@oneToMany field: 'user' key: 'user_key'`|
+|`key`| Optionally specifies the name of foreign key field on the foreign object. Accepts a string value. Defaults to `<typeName>Id` | `@oneToMany(field: 'user', key: 'user_key')`|
```
machi1990, line 154:
```suggestion
|`key`| Optionally specifies the name of foreign key field on the foreign object. Accepts a string value. Defaults to `<typeName>Id` | `@oneToOne(field: 'user', key: 'user_key')`|
```

packages/graphback-core/src/db/buildModelTableMap.ts
```diff
@@ -37,15 +37,15 @@ export function getTableName(model: GraphQLObjectType): string {
...
+ * Checks for the `@db(name)` annotation for a customised name
+ *
+ * @param field
  */
 export function getColumnName(field: GraphQLField<any, any>): string {
   let columnName = field.name;
 
   const dbAnnotations = parseDbAnnotations(field);
-  if (dbAnnotations.name) {
+  if (dbAnnotations && dbAnnotations.name) {
```
machi1990, line 48:
```suggestion
  if (dbAnnotations?.name) {
```

wtrocki:
@craicoverflow Is that changing also dbmigrations (OK I see it does)

wtrocki:
I think we need to add those to:
https://github.com/aerogear/graphback/tree/use-new-annotations/packages/graphback-cli/src/templates/resources/models

Most of the models there do not even work without this change

wtrocki:
I think we need to add those to:
https://github.com/aerogear/graphback/tree/use-new-annotations/packages/graphback-cli/src/templates/resources/models

Most of the models there do not even work without this change

craicoverflow:
> I think we need to add those to:
> [https://github.com/aerogear/graphback/tree/use-new-annotations/packages/graphback-cli/src/templates/resources/models](https://github.com/aerogear/graphback/tree/use-new-annotations/packages/graphback-cli/src/templates/resources/models?rgh-link-date=2020-06-24T10%3A35%3A02Z)
> 
> Most of the models there do not even work without this change

Wow, yeah good spot. Will address.

wtrocki:
Not all docs are updated. Relationships.md for example - Would you like to have follow up PR on this?

packages/graphback-core/src/db/buildModelTableMap.ts
wtrocki: Using oportunity. Do we have the same key for the field name and table name?

craicoverflow: Yep, `@db(name: 'fieldOrTableName')`.

craicoverflow:
> Not all docs are updated. Relationships.md for example - Would you like to have follow up PR on this?

Docs are being rewritten in #1526 so there is not much point in updating it here.

packages/graphql-migrations/src/util/parseAnnotations.ts
```diff
@@ -0,0 +1,14 @@
+import { parseMetadata } from 'graphql-metadata'
+
+/**
+ * Wraps `parseMetadata` to return empty object instead of undefined
+ * As lots of graphql-migration uses this.
+ * 
+ * @param namespace
+ * @param description
+ */
+export function parseAnnotations(namespace: string, description: string): any {
```
machi1990, line 10: A good candidate for the core, we already have https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/annotations/parser.ts#L4

craicoverflow: Actually I will remove `parseDbAnnotations`. I added this in `graphql-migrations` because the library heavily relies on the assumption that an empty object will be returned for `@db` so it does not do any checking for undefined. `parseMetadata` returns `true` so these approaches are incompatible. This function wraps `parseMetadata` to ensure the output stays the same in graphql-migrations. I will change the name to make it clear.

PR #1544, opened 2020-06-24
feat: add proxy service for simplicity of extending service chain

wtrocki:
Idea here is to simplify creation of the service chains even further. 
Currently you need to creat entire class witch is very verbose. 
Adding proxy service to the core and extending it is very simple and can be reused to build long chain of services without implementing entire API

### DIFF
--- a/packages/graphback-core/src/runtime/GraphbackProxyService.ts
+++ b/packages/graphback-core/src/runtime/GraphbackProxyService.ts
@@ -1,0 +1,54 @@
+import { GraphbackCRUDService, ResultList } from './GraphbackCRUDService';
+import { GraphbackOrderBy, GraphbackPage } from '.';
+
+/**
+ * This custom CRUD Service shows another potential way to add auth
+ *
+ * This is actually quite nice and clean but it does not allow for field level auth.
+ * It's still a possibility that we could go with though!
+ */
+export class GraphbackProxyService<T = any> implements GraphbackCRUDService<T> {
+
+  protected proxiedService: GraphbackCRUDService;
+
+  public constructor(service: GraphbackCRUDService) {
+    this.proxiedService = service;
+  }
+
+  public create(data: T, context: any): Promise<T> {
+    return this.proxiedService.create(data, context);
+  }
+
+  public update(data: T, context: any): Promise<T> {
+    return this.proxiedService.update(data, context);
+  }
+
+  public delete(data: T, context: any): Promise<T> {
+    return this.proxiedService.delete(data, context);
+  }
+
+  public findOne(args: any, context: any): Promise<T> {
+
+    return this.proxiedService.findOne(args, context);
+  }
+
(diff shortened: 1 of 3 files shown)


packages/graphback-core/src/runtime/index.ts
```diff
@@ -4,4 +4,5 @@ export * from './GraphbackDataProvider'
 
 export * from "./CRUDService"
 export * from "./createCRUDService"
-export * from "./NoDataError"
\ No newline at end of file
+export * from "./NoDataError"
+export * from "./GraphbackProxyService"
```
machi1990, line 8: Hmm where does the ProxyService class come from?

machi1990: CI failure https://app.circleci.com/pipelines/github/aerogear/graphback/5195/workflows/47abd466-58b7-4997-9920-616de11ad610/jobs/5296/steps related.

wtrocki: Yep.. It wasn't staged. My bad

packages/graphback-core/src/runtime/GraphbackProxyService.ts
```diff
@@ -0,0 +1,54 @@
...
+  public update(data: T, context: any): Promise<T> {
+    return this.proxiedService.update(data, context);
+  }
+
+  public delete(data: T, context: any): Promise<T> {
+    return this.proxiedService.delete(data, context);
+  }
+
+  public findOne(args: any, context: any): Promise<T> {
+
```
machi1990, line 31:
```suggestion
```

packages/graphback-core/src/runtime/GraphbackProxyService.ts
```diff
@@ -0,0 +1,54 @@
+import { GraphbackCRUDService, ResultList } from './GraphbackCRUDService';
+import { GraphbackOrderBy, GraphbackPage } from '.';
+
+/**
+ * This custom CRUD Service shows another potential way to add auth
+ *
+ * This is actually quite nice and clean but it does not allow for field level auth.
+ * It's still a possibility that we could go with though!
+ */
```
machi1990, line 9: Comments needs update

packages/graphback-core/src/runtime/GraphbackProxyService.ts
```diff
@@ -0,0 +1,53 @@
...
```
wtrocki, line 9:
```suggestion
/**
 * ProxyService that can be used by any services that wish to extend 
 * Graphback functionality.
 * Service works by proxying method requests to another service or 
 * datastore.
 */
```

wtrocki:
Note I do not want to document this yet. Going to try this approach with multiple services and see if that will be useful. This stuff is more for us as maintainers to be able to build another set of services like AMQ and build some wrappers etc.

PR #1545, opened 2020-06-24
fix: Remove invalid typings from DataSync service

wtrocki:


packages/graphback-datasync/src/services/createDataSyncService.ts
```diff
@@ -15,7 +15,7 @@ export interface CreateDataSyncCRUDServiceOptions {
  *
  * @param config
  */
-export function createDataSyncCRUDService(config: CreateDataSyncCRUDServiceOptions): (...args: any[]) => GraphbackCRUDService {
```
craicoverflow, line 18: So is it a non-typed function now?

wtrocki: Proper types are inferred - specifying return types became antipattern

Issue #1555, opened 2020-06-25
graphql-compose not found when using migrations


lastmjs:
On alpha 4, when trying to use the migration API, an error was thrown saying that `graphql-compose` could not be found. I had to manually install it to get the migrations to progress.

PR #1559, opened 2020-06-25
fix(migrations): graphql-compose not found when using migrations

machi1990:
Make the dependency a prod one since it is required in the SchemaTransformer.ts file

Fixes https://github.com/aerogear/graphback/issues/1555


packages/graphql-migrations/src/plugin/SchemaTransformer.ts
```diff
@@ -15,12 +15,8 @@ export const removeNonModelTypesFromSchema: SchemaTransformerPlugin = {
     }
 
     const schemaComposer = new SchemaComposer(schema)
-    for (const userType of nonModelTypes) {
-      schemaComposer.delete('StringInput')
-    }
```
machi1990, line 20: @wtrocki @craicoverflow this was basically repeating the same operation, so no need to use the loop

Issue #1556, opened 2020-06-25
Alpha 7 resolvers is an array


lastmjs:
Is there a reason that the resolvers returned from buildGraphbackAPI are now an array in alpha 7? This is another breaking change that was not documented in the release notes (unless I'm missing where the release information is).

ssd71:
Thank you for the issue! Sorry for the lack of release notes! I think this is also under discussion in #1546 feel free to shift this discussion there if you feel it is relevant

craicoverflow:
Fixed in #1560 - will be available in the next alpha release.

PR #1560, opened 2020-06-25
fix(core): return resolvers as object

### LINKED ISSUE #1546: Graphback returns resolvers as array
When working with resolves we see that graphback resolvers are returned as array where most of the people will expect this to be just single object. Any reason why we have this. 

Most of the frameworks will return error when dealing with array.
[Error: "0" defined in resolvers, but not in schema]

craicoverflow:
Fixes #1546 #1556 

Returns resolvers as an object instead of an array.

### DIFF
--- a/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
+++ b/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
@@ -2,6 +2,7 @@ import { GraphQLObjectType, GraphQLSchema } from 'graphql'
 import { parseMetadata } from 'graphql-metadata'
 import { getUserTypesFromSchema } from '@graphql-toolkit/common'
 import { IResolvers } from '@graphql-tools/utils'
+import { mergeResolvers } from '@graphql-tools/merge';
 import { RelationshipMetadataBuilder, FieldRelationshipMetadata } from '../relationships/RelationshipMetadataBuilder'
 import { GraphbackCRUDGeneratorConfig } from './GraphbackCRUDGeneratorConfig'
 import { GraphbackGlobalConfig } from './GraphbackGlobalConfig'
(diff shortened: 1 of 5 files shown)


templates/ts-apollo-postgres-backend/src/index.ts
```diff
@@ -26,7 +27,10 @@ const db = connectDB()
 const dbConfig = loadDBConfig()
 
 const { typeDefs, resolvers, contextCreator } = buildGraphbackAPI(modelDefs, {
-  dataProviderCreator: createKnexDbProvider(db)
+  dataProviderCreator: createKnexDbProvider(db),
+  plugins: [
```
wtrocki, line 31: Let's remove it.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
```diff
@@ -44,11 +44,15 @@ export class GraphbackCoreMetadata {
 
   public addResolvers(resolvers: IResolvers) {
     if (resolvers) {
-      this.resolvers.push(resolvers);
+      const mergedResolvers = [
+        this.resolvers,
+        resolvers
+      ]
+      this.resolvers = mergeResolvers(mergedResolvers)
```
wtrocki, line 51: A kinda fancy version of Object.assign ;P

templates/ts-apollo-postgres-backend/src/index.ts
craicoverflow: Done, put it in to test multiple sets of resolvers.

packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
craicoverflow: Object.assign does not merge deep, neither does the spread operator. So you would have to do:

```ts
this.resolvers = {
  Query: {
    ...this.resolvers.Query,
    ...resolvers.Query
  },
  Mutation: {
    ...this.resolvers.Mutation,
    ...resolvers.Mutation
  }
}
```

And it would get trickier trying to merge custom root resolvers.

PR #1564, opened 2020-06-25
fix: remove schema transformer plugin

craicoverflow:
Follow up from https://github.com/aerogear/graphback/pull/1559

This plugin engine was never fully implemented, and is not used anywhere. It was a spike that got merged.


machi1990:
Was about to do the same but was not so sure.

Issue #1547, opened 2020-06-24
Requiring _id field for mongo instead of `id`


wtrocki:
`Graphback was unable to create the specified indexes: E11000 duplicate key error collection: showcase.comment index: id_1 dup key: { : null }. If all else fails, try recreating the index manually.
`

Schema:

```graphql
"""
@model
@datasync
"""
type Task {
  """@id"""
  id: ObjectID!
}
```

Can be replicated in datasync starter.

wtrocki:
CC @ssd71 

It is kinda critical issue as it is not possible to issue any query on this:

```
 "message": "E11000 duplicate key error collection: showcase.task index: id_1 dup key: { : null }",
```

ssd71:
Looked around a bit, maybe it couldn't create index because there were multiple `id: null` documents in the collection?

I will have to take closer look if that's not it

ssd71:
Either way, I think graphback should handle this error a bit better

wtrocki:
That is what I have in db. Basically it fails after creating single document

<img width="557" alt="Screenshot 2020-06-24 at 17 33 48" src="https://user-images.githubusercontent.com/981838/85595606-ec7d0580-b640-11ea-871a-425e179161b1.png">

ssd71:
So I replicated this and I think I am onto something here:

- The error in insert occurs because the document that is inserted has no id field, effectively treated as `id: null`, As an existing document also does not have an id field (hence `id: null`) the insertion fails as `@id` creates a unique index
- I am sure you already know this, but the `id` field in the GraphQL SDL is being mapped to `_id` for everything from inserts to deletes, but the index is being created on the `id` field, so the thing is even if we have a custom primary key, in the current state of Graphback Mongo support, the name of custom primary key *has* to be other than `id`
[...]

PR #1565, opened 2020-06-25
fix: remove id indexes from mongo

ssd71:
Towards #1547

### DIFF
--- a/packages/graphback-runtime-mongodb/package.json
+++ b/packages/graphback-runtime-mongodb/package.json
@@ -29,7 +29,7 @@
     "graphql": "15.1.0",
     "jest": "26.1.0",
     "jest-date-mock": "1.0.8",
-    "mongodb-memory-server": "6.6.1",
+    "mongodb-memory-server-global": "6.6.1",
     "rimraf": "3.0.2",
     "ts-jest": "26.1.1",
     "ts-node": "8.10.2",
--- a/packages/graphback-runtime-mongodb/src/utils/createIndexes.ts
+++ b/packages/graphback-runtime-mongodb/src/utils/createIndexes.ts
@@ -42,14 +42,6 @@ export function getIndexFields(baseType: GraphQLObjectType): IndexSpecification[
   Object.keys(fields).forEach((k: string) => {
     const field = fields[k];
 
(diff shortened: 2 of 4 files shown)


wtrocki:
Build failing

wtrocki:
`can create default indexes` needs to be changed as well. Minor fix

ssd71:
The tests are still failing it seems, same as before, it's downloading Mongo memory server:
Screenshot_2020-06-25 build (5345) - aerogear graphback

wtrocki:
We just have removed stuff so this is really strange. 
On long run we need to think if there is Mongo wrapper that do not require Mongo binary.

wtrocki:
Also see: https://github.com/aerogear/graphback/pull/1567/files

craicoverflow:
So strange, this is a very intermittent issue that I was seeing for months when running tests locally, but it all of a sudden went away a couple of weeks ago.

ssd71:
How about this:
https://github.com/nodkz/mongodb-memory-server#mongodb-memory-server-global
> Auto-downloads the latest mongod binary on npm install to: `%HOME%/.cache/mongodb-binaries` / `~/.cache/mongodb-binaries.`

I am not seeing this error locally, perhaps because I have copied the mongo binary to the packages' node_modules

ssd71:
I think this error occurs because mongo-memory-server cannot find the mongod binary in the node_modules of the package folder, because it is in the root node_modules

ssd71:
Still doesn't explain why we weren't seeing it for a few weeks, and it just keeps popping up every now and then

ssd71:
image

Please verify @wtrocki @machi1990 @craicoverflow

wtrocki:
@ssd71 I see changes but I have not way to verify it without sample app. Let's merge and invest time into integration tests and datasync sample. 

This works for vanilla mongo app now.

wtrocki:
Packages should be cached and this problem should never happen on CI/CD

ssd71:
> @ssd71 I see changes but I have not way to verify it without sample app. Let's merge and invest time into integration tests and datasync sample.

+9000 to DataSync Sample

PR #1567, opened 2020-06-25
test: add mongo db integration test

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1562

- [x] fix failing tests


integration/.graphqlrc.yml
```diff
@@ -18,7 +18,4 @@ extensions:
     ## Codegen plugins
     plugins:
```
machi1990, line 19: This should not be needed but kind of added it here to not ran through this issuehttps://github.com/aerogear/graphback/issues/1296

integration/tests/runtime-workflow-mongo.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    mongoClient = new MongoClient('mongodb://mongodb:mongo@localhost:27017/users?authSource=admin', { useUnifiedTopology: true });
+    await mongoClient.connect();
+    db = mongoClient.db('users');
+    graphbackApi = buildGraphbackAPI(modelText, {
+      dataProviderCreator: createMongoDbProvider(db),
+      plugins: [
+        new SchemaCRUDPlugin({outputPath: "./output-mongo/schema/schema.graphql"}),
+        new ClientCRUDPlugin({format: 'graphql', outputFile: './output-mongo/client/graphback.graphql'})
+      ]
```
machi1990, line 50: Later on we could have the datasync plugin and add related tests here
/cc @ssd71

ssd71: Sounds good

machi1990:
For now the file https://github.com/aerogear/graphback/blob/5b5be08e03f361ebc389eaa038a137f4ba914da0/integration/tests/runtime-workflow-mongo.ts is just a replica of the postgres tests. But I think later on we could generate most of these tests and have them ran against different DB flavours we support.

integration/.graphqlrc.yml
machi1990: Remove this hack once https://github.com/aerogear/graphback/pull/1569 is merged.

wtrocki:
I would recomend to rename runtime-workflow to runtime workflow pg or something

machi1990:
> I would recomend to rename runtime-workflow to runtime workflow pg or something

Good recommendation and this is already the case https://github.com/aerogear/graphback/blob/5b5be08e03f361ebc389eaa038a137f4ba914da0/integration/tests/runtime-workflow-postgres.ts

integration/.graphqlrc.yml
```diff
@@ -15,10 +15,3 @@ extensions:
       subCreate: true
```
wtrocki, line 15: With that in mind we can also have crud config and location in js

wtrocki:
Works for me.
@ssd71 comments?

integration/.graphqlrc.yml
machi1990: To be fair, the file can be removed all together now. Let me do it.

machi1990: Done.

ssd71:
Looks great! :smile:

Issue #1543, opened 2020-06-24
Running `yarn graphback generate` from a fresh generated project using `graphback-cli init` fails


machi1990:
After successful generating a new project using our cli, we have the next steps printed in console 
```
Next Steps:
1. Change directory into project folder - cd test
3. Review Graphback graphqlrc.yml config file
2. Edit the .graphql file inside your model folder.
3. Depending on template you might also need to run generator:
   "graphback generate" command to generate source code
```

However, when running the `yarn graphback generate` command it fails. 

I think it is because the templates are missing the `graphback-cli` package

https://github.com/aerogear/graphback/blob/master/templates/ts-apollo-mongodb-backend/package.json#L14

https://github.com/aerogear/graphback/blob/master/templates/ts-apollo-postgres-backend/package.json#L14

https://github.com/aerogear/graphback/blob/master/templates/ts-react-apollo-client/package.json#L16

/cc @wtrocki @craicoverflow

craicoverflow:
This is because it looks for a `.graphqlrc` file. We have removed that since server generation is completely runtime, however if you want to generate client documents you need a config file.

However, I believe work has begun to create an offix-cli (https://github.com/aerogear/offix/issues/525)
So the `generate` command will be removed I think.

machi1990:
The error message I am seeing is 
```
yarn graphback generate
error Command "graphback" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

craicoverflow:
Oh. Well then you are right about `I think it is because the templates are missing the graphback-cli package`. We should add them in.

wtrocki:
Yep... Amazing catch!

PR #1568, opened 2020-06-25
chore(templates): add graphback-cli as dev dependency

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1543

### DIFF
--- a/templates/ts-apollo-postgres-backend/package.json
+++ b/templates/ts-apollo-postgres-backend/package.json
@@ -31,6 +31,7 @@
     "@types/glob": "7.1.2",
     "@types/graphql": "14.2.3",
     "@types/node-fetch": "2.5.7",
+    "graphback-cli": "0.14.0-alpha4",
     "ts-node": "8.10.2",
     "tslint": "6.1.2",
     "typescript": "3.9.5"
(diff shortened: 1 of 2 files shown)


templates/ts-apollo-postgres-backend/package.json
```diff
@@ -31,6 +31,7 @@
     "@types/glob": "7.1.2",
     "@types/graphql": "14.2.3",
     "@types/node-fetch": "2.5.7",
+    "graphback-cli": "0.14.0-alpha4",
```
craicoverflow, line 34: Actually, now that releases are automated through GitHub, templates will need to be manually adjusted (since I don't think renovate bumps alphas).

So this should be 0.14.0-alpha7?

machi1990: Thought so but I decided to align the version with other graphback packages version. 

Maybe @wtrocki can help here about how the process works?

craicoverflow: I imagine this problem will rectify itself once we are back on stable release, because templates probably should not be alpha/dev under production circumstances.

wtrocki: Templates are par of the lerna and they are symlinked. Make sure you do lerna bootstrap and verify it to confim/

wtrocki: Actually they are symlinked only when they are aligned so @machi1990 is perfect! fixing this issue

Issue #1411, opened 2020-06-09
Cannot use sync from DataSync


ssd71:
On trying the 'sync' query:
```
Please use DataSync provider for delta queries
```

It seems the sync resolver is not actually returning anything:
https://github.com/aerogear/graphback/blob/b7c74ed0515d9e3e1771cbf9cb8db47732ddb98c/packages/graphback-runtime/src/resolvers/LayeredRuntimeResolverCreator.ts#L89-L100

Was this removed to discuss further ways of adding this resolver outside of @graphback/runtime :thinking: ?

I think this is a good opportunity regardless, do let me know what you think :)

ssd71:
CC @craicoverflow

wtrocki:
Yep. We need to have chat about this
Generally it is pointless for us to have separate package with datastore if we have embedded DataSync in resolvers.

ssd71:
I was thinking if it would be useful to be able to add resolvers from a plugin (or something similar).
In current state the plugins can only modify the schema, leaving graphback to fill in the gaps, therefore there is not much point to other people building plugins if they cannot do anything other than modify schema without adding any extra resolvers. Correct me if I am wrong, if plugins can add more resolvers, doesn't that also mean we could add custom scalars with a plugin? Sounds overkill I guess but just wanted to poke around and discuss about this

wtrocki:
You are right. We kinda need that. Previously we had plugins generate files that could be resolvers now we have this runtime resolver creator that do not have ir

craicoverflow:
> It seems the sync resolver is not actually returning anything:

That was a mistake by me, when merging conflicts! Sorry about that.

ssd71:
> > It seems the sync resolver is not actually returning anything:
> 
> That was a mistake by me, when merging conflicts! Sorry about that.

Ah cool :smile: , though what do you think about the rest of the discussion above

craicoverflow:
I do believe we need some form of plugin for resolvers. I was originally thinking this might require a new plugin engine specifically for resolvers, but it should work just fine in the regular plugins. 

`GraphQLSchema` object can contain all resolver methods. I've just tested the following example in `SchemaCRUDPlugin` and it works.

```
schemaComposer.addResolveMethods({
  Query: {
    hello: (source: any, args: object, context: unknown, info: any) => {
      return 'Hello world!'
    }
  }
})

const resolvers = schemaComposer.getResolveMethods()
```

[...]

PR #1570, opened 2020-06-25
Datasync template

ssd71:
This took a while to push :smile:

### DIFF
--- a/packages/graphback-cli/src/init/starterTemplates.ts
+++ b/packages/graphback-cli/src/init/starterTemplates.ts
@@ -51,6 +51,15 @@ export const allTemplates: Template[] = [
     }]
   },
   {
+    name: 'apollo-mongo-datasync-server-ts',
+    description: 'Apollo GraphQL Server connecting to Mongo database using TypeScript. Contains Data Synchronization features.',
+    repos: [{
+      uri: 'https://github.com/aerogear/graphback',
+      branch: 'templates-0.14.0',
+      path: '/templates/ts-apollo-mongodb-datasync-backend',
+    }]
+  },
+  {
     name: 'apollo-postgres-server-ts',
     description: 'Apollo GraphQL Server connecting to Postgres database using TypeScript',
     repos: [{
--- a/templates/ts-apollo-mongodb-datasync-backend/README.md
+++ b/templates/ts-apollo-mongodb-datasync-backend/README.md
@@ -22,13 +22,20 @@ docker-compose up -d
 - Define your schema in the `model/datamodel.graphql` file. Or use the default:
 
 ```
-type User {
+"""
+@model
+@datasync
+"""
+type Note {
   id: ID!
-  name: String
+  title: String!
+  description: String
+  """
+  @oneToMany(field: 'note')
+  """
+  comments: [Comment]!
 }
 ```
-
-> **NOTE**: for Data Synchonization to work, you must use the `@datasync` annotation on required model(s)
 
 - Start the server
 
--- a/templates/ts-apollo-mongodb-datasync-backend/README.md
+++ b/templates/ts-apollo-mongodb-datasync-backend/README.md
@@ -1,0 +1,47 @@
+## Graphback Apollo+MongoDB Data Synchronization Runtime Template
+
+Template showcases Graphback Data Synchronization capabilities using
+resolver layer created in-memory as opposed to codegeneration.
+
+Serverless example can be build as docker image with configurable volume for models. 
+This models will be processed at application at startup giving fully featured GraphQL Server 
+following Graphback CRUD specification.
+
+For more information please refer to: 
+
+### Running example
+
+The project has been created using `graphback`. Run the project using the following steps.
+
+- Start the database
+
+```
+docker-compose up -d
+```
+
+- Define your schema in the `model/datamodel.graphql` file. Or use the default:
+
+```
+type User {
+  id: ID!
+  name: String
+}
+```
+
+> **NOTE**: for Data Synchonization to work, you must use the `@datasync` annotation on required model(s)
+
+- Start the server
+
+```
+npm run develop
+```
+
+Or, if using yarn
+
+```
+yarn develop
+```
+
+For more on customising your Graphback application, check out [our docs](https://graphback.dev/docs/gettingstarted)
+
+If your project contains client application please follow `./client/README.md` for running client side application
--- a/templates/ts-apollo-mongodb-datasync-backend/README.md
+++ b/templates/ts-apollo-mongodb-datasync-backend/README.md
@@ -3,15 +3,15 @@
 Template showcases Graphback Data Synchronization capabilities using
 resolver layer created in-memory as opposed to codegeneration.
 
-Serverless example can be build as docker image with configurable volume for models. 
-This models will be processed at application at startup giving fully featured GraphQL Server 
-following Graphback CRUD specification.
+Serverless example can be built as docker image with configurable volume for models. 
+This models will be processed at application startup giving fully featured GraphQL Server 
+following [GraphQL CRUD specification.](https://graphqlcrud.org/)
 
 For more information please refer to: 
 
 ### Running example
 
-The project has been created using `graphback`. Run the project using the following steps.
+The project has been created using [`Graphback`](https://graphback.dev/). Run the project using the following steps.
 
 - Start the database
 
--- a/templates/ts-apollo-mongodb-datasync-backend/README.md
+++ b/templates/ts-apollo-mongodb-datasync-backend/README.md
@@ -35,6 +35,17 @@ type Note {
   """
   comments: [Comment]!
 }
+
+""" 
+@model
+@datasync
+"""
+type Comment {
+  id: ID!
+  text: String
+  description: String
+}
+
 ```
 
 - Start the server
--- a/templates/ts-apollo-mongodb-datasync-backend/package.json
+++ b/templates/ts-apollo-mongodb-datasync-backend/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "graphback-templates-mongodb-datasync-runtime",
+  "version": "0.14.0-alpha4",
+  "description": "Graphback with Data Synchronization runtime template with Apollo Server and MongoDb",
+  "private": true,
+  "main": "dist/src/index.js",
+  "scripts": {
+    "build": "tsc",
+    "develop": "ts-node src/index.ts",
+    "start": "node dist/src/index.js",
+    "watch": "tsc -w"
+  },
+  "license": "Apache 2.0",
+  "dependencies": {
+    "apollo-server-express": "2.14.3",
+    "@graphql-tools/graphql-file-loader": "6.0.9",
+    "@graphql-tools/load": "6.0.9",
+    "@graphback/datasync": "0.14.0-alpha4",
+    "cors": "2.8.5",
+    "dotenv": "8.2.0",
+    "express": "4.17.1",
+    "graphback": "0.14.0-alpha4",
+    "graphql": "14.6.0",
+    "graphql-config": "3.0.2",
+    "graphql-subscriptions": "1.1.0",
+    "graphql-tag": "2.10.3",
+    "mongodb": "3.5.8"
+  },
+  "devDependencies": {
+    "@types/glob": "7.1.2",
+    "@types/graphql": "14.2.3",
+    "@types/node-fetch": "2.5.7",
+    "ts-node": "8.10.2",
+    "tslint": "6.1.2",
+    "typescript": "3.9.5"
+  }
+}
--- a/templates/ts-apollo-mongodb-datasync-backend/package.json
+++ b/templates/ts-apollo-mongodb-datasync-backend/package.json
@@ -24,7 +24,7 @@
     "graphql-config": "3.0.2",
     "graphql-subscriptions": "1.1.0",
     "graphql-tag": "2.10.3",
-    "mongodb": "3.5.8"
+    "mongodb": "3.5.9"
   },
   "devDependencies": {
     "@types/glob": "7.1.2",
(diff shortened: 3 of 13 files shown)


templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
+## Graphback Apollo+MongoDB Data Synchronization Runtime Template
+
+Template showcases Graphback Data Synchronization capabilities using
+resolver layer created in-memory as opposed to codegeneration.
+
+Serverless example can be build as docker image with configurable volume for models.
```
machi1990, line 6:
```suggestion
Serverless example can be built as docker image with configurable volume for models. 
```

templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
+## Graphback Apollo+MongoDB Data Synchronization Runtime Template
+
+Template showcases Graphback Data Synchronization capabilities using
+resolver layer created in-memory as opposed to codegeneration.
+
+Serverless example can be build as docker image with configurable volume for models. 
+This models will be processed at application at startup giving fully featured GraphQL Server
```
machi1990, line 7:
```suggestion
This models will be processed at application startup giving fully featured GraphQL Server 
```

templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
+## Graphback Apollo+MongoDB Data Synchronization Runtime Template
+
+Template showcases Graphback Data Synchronization capabilities using
+resolver layer created in-memory as opposed to codegeneration.
+
+Serverless example can be build as docker image with configurable volume for models. 
+This models will be processed at application at startup giving fully featured GraphQL Server 
+following Graphback CRUD specification.
```
machi1990, line 8:
```suggestion
following [GraphQL CRUD specification.](https://graphqlcrud.org/)
```

Let's add a link to the spec maybe?

templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+For more information please refer to: 
+
+### Running example
+
+The project has been created using `graphback`. Run the project using the following steps.
```
machi1990, line 14:
```suggestion
The project has been created using [`Graphback`](https://graphback.dev/). Run the project using the following steps.
```

templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+- Start the database
+
+```
+docker-compose up -d
+```
+
+- Define your schema in the `model/datamodel.graphql` file. Or use the default:
+
+```
```
machi1990, line 24: should we add the annotations here

I am thinking of `@model`, `@datasync` and friends?

templates/ts-apollo-mongodb-datasync-backend/README.md
```diff
@@ -0,0 +1,47 @@
...
+type User {
+  id: ID!
+  name: String
+}
+```
+
+> **NOTE**: for Data Synchonization to work, you must use the `@datasync` annotation on required model(s)
```
machi1990, line 31: Let's add the annotation directly to the default model

templates/ts-apollo-mongodb-datasync-backend/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@graphback/datasync": "0.14.0-alpha4",
+    "cors": "2.8.5",
+    "dotenv": "8.2.0",
+    "express": "4.17.1",
+    "graphback": "0.14.0-alpha4",
+    "graphql": "14.6.0",
+    "graphql-config": "3.0.2",
+    "graphql-subscriptions": "1.1.0",
+    "graphql-tag": "2.10.3",
+    "mongodb": "3.5.8"
```
machi1990, line 27:
```suggestion
    "mongodb": "3.5.9"
```

I think 3.5.9 is around, let's use it then

templates/ts-apollo-mongodb-datasync-backend/src/db.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  if (DB_USER && DB_PASSWORD) {
+    url = `mongodb://${DB_USER}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_DATABASE}?authSource=${DB_AUTHSOURCE}`;
+  } else {
+    url = `mongodb://${DB_HOST}:${DB_PORT}/${DB_DATABASE}`
+  }
+
+  const mongoClient = await MongoClient.connect(url, { useUnifiedTopology: true });
+
+  // get db instance
```
machi1990, line 16: Do we need this comment?

templates/ts-apollo-mongodb-datasync-backend/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "devDependencies": {
+    "@types/glob": "7.1.2",
+    "@types/graphql": "14.2.3",
+    "@types/node-fetch": "2.5.7",
+    "ts-node": "8.10.2",
+    "tslint": "6.1.2",
+    "typescript": "3.9.5"
+  }
```
machi1990, line 36: Let's add graphback-cli - see https://github.com/aerogear/graphback/pull/1568 and the issue fixed by it for reference

templates/ts-apollo-mongodb-datasync-backend/README.md
ssd71: Cool :+1:

templates/ts-apollo-mongodb-datasync-backend/package.json
ssd71: understood, thanks!

templates/ts-apollo-mongodb-datasync-backend/src/db.ts
ssd71: Good point :smile:

machi1990:
A general question, should we also update the [`graphback-cli init` command templates](https://github.com/aerogear/graphback/blob/master/packages/graphback-cli/src/init/starterTemplates.ts#L13) in this PR or we prefer to do that in followup PR?

wtrocki:
Does this work for you guys? We can use latest release here

wtrocki:
Also we need to add this to graphback-cli templates as option on init

wtrocki:
No followups for me :)

templates/ts-apollo-mongodb-datasync-backend/README.md
ssd71: Added now,  but the example is a bit janky now, I couldn't decide between putting in both models versus putting a different model here, so I only put one model, what do you think?

packages/graphback-cli/src/init/starterTemplates.ts
```diff
@@ -50,6 +50,15 @@ export const allTemplates: Template[] = [
...
   },
+  {
+    name: 'apollo-mongo-datasync-server-ts',
+    description: 'Apollo GraphQL Server connecting to Mongo database using TypeScript. Contains Data Synchronization features.',
+    repos: [{
+      uri: 'https://github.com/aerogear/graphback',
+      branch: 'templates-0.14.0',
+      path: '/templates/ts-apollo-mongodb-datasync-backend',
+    }]
+  },
```
ssd71, line 61: Is this okay?

machi1990: Also a fullstack entry will be required as in https://github.com/aerogear/graphback/blob/master/packages/graphback-cli/src/init/starterTemplates.ts#L29

machi1990:
> Does this work for you guys? We can use latest release here

Yes, I tested this but I had to do minor changes 

```patch
diff --git a/templates/ts-apollo-mongodb-datasync-backend/package.json b/templates/ts-apollo-mongodb-datasync-backend/package.json
index 7b69105e..4d82f5dd 100644
--- a/templates/ts-apollo-mongodb-datasync-backend/package.json
+++ b/templates/ts-apollo-mongodb-datasync-backend/package.json
@@ -15,11 +15,11 @@
     "apollo-server-express": "2.14.3",
     "@graphql-tools/graphql-file-loader": "6.0.9",
     "@graphql-tools/load": "6.0.9",
-    "@graphback/datasync": "0.14.0-alpha4",
+    "@graphback/datasync": "0.14.0-alpha8",
     "cors": "2.8.5",
     "dotenv": "8.2.0",
     "express": "4.17.1",
-    "graphback": "0.14.0-alpha4",
+    "graphback": "0.14.0-alpha8",
     "graphql": "14.6.0",
     "graphql-config": "3.0.2",
```  

to set the graphback version to `0.14.0-alpha8` since we had issues in alpha4.

packages/graphback-cli/src/init/starterTemplates.ts
ssd71: Was kinda deciding against this as full-stack equivalent would be with offix perhaps and do not want to add that here :smile: 
Does client plugin produce sync queries?

machi1990: Ah good question and remark. I do not think it does so maybe let's keep it as you were suggesting. 

And this way, maybe we can remove the [client sysmink](https://github.com/aerogear/graphback/blob/29a82d5f6f66111c5d56732309f76414e1da6e0a/templates/ts-apollo-mongodb-datasync-backend/client) ?

templates/ts-apollo-mongodb-datasync-backend/README.md
machi1990: Not bad to call it out in the readme. You can copy [Comment model]( https://github.com/aerogear/graphback/pull/1570/files#diff-e4d6686600db9bde9b351b6328b09c7eR19) here and it will be fine by me.

packages/graphback-cli/src/init/starterTemplates.ts
ssd71: ah yes, definitely :smile: , did not see that one , on it

wtrocki:
Yep. Sorry I assumed that those will be bumped but since we do auto updates they will not.

wtrocki:
@ssd71 do you have this working end to end - Are delta queries generated in schema?
I do not see plugin doing any work for me. Logged that yesterday - #1411 
It will be also worth to verify the delta queries working

ssd71:
I think it's working:
- Has sync query
image
- Type has updatedAt, createdAt
image
- can do sync queries
image
- can remove documents as well
image

@machi1990 Did all these work for you as well?

machi1990:
> @machi1990 Did all these work for you as well?

Yes, they did but the challenge is here when you have existing data without `updatedAt` field.

Since the field is required when doing an update (https://github.com/aerogear/graphback/blob/master/packages/graphback-datasync/src/DataSyncPlugin.ts#L81) it will fail since we do not have this data: this is an issue I encountered while testing.

@wtrocki are we talking about the same issue?

ssd71:
> @ssd71 Yes, they did but the challenge is here when you have existing data without updatedAt field.

Ah that, yes, We can simple put a check if the document stored in DB doesn't have the field, then we ignore it, this PR or the next one

machi1990:
> > @ssd71 Yes, they did but the challenge is here when you have existing data without updatedAt field.
> 
> Ah that, yes, We can simple put a check if the document stored in DB doesn't have the field, then we ignore it

Yeap.  You mean ignore the conflicts check right? 

> , this PR or the next one

I'd say a separate PR addressing this issue https://github.com/aerogear/graphback/issues/1573. But let's wait first for @wtrocki to see if we are talking about the same issue.

ssd71:
> > > @ssd71 Yes, they did but the challenge is here when you have existing data without updatedAt field.
> > 
> > 
> > Ah that, yes, We can simple put a check if the document stored in DB doesn't have the field, then we ignore it
> 
> @ssd71 Yeap. You mean ignore the conflicts check right?

Bingo

> 
> > , this PR or the next one
> 
> I'd say a separate PR addressing this issue #1573. But let's wait first for @wtrocki to see if we are talking about the same issue.

Agreed :+1:

wtrocki:
> Ah that, yes, We can simple put a check if the document stored in DB doesn't have the field, then we ignore it, this PR or the next one

Kinda goes against datasync. We need to have integrity field on every item so this should not happen unless someone want now to add datasync to existing type. That will cause problem :D but it will mean that if there is update we will just update this value so it technically should not happen :D

wtrocki:
DataSync means that you have delta queries and we also ensure consistency using conflicts checks. When working in datasync consistency check `updatedAt` is required to do any operation with object. 
If you delete, update object and there is no updatedAt field it means that you want to introduce datasync for this field and we should write lastUpdatedAt value

wtrocki:
Issue was related to wrong version being cached. All works. Happy to merge it.

machi1990:
> DataSync means that you have delta queries and we also ensure consistency using conflicts checks. When working in datasync consistency check `updatedAt` is required to do any operation with object.
> If you delete, update object and there is no updatedAt field it means that you want to introduce datasync for this field and we should write lastUpdatedAt value

Yeap, I think this will only happen once but to be able to actually add datasync to existing db, we need a way to write the `updateAt` field (which is done automatically already during create/ update/ delete). 

The challenge is how do we do it? 
- By ignoring the conflicts check for the first time during update/delete operation as @ssd71 proposed here  https://github.com/aerogear/graphback/pull/1570#issuecomment-649563729 meaning that the field will be updated afterward? 
How does one perform the update/delete mutation without the `updatedAt` since it is required and  the existing documents do not have it?
- or we run some sort of migrations to existing data (expensive)?

wtrocki:
Nr 1 please. Current timestamp if possible

Issue #1540, opened 2020-06-24
Missing mutation input fields for embedded relationships


craicoverflow:
If we supported embedded relationships on non-model GraphQL types, there should also be a way to insert that field or filter on it.

```gql
""" @model """
type Task {
  id: ID!
  version: Int
  title: String!
  description: String!
  status: String!
  """
  @db(type: 'json')
  """
  creationMetadata: CreationMetadata
}

type CreationMetadata {
  createdDate: String
  createdBy: User
}
```

There are no input fields for this embedded relationship.

```gql
input CreateTaskInput {
  id: ID
  version: Int
  title: String!
  description: String!
  status: String!
  creationMetadata: CreationMetadata # Let's add this
}
```

wtrocki:
Awesome! Is that blocking?

craicoverflow:
Alpha release - no
0.14.0 - maybe. It depends on how much of a priority it is.

It is not a common use case in Postgres (still good to support it)
But I can see it being commonly needed in Mongo.

The data can be retrieved if it already exists, but if it cannot be inserted then it is almostr useless to have it as a feature.

Relates to https://github.com/aerogear/graphback/issues/1432

craicoverflow:
This is needed in 0.14.0 otherwise it is useless to have embedded relationships.

PR #1574, opened 2020-06-25
feat: support create/update of embedded JSON objects

craicoverflow:
This adds support for create/update of embedded relationship documents in both PostgreSQL and MongoDB.

Before this change, it was not possible to CREATE/UPDATE/FIND fields which are GraphQL types which are not models because:

a) There were no input fields added for these types to the CRUD input types
b) `getSelectedFieldsFromResolverInfo` did not include GraphQL type fields in the selectable fields, so `.select(x)` would never return this data from the database.

TL;DR we added support for non-model type fields in out models in https://github.com/aerogear/graphback/issues/1432 but this data could not be inserted/updated/retrieved from the database.

```graphql
"""
@model
"""
type Note {
  id: ID!
  title: String
  """
  this annotation is only needed for Postgres
  @db(type: 'json') 
  """
  comments: [Comment]
}

type Comment {
  owner: String
  comment: String
}
```

You can perform a mutation like so:

```gql
mutation {
  createNote(input: {
    title:"my new note",
    comments: [
      {
        comment:"Awesome!",
        owner:"Enda"
      }
    ]
  }) {
    id
    title
    comments {
      owner
      comment
    }
  }
}
```

### Tasks

- [x] Create input types for non-model fields
- [x] Allow CREATE/UPDATE/FIND on non-model fields (Mongo/Postgres)
- [x] Integration tests

### DIFF
--- a/integration/tests/runtime-workflow-mongo.ts
+++ b/integration/tests/runtime-workflow-mongo.ts
@@ -42,8 +42,8 @@ beforeAll(async () => {
     graphbackApi = buildGraphbackAPI(modelText, {
       dataProviderCreator: createMongoDbProvider(db),
       plugins: [
-        new SchemaCRUDPlugin({outputPath: "./output-mongo/schema/schema.graphql"}),
-        new ClientCRUDPlugin({format: 'graphql', outputFile: './output-mongo/client/graphback.graphql'})
+        new SchemaCRUDPlugin({ outputPath: "./output-mongo/schema/schema.graphql" }),
+        new ClientCRUDPlugin({ format: 'graphql', outputFile: './output-mongo/client/graphback.graphql' })
       ]
     });
 
@@ -89,7 +89,15 @@ async function seedDatabase(db: Db) {
     },
     {
       title: 'Note B',
-      description: 'Note B Description'
+      description: 'Note B Description',
+      tasks: [
+        {
+          title: 'Task 1'
+        },
+        {
+          title: 'Task 2'
+        }
+      ]
     }
   ]
 
@@ -108,7 +116,7 @@ async function seedDatabase(db: Db) {
   ]
 
   for (const metadata of commentMetadata) {
-    const {ops} = await db.collection('commentmetadata').insertOne(metadata);
+    const { ops } = await db.collection('commentmetadata').insertOne(metadata);
     metadataId.push(ops[0]._id.toString());
   }
 
@@ -127,24 +135,10 @@ async function seedDatabase(db: Db) {
     }
   ]
 
-  for (const comment of comments)  {
-    const {ops} = await db.collection('comment').insertOne(comment);
+  for (const comment of comments) {
+    const { ops } = await db.collection('comment').insertOne(comment);
     commentId.push(ops[0]._id.toString())
   }
-}
-
-const getConfig = async () => {
-  const config = await loadConfig({
-    rootDir: process.cwd(),
-    extensions: [
-      () => ({ name: 'graphback' })
-    ]
-  });
-
-  const projectConfig = config.getDefault();
-  const graphbackConfig = projectConfig.extension('graphback');
-
-  return { projectConfig, graphbackConfig };
 }
 
 test('Find all notes', async () => {
@@ -174,7 +168,15 @@ test('Find all notes', async () => {
         id: notesId[1],
         title: 'Note B',
         description: 'Note B Description',
-        comments: []
+        comments: [],
+        tasks: [
+          {
+            title: 'Task 1'
+          },
+          {
+            title: 'Task 2'
+          }
+        ]
       }
     ],
     limit: null,
--- a/integration/tests/runtime-workflow-mongo.ts
+++ b/integration/tests/runtime-workflow-mongo.ts
@@ -85,7 +85,8 @@ async function seedDatabase(db: Db) {
   const notes = [
     {
       title: 'Note A',
-      description: 'Note A Description'
+      description: 'Note A Description',
+      tasks: []
     },
     {
       title: 'Note B',
@@ -143,7 +144,6 @@ async function seedDatabase(db: Db) {
 
 test('Find all notes', async () => {
   const { data } = await client.query({ operationName: 'findNotes', query: documents });
-
   expect(data).toBeDefined();
   expect(data.findNotes).toEqual({
     items: [
@@ -151,6 +151,7 @@ test('Find all notes', async () => {
         id: notesId[0],
         title: 'Note A',
         description: 'Note A Description',
+        tasks: [],
         comments: [
           {
             id: commentId[0],
@@ -199,7 +200,15 @@ test('Find all notes except the first', async () => {
         id: notesId[1],
         title: 'Note B',
         description: 'Note B Description',
-        comments: []
+        comments: [],
+        tasks: [
+          {
+            title: 'Task 1'
+          },
+          {
+            title: 'Task 2'
+          }
+        ]
       }
     ],
     limit: null,
@@ -222,6 +231,7 @@ test('Find at most one note', async () => {
         id: notesId[0],
         title: 'Note A',
         description: 'Note A Description',
+        tasks: [],
         comments: [
           {
             id: commentId[0],
@@ -291,6 +301,7 @@ test('Note 1 should be defined', async () => {
     id: notesId[0],
     title: 'Note A',
     description: 'Note A Description',
+    tasks: [],
     comments: [
       {
         id: commentId[0],
@@ -374,9 +385,12 @@ test('Should update Note 1 title', async () => {
 });
 
 test('Should create a new Note', async () => {
-  const response = await createNote(client, { title: 'New note', description: 'New note description' });
+  const response = await createNote(client, { title: 'New note', description: 'New note description', tasks: [{ title: "new task title" }] });
   expect(response.data).toBeDefined();
   expect(response.data.createNote).toEqual({ id: response.data.createNote.id, title: 'New note', description: 'New note description' });
+
+  const { data } = await getNote(response.data.createNote.id, client);
+  expect(data.getNote.tasks).toEqual([{ title: "new task title" }]);
 })
 
 test('Delete Note 1', async () => {
@@ -423,14 +437,3 @@ async function getNote(id: string | number, client: ApolloServerTestClient) {
 
   return response;
 }
-
-async function findNoteComments(noteId: string, client: ApolloServerTestClient) {
-  const response = await client.query({
-    operationName: "findComments",
-    query: documents,
-    variables: { filter: { noteId } }
-  });
-
-  return response;
-}
-
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -725,6 +725,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       const namedType = tc.getType();
       if (isScalarType(namedType) && !isSpecifiedScalarType(namedType) && !isSpecifiedGraphbackScalarType(namedType)) {
         schemaComposer.add(createInputTypeForScalar(namedType));
+
+        return;
       }
 
       const isRootType = ['Query', 'Subscription', 'Mutation'].includes(namedType.name)
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -727,7 +727,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
         schemaComposer.add(createInputTypeForScalar(namedType));
       }
 
-      if (isObjectType(namedType) && !isModelType(namedType)) {
+      const isRootType = ['Query', 'Subscription', 'Mutation'].includes(namedType.name)
+      if (isObjectType(namedType) && !isModelType(namedType) && !isRootType) {
         addCreateObjectInputType(schemaComposer, namedType)
         addUpdateObjectInputType(schemaComposer, namedType)
       }
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -1,13 +1,13 @@
 /* eslint-disable max-lines */
-import { existsSync, mkdirSync, writeFileSync, fstat } from 'fs';
+import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve, dirname, join } from 'path';
-import { getFieldName, metadataMap, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap, getInputTypeName, FieldRelationshipMetadata, getPrimaryKey, GraphbackContext, getSelectedFieldsFromResolverInfo } from '@graphback/core'
-import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt, GraphQLFloat, isScalarType, isSpecifiedScalarType, GraphQLResolveInfo } from 'graphql';
+import { getFieldName, metadataMap, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap, getInputTypeName, FieldRelationshipMetadata, getPrimaryKey, GraphbackContext, getSelectedFieldsFromResolverInfo, isModelType, isSpecifiedGraphbackScalarType } from '@graphback/core'
+import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt, GraphQLFloat, isScalarType, isSpecifiedScalarType, GraphQLResolveInfo, isObjectType } from 'graphql';
 import { SchemaComposer, NamedTypeComposer } from 'graphql-compose';
 import { IResolvers, IFieldResolver } from '@graphql-tools/utils'
 import { parseMetadata } from "graphql-metadata";
 import { gqlSchemaFormatter, jsSchemaFormatter, tsSchemaFormatter } from './writer/schemaFormatters';
-import { buildFilterInputType, createModelListResultType, StringScalarInputType, BooleanScalarInputType, SortDirectionEnum, buildCreateMutationInputType, buildFindOneFieldMap, buildMutationInputType, OrderByInputType, buildSubscriptionFilterType, IDScalarInputType, PageRequest, createInputTypeForScalar, createVersionedFields,createVersionedInputFields } from './definitions/schemaDefinitions';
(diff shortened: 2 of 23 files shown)


machi1990:
You rock @craicoverflow, looking at this now!

craicoverflow:
It is not possible to query the embedded relationship using Postgres with Knex (unless you do raw queries, which are not a quick fix)

```
findNotes(filter: {comment: {contains: "{hello: 1}"}}) {
    items {
      comment {
        description
        test {
          hello
        }
      }
    }
 }
```

This throws: `"select \"comment\" from \"note\" where \"comment\" like $1 - operator does not exist: json ~~ unknown""`

I am now going to do some testing with Mongo, and see if it works. We can decide whether to disable filtering on both for now until we have a solution for Postgres.

craicoverflow:
Filtering on the field for Mongo does nothing either, going to disable that.

wtrocki:
Did I missed some issue. Trying to recall when we talked about this being problematic etc?

craicoverflow:
@wtrocki see https://github.com/aerogear/graphback/issues/1540

wtrocki:
> It is not possible to query the embedded relationship using Postgres with Knex (unless you do raw queries, which are not a quick fix)


But aren't this coming from graphql resolvers ?

wtrocki:
What happening here - are we downgrading graphqlcrud features? Tell me like I'm 5 y old :D

craicoverflow:
The data exists in the database so it would be more performant to filter the JSON or embedded document in the database.

craicoverflow:
The CRUD has been upgraded - no features have been removed, because now it supports strongly-typed fields which are non-model objects.

Before this change, it was not possible to CREATE/UPDATE/FIND fields which are GraphQL types which are not models because:

a) There were no input fields added for these types to the CRUD input types
b) `getSelectedFieldsFromResolverInfo` did not include GraphQL type fields in the selectable fields, so `.select(x)` would never return this data from the database.

TL;DR we added support for non-model type fields in out models in https://github.com/aerogear/graphback/issues/1432 but this data could not be inserted/updated/retrieved from the database.

craicoverflow:
@wtrocki @machi1990 I need to add tests, but ready for review other than that.

wtrocki:
Mind blown! Calling out MachiMan hero to help here(@machi1990) 
I will need to have more time to understand this better.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -258,8 +281,64 @@ export const buildMutationInputType = (modelType: GraphQLObjectType) => {
...
+  return fields.map((field: GraphQLField<any, any>) => {
+    let namedType = getNamedType(field.type)
+    let typeName = namedType.name
+
+    let inputType
+    if (isObjectType(namedType)) {
+      typeName = getInputTypeName(typeName, GraphbackOperationType.CREATE)
+      namedType = schemaComposer.getOrCreateITC(typeName).getType()
+
+      inputType = copyWrappingType(field.type, namedType)
```
machi1990, line 298: Why is copyingWrapping type needed?

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -258,8 +281,64 @@ export const buildMutationInputType = (modelType: GraphQLObjectType) => {
...
+  schemaComposer.add(mutationInputObject)
+}
+
+function mapObjectInputFields(schemaComposer: SchemaComposer<any>, fields: GraphQLField<any, any>[]): GraphQLInputField[] {
+  return fields.map((field: GraphQLField<any, any>) => {
+    let namedType = getNamedType(field.type)
+    let typeName = namedType.name
+
+    let inputType
+    if (isObjectType(namedType)) {
```
machi1990, line 294: What happens if the type is a model?

packages/graphback-runtime-knex/src/knexQueryMapper.ts
```diff
@@ -101,7 +101,7 @@ function where(builder: Knex.QueryBuilder, filter: any, or: boolean = false, not
 
     // eslint-disable-next-line no-null/no-null
     if (exprEntry[1] === null) {
-      builder = builder[builderMethod('whereNull', or, not)](col)
+      builder = builder[builderMethod('whereNull', or, exprEntry[0] === 'ne')](col)
```
machi1990, line 104: Is this changed related?

craicoverflow: I found a bug while testing the querying.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
craicoverflow: An unhandled error is thrown. Thanks.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
craicoverflow:
```
"""@model"""
type Note {
  id: ID!
  comments: [Comment]!
}
```

`Note.comments` will need to map to a newly created input type, so this ensures that the required level is the exact same in `CreateNoteInput.comments` as `Note.comments`.

```
type CreateNoteInput {
  id: ID!
  comments: [CreateCommentInput]!
}
```

packages/graphback-runtime-knex/src/knexQueryMapper.ts
machi1990: Nice catch. What was it?

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: Gotcha.

templates/ts-apollo-postgres-backend/model/datamodel.graphql
```diff
@@ -4,14 +4,19 @@ type Note {
...
-  text: String
-  description: String
+  text: String!
+  description: String!
+  test: Test!
+  test2: [Test!]!
+}
+
+type Test {
+  hello: String
```
machi1990, line 21: Maybe we could add an [integration test](https://github.com/aerogear/graphback/tree/9622f7373cb59f3926b6a538ea4424de29af7d4e/integration) and revert the change on this template file?

packages/graphback-runtime-knex/src/knexQueryMapper.ts
craicoverflow: `ne` was being evaluated as `eq`, so {ne: null} in SQL gave `IS NULL`.

Added https://github.com/aerogear/graphback/pull/1574/commits/0ac97c2d25774a468f08e243c67d81dbfdab712b to cover it.

templates/ts-apollo-postgres-backend/model/datamodel.graphql
craicoverflow: Reverted file now, thanks. Have begun the process of adding integration test cases but not sure if I will get it complete in time (off tomorrow).

machi1990:
Tested this issue and it worked out of the box for Mongo. But I ran into issues for Postgres. 

- Will a custom type as an array  e.g `comments: [Comment]` work ? I remember having issues with storing arrays using the `json` Postgres type and had to store arrays as `{data: array}`
- Another issue which is a general one is:

with the model below 
```
"""
@model
"""
type Note {
  id: ID!
  title: String!
  description: String
  """
  @db(type: 'json')
  """
  comments: [Comment]!
}

type Comment {
  text: String
  description: String
}

```
 
If I issue the following create mutation
```graphql
mutation {
  createNote(
    input: { title: "test", description: "description", comments: [{text: "mambo"}] }
  ) {
    comments {
      text
      description
    }
    title
  }
}

```

I had the following error 
```log

{
  "errors": [
    {
      "message": "insert into \"note\" (\"comments\", \"description\", \"title\") values ($1, $2, $3) returning \"comments\", \"title\" - invalid input syntax for type json",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createNote"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "length": 171,
          "name": "error",
          "severity": "ERROR",
          "code": "22P02",
          "detail": "Expected \":\", but found \"}\".",
          "where": "JSON data, line 1: {\"{\\\"text\\\":\\\"mambo\\\"}\"}",
          "file": "json.c",
          "line": "1199",
          "routine": "report_parse_error",
          "stacktrace": [
            "error: insert into \"note\" (\"comments\", \"description\", \"title\") values ($1, $2, $3) returning \"comments\", \"title\" - invalid input syntax for type json",
            "    at Parser.parseErrorMessage 

```

craicoverflow:
Thanks for spotting that @machi1990 - it should be fixed now with 2b601d870d7751a5ac526482efbfbfece7b0c114

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -29,10 +30,19 @@ function getTableId(idField: string, data: any): TableID {
...
-  //TODO: Map fields to custom db names
-
   return {
     idField: getTableId(idField, data),
-    data
+    // eslint-disable-next-line @typescript-eslint/tslint/config
+    data: Object.keys(data).reduce((dataObj, key) => {
+      const value = data[key]
+
+      if (typeof value === 'object' || Array.isArray(value)) {
```
machi1990, line 39:
```suggestion
      if (typeof value === 'object') {
```

typeof value will aways be 'object' if it is an object or an array.

machi1990:
> Thanks for spotting that @machi1990 - it should be fixed now with [2b601d8](https://github.com/aerogear/graphback/commit/2b601d870d7751a5ac526482efbfbfece7b0c114). Knex did not handle insertion of an array into a JSON field (it worked fine for objects), so I just had to stringify the data in advance of insert.

Cool thanks. I'll verify tomorrow but that should fix the issue too. I'll a full verification and full review tomorrow.

integration/tests/runtime-workflow-mongo.ts
```diff
@@ -42,8 +42,8 @@ beforeAll(async () => {
     graphbackApi = buildGraphbackAPI(modelText, {
       dataProviderCreator: createMongoDbProvider(db),
       plugins: [
-        new SchemaCRUDPlugin({outputPath: "./output-mongo/schema/schema.graphql"}),
-        new ClientCRUDPlugin({format: 'graphql', outputFile: './output-mongo/client/graphback.graphql'})
+        new SchemaCRUDPlugin({ outputPath: "./output-mongo/schema/schema.graphql" }),
```
wtrocki, line 45: This should be gone. We do not need schema plugin

machi1990: It is needed, otherwise the integration tests fails with 
```
  ● Note 1 should be defined

    expect(received).toEqual(expected) // deep equality

    Expected: {"comments": [{"description": "Note A Comment Description", "id": "5ef5b6951f82b90668f8da94", "text": "Note A Comment"}, {"description": "Note A Comment Description", "id": "5ef5b6951f82b90668f8da95", "text": "Note A Comment 2"}], "description": "Note A Description", "id": "5ef5b6951f82b90668f8da90", "title": "Note A"}
    Received: null

      287 |   expect(response.data).toBeDefined();
      288 |   const notes = response.data.getNote;
    > 289 |   expect(notes).toEqual({
          |                 ^
      290 |     id: notesId[0],
      291 |     title: 'Note A',
      292 |     description: 'Note A Description',

      at tests/runtime-workflow-mongo.ts:289:17
      at fulfilled (../node_modules/tslib/tslib.js:112:62)

  ● Should create a new Note

    expect(received).toBeDefined()

    Received: undefined

      375 | test('Should create a new Note', async () => {
      376 |   const response = await createNote(client, { title: 'New note', description: 'New note description' });
    > 377 |   expect(response.data).toBeDefined();
          |                         ^
      378 |   expect(response.data.createNote).toEqual({ id: response.data.createNote.id, title: 'New note', description: 'New note description' });
      379 | })
      380 | 

      at tests/runtime-workflow-mongo.ts:377:25

```

packages/graphback-runtime-knex/src/knexQueryMapper.ts
machi1990: Amazing thanks.

integration/tests/runtime-workflow-mongo.ts
wtrocki: Looks like we found bug that is not related to this PR

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: Great. This works fine but it will fail when the user adds the `SchemaPlugin` to print out the schema 

```
    Error: Non-model NoteResultList.items maps to Graphback model (Note) which is not allowed.
        at /home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts:296:15
        at Array.map (<anonymous>)
        at mapObjectInputFields (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts:289:17)
        at Object.addCreateObjectInputType (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts:319:13)
        at /home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts:731:11
        at Map.forEach (<anonymous>)
        at SchemaComposer.forEach (/home/manyanda/Documents/Projects/aerogear/graphback/node_modules/graphql-compose/lib/TypeStorage.js:37:23)
        at SchemaCRUDPlugin.createSchemaCRUDTypes (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts:724:20)
        at SchemaCRUDPlugin.buildSchemaForModels (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts:135:10)
        at SchemaCRUDPlugin.transformSchema (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts:63:10)
        at GraphbackPluginEngine.createSchema (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts:66:32)
        at GraphbackPluginEngine.createResources (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts:50:10)
        at Object.buildGraphbackAPI (/home/manyanda/Documents/Projects/aerogear/graphback/packages/graphback/src/buildGraphbackAPI.ts:92:33)
        at /home/manyanda/Documents/Projects/aerogear/graphback/integration/tests/runtime-workflow-mongo.ts:42:20
        at fulfilled (/home/manyanda/Documents/Projects/aerogear/graphback/node_modules/tslib/tslib.js:112:62)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

      at tests/runtime-workflow-mongo.ts:59:13

```

I ran into this issue locally when executing our integration tests - where we add the schema plugin to print out the schema so that it gets picked out by gql client
https://github.com/aerogear/graphback/blob/master/integration/tests/runtime-workflow-postgres.ts#L48-L50

But we already have a schema plugin defined by default https://github.com/aerogear/graphback/blob/master/packages/graphback/src/buildGraphbackAPI.ts#L82

/cc @wtrocki we kind of chatted about it today.

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -1,4 +1,5 @@
 import { ModelTableMap } from './buildModelTableMap';
+import { type } from 'os';
```
machi1990, line 2:
```suggestion
```

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -682,14 +717,21 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
+    schemaComposer.add(JSONScalarInputType);
     schemaComposer.add(createInputTypeForScalar(GraphQLInt));
     schemaComposer.add(createInputTypeForScalar(GraphQLFloat));
 
     schemaComposer.forEach((tc: NamedTypeComposer<any>) => {
       const namedType = tc.getType();
-      if (isScalarType(namedType) && !isSpecifiedScalarType(namedType)) {
+      if (isScalarType(namedType) && !isSpecifiedScalarType(namedType) && !isSpecifiedGraphbackScalarType(namedType)) {
         schemaComposer.add(createInputTypeForScalar(namedType));
       }
```
machi1990, line 728:
```suggestion
      return;
      }
```

let's short-circuit the flow to avoid the checks below.

packages/graphback-core/src/plugin/getSelectedFieldsFromResolverInfo.ts
```diff
@@ -21,6 +21,8 @@ export const getSelectedFieldsFromResolverInfo = (info: GraphQLResolveInfo, mode
 
         if (foundRelationship) {
           acc.push(foundRelationship.relationForeignKey)
+        } else {
```
machi1990, line 24: We'll need to check if it is not a `oneToMany` relationship before adding it.

Without it a relationship like this one below 
```
""" @model """
type Note {
  id: ID!
  title: String!
  description: String
  """
  @oneToMany(field: 'note')
  """
  comments: [Comment]!
}

""" @model """
type Comment {
  id: ID!
  text: String
  description: String
}
```

Would fail when issue a graphql request of this sort
```
findNotes {

 items {
  id, 
  comments {
    id
  }
}
}

```

it will fail consider `comments` as selectable field while it should not.

machi1990: Pushed a https://github.com/aerogear/graphback/commit/34365254d29a4146231169e1d409f998ab11968d to fix this issue and make postgres integration tests to pass. Working on the Mongo test cases now.

packages/graphback-core/src/db/dataMapper.ts
```diff
@@ -29,10 +29,19 @@ function getTableId(idField: string, data: any): TableID {
 export const getDatabaseArguments = (modelMap: ModelTableMap, data?: any, fieldMap?: any): TableDataMap => {
   const idField = modelMap.idField;
 
-  //TODO: Map fields to custom db names
-
   return {
     idField: getTableId(idField, data),
-    data
+    // eslint-disable-next-line @typescript-eslint/tslint/config
+    data: Object.keys(data).reduce((dataObj, key) => {
```
machi1990, line 35: We have to check for if data is defined here. Otherwise it will crash. 

Since from https://github.com/aerogear/graphback/pull/1574/files#diff-5a1ded10109278ede19b3429e917d253R29 data can be undefined.

machi1990: I'll push a commit to only transform data if defined.

machi1990: This was the case for the MongoDB provider https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L130 causing the Mongo integration tests to fail.

wtrocki:
No worries. I will do another verification soon

wtrocki:
That will be actually very cool to have. Doing some workarounds for this feature in OVP at the moment

wtrocki:
That will be actually very cool to have. Doing some workarounds for this feature in OVP at the moment

wtrocki:
Would this work for scalars as well (along with the types)

wtrocki:
Latest release doesn't seem to be working for querying data:
Screenshot 2020-06-28 at 22 06 34
Screenshot 2020-06-28 at 22 06 02

machi1990:
> Latest release doesn't seem to be working for querying data:
> Screenshot 2020-06-28 at 22 06 34
> Screenshot 2020-06-28 at 22 06 02

I saw a fix for this in this PR.

wtrocki:
Do we have integration test for this case?

machi1990:
> Do we have integration test for this case?

I think the added integration tests should reflect that. 

The `tasks` custom type.

machi1990:
@craicoverflow Thanks for merging this and for the amazing work.

PR #1576, opened 2020-06-25
enable datasync for graphql-serve

wtrocki:
Motivation here is to enable datasync in graphql serve.
We need that to have headless backend for offix that works with conflicts etc. 
This will be only for example purposes but it will be really powerful for the demos.

## Verification

1. Link cli
2. Run graphql-serve serve task.graphql -p 8080 --datasync

### DIFF
--- a/packages/graphql-serve/src/runtime.ts
+++ b/packages/graphql-serve/src/runtime.ts
@@ -35,16 +35,16 @@ export const createRuntime = async (modelDir: string, db: Db, datasync: boolean)
   let graphbackAPI;
   if (datasync) {
     graphbackAPI = buildGraphbackAPI(model, {
-      dataProviderCreator: createMongoDbProvider(db)
-    })
-  } else {
-    graphbackAPI = buildGraphbackAPI(model, {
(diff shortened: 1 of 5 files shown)


packages/graphql-serve/src/runtime.ts
```diff
@@ -27,12 +29,22 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
...
-export const createRuntime = async (modelDir: string, db: Db): Promise<GraphbackAPI> => {
+export const createRuntime = async (modelDir: string, db: Db, datasync: boolean): Promise<GraphbackAPI> => {
   const model = await loadModel(modelDir);
 
-  const graphbackAPI = buildGraphbackAPI(model, {
-    dataProviderCreator: createMongoDbProvider(db)
-  })
+  let graphbackAPI;
+  if (datasync) {
+    graphbackAPI = buildGraphbackAPI(model, {
```
wtrocki, line 37: @ssd71 I think it would be worth to wrap this API into a single helper object as this will rarely will change

ssd71: Agreed! Sounds great!

PR #1585, opened 2020-06-26
fix(plugin): choose custom crud schema plugin over default one

machi1990:
User can supply a custom instance of the schema plugin, let's use it over the default plugin.


wtrocki:
Can we merge this? Planning release

PR #1586, opened 2020-06-26
refactor: avoid creating a new context object when adding options

machi1990:


machi1990:
Thanks @wtrocki for spotting these.

Issue #1590, opened 2020-06-28
[DataSync] Do we need createdAt field for datasync?


ssd71:
Was just wondering if we need 'createdAt' for datasync? I am kinda leaning towards making it optional for datasync
This argument could be extended to creating some functionality for `@versioned` as well, where we could specify which fields we want, for example: `@versioned(updatedAt: false)` for someone who only wants the creation timestamp

wtrocki:
It can be confusion. DataSync offers delta synchronization and conflict resolution. Default strategy for conflict resolution relies on lastUpdated.

I think we both know that this ended up little bit convoluted, but I will need to think of we need:

- deltasync
- conflict 

And how to represent those two separately. 
I wanted to keep them together to minimize amount of DataProviders we will need. I think you are right and we do not need it but we need to get some thoughts to avoid yet another deltadatasync. I think we need to avoid confusion for end users and also match some internal simplicity when maintaining data providers.

I think we need to push another strategy and then decide on what we will do here

wtrocki:
I think we can sacrifice flexibility towards simplicity, but the challenge I have is that there are two ways to set this up: Conflicts are setup in code and versioning thing is set in schema. 
Pulling strategies into schema makes less sense because you might want to have custom one anyway. 
As I said it is easy to operate on the existing API once we bring the second strategy, but it looks like we going to have convoluted API so better to react quickly. 

Let's dedicate some time next week to chat about this.

ssd71:
Hmm I completely agree, in hindsight the resulting API does seem rather convoluted but like you said let's bring in the strategy first maybe it will be better for us to look at it in entirety and then decide on improvements rather than ruminating before the fact hehe

wtrocki:
I think if you work on strategy and conflict on separate branch without merging we will get 100 of the capacity we need and from there we going to work on improving this

ssd71:
This is outdated now because we have specific `_lastUpdatedAt` field for DataSync that's different from the one used with `@versioned`, added in #1771 

Closing

PR #1591, opened 2020-06-28
[WIP]Custom conflicts for DataSync

ssd71:
This is a very (**very**) rough implementation of custom conflicts in DataSync, need some serious feedback on this before we go ahead with this :smile:
### What is this?
This PR adds the ability for checking custom fields for conflicts and resolving conflicts as well. Similar to what's in Offix conflict-server.

### What's done?
A ConflictEngine interface which could be used to create custom Conflict Engines, not cleanly incorporated into the workflow yet, I am thinking of something like passing a custom conflict engine class to `createDataSyncMongoDbProvider` but I doubt it's optimal because I cannot see any way to extend this other than basic stuff like the engine used by default

### What's left?
- [ ] Finalize approach while optimising for extensiblity
- [ ] Writing tests
- [ ] Writing documentation

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -18,7 +18,7 @@ export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
 /**
  * DataSync plugin
  *
- * Plugin is enabled by """ @delta """ annotation
+ * Plugin is enabled by """ @datasync """ annotation
  * It will generate delta queries
  */
 export class DataSyncPlugin extends GraphbackPlugin {
@@ -32,7 +32,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
       return schema;
     }
-
+    let dataSyncModelCount = 0;
     models.forEach((model: ModelDefinition) => {
 
       this.addDataSyncMetadataFields(schemaComposer, model);
@@ -60,6 +60,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
       // Diff queries
       if (isDataSyncModel(model)) {
 
+        dataSyncModelCount+= 1;
+
         const deltaQuery = getDeltaQuery(model.graphqlType.name)
         schemaComposer.Query.addFields({
           [deltaQuery]: `${getDeltaListType(modelName)}!`
@@ -84,6 +86,10 @@ export class DataSyncPlugin extends GraphbackPlugin {
         }
       }
     })
+
+    if (dataSyncModelCount === 0) {
+      console.error("No DataSync Models detected, ensure that your models are properly annotated.")
+    }
 
     return buildSchema(schemaComposer.toSDL())
   }
--- a/packages/graphback-datasync/src/conflict/conflictEngine.ts
+++ b/packages/graphback-datasync/src/conflict/conflictEngine.ts
@@ -1,0 +1,48 @@
+import { GraphbackContext, metadataMap, FieldTransform } from '@graphback/core';
+import { ConflictStateMap } from "../util";
+
+const {fieldNames} = metadataMap;
+/**
+ * API for conflict detection and resolution engine
+ */
+export interface ConflictEngine {
+  /**
+   * checkForConflicts
+   * @param serverState document in DB
+   * @param clientState document received from client
+   * @returns ConflictStateMap if there is a conflict, undefined if not
+   */
+  checkForConflicts(serverState: any, clientState: any): ConflictStateMap | undefined 
+
+  /**
+   * checkForConflicts
+   * @param conflict ConflictStateMap received from checkForConflicts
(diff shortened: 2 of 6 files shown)


packages/graphback-datasync/src/conflict/conflictEngine.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  resolveConflicts?(conflict: ConflictStateMap): any
+
+  /**
+   * updateState
+   * Supplies the updated At
+   * @param document document to be updated
+   * @returns document with updated state
+   */
+  updateState?(): FieldTransform[]
```
wtrocki, line 30: This implementation was based on the implementations managing state (for example updating counter), but that is not reliable especially for providers like Mongo. We will not be able to `updateState` here as we actually do that outside using transformer for lastUpdated. 

If we going to do this we might actually need datasync to use it's own updatedAt implementation in updateState. This probably requires discussion but flagging it early

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -21,6 +23,17 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
     ], this.db.collection(this.collectionName)).catch((e: any) => {
       throw e;
     });
+    if (customConflictEngine !== undefined) {
```
wtrocki, line 26: Redundant check :D

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -21,6 +23,17 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
+      this.conflictEngine = customConflictEngine;
+    } else {
+      this.conflictEngine = new TimestampConflictEngine();
```
wtrocki, line 29: Discussion - conflicts enabled by default.
I think we should simply assign value and do not setup default conflict engine. Conflict engine will affect queries etc. so not anyone will want to have it when not needing this feature.

wtrocki:
Looks good so far. I would love to get your opinion about next state - we also need ability to fetch base (I send link to datasync repository that had that implemented) as current production conflict implementation is only client side

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: Agreed :+1:

packages/graphback-datasync/src/conflict/conflictEngine.ts
ssd71: Yes, I agree, my thoughts on this are pretty jumbled at the moment,
> We will not be able to updateState here as we actually do that outside using transformer for lastUpdated.

I was thinking about just that, what about using a function to transform document here:
https://github.com/aerogear/graphback/pull/1591/files/390b9b642048fddf3b586cc6ce24fd1480f47ce6#diff-be2392f5aaf3d58e02d7e9c6429f279cR151
I think it looks fair, will try it out now, but I have some concerns towards:
> If we going to do this we might actually need datasync to use it's own updatedAt implementation in updateState. 

So, it was kind of a mistake coupling that with datasync as it is vanilla provider stuff I guess, but anyhow datasync still needs updatedAt for delta queries, so what I'm thinking is that for the default strategy with timestamps, we might use a dummy updateState function which will not actually set anything, this way custom conflict engines can update state while not affecting `updatedAt`

What do you think about settling for this until we sort out #1590 ?

ssd71:
> we also need ability to fetch base (I send link to datasync repository that had that implemented) as current production conflict implementation is only client side

Definitely! I am will keep that in mind as well

wtrocki:
I think we need:
- yourself spending another day thinking about this
- have call to decide if conflicts and deltasync are separate or together.

If I add now deltasync my create queries will require timestamp but that is only required for conflicts, which is very confusing. We need to clear this as soon as possible.

Technical capabilities will follow - at first this needs to make sense from the functionality point of view

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -85,6 +87,10 @@ export class DataSyncPlugin extends GraphbackPlugin {
       }
     })
 
+    if (dataSyncModelCount === 0) {
+      console.error("No DataSync Models detected, ensure that your models are properly annotated.")
```
craicoverflow, line 91: Maybe use `this.logWarning` to keep consistent with the other plugins ([including this one](https://github.com/aerogear/graphback/blob/071c2744aed20eea9414ac9c8e445c620e1545d9/packages/graphback-datasync/src/DataSyncPlugin.ts#L31))

ssd71:
Branch is stale now, going to push prototype for this in the coming days

PR #1596, opened 2020-06-29
Revert "refactor: avoid creating a new context object when adding options"

machi1990:
Reverts aerogear/graphback#1586

### DIFF
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -441,9 +441,12 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
       }
 
       const selectedFields = getSelectedFieldsFromResolverInfo(info, model);
-      context.graphback.options = { selectedFields };
+      const graphback = {
+        services: context.graphback.services,
+        options: { selectedFields }
+      };
 
-      return context.graphback.services[modelName].create(args.input, context);
+      return context.graphback.services[modelName].create(args.input, {...context, graphback});
     }
   }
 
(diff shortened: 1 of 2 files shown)


machi1990:
When I was testing OVP I noticed that the DistributionCentre page was broken

While debugging I noticed that wrong fields where sent to the MongoDB provider 

e.g 
```
[
  'id',
  'title',
  'description',
  'status',
  'createdAt',
  'assignedAt',
  'completedAt',
  'updatedAt',
  '__typename',
  'distributionCentreId'
] product
[
  'id',
  'title',
  'description',
  'status',
  'createdAt',
  'assignedAt',
  'completedAt',
  'updatedAt',
  '__typename',
  'distributionCentreId'
] volunteeraction
```

We can see that the `product` collection received the same array of selected fields as the `volunteeraction` (indicating that it was overridden.) 

The two models are 
```

"""
@model(delete: false)
@versioned
"""
type Product {
  id: ID!
  label: String!
  description: String
  """
  @oneToMany(field: 'product')
  """
  volunteerActionProducts: [VolunteerActionProduct]
}


"""
Represents action that is assigned to volunteer

@model
@versioned
"""
type VolunteerAction {
  id: ID!
  title: String!
  description: String
  status: ActionStatus
  assignedAt: DateTime
  completedAt: DateTime
  """
  @manyToOne(field: 'actions', key: 'volunteerId')
  """
  volunteer: Volunteer
  """
  @manyToOne(field: 'actions', key: 'distributionCentreId')
  """
  distributionCentre: DistributionCentre
  """
  @oneToMany(field: 'volunteerAction')
  """
  products: [VolunteerActionProduct]
}

"""
Represents a join model between a volunteer action and product

@model(create: true, update: false, delete: false, subCreate: false, subUpdate: false, subDelete: false)
@versioned
"""
type VolunteerActionProduct {
  id: ID!
}
```

machi1990:
It appears to me that even though the context is unique (fresh created) for each incoming request but it is a shared one (singleton) for all resolvers functions within the same request meaning that we can easily override information in between resolvers (what was happening here). So let's revert this please. @wtrocki

wtrocki:
I think that:
- We need to fix this problem now by merging it (not sure if that fixes is or masks it).
- We need to figure out how to avoid multiple operations overriding the context

wtrocki:
@machi1990 Let's create follow up issue for this

PR #1607, opened 2020-06-29
fix(schema): make AND/OR fields non-nullable

craicoverflow:
This change makes filter field lists `and` and `or` non-nullable

```
and: [CommentFilter!]
or: [CommentFilter!]
```

Implements https://github.com/graphqlcrud/spec/pull/26


craicoverflow:
Merging to get this into release.

PR #1611, opened 2020-06-29
fix(graphql-migrations): map list to JSON

craicoverflow:
This fixes an issue where the `mapListToJson` migration option (which is true by default) is not passed from `migrateDB` to `generateAbstractDatabase`, resulting in the field being ignored during migrations.

To verify that this works, add this model to your PG template:

```gql
"""
@model
"""
type User {
  id: ID!
  names: [String]
}
```

You should be able to perform mutations on `names`:

```gql
mutation {
  createUser(input: {names: ["Enda"]}) {
    id
    names
  }
}
```


wtrocki:
Psttt. Integration test please ;)

craicoverflow:
Added now @wtrocki

wtrocki:
Late to the party. Can you reference issues like this in the docs so we will document this. 
There are tons of stuff that we will most likely to forget to document if not cross referenced.

craicoverflow:
This is already referenced in the docs. It is a bug as this was intended behaviour which was not working.

PR #1612, opened 2020-06-29
fix: add query options without metadata

wtrocki:


wtrocki:
Needs discussion

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -83,15 +83,19 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
...
-
-    if (context.graphback.options.selectedFields.length) {
+    let count: number; let items: T[];
+    const options = context.graphback.options;
+
+    if (options) { // query options defined, let's query db based on available informations
+      if (options.aggregations?.count) {
+        count = await this.db.count(filter);
+      }
+      if (options.selectedFields?.length) {
```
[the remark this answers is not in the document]
craicoverflow: This additional check seems redundant since it is checked in the data provider aswell:

https://github.com/aerogear/graphback/blob/4bd810161a09714e4d6cae7081518960b505e843/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts#L120-L124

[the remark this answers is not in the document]
machi1990: > What happens if `options` is defined but `options.selectedFields?.length` is not?
> 
> it looks like `findBy` won't be called.

Yeap. This should be gone.

wtrocki:
@craicoverflow Can you help me with this? I'm creator of the PR but @machi1990 done 100% work.
This works and looks like (temporary) solution to our problems

craicoverflow:
Yep, to get my attention to re-review though please use the Re-request review button so I will be pinged.

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -84,15 +84,12 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
...
+    if (context.graphback.options?.aggregations?.count) {
       count = await this.db.count(filter);
     }
 
-    let items: T[];
-
-    if (context.graphback.options.selectedFields.length) {
-      items = await this.db.findBy(filter, context, page, orderBy);
-    }
+    const items: T[] = await this.db.findBy(filter, context, page, orderBy);
```
wtrocki, line 92: 👍

wtrocki:
I think this now affects datasync stuff in generic terms. 
@ssd71 WDYT about this PR? All good

wtrocki:
Actually we should have follow up for this by adding custom operation to every template we have. 
Without that people will not know how to add their own resolvers that will be using our db or context. 
I will create follow up issue

packages/graphback-runtime-knex/src/PgKnexDBDataProvider.ts
```diff
@@ -20,7 +20,7 @@ export class PgKnexDBDataProvider<Type = any> extends KnexDBDataProvider<Type> {
     const { data: createData } = getDatabaseArguments(this.tableMap, data);
 
     //tslint:disable-next-line: await-promise
-    const dbResult = await this.db(this.tableName).insert(createData).returning(context.graphback.options.selectedFields);
+    const dbResult = await this.db(this.tableName).insert(createData).returning(this.getSelectedFields(context));
```
craicoverflow, line 23: We can actually delete this file (no longer used, moved to KnexDBDataProvider)

wtrocki: You mean this file or method?

wtrocki:
Verified locally. I was able to execute a query using service.

craicoverflow:
> Actually we should have follow up for this by adding custom operation to every template we have.

I dunno about this. If I am a developer I would expect the template I start with the contain the minimal config that I need - I did not ask for some extra resolvers which I will only remove. Wouldn't this be better served as an example on the website?

packages/graphback-runtime-knex/src/PgKnexDBDataProvider.ts
craicoverflow: File/class.

craicoverflow: It's deprecated: https://graphback.dev/docs/next/releases#deprecated

But we can probably just remove it since its pre-v1.

wtrocki:
> I dunno about this. If I am a developer I would expect the template I start with the contain the minimal config that I need - I did not ask for some extra resolvers which I will only remove. Wouldn't this be better served as an example on the website?

It is guaranteed you will need then and it is very very hard to write to them for graphback. 
As maintained I had troubles and things can break like here. 

No one reads docs/website

packages/graphback-runtime-knex/src/PgKnexDBDataProvider.ts
machi1990: I prefer if we have a separate issue "technical debt" to remove it.

craicoverflow: Absolutely.

ssd71:
> @ssd71 WDYT about this PR? All good

Taking a look

wtrocki:
Sorry @ssd71 Missed your message.

ssd71:
@wtrocki ah, it works anyway, so cool :smile:

PR #1614, opened 2020-06-29
fix: add ability to keep custom updatedAt and createdAt

### LINKED ISSUE #1603: Mongo queryBuilder not playing well with collection having a custom createdAt and updatedAt field.
The [queryBuilder.ts](https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-mongodb/src/queryBuilder.ts#L88) tries to coerce custom `createdAt` field to a timestamp even when not using the datasync plugin. 

I can have a collection with a custom `createdAt` and `updatedAt` of type DateTime as in https://github.com/aerogear/OpenVolunteerPlatform/blob/e604010b09aadbac59a2099b32e7eef183af11c3/platform/server/model/main.graphql#L82  (it is a `_createdAt` and not `createdAt` to work around this issue)

we have this logic that tries to coerce filtering values we have for these fields to timestamps - https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-mongodb/src/queryBuilder.ts#L88

which does not play well with a graphql query like https://github.com/aerogear/OpenVolunteerPlatform/blob/e604010b09aadbac59a2099b32e7eef183af11c3/platform/client-admin/src/graphql/queries/getTodayActionReport.graphql#L90

Since the value sent e.g `2020-06-29T22:00:00.000Z` is not a number.

/cc @ssd71

ssd71:
fixes #1603 
Please take a look :D

### DIFF
--- a/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
+++ b/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
@@ -1,6 +1,7 @@
 import { GraphQLObjectType } from 'graphql';
 import { ObjectId, Db, Cursor } from "mongodb"
 import { QueryFilter, ModelTableMap, buildModelTableMap, getDatabaseArguments, GraphbackContext, GraphbackDataProvider, FieldTransformMap, getFieldTransformations, TransformType, FieldTransform, GraphbackOrderBy, GraphbackPage, NoDataError } from '@graphback/core';
+import { parseMetadata } from "graphql-metadata";
 import { buildQuery } from './queryBuilder'
 import { findAndCreateIndexes } from "./utils/createIndexes";
 
@@ -16,12 +17,14 @@ export class MongoDBDataProvider<Type = any> implements GraphbackDataProvider<Ty
   protected collectionName: string;
   protected tableMap: ModelTableMap;
   protected fieldTransformMap: FieldTransformMap;
+  protected coerceTSFields: boolean;
 
   public constructor(baseType: GraphQLObjectType, db: any) {
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
     this.fieldTransformMap = getFieldTransformations(baseType);
+    this.coerceTSFields = parseMetadata("versioned", baseType)
     findAndCreateIndexes(baseType, this.db.collection(this.collectionName)).catch((e: Error) => {
       throw e
     })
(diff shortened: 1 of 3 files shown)


packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -16,12 +17,14 @@ export class MongoDBDataProvider<Type = any> implements GraphbackDataProvider<Ty
...
   protected tableMap: ModelTableMap;
   protected fieldTransformMap: FieldTransformMap;
+  protected coerceTSFields: boolean;
 
   public constructor(baseType: GraphQLObjectType, db: any) {
     this.db = db;
     this.tableMap = buildModelTableMap(baseType);
     this.collectionName = this.tableMap.tableName;
     this.fieldTransformMap = getFieldTransformations(baseType);
+    this.coerceTSFields = parseMetadata("versioned", baseType)
```
machi1990, line 27: Instead of doing this maybe you could define a method `protected coerceVersionFields()`  which for this base class would return false and for the datasync one it returns true.

wtrocki:
Intentions here were good but actual issue is related to missunderstanding. I think this is feature and do not need much fix. Lets chat about general goals for DataSync first so we will be all the same spot.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
ssd71: Sorry for late response! Do you think it should always return false for base class? We have the `@versioned` annotation so was wondering if that would become non-functional for plain mongo if we do that?

machi1990: ah right, I thought the `@versioned` was handled by the `DataSyncProvider`. My concern is doing the parsing here which could later impact performance.

ssd71: Hmm fair point about performance, if we do a function for this, I was thinking the parser would be called everytime a query is executed, which kinda seems like a heavier hit wdyt

machi1990: Ideally this information should be carried in a `model: ModelDefinition`. But we do not have that at the moment, so let's keep it this way for now while waiting for https://github.com/aerogear/graphback/pull/1614#issuecomment-651329839 chat

wtrocki:
As explained on issue. I was wrong here. We need to ignore user provided updatedAt

wtrocki:
Kinda missing how this works and how this address the problem. Any description will be great

ssd71:
@wtrocki This PR makes it so that queryBuilder takes a coerceTSFields parameter which it uses to decide if it should convert the `updatedAt` etc field values (in a filter object) to numeric values for comparison with documents in the database

wtrocki:
I would change name of the fields to reflect things. Also holding this in model could be good option - but that it separate PR. I guess ok to merge it but the name/field seems overly complex /redundant - we should just read annotation and act on it.

ssd71:
Hmm that sounds cool, I will see if I can manage something, but not sure I can make it for this release

wtrocki:
I meant let's rename this cryptic field to something like - generateUpdatedAt flag or something.

wtrocki:
Happy to merge it though for current release. Let's leave rest for discussion in the issue that @machi1990 created

PR #1620, opened 2020-06-30
feat(core): add JSON scalar support

craicoverflow:
This is the result of an investigation into how we could implement out of the box JSON scalar support in Graphback.

Things to note:

It works very well in all cases EXCEPT we would only be able to offer very minimal field filtering capabilities on it (in Postgres anyway)

Knex.js does not have an operator for comparing values in a JSON column. Using the query mapper to do so throws an error like this:

`select \"id\", \"testObj\" from \"comment\" where \"testObj\" = $1 - operator does not exist: json = unknown`.

We could offer very limited filtering on JSON objects:

```graphql
input GraphbackJSONInput {
   null: Boolean 
}
```

Or else don't give any filtering capabilities on this field. The same idea can be replicated for embedded types also.

## Using

To use the JSON scalar, simply add a JSON scalar to your data model:

```gql
""" @model """
type Comment {
  id: ID!
  text: String
  description: String
  test: JSON
}

scalar JSON
```

```gql
 createComment(input: {test: {
    some:"some"
    random:"random",
    props:"props"
  }}) {
    test
  }
```


craicoverflow:
I've done this quick spike in between issues

Basically we can add JSON scalars easily but there will be problems with filtering (but we can disable filtering on that field or limit it).

Not going to add tests or anything for this yet.

wtrocki:
> Basically we can add JSON scalars easily but there will be problems with filtering (but we can disable filtering on that field or limit it).

For JSON scalars filtering makes no sense for me. I would expect every scalar work on field level. 
Can you propose list of the operations possible on JSON scalar - it could be seen as string right?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -75,7 +75,8 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     const resolvers: IResolvers = {
       Query: {},
       Mutation: {},
-      Subscription: {}
+      Subscription: {},
+      JSON: GraphbackJSON
```
wtrocki, line 79: This will clash with people having JSON. We need to introduce our own namespace scalar - not 100% sure but in all client apps we have JSON already defined

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable max-lines */
 import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve, dirname, join } from 'path';
-import { getFieldName, metadataMap, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap, getInputTypeName, FieldRelationshipMetadata, GraphbackContext, getSelectedFieldsFromResolverInfo, isModelType, isSpecifiedGraphbackScalarType, getPrimaryKey } from '@graphback/core'
+import { getFieldName, metadataMap, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, buildGeneratedRelationshipsFieldObject, buildModifiedRelationshipsFieldObject, buildRelationshipFilterFieldMap, getInputTypeName, FieldRelationshipMetadata, GraphbackContext, getSelectedFieldsFromResolverInfo, isModelType, isSpecifiedGraphbackScalarType, getPrimaryKey, GraphbackJSON } from '@graphback/core'
```
wtrocki, line 4: There are certain differences between JSON implementations. Any reason for using current scalar?

wtrocki:
Do you plan to use JSONObject scalar as well?

wtrocki:
I checked OVP and I think that this stuff is redundant. 
First, is feature where we are at code freeze and doing mostly docs and bugfixes.
In all sample apps we already using JSON scalars and they work + filtering is generated automatically. 
Tell me like I'm 5 - is that only convenience method - because if it is it might backfire with people getting their JSON scalars (commonly JSONObject) being swapped and getting some issues with arrays.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Current scalar instead of what?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
craicoverflow: Yeah, thing is that the name of the scalar comes from `graphql-type-json`, so even if we changed the resolver name, the scalar being used is still the same and it affects things like FilterInput generation.

There are ways around it, by adapting whats in https://github.com/aerogear/graphback/pull/1620/files#diff-f3d890fe21a2f3c11be271e48b1450f0R12 to not check the scalar name, but it seemed a bit hacky to me.

craicoverflow:
>  Can you propose list of the operations possible on JSON scalar - it could be seen as string right?

You would like to be able to to return all items where a JSON field is null/not null.

wtrocki:
> Are you sure filtering works on JSON fields? I believe the filter will not work.

I can do eq and not eq with full value on Mongo. Kinda leaving postgress at the moment.
If we see this way this is actually an bug - generated filter types for scalars do not work, but I would love to chat about it on issue to see how this spreads across mongo/pg etc.

craicoverflow:
> generated filter types for scalars do not work

I think it is really only for JSON scalars and should work for simpler scalars like BigInt etc.

craicoverflow:
Also, that is only 2 filter fields for the generated filter type - the others will not work so it is not appropriate to generate these extra filters like `lt`, `gt` etc. We need to take control of the JSON scalar.

wtrocki:
100% agree. Thanks for explaining

craicoverflow:
> Do you plan to use JSONObject scalar as well?

I did not want to fully implement this yet as this PR is mainly from investigation into how we can add some default scalars. I had checked AppSync though and they only have a `AWSJSON` scalar:

https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

craicoverflow:
I checked this morning how well it would play by naming this scalar `GraphbackJSON` and also adding a `JSON` scalar from the template.

```gql
""" @model """
type Comment {
  id: ID!
  text: String
  description: String
  test1: JSON
  test: GraphbackJSON
}

scalar GraphbackJSON 
scalar JSON # from graphql-scalars
```

Since both scalars have a real name of "JSON", Graphback will assign them to the single scalar JSON in the generated schema:

```gql
type Comment {
  id: ID!
  text: String
  description: String
  test1: JSON
  test: JSON
  note: Note
}
```

And will use the single scalar in the filters:

```
input MutateCommentInput {
  id: ID!
  text: String
  description: String
  test1: JSON
  test: JSON
  noteId: ID
}
```

While in theory both scalars are different implementations of the same result, there could be a divergence in the future, but also there is a greater chance of difference in other scalars like DateTime which is less straightforward.

I see two options:

1. Bring the code from [graphql-type-json](https://github.com/taion/graphql-type-json) (or whichever we decide to use) to gain full control over the scalar. This is actually not a bad idea because we can control changes to its design in the future - but it means more maintenance on us.
2. Try to rename the scalar dynamically using [graphql-compose](https://graphql-compose.github.io/docs/api/ScalarTypeComposer.html).

I prefer option 2: it should be straightforward and reduce our maintenance cost.

wtrocki:
Vote nr2. Also I wote to use graphql/scalars package as we will need ObjectID exposed as well

machi1990:
> Vote nr2. Also I vote to use graphql/scalars package as we will need ObjectID exposed as well

+100

craicoverflow:
Using graphql-scalars now as of 82504a0f41226e55ebf0f2501c1776ef1f6ebbc6

Although I don't see any difference between JSON and JSONObject right now (in input/output)

craicoverflow:
I am leaving this in draft, as it is far from final, but is facilitating good discussion and foundations into how we can add default Graphback scalars.

wtrocki:
TL;DR `JSONObject: {} `, `JSON: []`

craicoverflow:
> TL;DR `JSONObject: {} `, `JSON: []`

I know what they are _supposed_ to represent but I am able to insert an array into `JSONObject`

wtrocki:
ROFL :D

packages/graphback-core/src/scalars/scalars.ts
```diff
@@ -1,12 +1,34 @@
...
+const graphbackScalarTypes: GraphQLScalarType[] = [GraphbackJSON, GraphbackJSONObject]
 
 /**
  * Checks if the type is on the the default Graphback supported scalars
  *
- * @param type - GraphQL type
+ * @param type - any type
  */
 export function isSpecifiedGraphbackScalarType(type: GraphQLNamedType): boolean {
   return graphbackScalarTypes.some(({ name }: GraphQLScalarType) => type.name === name);
```
machi1990, line 34: If we had 
```
const graphbackScalarTypes = {
       GraphbackJSON: true,
       GraphbackJSONObject: true
}
```

This check could be re-written to 
```
return graphbackScalarTypes[type.name];
```

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -84,6 +75,30 @@ export const BooleanScalarInputType = new GraphQLInputObjectType({
...
 
+/**
+ * Custom GraphbackJSON scalar
+ */
+export const GraphbackJSONScalarInputType = new GraphQLInputObjectType({
+  name: getInputName(GraphbackJSON),
+  fields: {
+    ne: { type: GraphQLString },
+    eq: { type: GraphQLString },
+    contains: { type: GraphQLString }
```
machi1990, line 86: Curious to see how this filtering works for PG.

packages/graphback-core/src/scalars/scalars.ts
craicoverflow: True, but then you have to maintain two lists referencing the Graphback scalars (the actually scalars + `graphbackScalarTypes`.

I copied the same function from graphql-js:

https://github.com/graphql/graphql-js/blob/58aaa1df30b501b67d6986ea2fc1a8b95ac9fbca/src/type/scalars.js#L286

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
craicoverflow: It doesn't :( `contains` will need to go.

packages/graphback-core/src/scalars/scalars.ts
machi1990: > True, but then you have to maintain two lists referencing the Graphback scalars (the actually scalars + `graphbackScalarTypes`.
> 
> I copied the same function from graphql-js:
> 
> https://github.com/graphql/graphql-js/blob/58aaa1df30b501b67d6986ea2fc1a8b95ac9fbca/src/type/scalars.js#L286

I might have missed something. Why maintain two lists?

craicoverflow: Maybe not two "lists" but you would no longer have a single source of truth. If the actual scalar gets renamed to GBJSON you will also need to update the key `graphbackScalarTypes.GraphbackJSON` to match it.

Whereas with the current implementation, The list is created from the scalar objects and referenced in this function.

https://github.com/aerogear/graphback/blob/json-scalar/packages/graphback-core/src/scalars/scalars.ts#L9-L26

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: It is more a problem with knex.js rather than pg right? I mean, we'll have to send the right json operator in the built query since knex cannot do that for us automatically as for a string?

packages/graphback-core/src/scalars/scalars.ts
machi1990: > Maybe not two "lists" but you would no longer have a single source of truth. If the actual scalar gets renamed to GBJSON you will also need to update the key `graphbackScalarTypes.GraphbackJSON` to match it.

I see what you mean, I guess this would limit such an issue
```js
const graphbackScalarTypes = {
       [GraphbackJSON.name]: true,
       [GraphbackJSONObject.name]: true
}
```

It is a nitpick. I've no strong opinion on this.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
craicoverflow: Knex does not map operators for querying JSON.

It's possible, but not a quick feature to add. [knexQueryMapper]() does not know the column type when mapping operators, and the syntax for querying JSON columns is completely different:

```
select comments->>'name' from note where (comments->'complete')::jsonb ? TRUE;
```

craicoverflow:
> I can do eq and not eq with full value on Mongo. Kinda leaving postgress at the moment.

@wtrocki Did you stringify the full value before applying it to the filter?

craicoverflow:
Closing and can be reopened at a later date if we decide to follow up.

packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
machi1990: Yes, this kind of query mapping is not a quick feature to add. For example, to do this we'll need to have the column info in the `model: ModelDefinition`.

PR #1621, opened 2020-06-30
fix: start server if schema has no models

### LINKED ISSUE #1427: Graphback server will not start if there are no model-annotated types.
Here is my schema:

```
type User {
  id: ID!
  """
  @id
  """
  email: String
}
```

When trying to start the server I get an error and the server does not start.

```
Warning - SchemaCRUD: Provided schema has no models. Returning original schema without any changes.
GraphbackPlugin.ts:29
Error: "Query" defined in resolvers, but not in schema
```

Although there is no value in using Graphback without any model types, the server should still start.

craicoverflow:
Fixes #1427 

This allows the server to start if there are no types annotated by `@model`.

To verify, start a server with this model:

```
type User {
  name: String
}

type Query {
  hello: String
}
```


craicoverflow:
I thought about integrations tests and it would require a whole new intgration suite without any models to verify this works.

I suppose that would be quick to add as it would be very minimal.

wtrocki:
I think what we talking here is just feature to have something like graphql-compose guy. 
Integration tests with various inputs. We have now single input that over long term with restrict our use cases or make this schema large. However this might be very very large amount of work. 
I will try to see how we can pull that off easily - it will be cool but then it can take time and be huge maintenance issue

PR #1631, opened 2020-07-01
docs: add new site layout

craicoverflow:
See #1526 

This PR adds the basic new docs layout/site navigation as defined in #1526 

I have not added every possible page that will exist in these sections but I have added most of them and we can fill them out as we go.

I have backed up the old docs files in `docs/.old` in 20151e7a7b34ce1c345422dce53dad7a785e8fa4 to help with writing the new docs.

## Verification

1. `cd website`
2. `yarn start`


craicoverflow:
I'm going to merge this, since it is blocking me and it is just docs - so the library will not be affected.

docs/.old/authentication/keycloak.md
```diff
@@ -0,0 +1,114 @@
+---
```
wtrocki, line 1: Why this file is in .old?

wtrocki: Ignore ignore me

PR #1633, opened 2020-07-01
Keycloak fix

wtrocki:
Fixing critical security issues for keycloak:

- You can still subscribe to data despite not having permission
- You can fetch object using relationships.

Verified in datasync starter

Added new features:

- auth by input fields
- auth by output fields
- filtering added on server side


wtrocki:
Still working on integration tests and adding more features.

wtrocki:
Early review for feedback. Working on docs and testing

wtrocki:
Based on the chat with existing users I have collected some requirements and implemented them. All this list concludes list of the features that people would like to have for Auth

packages/graphback-keycloak-authz/src/KeycloakCrudService.ts
```diff
@@ -66,41 +69,116 @@ export class KeycloakCrudService<T = any> extends GraphbackProxyService<T> {
...
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Checks if user is allowed to request particular field
+   */
+  private checkAuthRulesForSelections(context: any) {
```
craicoverflow, line 170:
```suggestion
  private checkAuthRulesForSelections(context: GraphbackContext) {
```

packages/graphback-keycloak-authz/src/KeycloakCrudService.ts
```diff
@@ -66,41 +69,116 @@ export class KeycloakCrudService<T = any> extends GraphbackProxyService<T> {
...
+        throw new UnauthorizedError()
+      }
+    }
+
     return super.subscribeToDelete(filter, context)
   }
 
-  public batchLoadData(relationField: string, id: string | number, filter: any, context: any) {
+  public batchLoadData(relationField: string, id: string | number, filter: any, context: GraphbackContext) {
+    if (this.authConfig?.relations[relationField]?.roles.length > 0) {
```
machi1990, line 141: I see that this `relations` rule is only applied here, will it also work for `oneToOne` or `manyToOne` relationship?

wtrocki: Good point. It will actually require us to refactor it. We seem to do queries on resolver which is bad.

wtrocki:
Challenge I found out here is that filter is untyped. I will try to write typing for filter as it is really backward to work with graphback context now without it. Also we can provide better typings for graphback context to serve customized code :D

machi1990:
> Challenge I found out here is that filter is untyped. I will try to write typing for filter as it is really backward to work with graphback context now without it. Also we can provide better typings for graphback context to serve customized code :D

A typed filter will be a huge plus. Thanks!

wtrocki:
Documentation requires major revamp. I want to do it slowly here: https://github.com/aerogear/graphback/issues/1642

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -105,7 +104,7 @@ export class CRUDService<T = any> implements GraphbackCRUDService<T>  {
     }
   }
 
-  public subscribeToCreate(filter: any, context: GraphbackContext): AsyncIterator<T> | undefined {
+  public subscribeToCreate(filter: any, context: GraphbackContext): AsyncIterator<Type> | undefined {
```
machi1990, line 107:
```suggestion
  public subscribeToCreate(filter: QueryFilter<Type>, context: GraphbackContext): AsyncIterator<Type> | undefined {
```

filter missing type?

packages/graphback-keycloak-authz/src/KeycloakCrudService.ts
machi1990: Is there a way to solve it?

packages/graphback-core/src/runtime/CRUDService.ts
wtrocki: I would not put this here as we do not support any filtering. So if I put anything it will be instantly breaking change where we do not know what to filter and how

machi1990: Okay, very good point.

wtrocki:
@machi1990 I have verified this by testing it with datasync starter.
Keycloak-authz is still not officially part of the platform however it will be documented as beta.

Would you consider giving me credit here and accept this changes?

packages/graphback-keycloak-authz/src/KeycloakCrudService.ts
wtrocki: Hmm.. I have opened issue for this and explained that this is documentation issue

#1638

machi1990: Consider the following scenarios: 

I have the following business models, showing relationships between between `One` and `Two`.

```graphql
  type One {
      id: ID
      """ @oneToOne(field: 'prev') """
      next: Two
    }

    """
    @model 
    """
    type Two {
      id: ID!
      prev: One
    }
```

I have two roles: `basic`, `advanced`.

If I have a `auth` config like this:

```ts
  const authConfig = {
      One: {
        relations: { prev: { roles: ['advanced'] } }
      }
  }
```  

With this configuration, I want the user to retrieve only the `next` only if they have advanced level. 

The typical graphql query I'll be using for that is: 
```graphql
query {
    findOnes {
        items {
            id, 
            next {
                id                
            }
        }
    }
}

```

This query will land me to this resolver https://github.com/aerogear/graphback/blob/6d9538776f64393b459401f5f4fdbaf730439c11/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L697 

which will land us here https://github.com/aerogear/graphback/blob/6d9538776f64393b459401f5f4fdbaf730439c11/packages/graphback-keycloak-authz/src/KeycloakCrudService.ts#L74

which is current missing the the relations authz check. 

The solution to this could to find a way to add those checks or documentating that this could be achieved by using the https://github.com/aerogear/graphback/blob/6d9538776f64393b459401f5f4fdbaf730439c11/packages/graphback-keycloak-authz/src/KeycloakCrudService.ts#L82 (field selection on the joining column )

wtrocki: Actually that is the problem. @machi1990 Let's port that to the new issue

PR #1643, opened 2020-07-02
fix: add example of the custom resolvers

wtrocki:
Openining for people to play with. Generally experience is still pretty.. bad :)
Unless we do type generation we cannot do much.

### DIFF
--- a/templates/ts-apollo-mongodb-backend/model/Note.graphql
+++ b/templates/ts-apollo-mongodb-backend/model/Note.graphql
@@ -1,4 +1,6 @@
-""" @model """
+"""
+ @model
+"""
 type Note {
   id: ID!
   title: String!
@@ -9,9 +11,15 @@ type Note {
   comments: [Comment]!
 }
 
-""" @model """
+"""
+ @model
+"""
 type Comment {
   id: ID!
   text: String
   description: String
+}
+
+type Query {
+  printNote: String
 }
--- a/templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
+++ b/templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
@@ -1,0 +1,25 @@
+import { GraphbackContext, GraphbackCRUDService, Maybe, Scalars } from "graphback";
+
+/**
+ * Types can be generated using GraphQL-Code-Generator
+ */
+export type Note = {
+  id?: Maybe<Scalars['String']>;
+  title?: Maybe<Scalars['String']>;
+  description?: Maybe<Scalars['String']>;
+};
+
+const getNoteService = (context: GraphbackContext) => {
+  return context.graphback?.services?.Note as GraphbackCRUDService<Note>
+}
+
+export const customResolvers = {
+  Query: {
+    printNote: async (parent: any, variables: any, context: GraphbackContext, info: any) => {
+      const noteService = getNoteService(context);
+      const notes = await noteService.findBy({}, { graphback: {} });
+
+      return JSON.stringify(notes);
+    }
+  }
+}
--- a/templates/ts-apollo-mongodb-backend/src/index.ts
+++ b/templates/ts-apollo-mongodb-backend/src/index.ts
@@ -1,15 +1,17 @@
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 require('dotenv').config()
 import path from 'path'
+import http from "http"
+import express from "express"
+import cors from "cors"
 import { ApolloServer } from "apollo-server-express"
 import { buildGraphbackAPI } from 'graphback'
 import { loadSchemaSync } from '@graphql-tools/load'
 import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader'
 import { createMongoDbProvider } from '@graphback/runtime-mongo'
-import cors from "cors"
-import express from "express"
-import http from "http"
+
 import { connectDB } from './db'
+import { customResolvers } from "./custom-resolvers"
 
 async function start() {
   const app = express()
@@ -30,7 +32,7 @@ async function start() {
 
   const apolloServer = new ApolloServer({
     typeDefs,
-    resolvers,
+    resolvers: [resolvers, customResolvers],
     context: contextCreator
   })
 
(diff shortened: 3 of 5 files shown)


templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const getNoteService = (context: GraphbackContext) => {
+  return context.graphback?.services?.Note as GraphbackCRUDService<Note>
+}
+
+export const customResolvers = {
+  Query: {
+    printNote: async (parent: any, variables: any, context: GraphbackContext, info: any) => {
+      const noteService = getNoteService(context);
+      const notes = await noteService.findBy({}, { graphback: {} });
```
machi1990, line 20:
```suggestion
      const notes = await noteService.findBy({}, context);
```

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const getNoteService = (context: GraphbackContext) => {
+  return context.graphback?.services?.Note as GraphbackCRUDService<Note>
+}
+
+export const customResolvers = {
+  Query: {
+    printNote: async (parent: any, variables: any, context: GraphbackContext, info: any) => {
+      const noteService = getNoteService(context);
```
craicoverflow, line 19: How about a helper on the context to fetch a service?

```suggestion
      const noteService: GraphbackCRUDService<Note> = context.getService(Note)
```

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
```diff
@@ -0,0 +1,25 @@
...
+  description?: Maybe<Scalars['String']>;
+};
+
+const getNoteService = (context: GraphbackContext) => {
+  return context.graphback?.services?.Note as GraphbackCRUDService<Note>
+}
+
+export const customResolvers = {
+  Query: {
+    printNote: async (parent: any, variables: any, context: GraphbackContext, info: any) => {
```
craicoverflow, line 18:
```suggestion
    printNotes: async (parent: any, variables: any, context: GraphbackContext, info: any) => {
```

templates/ts-apollo-mongodb-backend/src/index.ts
```diff
@@ -30,7 +32,7 @@ async function start() {
 
   const apolloServer = new ApolloServer({
     typeDefs,
-    resolvers,
+    resolvers: [resolvers, customResolvers],
```
craicoverflow, line 35: It would be weird to call your resolvers `customResolvers` in a regular API. We should maybe change `resolvers` to `graphbackResolvers` or call `customResolvers` `noteResolvers`

templates/ts-apollo-mongodb-backend/model/Note.graphql
```diff
@@ -9,9 +11,15 @@ type Note {
   comments: [Comment]!
 }
 
-""" @model """
+"""
+ @model
```
craicoverflow, line 15:
```suggestion
@model
```

templates/ts-apollo-mongodb-backend/model/Note.graphql
```diff
@@ -1,4 +1,6 @@
...
```
craicoverflow, line 2:
```suggestion
@model
```

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
craicoverflow: I feel like we are just patching the symptoms of a clunky API though.

craicoverflow:
Noticed we are actually returning the services as a top-level value in the main API also:

```
 return {
  schema: schemaWithResolvers,
  typeDefs,
  resolvers,
  services,
  contextCreator
}
```

We should probably take that out.

https://github.com/aerogear/graphback/blob/master/packages/graphback/src/buildGraphbackAPI.ts#L47
https://github.com/aerogear/graphback/blob/master/packages/graphback/src/buildGraphbackAPI.ts#L142
```

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
machi1990: I was leading this https://graphql-modules.com/docs/introduction/dependency-injection#providers this morning and I like their API: Not to bring in a full fledged DI container but we kind of have it already and having something like `context.services.get(serviceName)` would be a big plus.

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
wtrocki: Try doing it :) It will fail with `graphback is undefined`

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
wtrocki: That will be kinda cool, but you probably meant this?

```ts
context.service<User>("User")
````

templates/ts-apollo-mongodb-backend/src/index.ts
wtrocki: yep!

wtrocki:
Actually us returning services could be really cool alternative. Let's chat about it. It will mean that you will need to inject services yourself to resolvers but context is really hard to make strong typed so best to spent more time on this and improve this experience for 0.15.0.

This is not instant priority - We can try things like SQLmancer etc.

templates/ts-apollo-mongodb-backend/src/custom-resolvers.ts
machi1990: Hmm something fishy might be going on then. Let me re-check. We've it here https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/src/resolvers/custom-resolvers.ts#L25 without any issue.

wtrocki:
we need little bit more work in graphback to improve this

wtrocki:
This got outdated. makes no sense to keep this

Issue #1445, opened 2020-06-15
Investigate conflict resolution capabilities on top of the delta store


wtrocki:
When building delta store (separate table for deltas) we should take into consideration abilities to provide conflict resolution based on history etc.

ssd71:
If we would need a separate delta store anyway, what do you think about using debezium for this, I would need to look around a bit to see if it is viable for our needs. The way I see it is that if we are going to do a separate delta store when we use datasync with postgres, I wonder if it could be worthwhile to use debezium in place of that

wtrocki:
Trick is to have:
1) one constant delta schema standard that works
2) Default implementation for DataSync involving versioned
3) ability to plug in different field or strategy

We have 1 and 2 at the moment

wtrocki:
Datastore is different to using different field for specific datastore/strategy. You can use soft deletes but still use version field to ensure consistency and different field to partition data

ssd71:
Hi, sorry for being late here, but I think I see what you mean: It is entirely different thing, so I should focus on soft deletes for now, is that it?

wtrocki:
Not sure what you mean. We need to be able to control consistency field for soft deletes. Let's create issue

ssd71:
Yes, done #1524

wtrocki:
Also worth to mention. WE had this really early versions of the other strategies that involve different tables. More info here:

https://github.com/aerogear/datasync/issues/46
https://github.com/aerogear/datasync/pull/45/files

machi1990:
@ssd71 what's the update on this one, is the investigation concluded with https://github.com/aerogear/graphback/issues/1738 ?

ssd71:
@machi1990 We do not yet have the ability to specify a different kind of State management(we only check `_version`) (drawing parallels from [here](https://github.com/aerogear/offix/blob/master/packages/offix/conflicts-server/src/api/ObjectState.ts)), but I believe that needs another issue?

wtrocki:
Let's close this

Issue #1381, opened 2020-05-20
DataSync - handling properly data deletion - Phase2


wtrocki:
## Feature Request

Offix is not handling data deletion properly. When data is deleted we are loosing ability to restore the conflict using common strategies. 

What we need to do is to introduce concept of the soft deletes for offix and bring more comprehensive data handlers. 

## How 

Graphback will define number of the patterns that developers can apply on the server:

- Soft deletes and order by timestamp
- Separate tables with soft deletes
- Separate tables with event driven approach (as log) 

The reference implementation for those patterns will be available thru separate package offered by https://graphback.dev so anyone can use it as simple CRUD interface for typescript and reimplement it for other languages of the choice

Protocol will be documented in offix (as we have it done now)

wtrocki:
CC @ssd71 

We have moved this issue from the offix. It is very closely relying on the work we doing now

wtrocki:
Let's reserach delta tables with event driven approach

PR #1647, opened 2020-07-03
Add delta table

ssd71:
This is a spike towards investigating Delta Tables for datasync, can be tested out as follows:

- for delta Table

This is enabled in schema
Use the following schema with datasync template

```graphql
""" 
@model
@datasync(deltaTable: true)
"""
type Comment {
  id: ID!
  text: String
  description: String
}
```
It should create a comment_delta collection in the database

- conflict

Conflicts are enabled in code
Import TimestampConflictEngine from datasync package:
```typescript
import { TimestampConflictEngine } from '@graphback/datasync'
```

Provide a map of model name to conflict strategy(pardon the horrendous inconsistency in the names, I would love some feedback about what to call them) in createDataSyncMongoDbProvider as follows:
```typescript
createDataSyncMongoDbProvider(db, { Comment: TimestampConflictEngine })
```
This tells it to use Timestamp based conflicts for the Comment model, there is also another strategy intended for use with delta tables called DeltaConflictEngine, it checks if the fields to be updated have changed since the client passed `updatedAt` timestamp, and raises a conflict if they have, I would love some feedback about this as well.

Note, if using delta table, sync query only returns fields that have changed, not sure this is optimal so feedback needed

### DIFF
--- a/packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
+++ b/packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
@@ -3,14 +3,16 @@ import { getDatabaseArguments, metadataMap, GraphbackContext, NoDataError, Trans
 import { ObjectId } from 'mongodb';
 import { MongoDBDataProvider, applyIndexes } from '@graphback/runtime-mongo';
 import { ConflictError, ConflictStateMap } from "../util";
+import { ConflictEngine, TimestampConflictEngine } from "../conflict";
 import { DataSyncProvider } from "./DataSyncProvider";
 
 /**
  * Mongo provider that attains data synchronization using soft deletes
  */
 export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider<Type> implements DataSyncProvider {
+  protected conflictEngine: ConflictEngine;
 
-  public constructor(baseType: GraphQLObjectType, client: any) {
+  public constructor(baseType: GraphQLObjectType, client: any, ConflictStrategy?:new () => ConflictEngine) {
     super(baseType, client);
     applyIndexes([
       {
@@ -21,6 +23,10 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
     ], this.db.collection(this.collectionName)).catch((e: any) => {
       throw e;
     });
+
+    if (ConflictStrategy !== undefined){
+      this.conflictEngine = new ConflictStrategy();
+    }
   }
 
   public async create(data: any, context: GraphbackContext): Promise<Type> {
@@ -36,6 +42,10 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
       throw new ConflictError(conflict);
     }
 
+    if (this.conflictEngine !== undefined) {
+      Object.assign(data, this.conflictEngine.updateState(data));
+    }
+
     // TODO use findOneAndUpdate to check consistency afterwards
     return super.update(data, context);
   }
@@ -47,7 +57,7 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
     }
 
     const { idField } = getDatabaseArguments(this.tableMap, data);
-    data = {};
+    delete data.id
 
     this.fieldTransformMap[TransformType.UPDATE]
       .forEach((f: FieldTransform) => {
@@ -113,33 +123,38 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
   }
 
   protected async checkForConflicts(clientData: any, context: GraphbackContext): Promise<ConflictStateMap> {
-    const { idField } = getDatabaseArguments(this.tableMap, clientData);
-    const { fieldNames } = metadataMap;
+    if (this.conflictEngine !== undefined) {
+
+      const queryResult = await this.getServerState(clientData);
+      if (queryResult) {
+        const conflict = this.conflictEngine.checkForConflicts(queryResult, clientData)
+
+        if (conflict !== undefined && this.conflictEngine.resolveConflicts !== undefined) {
+          const resolved = this.conflictEngine.resolveConflicts(conflict);
+
+          if (resolved !== undefined) {
+            Object.assign(clientData, this.conflictEngine.resolveConflicts(conflict));
+
+            return undefined
+          }
+        }
+
+        return conflict
+      }
+
+      throw new NoDataError(`Could not find any such documents from ${this.collectionName}`);
+    }
+
+    return undefined;
+  }
+
+  protected async getServerState(clientState: any) {
+    const { idField } = getDatabaseArguments(this.tableMap, clientState);
 
     if (!idField.value) {
       throw new NoDataError(`Couldn't get document from ${this.collectionName} - missing ID field`)
     }
-
-    const projection = this.buildProjectionOption(context);
-
-    if (projection) {
-      projection[fieldNames.updatedAt] = 1;
-      projection._deleted = 1
-    }
-
-    const queryResult = await this.db.collection(this.collectionName).findOne({ _id: new ObjectId(idField.value), _deleted: { $ne: true } }, { projection });
-    if (queryResult) {
-      queryResult[idField.name] = queryResult._id;
-      if (
-        queryResult[fieldNames.updatedAt] !== undefined &&
-        clientData[fieldNames.updatedAt].toString() !== queryResult[fieldNames.updatedAt].toString()
-      ) {
-        return { serverState: queryResult, clientState: clientData };
-      }
-
-      return undefined;
-    }
-
-    throw new NoDataError(`Could not find any such documents from ${this.collectionName}`);
+    
+    return this.db.collection(this.collectionName).findOne({ _id: new ObjectId(idField.value), _deleted: { $ne: true } });
   }
 }
--- a/packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
+++ b/packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
@@ -1,0 +1,178 @@
+import { GraphQLObjectType } from 'graphql';
+import { getDatabaseArguments, metadataMap, GraphbackContext, NoDataError, TransformType, FieldTransform, GraphbackOrderBy, GraphbackPage } from '@graphback/core';
+import { ObjectId } from 'mongodb';
+import { MongoDBDataProvider, applyIndexes } from '@graphback/runtime-mongo';
+import { ConflictError, ConflictStateMap } from "../util";
+import { DataSyncProvider } from "./DataSyncProvider";
+import { DeltaProvider } from "./DeltaProvider";
+
+/**
+ * Mongo provider that attains data synchronization using soft deletes
+ */
+export class DeltaDBDataProvider<Type = any> extends MongoDBDataProvider<Type> implements DataSyncProvider {
+  protected deltaProvider: DeltaProvider;
+
+  public constructor(baseType: GraphQLObjectType, client: any) {
+    super(baseType, client);
+    applyIndexes([
+      {
+        key: {
+          _deleted: 1
+        }
+      }
+    ], this.db.collection(this.collectionName)).catch((e: any) => {
+      throw e;
+    });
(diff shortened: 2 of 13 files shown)


packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    ], this.db.collection(this.collectionName));
+  }
+
+
+  public async insertDiff(data: any) {
+    delete data.id;
+
+    const res = await this.db.collection(this.collectionName).insertOne({
+      doc: data
```
wtrocki, line 56: We might need more information here. Like what operation was executed etc.

packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    filter._deleted = { ne: true};
+
+    return super.count(filter);
+  }
+
+  public sync(lastSync: string, context: GraphbackContext, filter?: any): Promise<Type[]> {
+
+    return this.deltaProvider.sync(parseInt(lastSync, 10)).then((res: any[]) => {
```
wtrocki, line 144: So we went to the 1.5 - WE have separate delta table with duplicate of the data. (Not really delta table itself.)?

packages/graphback-datasync/src/providers/DeltaProvider.ts
ssd71: Agreed :+1:

packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
ssd71: It is open for discussion, really, I just wanted get some feedback, I haven't hacked around with it that much to make a decisive move *yet*

wtrocki: yep. No worries. This works and it is very very well written. But we cannot get conflict information from it. Let's drop some short research to see how we can get that done

packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+        data[f.fieldName] = f.transform(f.fieldName);
+      });
+
+    const objectId = new ObjectId(idField.value);
+    const result = await this.db.collection(this.collectionName).findOneAndUpdate({ _id: objectId }, { $set: data }, { returnOriginal: false });
+    if (result.ok) {
+      
+      this.deltaProvider.insertDiff(result.value).catch((e: any) => {
+        console.error(`Error: Couldn't insert into delta table: ${e}`);
+      })
```
craicoverflow, line 66: This block in `update`, `delete` and `create` seem to be the only things that make this provider different to `DatasyncMongoDBDataProvider`.

Could this just extend that instead, to save repeating code in two providers and increasing maintenance costs.

Or would it be feasible to use the same provider, and only execute this block if `this.deltaProvider` has been created, which could be conditionally created based on the existence of the the `deltaTable` annotation key.

This got me thinking also - maybe we should be passing the `ModelDefinition` to the provider instead of the base GraphQL type, as a lot of the preparsed metadata information is not available from the base type and you would have to re-parse it.

machi1990: > This got me thinking also - maybe we should be passing the ModelDefinition to the provider instead of the base GraphQL type, as a lot of the preparsed metadata information is not available from the base type and you would have to re-parse it.

+1, this rejoins this comment https://github.com/aerogear/graphback/pull/1614#discussion_r448159331

packages/graphback-datasync/src/providers/DeltaDbDataProvider.ts
machi1990: > So we went to the 1.5 - WE have separate delta table with duplicate of the data. (Not really delta table itself.)?

This should be more of a changelog + auditing table. This article might be of help https://ludwigstuyck.wordpress.com/2013/04/04/history-tracking/

ssd71:
Hi @craicoverflow , sorry for the late response
Some background can be found on #1381 #1445 
Sorry if there isn't much to go on, I definitely should have discussed about it before spiking :stuck_out_tongue:

ssd71:
- I have implemented a more meaningful delta table where we only keep the changes
- I am using this table for sync, as well as conflicts
- Consolidated custom conflicts(#1591) into this 

To test out with datasync template:
```graphql
""" 
@model
@datasync(deltaTable: true)
"""
type Comment {
  id: ID!
  text: String
  description: String
}
```

and in `index.ts`:
```typescript
import { createDataSyncCRUDService, DataSyncPlugin, createDataSyncMongoDbProvider, DeltaConflictEngine } from '@graphback/datasync';
...
    dataProviderCreator: createDataSyncMongoDbProvider(db, { Comment: DeltaConflictEngine }),
    serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
    plugins: [
      new DataSyncPlugin()
    ]
  });
```

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,164 @@
...
+        }
+      }, {
+        $match: filter
+      }];
+
+    return this.db.collection(this.collectionName).aggregate(syncPipeline).toArray()
+  }
+
+  public async checkForConflicts(updateData: any): Promise<boolean> {
+    if (updateData === {} || updateData === undefined) {
```
craicoverflow, line 109: Because of object references, I am not sure if `if (updateData === {})` would work.

A search of "how do check if object is empty in javascript" shows no examples using this method. Best to look into it a bit more

https://stackoverflow.com/questions/4994201/is-object-empty

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,164 @@
...
+
+export interface DiffEntry {
+  docId: ObjectId,
+  sets: { [fieldName: string]: any },
+  timestamp: Date
+}
+
+export function getDiffEntry(data: any): DiffEntry {
+  const sets = Object.assign({}, data);
+  const objectId = sets.id instanceof ObjectId ? sets.id : new ObjectId(sets.id);
```
craicoverflow, line 151: I've seen this used somewhere else too. Would be nice in a helper function.

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
```diff
@@ -2,17 +2,32 @@ import { Db } from 'mongodb';
...
  *
  * @param {Db} db - MongoDb connection
  */
-export function createDataSyncMongoDbProvider(db: Db): (...args: any[]) => GraphbackDataProvider {
+export function createDataSyncMongoDbProvider(db: Db, conflictStrategyMap?: ConflictStrategyMap): (...args: any[]) => GraphbackDataProvider {
   return (model: ModelDefinition): GraphbackDataProvider => {
-    if (isDataSyncModel(model)) {
-      return new DataSyncMongoDBDataProvider(model.graphqlType, db)
+    const annotationData = isDataSyncModel(model);
+    if (annotationData) {
```
craicoverflow, line 21: You could improve readibility if you try to avoid nested ifs (where possible):

```ts
    if (!annotationData) {
       return new MongoDBDataProvider(model.graphqlType, db)
    }
    
    let conflictStrategy: new() => ConflictEngine;
    if (conflictStrategyMap !== undefined) {
      conflictStrategy = conflictStrategyMap[model.graphqlType.name];
    }
    if (annotationData.deltaTable === true) {
      return new DeltaDBDataProvider(model.graphqlType, db, conflictStrategy);
    }
      
    return new DataSyncMongoDBDataProvider(model.graphqlType, db, conflictStrategy);
```

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
```diff
@@ -2,17 +2,32 @@ import { Db } from 'mongodb';
...
+      let conflictStrategy: new() => ConflictEngine;
+      if (conflictStrategyMap !== undefined) {
+        conflictStrategy = conflictStrategyMap[model.graphqlType.name];
+      }
+      if (annotationData.deltaTable === true) {
```
craicoverflow, line 26: `annotationData` can be boolean, so this would throw an error.

packages/graphback-datasync/src/util.ts
```diff
@@ -2,10 +2,10 @@ import { ModelDefinition, GraphbackCRUDService } from '@graphback/core';
 import { parseMetadata } from 'graphql-metadata';
 import { DataSyncCRUDService } from "./services";
 
-export function isDataSyncModel(model: ModelDefinition): boolean {
+export function isDataSyncModel(model: ModelDefinition): any {
```
craicoverflow, line 5: If this can now return the annotation data the function name should change too.

packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts
```diff
@@ -50,18 +50,20 @@ export class MongoDBDataProvider<Type = any> implements GraphbackDataProvider<Ty
 
   public async update(data: Type, context: GraphbackContext): Promise<Type> {
     const { idField } = getDatabaseArguments(this.tableMap, data);
+    const updateSets = Object.assign({}, data as any);
+    delete updateSets.id;
```
craicoverflow, line 54: `getDatabaseArguments` will return an ID field different to `id` if the `@id` annotation is different. So should this be deleting that key instead?

ssd71:
@craicoverflow Sorry for lack of proper instructions, I have updated PR description a bit, please test it out at your leisure

craicoverflow:
No problem, thanks for updating :)

Should this still be in draft btw? Or are you seeking more implementation feedback first.

ssd71:
Looking for implementation feedback actually, not very confident about it, would definitely appreciate it 😄

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -113,33 +123,38 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
-    const { idField } = getDatabaseArguments(this.tableMap, clientData);
-    const { fieldNames } = metadataMap;
+    if (this.conflictEngine !== undefined) {
 
-    if (!idField.value) {
-      throw new NoDataError(`Couldn't get document from ${this.collectionName} - missing ID field`)
-    }
+      const queryResult = await this.getServerState(clientData);
+      if (queryResult) {
+        const conflict = this.conflictEngine.checkForConflicts(queryResult, clientData)
```
wtrocki, line 130: I think while this made sense for the Postgress when we were building it for Mongo this approach makes less sense. There could be some operation happening in the middle of those two. 

Approach here could be that we can research `mongodb optimistic locking`
We should check if we can have conditional errors in mongo - error that is returned if query do not accept some condition. 

While this approach is ok.. it is definitely not production ready as consistency is not guaranteed here.

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,164 @@
...
+  }, {})
+}
+
+export interface DiffEntry {
+  docId: ObjectId,
+  sets: { [fieldName: string]: any },
+  timestamp: Date
+}
+
+export function getDiffEntry(data: any): DiffEntry {
```
wtrocki, line 149: I would love to get more explaination what you creating here. 
I see couple fields removed that I would consider essential.

wtrocki:
@ssd71 We agreed on investigation. Results of the investigations are usually presented in the form of the write-off. It is nice to have working code for it, but if we getting just code and no info about the approach or strategy you are taking here it is not possible to really asses this. 

It feels that approach here is wrong. We want to investigate proper patterns and document how we see this working. Step by step kinda approach. Maybe some quick diagram etc.
I see some changes - some of those sound good, some of those might need more explanation as they look controversial. Let's put some document or comment in issue first.

ssd71:
Results of investigation and spike documented in #1738

Issue #1648, opened 2020-07-03
GraphQL-Serve has invalid doc command


wtrocki:
https://www.npmjs.com/package/graphql-serve

gqls is mixed with gqlserve - not sure what to use.

wtrocki:
gqlserve is much better. gqls is clashing in my system with git aliases from zsh

wtrocki:
Also we renamed it without notice so we should restore to gqlserve

wtrocki:
Datasync option is not documented :)

machi1990:
> gqlserve is much better. gqls is clashing in my system with git aliases from zsh

We'll have `gqlserve serve models --port=8080`, won't that make the `serve` redundant? 

> Also we renamed it without notice so we should restore to gqlserve

Where was it renamed? For me to grab some context.

wtrocki:
if you look into history it should show all details. 
Renaming CMD is quite bad approach as people might already have it installed etc.

machi1990:
> if you look into history it should show all details.

Found this https://github.com/aerogear/graphback/pull/865/commits/23cf7be73b76e52d8350464a18cda6d5fea82588 but does not say much. 
 
> Renaming CMD is quite bad approach as people might already have it installed etc.

This can be easily fixed.

craicoverflow:
> We'll have `gqlserve serve models --port=8080`, won't that make the `serve` redundant? 

I think this is the original reason `gqls` was chosen.

wtrocki:
Hi
Anything would worki for me as long we are consistent

PR #1649, opened 2020-07-03
fix: graphqlserve has invalid doc command

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1648


wtrocki:
Amazing work! Thank you for doing it so fast!

PR #1657, opened 2020-07-06
docs(website): add database migrations documentation

machi1990:
TODO

- [x] Show an example generate table
- [x] Show an example of generated tables with advance options e.g database type, length, default value, foreign keys 
- [x] Indexing and unique columns 
- [x] Document various main options e.g ScalarMap

### DIFF
--- a/docs/databases/migrations.md
+++ b/docs/databases/migrations.md
@@ -1,7 +1,0 @@
----
-id: graphql-migrations
-title: GraphQL Migrations
-sidebar_label: GraphQL Migrations
----
-
-## TODO
--- a/docs/graphql-migrations/filter.md
+++ b/docs/graphql-migrations/filter.md
@@ -1,0 +1,7 @@
+---
+id: filter
+title: GraphQL Migrations Filters
+sidebar_label: Operation Filter
+---
+
+## TODO
--- a/docs/graphql-migrations/plugins.md
+++ b/docs/graphql-migrations/plugins.md
@@ -1,0 +1,7 @@
+---
+id: plugins
+title: GraphQL Migrations Plugins
+sidebar_label: Plugins
+---
+
+## TODO
--- a/docs/graphql-migrations/api-reference.md
+++ b/docs/graphql-migrations/api-reference.md
@@ -1,0 +1,140 @@
+---
+id: api
+title: API Reference
+sidebar_label: API Reference
+---
+
+The `migrateDB` method takes the three arguments, listed below.
+- `config`: [Knex](https://knexjs.org/#Installation-client) database configuration options.
+- `schemaText`: GraphQL schema text.
+- `options`: 
+   - `dbSchemaName` (default: `'public'`): table schema: `<schemaName>.<tableName>`.
+  - `dbTablePrefix` (default: `''`): table name prefix: `<prefix><tableName>`.
+  - `dbColumnPrefix` (default: `''`): column name prefix: `<prefix><columnName>`.
+  - [`updateComments`](#table-and-column-comments) (default: `false`): by default, `migrateDB` will not create or update comments on table and columns.
+  - [`scalarMap`](#scalar-mapping) (default: `undefined`): Custom Scalar mapping
+  - `mapListToJson` (default: `true`): Map scalar/enum lists to json column type by default.
+  - [`plugins`](plugins.md) (default: `[]`): List of [graphql-migrations plugins](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigratePlugin.ts) which describes queries that can be executed during migrations. 
+  - `debug` (default: `false`): displays debugging informations and SQL queries.
+  - `removeDirectivesFromSchema`: (default: `true`): Strips all directives from schema before processing.
+  - [`operationFilter`](./filter.md): Filter out database operations that we don't want, e.g. [prevent table deletion](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigrateOperationFilter.ts).
+
+## Table and Column Comments
+
+When `updateComments` is set to `true` it will automatically create/update table and column comments. 
+Table / column comments are type or field description parsed from the model schema. 
+The parsing logic takes care of stripping off all annotations leaving only business related comments.    
+
+The below model:
+
+```graphql
+"""
+Note business model table comment.
+@model
+"""
+type Note {
+  id: ID!
+  """
+  Note title.
+  """
+  title: String!
+}
+```
+
+Execution with the `updateComments` options set to true, as shown below
+```ts
+....
+migrateDB(dbConfig, schemaText, {
+  updateComments: true
+}).then((ops) => {
+    console.log(ops);
+});
+...
+```
+
+Will create / update the `note` table comments as shown below: 
+
+```sql
+\d+
+                                       List of relations
+ Schema |    Name     |   Type   |   Owner    |    Size    |            Description             
+--------+-------------+----------+------------+------------+------------------------------------
+ public | note        | table    | postgresql | 8192 bytes | Note business model table comment.
+ public | note_id_seq | sequence | postgresql | 8192 bytes | 
+(2 rows)
+```
+
+```sql
+\d+ note
+                                                     Table "public.note"
+ Column |          Type          |                     Modifiers                     | Storage  | Stats target | Description 
+--------+------------------------+---------------------------------------------------+----------+--------------+-------------
+ title  | character varying(255) | not null                                          | extended |              | Note title.
+ id     | integer                | not null default nextval('note_id_seq'::regclass) | plain    |              | 
+Indexes:
+    "note_pkey" PRIMARY KEY, btree (id)
+```
+ 
+> NOTE: This options may not be supported by some database. 
+
+## Scalar Mapping
+
+This option is a function used to specify custom Scalar mapping.
+It takes three arguments `(field: GraphQLField, fieldType: GraphQLScalarType, dbAnnotation: any)`, depending on the field type you could return the `TableColumnTypeDescriptor`: 
+
+```ts
+interface TableColumnTypeDescriptor {
+  /**
+   * Knex column builder function name.
+   */
+  type: TableColumnType | string
+  /**
+   * Builder function arguments.
+   */
+  args: any[]
+}
+```
+
+For example, if you want a custom `CustomDate` scalar to be have a `date` type in the database:
+
+```ts
+import { migrateDB } from 'graphql-migrations';
+
+const schemaText = ```
+""" @model """
+type Note {
+  id: ID!
+  title: String!
+  createdAt: CustomDate
+}
+
+scalar CustomDate
+```;
+
+const dbConfig = {
+   // Knex.js db configuration
+};
+
+migrateDB(dbConfig, schemaText, {
+  scalarMap: (
+  field: GraphQLField<any, any>,
+  scalarType: GraphQLScalarType,
+  annotations: any,
+): TableColumnTypeDescriptor | null => {
+
+  // custom date converted to date type
+  if (scalarType.name === 'CustomDate') {
+    return {
+      type: 'date',
+      args: []
+    }
+  }
+
+  return undefined
+} 
+}).then((ops) => {
+    console.log(ops);
+});
+```
+
+A good example of this function is the [default scalar type to database column](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/abstract/getColumnTypeFromScalar.ts#L18-L129) function, that Graphback uses to convert scalar types based on certain creteria.
--- a/docs/graphql-migrations/db-design.md
+++ b/docs/graphql-migrations/db-design.md
@@ -1,0 +1,464 @@
+---
+id: db-design
+title: Database Design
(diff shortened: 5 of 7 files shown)


docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
 sidebar_label: GraphQL Migrations
 ---
 
-## TODO
\ No newline at end of file
+Graphback uses [graphql-migrations](https://www.npmjs.com/package/graphql-migrations) to automatically create and update tables from a GraphQL schema.
+The package is able to calculate diff between your schema and your database to only executes required changes and thus keeping your database and schema in sync.
```
craicoverflow, line 8:
```suggestion
The library compares your database schema to your GraphQL schema and executes the required changes to keep the database structure synchronised with the GraphQL schema.
```

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+
+## Compatibility
+
+- PostgreSQL (create and update database)
+- SQLite (create database only)
+
+## Usage
+
+Graphback migrations operates on business models defined in your schema: These are graphql types decorated with a [`@model`](../model/datamodel#Model) annotation.
```
craicoverflow, line 17:
```suggestion
GraphQL Migrations operates on business models defined in your schema: These are GraphQL types decorated with a [`@model`](../model/datamodel#Model) annotation. 
```

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+
+## Compatibility
+
+- PostgreSQL (create and update database)
+- SQLite (create database only)
+
+## Usage
+
+Graphback migrations operates on business models defined in your schema: These are graphql types decorated with a [`@model`](../model/datamodel#Model) annotation. 
+The package exports an entry point which you can programatically setup in your source code and have it peform the migrations.
```
craicoverflow, line 18:
```suggestion
The package expose an API which you can programmatically set up in your source code and have it perform the migrations. 
```

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+## Usage
+
+Graphback migrations operates on business models defined in your schema: These are graphql types decorated with a [`@model`](../model/datamodel#Model) annotation. 
+The package exports an entry point which you can programatically setup in your source code and have it peform the migrations. 
+
+### API
+
+The `migrateDB` method creates and updates your tables and columns to match your GraphQL schema.
+All the database operations are wrapped in a single transaction, so your database will be fully rolled back to its initial state if an error occurs.
+The method takes three arguments as descripted in [migrations options](./migrations#Options) section.
```
craicoverflow, line 24:
```suggestion
The method takes three arguments as described in [migrations options](#Options) section.
```

wtrocki:
I went thru the docs and checked use cases. I did not had chance to verify validity of the info.

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+                                 Table "public.note"
+ Column |          Type          |                     Modifiers                     
+--------+------------------------+---------------------------------------------------
+ title  | character varying(255) | not null
+ id     | integer                | not null default nextval('note_id_seq'::regclass)
+Indexes:
+    "note_pkey" PRIMARY KEY, btree (id)
+```
+
+## Designing your data model
```
craicoverflow, line 74: I think a lot of this is duplicated from the Data Model page and can quickly go out of sync with each other. Though this does have only the migration specific information. Up for discussion what we do with this.

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+
+Sometimes our business model can contain more fields and some of these fields may only be there for representation purposes.
+We can safely ignore generating columns for these fields using the `@db(skip: true)` annotation on the corresponding field. 
+
+```graphql
+type Note {
+  id: ID!
+  """
+  Define custom column length
+  @db(skip: true)
```
craicoverflow, line 393: Though GraphQL Migrations supports this, Graphback does not recognise it (in that it will still create input fields.

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+type Comment {
+  id: ID!
+  """
+  @unique
+  """
+  text: String
+}
+```
+
+The above model will result in the following constrant being created
```
craicoverflow, line 484:
```suggestion
The above model will result in the following constraint being created
```

docs/databases/migrations.md
```diff
@@ -4,4 +4,568 @@ title: GraphQL Migrations
...
+ text        | character varying(255) | 
+ id          | integer                | not null default nextval('comment_id_seq'::regclass)
+Indexes:
+    "comment_pkey" PRIMARY KEY, btree (id)
+    "comment_text_unique" UNIQUE CONSTRAINT, btree (text)
+```
+
+### Options
+
+The `migrateDB` takes the three arguments, listed below.
```
craicoverflow, line 500:
```suggestion
`migrateDB` takes the three arguments, listed below.
```

wtrocki:
Actually after seeing comment. Should we remove things that we graphback doesn't support to not cause problems/unexpected results?

docs/databases/migrations.md
machi1990: Yeap, I was only focused on the migration part. 

I guess it is worth to add a note. WDYT?

machi1990:
> Actually after seeing comment. Should we remove things that we graphback doesn't support to not cause problems/unexpected results? Readme.md of the package can have all of that anyway.

I think adding a NOTE for under each thing that Graphback CRUD does not. Unless otherwise we are planning to remove those.

docs/databases/migrations.md
machi1990: I could not find a better to do this part. I have added links above and I think it is important to note that when modifying docs for a model / annotation etc one has to verify that the migration package is not broken and this includes the docs. The chance of the two docs being out of sync are there but very minimal considering that we do not go into details about what `@model` contains for example. Having said that, any suggestions welcomed about how we can re-restructure this part.

machi1990:
> Will do visual check on website tomorrow. Left some comments.

Thanks for the review @craicoverflow 
 
> One thing we need to watch out for is the Migration docs misleading the reader into thinking Graphback supports this.

Good point. I think we should add NOTE on each section that's not supported by Graphback CRUD. Thoughts? 
 
> There are a couple of things mentioned that Graphback does not really support (field naming, column skipping) but GraphQL Migrations.
> 
> Likewise there are many things not mentioned her supported by GraphQL Migrations that Graphback does not support, but would be useful if someone were to use the library upstream, detached from Graphback. See https://graphql-migrate.netlify.app/guide/cookbook.html

Good catch. I forgot about comments and non nullability (Anything else?). 
Will add those tomorrow.

craicoverflow:
image

We have GraphQL Migrations as a first-class section too. I was thinking to have it here since it is an upstream project that does not require Graphback, so it should have more than one page.

craicoverflow:
image

I think because GraphQL Migrations is a large library by itself it should have its own full section. The ToC/secondary navigation only goes in 2 levels so it would be very hard for users to quickly find what they want without having to scroll through everything.

docs/databases/migrations.md
craicoverflow: I think so yeah.

docs/databases/migrations.md
craicoverflow: What I meant by "going out of sync" was that because the same content needs to be duplicated in multiple places, the likelihood of one not getting updated by accident and becoming out of sync with the other is possible.

Though you are right, I think here there needs to be duplication, especially considering this library can be used without Graphback.

craicoverflow:
Have a look at [graphql-migrate](https://graphql-migrate.netlify.app) docs (which this is forked from) to see if there are things missing - for example plugins.

machi1990:
> image
> 
> We have GraphQL Migrations as a first-class section too. I was thinking to have it here since it is an upstream project that does not require Graphback, so it should have more than one page.

+1. I was working on it just now.

machi1990:
> image
> 
> I think because GraphQL Migrations is a large library by itself it should have its own full section. The ToC/secondary navigation only goes in 2 levels so it would be very hard for users to quickly find what they want without having to scroll through everything.

Good catch! Something was messed up. Here is what I've now.
Screenshot from 2020-07-07 10-15-14

machi1990:
@craicoverflow can you re-review? I have a few words about the plugins options but I think it is missing an example etc (this can be added later). Let me know what you think.

docs/databases/migrations.md
```diff
@@ -4,4 +4,661 @@ title: GraphQL Migrations
...
+  id: ID!
+  title: String!
+  """
+  @oneToMany(field: 'note')
+  """
+  comments: [Comment]
+}
+```
+
+This creates a relationship between `Note.comments` and `Comment.note` and maps to `comment.noteId` in the database. If `Comment.note` does not exist it will be generated for you.
```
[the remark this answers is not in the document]
machi1990: Very good comment.

docs/databases/migrations.md
```diff
@@ -4,4 +4,661 @@ title: GraphQL Migrations
...
+  - `plugins` (default: `[]`): List of [graphql-migrations plugins](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigratePlugin.ts) which describes queries that can be executed during migrations. 
+  - `debug` (default: `false`): displays debugging informations and SQL queries.
+  - `removeDirectivesFromSchema`: (default: `true`): Strips all directives from schema before processing.
+  - `operationFilter`: Filter out database operations that we don't want, e.g. [prevent table deletion](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigrateOperationFilter.ts).
+
+### Tabel and Column Comments
+
+The `updateComments` once set to `true` will automatically create/update table and column comments. 
+Table / column comments are type or field description parsed from the model schema. 
+The parsing logic takes care of stripping off all Graphback annotations leaving only business related comments.
```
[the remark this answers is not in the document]
machi1990: I applied the suggestion. 

Do we have an issue for that in graphql-metadata?

docs/databases/migrations.md
```diff
@@ -4,4 +4,661 @@ title: GraphQL Migrations
...
+```graphql
+type Note {
+  id: ID!
+  """
+  @db(name: 'note_title')
+  """
+  title: String!
+}
+```
+> NOTE: This annotation is not supported by Graphback CRUD
```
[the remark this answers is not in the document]
machi1990: I was thinking of caution.

docs/databases/migrations.md
```diff
@@ -4,4 +4,661 @@ title: GraphQL Migrations
...
+--------+------------------------+---------------------------------------------------+----------+--------------+-------------
+ title  | character varying(255) | not null                                          | extended |              | Note title.
+ id     | integer                | not null default nextval('note_id_seq'::regclass) | plain    |              | 
+Indexes:
+    "note_pkey" PRIMARY KEY, btree (id)
+```
+ 
+> NOTE: This options may not be supported by some database. 
+
+### Scalar Mapping
```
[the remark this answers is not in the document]
machi1990: +1. This crossed over my mind too but I had no strong opinion about it. Will do it!

machi1990:
> I think the Migration section should be split into pages:
> 
> ```
> - GraphQL Migrations
> 	- Installation - missing an installation section. Imagine using this without Graphback or if you want to install it at a later point to your own Graphback project.
> 	- **Data models** is a Graphback concept, So we should refrain from using it in migrations (GraphQL Migrations uses the generated schema anyway). Let's try to think of another page name. Usage? Configure your schema?
> 	- API Reference. Move The Options section to a new page and we can maybe call it an API Reference (to stay consistent with the lexicon of the new website)
> 	- Plugins. Can be added later.
> 	- Filters. Can be added later.
> ```

Thanks for review. Good one, I've go on with splitting the page as suggested. 

As for data models, maybe `db-design`?

craicoverflow:
> As for data models, maybe `db-design`?

Sounds good. But maybe with the full words "Database design"

docs/graphql-migrations/filter.md
```diff
@@ -0,0 +1,7 @@
+---
+id: filter
+title: GraphQL Migrations Filters
+sidebar_label: Operation Filter
+---
+
+## TODO
```
machi1990, line 7: Followup PR.

docs/graphql-migrations/plugins.md
```diff
@@ -0,0 +1,7 @@
+---
+id: plugins
+title: GraphQL Migrations Plugins
+sidebar_label: Plugins
+---
+
+## TODO
```
machi1990, line 7: To be done in followup PR. Do we have a use case in mind for this options? @craicoverflow

craicoverflow: Copy it from the forked project:

https://graphql-migrate.netlify.app/guide/plugins.html

machi1990: Okay. That could be a very good start.

PR #1664, opened 2020-07-07
build: use graphql-tools to load files

craicoverflow:
Fixes failing integration tests in [jobs/5744](https://app.circleci.com/pipelines/github/aerogear/graphback/5629/workflows/07663b62-8eff-4f4a-8c1d-778e65db871b/jobs/5744/steps)

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -20,6 +20,7 @@
     "ts-node": "8.10.2",
     "tsutils": "3.17.1",
     "typescript": "3.9.5",
+    "@graphql-tools/graphql-file-loader": "6.0.12",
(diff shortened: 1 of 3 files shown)


machi1990:
Thanks for the fix, did anything to to make the test fail?

craicoverflow:
I noticed the build is failing in https://github.com/aerogear/graphback/pull/1616. So it is blocking dependency updates.

Though I see this is failing for a different reason now, will do some more checks.

machi1990:
> I noticed the build is failing in #1616. So it is blocking dependency updates.
> 
> Though I see this is failing for a different reason now, will do some more checks.

Thanks for the quick fix!!

PR #1666, opened 2020-07-07
docs(website): add how to use plugins in graphql migrations

machi1990:
- [x] Test the example that it actually works.


machi1990:
Merged, thanks @craicoverflow for the quick review.

PR #1669, opened 2020-07-07
fix: ensure that updatedAt is returned as string value

wtrocki:
In db we store it as number and we return invalid value to client.


packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -134,6 +134,8 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
         queryResult[fieldNames.updatedAt] !== undefined &&
         clientData[fieldNames.updatedAt].toString() !== queryResult[fieldNames.updatedAt].toString()
```
wtrocki, line 135: Client side will always will be string as it comes thru scalar.

wtrocki:
Verified in datasync-starter

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
machi1990: Should the fix be done here https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/utils/fieldTransformHelpers.ts?

Issue #1656, opened 2020-07-06
Graphback overwrites custom query with same name


craicoverflow:
Given this model:

```gql
type Query {
  getNote(id: ID!, test: String): Note
}
```

Graphback will ignore it and still use the generated query:

```gql
type Query {
  getNote(id: ID!): Note
}
```

machi1990:
Discussion we had in https://github.com/aerogear/graphback/issues/1597 could be useful here too.

craicoverflow:
See #1670 

The problem is that `buildModel` is called multiple times, so the approach taken here will not work.

The first time `buildModel` is called it is using the original data model schema, which does not have any generated CRUD API yet. At this point it is easy to identify custom queries that have a clashing name. This PR uses this information to update the CRUD options on the ModelDefinition. The main benefit of this is there is a single source of truth for the model's CRUD config.

The next time `buildModel` is called, it is using the schema which has the Graphback queries generated. It sees this as custom user queries and updates the CRUD config accordingly, so no resolvers are created.

The obvious quick solution is to not update the model CRUD config from the custom queries, instead calculating this when needed. The issue I have with this is there are now two truths - The model CRUD config will say that the `create` CRUD is enabled (as an example), but Graphback does not generate this because the user supplied their own custom create query in the schema.

machi1990:
> See #1670
> 
> The problem is that `buildModel` is called multiple times, so the approach taken here will not work.
> 
> The first time `buildModel` is called it is using the original data model schema, which does not have any generated CRUD API yet. At this point it is easy to identify custom queries that have a clashing name. This PR uses this information to update the CRUD options on the ModelDefinition. The main benefit of this is there is a single source of truth for the model's CRUD config.
> 
> The next time `buildModel` is called, it is using the schema which has the Graphback queries generated. It sees this as custom user queries and updates the CRUD config accordingly, so no resolvers are created.

Instead of updating the CRUD config, I was thinking maybe we can throw an error if the a model's crud config is activated for a certain query/mutation and the user provides a custom one? The error message could be indicative proposing either changing the name of the custom query/mutation or suggesting deactivating the corresponding crud operation on the model config.

craicoverflow:
That sounds like a better approach. The current approach silently disables generation giving no explanation to why the CRUD query is not working for them.

This PR is re-implementing the same behaviour that existed in an early version of Graphback (pre-plugins). That is not to say it should be the case.

[...]

PR #1670, opened 2020-07-07
[WIP] fix(schema): prevent overwriting user queries

craicoverflow:
Fix for https://github.com/aerogear/graphback/issues/1656

**This PR is incomplete, requires discussion**

**Description**

The problem is that `buildModel` is called multiple times, so the approach taken here will not work.

The first time `buildModel` is called it is using the original data model schema, which does not have any generated CRUD API yet. At this point it is easy to identify custom queries that have a clashing name. This PR uses this information to update the CRUD options on the ModelDefinition. The main benefit of this is there is a single source of truth for the model's CRUD config.

The next time `buildModel` is called, it is using the schema which has the Graphback queries generated. It sees this as custom user queries and updates the CRUD config accordingly, so no resolvers are created.

The obvious quick solution is to not update the model CRUD config from the custom queries, instead calculating this when needed. The issue I have with this is there are now two truths - The model CRUD config will say that the `create` CRUD is enabled (as an example), but Graphback does not generate this because the user supplied their own custom create query in the schema.

### DIFF
--- a/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
+++ b/packages/graphback-core/src/plugin/GraphbackCoreMetadata.ts
@@ -5,19 +5,20 @@ import { IResolvers } from '@graphql-tools/utils'
 import { mergeResolvers } from '@graphql-tools/merge';
 import { RelationshipMetadataBuilder, FieldRelationshipMetadata } from '../relationships/RelationshipMetadataBuilder'
 import { getPrimaryKey } from '../db';
+import { GraphbackOperationType, getFieldName, getSubscriptionName } from '../crud';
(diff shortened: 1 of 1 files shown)


wtrocki:
I see overriding user queries as feature. If you add model you will be expect some methods to be created. If you create method with the same name (which is unusual) you should probably get validation error?

wtrocki:
Could we document this instead for 0.14.0?

craicoverflow:
Yeah, there is a good discussion over in https://github.com/aerogear/graphback/issues/1656 about this.

This PR is trying to restore the original feature, whereby Graphback did not overwrite custom queries with the same name.

craicoverflow:
Closing as this is documentation issue resolved in #1685

PR #1674, opened 2020-07-08
docs(website): add operations filter in graphql migrations docs

machi1990:


docs/graphql-migrations/operation-filters.md
```diff
@@ -0,0 +1,42 @@
...
+....
+
+migrateDB(dbConfig, schema, {
+  operationFilter: removeDeletionOperations
+})
+
+....
+```
+
+GraphQL Migrations has you covered and provides a [removeNonSafeOperationsFilter](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigrateOperationFilter.ts) to avoid doing non safe operations such as table deletion and renaming.
```
craicoverflow, line 42:
```suggestion
GraphQL Migrations has you covered and provides a [`removeNonSafeOperationsFilter`](https://github.com/aerogear/graphback/blob/master/packages/graphql-migrations/src/plugin/MigrateOperationFilter.ts) to avoid doing non safe operations such as table deletion and renaming. 
```

machi1990: Applied thanks!

PR #1693, opened 2020-07-09
fix(cli): restore generate command

craicoverflow:
The commands directory was removed in a previous commit, which means the CLI has not been working for a while.

```bash
Error: ENOENT: no such file or directory, scandir '/code/aerogear/graphback/packages/graphback-cli/dist/commands'
```

This restores the `generate` command, which is required for generation of client documents.

## Verification

1. Add a `.graphqlrc` file to your project.

```yaml
schema: './src/schema.graphql'
documents: './client/src/graphql/**/*.graphql'
extensions:
  graphback:
    model: ./model/datamodel.graphql
    plugins:
      graphback-client:
        # format: 'graphql'
        outputFile: ./client/src/graphql/graphback.graphql
```

2. Run `yarn graphback generate`. 
3. The command should run successfully.

## Tasks

- [x] Fix issue
- [x] Add integration test

## Notes

I tried to run the CLI through yarn in the integration tests. This worked fine locally but I had issues with CircleCI, so I changed it to call it with Node.js directly: https://github.com/aerogear/graphback/pull/1693/commits/349272c867f6241808cdfac6e5bf9d4e37eeb38a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
           command: yarn test
       - run:
           name: run integration tests
-          command: cd integration && yarn && yarn integration
+          command: yarn test:integration
   npm_publish:
     working_directory: ~/aerogear/graphback
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
           command: yarn test
       - run:
           name: run integration tests
-          command: yarn test:integration
+          command: cd integration && yarn && yarn integration
   npm_publish:
     working_directory: ~/aerogear/graphback
     docker:
--- a/integration/tests/cli-workflow.ts
+++ b/integration/tests/cli-workflow.ts
@@ -1,0 +1,26 @@
+import { mkdirSync, readFileSync, rmdirSync } from 'fs';
+import * as util from 'util';
+
+beforeAll(() => {
+  mkdirSync('./output-cli/client', { recursive: true })
+})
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           command: yarn test
       - run:
           name: run integration tests
-          command: yarn test:integration
+          command: cd integration && yarn && yarn integration
```
machi1990, line 45: Is this change related? 

Seems to be the same as the [yarn test:integration](https://github.com/aerogear/graphback/blob/master/package.json#L35) script.

integration/tests/cli-workflow.ts
```diff
@@ -0,0 +1,29 @@
...
+
+beforeAll(() => {
+  mkdirSync('./output-cli/client', { recursive: true })
+})
+
+afterAll(() => {
+  rmdirSync('./output-cli', { recursive: true })
+})
+
+describe('CLI tests', () => {
```
machi1990, line 13: +1000 for these tests! Thanks.

.circleci/config.yml
craicoverflow: actually that should have been reverted in 349272c - thanks.

craicoverflow:
> Thanks for catching this and for adding related IT tests.
> 
> Maybe we should squash when merging?

Yep I usually squash because I like to create lots of commits. Thanks for the review.

PR #1698, opened 2020-07-10
docs(website): add postgres database documentation

machi1990:


docs/databases/postgres.md
```diff
@@ -4,4 +4,118 @@ title: Using PostgreSQL with Graphback
...
+Additionally, Graphback comes with the [GraphQL Migrations](../graphql-migrations/intro.md) package that can be used to perform migrations against your database using the business models defined in the GraphQL Schema. 
+
+The provider is built on top of [Knex.js](http://knexjs.org/), a flexible SQL query builder.
+
+# Installation
+
+Install with npm:
+
+```bash
+npm install @graphback/runtime-knex knex
```
wtrocki, line 18: https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-knex/package.json#L41

Knex is there, but I know @craicoverflow asked to remove it.

docs/databases/postgres.md
```diff
@@ -4,4 +4,118 @@ title: Using PostgreSQL with Graphback
...
+// create the postgres data provider
+const dataProviderCreator = createKnexDbProvider(db);
+
+// highlight-end
+
+// Use the dataProvider in buildGraphbackAPI
+const { resolvers, services, contextCreator } = buildGraphbackAPI(userModel, { dataProviderCreator });
+```
+
+The highlighted code does the following:
```
wtrocki, line 112: WOW. That is very nice trick :D

wtrocki:
Grammar/Word: Leaving for experts
Rendering: 10/10
Content: 10/10

docs/databases/postgres.md
```diff
@@ -4,4 +4,118 @@ title: Using PostgreSQL with Graphback
 sidebar_label: PostgreSQL
 ---
 
-## TODO
\ No newline at end of file
+[PostgreSQL](https://www.postgresql.org/) is a powerful, open source object-relational database system with over 30 years of active development that has earned it a strong reputation for reliability, feature robustness, and performance. [Graphback Knex Provider](https://www.npmjs.com/package/@graphback/runtime-knex) package provides instant integration to your PostgreSQL database, giving you a full implementation of the [CRUD API](../crud/introduction.md). 
+
+Additionally, Graphback comes with the [GraphQL Migrations](../graphql-migrations/intro.md) package that can be used to perform migrations against your database using the business models defined in the GraphQL Schema.
```
craicoverflow, line 9: Technically, they are separate packages that work well together. But it need to be installed separately as `graphql-migrations` to be used, unless you are using a template.

docs/databases/postgres.md
```diff
@@ -4,4 +4,118 @@ title: Using PostgreSQL with Graphback
...
+npm install @graphback/runtime-knex knex
+```
+
+Install with yarn:
+
+```bash
+yarn add @graphback/runtime-knex knex
+```
+
+Running either of the two commands will install the `Graphback Knex Provider` ([`@graphback/runtime-knex`](https://www.npmjs.com/package/@graphback/runtime-knex)) and the Knex.js library ([`knex`](https://www.npmjs.com/package/knex)).
```
craicoverflow, line 27: Link to Knex.js website instead?

docs/databases/postgres.md
```diff
@@ -4,4 +4,118 @@ title: Using PostgreSQL with Graphback
...
+      host: "localhost",
+      port: 5432
+    },
+    pool: { min: 5, max: 30 }
+};
+
+const db = Knex(dbConfig);
+```
+
+The above code initiates a connection to a local running PostgreSQL database, using the `postgresql:password` credentials. We also create a database connection pool to a minimun of `5` and maximum of `30`. You can modify the configurations depending on your setup.
```
craicoverflow, line 55:
```suggestion
The above code initiates a connection to a local running PostgreSQL database, using the `postgresql:password` credentials. We also create a database connection pool to a minimum of `5` and maximum of `30`. You can modify the configurations depending on your setup. 
```

docs/databases/mongodb.md
```diff
@@ -25,7 +25,7 @@ yarn add @graphback/runtime-mongo mongodb
 
 Running either of the two commands will install the `Graphback MongoDB Provider` ([`@graphback/runtime-mongo`](https://www.npmjs.com/package/@graphback/runtime-mongo)) and the MongoDB Driver ([`mongodb`](https://www.npmjs.com/package/mongodb)).
 
-# Creating the Database Connection
```
craicoverflow, line 28: Hmm I have been reserving H1 titles "#" for the main page title and always start with "##".

I just checked and all the titles with # are the same size as the main page title.

I recommend starting with ##

machi1990: Good catch.

I'll change them throughout.

docs/databases/postgres.md
machi1990: We have the link here https://github.com/aerogear/graphback/pull/1698/files#diff-dd237a6c711114adc64df9e89cb0f528R11 since here we are talking about the package thought we could have the npm link insted

docs/databases/postgres.md
machi1990: I see what you mean. How would you re-phrase this to avoid any confusion?

docs/databases/mongodb.md
machi1990: Done for this PR and also other changes https://github.com/aerogear/graphback/pull/1700 to some of the docs which I did earlier.

docs/databases/postgres.md
craicoverflow: How about:

Additionally you can use GraphQL Migrations to perform database migrations using the models defined in the GraphQL schema created by Graphback.

machi1990: Sold!

craicoverflow:
One thing that is missing from the databases docs so far is an API reference for the providers themselves like `KnexDBDataProvider`. We can do this later.

machi1990:
> One thing that is missing from the databases docs so far is an API reference for the providers themselves like `KnexDBDataProvider`. We can do this later.

Good point. Would it be possible to use something like https://typedoc.org and have an API section linking to the generated docs. I see we have the package here https://github.com/aerogear/graphback/blob/master/package.json#L30

craicoverflow:
> > One thing that is missing from the databases docs so far is an API reference for the providers themselves like `KnexDBDataProvider`. We can do this later.
> 
> Good point. Would it be possible to use something like [typedoc.org](https://typedoc.org) and have an API section linking to the generated docs. I see we have the package here [/package.json@`master`#L30](https://github.com/aerogear/graphback/blob/master/package.json?rgh-link-date=2020-07-10T12%3A48%3A18Z#L30)

That would be a good thing to make use of, rather than maintaing docs in two places.

machi1990:
> > > One thing that is missing from the databases docs so far is an API reference for the providers themselves like `KnexDBDataProvider`. We can do this later.
> > 
> > 
> > Good point. Would it be possible to use something like [typedoc.org](https://typedoc.org) and have an API section linking to the generated docs. I see we have the package here [/package.json@`master`#L30](https://github.com/aerogear/graphback/blob/master/package.json?rgh-link-date=2020-07-10T12%3A48%3A18Z#L30)
> 
> That would be a good thing to make use of, rather than maintaing docs in two places.

Created https://github.com/aerogear/graphback/issues/1702 to keep track.

PR #1706, opened 2020-07-10
docs(plugins): add creating your own plugin guide

craicoverflow:


craicoverflow:
cc @machi1990 @wtrocki I did not get this fully finished. As I am away next week I have pushed it for you guys to take over.

The example plugin actually works, what is left is mainly to show an example query

```ts
import { GraphbackPlugin, GraphbackCoreMetadata, GraphbackContext, QueryFilter } from 'graphback';
import { GraphQLSchema, GraphQLList, GraphQLNonNull, GraphQLResolveInfo } from 'graphql';
import { SchemaComposer } from 'graphql-compose';
import { IResolvers, IObjectTypeResolver } from '@graphql-tools/utils';

export class MyGraphbackPlugin extends GraphbackPlugin {
  transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
    const schema = metadata.getSchema()
    const graphbackModels = metadata.getModelDefinitions()

    for (const model of graphbackModels) {
      schemaComposer.Query.addFields({
        [`getArchived${model.graphqlType.name}s`]: {
          type: GraphQLNonNull(GraphQLList(model.graphqlType))
        }
      })
    }

    return schemaComposer.buildSchema()
  }

  createResolvers(metadata: GraphbackCoreMetadata): IResolvers {
    const resolvers: IResolvers = {}
    const queryObj: IObjectTypeResolver = {}

    for (const model of metadata.getModelDefinitions()) {
      const modelName = model.graphqlType.name

      queryObj[`getArchived${modelName}s`] = async (_: any, args: any, context: GraphbackContext, info: GraphQLResolveInfo) => {
        const crudService = context.graphback.services[modelName]

        const filter: QueryFilter = {
          archived: {
            eq: true
          }
        }

        const results = await crudService.findBy(filter, context)

        return results.items
      }
    }

    resolvers.Query = queryObj

    return resolvers
  }

  createResources() {

  }

  getPluginName() {
    return 'MyGraphbackPlugin';
  }
}

```

machi1990:
@craicoverflow sure, I'll take over and drive this baby home ;-)

machi1990:
Going through the [GraphbackPlugin#createResources](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/GraphbackPlugin.ts#L52) and I feel like this method can be non-abstract, in a sense that a developer does not have to implement it. As maybe, they will be looking at [adding resolvers](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/GraphbackPlugin.ts#L34) or [transforming the schema](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/GraphbackPlugin.ts#L24)

/cc @wtrocki @craicoverflow WDYT?

wtrocki:
Makes sense now

Issue #1624, opened 2020-07-01
Move `init` command to new micro-CLI `create-graphback`


craicoverflow:
`graphback-cli` requires a peer dependency of `graphql` to execute the generate command, which is used to generate GraphQL resources such as queries, and schema.

The CLI also executes `init`, which would typically be executed outside of a folder with any grapHQL dependency. This causes #1530 to happen.

These two commands are incompatible to have in the same CLI package. It makes sense to move the `init` command to a different package which will not require `graphql` to be installed.

Let's move the `init` command/code to [aerogear/create-graphql](https://github.com/aerogear/create-graphql)

craicoverflow:
This will fix #1530

craicoverflow:
`create-graphql` does a lot of cool file generation with Yeoman. It would be a lot of work to retrofit this to use Graphback.

What about extracting the init command code to a package `create-graphback-app`?

Then you can do: 

```bash
npm init create-graphback-app my-cool-project
```

Or

```bash
npx create-graphback-app my-cool-project
```

Then work could be done on create-graphql post v1 release to make it compatible with Graphback while staying true to its original file generation.

wtrocki:
I have no strong opinion, but this is going to create yet another CLI. 
Less then better IMHO. Can we just add graphback to existing stuff?

craicoverflow:
Are we prepared to also support what currently exists in create-graphql?
Or is the plan to remove it all and just have Graphback options like from Graphback CLI?

I know create-graphql does not get much current usage, so maybe it is fine to scrap it all.

wtrocki:
> Or is the plan to remove it all and just have Graphback options like from Graphback CLI?

We give the same features by different way. Cli also generates resolves.. 
I would simply copy paste our current CLI and remove non init stuff there. 
That is all.

craicoverflow:
> We give the same features by different way. CLI also generates resolves..

The Graphback CLI does not currently generate resolvers.

create-graphql does not generate a schema file, it generates several JavaScript files that export `GraphQLObjectType` with mutation and queries as fields on the object (not separate like the format from graphql-tools).

Check its commands for more: https://github.com/aerogear/create-graphql/blob/master/docs/Commands.md

Also, I could not find anything to suggest that it currently generates TypeScript files - only JS.

wtrocki:
Agrement - No create-graphql for 0.14.0. We can have create-graphback for the moment

PR #1717, opened 2020-07-13
feat(create-graphback): add a create-graphback package

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1624


packages/create-graphback/src/index.ts
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env node
+
+import * as yargs from 'yargs';
```
wtrocki, line 3: I think knowing create has only single command. It makes no sense to have any cmd support here. but it might have sense in future if we want to extend it.

wtrocki:
Whilke this works we need to clean up original command from init and add release notes.

packages/create-graphback/src/index.ts
machi1990: I agree with you. And because of this I needed to do this https://github.com/aerogear/graphback/pull/1717/files#diff-2358e1b515242085670a028b71d9fa9cR6 to have a default command. 

let me clean this one up.

machi1990:
> Whilke this works we need to clean up original command from init and add release notes.

@wtrocki should be good now. I've cleaned up some left overs too from the `graphback-cli` package.

wtrocki:
yes.. All good now. Can we do proper beta release after merge.

Issue #1728, opened 2020-07-15
selectedField Issue with queries generated by codegen


machi1990:
codegen generated queries are causing issues like one below since the queries contains the `__typename` field which is not a model specific field. This field will be retrieved by our [getSelectedFieldsFromResolverInfo](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/getSelectedFieldsFromResolverInfo.ts)
logic and sent to the db.  

```
GraphQL error: insert into "note" ("description", "title") values ($1, $2) returning "id", "title", "description", "__typename" - column "__typename" does not exist
```

Do you think it is safe to ignore this field? 

/cc @wtrocki @craicoverflow

machi1990:
Opened a quick fix #1729 but I think the most reliable way is having the model fields information in [`ModelDefinition`](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/ModelDefinition.ts#L8). This will also include foreign keys / relationships join key.  Afterwards we can safely ignore any non model field.

Issue #1675, opened 2020-07-08
parse metadata and store them in `ModelDefinition` as much as possible


machi1990:
The idea has been raised a couple of times:
- In https://github.com/aerogear/graphback/pull/1614#discussion_r448159331
- And https://github.com/aerogear/graphback/pull/1647#discussion_r449454898

Opening it here to keep track and to drive potential discussion.

craicoverflow:
This will work really well in the service and provider creator abstractions. But it would make it extremely difficult for a user to make a one-off service that extends CRUDService outside of the service creator.

So the trade-off will be that there will be only one way to create services and providers, which I am fine with.

machi1990:
Additionally, we should parse model's field names and store them in an array [`fields`]. One particular usecase I am thinking of is described here https://github.com/aerogear/graphback/pull/1729#discussion_r455017887

craicoverflow:
Some thoughts from a discussion between myself and @machi1990 

in best practices the data providers should not have any reference to a GraphQLObjectTypes at all. By abstracting ModelDefinition completely away from GraphQLObjectType the providers could be used outside of a GraphQL application - which would be awesome.

However, the `ModelDefinition` does still need a reference to the GraphQLObjectType for use elsewhere, such as in GraphbackCoreMetadata when building a plugin.

Perhaps these interfaces should be separate, where the Data Provider accepts a different interface from ModelDefinition containing only data valuable to the dataprovider.

I am a little bit against having an abstraction of GraphQLObjectType that can be used in both places (meaning it would no longer have any reference to GraphQLObjectType) because in older versions of Graphback we had this, called `InputModelTypeContext` and it was difficult to work with. See https://github.com/aerogear/graphback/blob/0.10.1/packages/graphback-core/src/api/ContextTypes.ts#L79-L86

wtrocki:
Amazing writeoff. Please no `InputModelTypeContext` anymore :D

wtrocki:
Challenge is that we do not know what metadata people might add. And most of the people still will be using directives (sadly) so we can collect as much metadata as much we have inside core. (Stuff from datasync cannot be there as will polute core with concepts that are separate).

craicoverflow:
> Challenge is that we do not know what metadata people might add. 

Graphback could parse all metadata and add it the the model definition. It would not be typed but the user could then work with it in their providers.

```ts
export type ModelDefinition = {
  primaryKey: string
  annotations: any
}
```

[...]

PR #1729, opened 2020-07-15
fix: ignore __typename when retrieving user selected fields

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1728


wtrocki:
Unit test please

machi1990:
> Unit test please

Could not find an easy way to add them without mocking, chose to rely on the integrations tests we've.

wtrocki:
All good then

packages/graphback-core/src/plugin/getSelectedFieldsFromResolverInfo.ts
```diff
@@ -16,23 +22,28 @@ export const getSelectedFieldsFromResolverInfo = (info: GraphQLResolveInfo, mode
...
           }
         } else {
-          acc.add(entry[0])
+          selectedField = entry[0];
         }
       }
 
+      if (selectedField && !INTERNAL_NON_USER_FIELDS.includes(selectedField)) {
+        acc.add(selectedField);
+      }
```
machi1990, line 45: I think the most reliable way is having the model fields information in [`ModelDefinition`](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/plugin/ModelDefinition.ts#L8). This will also include foreign keys / relationships join key.  Afterwards we can safely ignore any non model field. 

I'll add this info in https://github.com/aerogear/graphback/issues/1675 so that we parse the fields info there too.

Issue #1023, opened 2020-04-01
[Model] Ability to exclude fields from generated input types


craicoverflow:
There should be additional annotations to exclude certain fields from being added to input types in the final schema.

```gql
""@model"""
type User {
  id: ID!
  name: String
  """
  @fields.ignore
  """
  password: String
  """
  @input.ignore
  """
  fieldIDontWantToEdit: Int
}
```

would generate the following input types:

```gql
input UserFields {
  id: ID
  name: String
  fieldIDontWantToEdit: Int
}

input UserInput {
  id: ID
  name: String
  password: String
}
```

machi1990:
With the `@ignore` annotation, we'd ignore the field:
- From generated mutation inputs 
- From generated filter inputs
- From query selection set
- From database migration
  
Anything missing? @craicoverflow 

Also cc @wtrocki for diversity of ideas.

craicoverflow:
> From generated mutation inputs

It should be included in the generated mutation inputs. Think of a `password` field - you want to create and update this, but do not want to necessarily be able to retrieve or filter on it.

> From database migration

It should still be created in the database - we have `@db(skip: true)` to ignore it in migrations and these can be combined to that effect.

Think of a `password` field as an example. You want it to be created in the database but do not want it to be selectable by the user at any stage.

We may need a combination of annotations - I think we should check out how other libraries do this, as it is probably quite a common use case in Entity Framework, Hibernate, JPA etc

machi1990:
> > From generated mutation inputs
> 
> It should be included in the generated mutation inputs. Think of a `password` field - you want to create and update this, but do not want to necessarily be able to retrieve or filter on it.

Makes sense for the password field. But generally, If field is not excluded from the input, it will be part of the payload that's going to be stored in db. This will likely crash as the column might be missing from database.   
 
> > From database migration
> 
> It should still be created in the database - we have `@db(skip: true)` to ignore it in migrations and these can be combined to that effect.
> Think of a `password` field as an example. You want it to be created in the database but do not want it to be selectable by the user at any stage.

Instead of having several annotations that's does the same thing, I'd very much see us offering one annotation. Possibly we can make it take some arguments. 

> We may need a combination of annotations - I think we should check out how other libraries do this, as it is probably quite a common use case in Entity Framework, Hibernate, JPA etc

I'd be much in favour of having one annotation.

craicoverflow:
Something like this?

```gql
type User {
[...]

PR #1730, opened 2020-07-15
parse fields info into ModelDefinition

machi1990:
Following
https://github.com/aerogear/graphback/issues/1675#issuecomment-658743262
and https://github.com/aerogear/graphback/issues/1728#issuecomment-658741379

Also use this information to improve the selectedFields retrieval logic as suggested in
https://github.com/aerogear/graphback/pull/1729#discussion_r455017887

### DIFF
--- a/packages/graphback-core/src/plugin/ModelDefinition.ts
+++ b/packages/graphback-core/src/plugin/ModelDefinition.ts
@@ -1,20 +1,25 @@
 import { GraphQLObjectType } from "graphql";
 import { FieldRelationshipMetadata } from '../relationships/RelationshipMetadataBuilder';
-import { GraphbackCRUDGeneratorConfig } from "./GraphbackCRUDGeneratorConfig"
+import { GraphbackCRUDGeneratorConfig } from "./GraphbackCRUDGeneratorConfig";
 
 /**
  * Describe the name and type of primary key
  */
-export type PrimaryKeyDescriptor = {
+export type FieldDescriptor = {
   name: string,
   type: string
+}
+
+export type ModelFieldMap = {
+  [key : string]: FieldDescriptor
 }
 
 /**
  * Used to encapsulate configuration for the type
  */
 export type ModelDefinition = {
-  primaryKey: PrimaryKeyDescriptor,
+  primaryKey: FieldDescriptor,
+  fields: ModelFieldMap
   graphqlType: GraphQLObjectType,
   relationships: FieldRelationshipMetadata[]
   crudOptions: GraphbackCRUDGeneratorConfig,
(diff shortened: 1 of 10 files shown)


machi1990:
Improves on https://github.com/aerogear/graphback/pull/1729

packages/graphback-core/src/plugin/ModelDefinition.ts
```diff
@@ -1,12 +1,20 @@
-import { GraphQLObjectType, GraphQLField } from "graphql";
+import { GraphQLObjectType } from "graphql";
 import { FieldRelationshipMetadata } from '../relationships/RelationshipMetadataBuilder';
 import { GraphbackCRUDGeneratorConfig } from "./GraphbackCRUDGeneratorConfig"
 
+/**
+ * Maps type fields to corresponding field in the database
+ */
+export type ModelField = {
```
craicoverflow, line 8:
```suggestion
export type ModelFieldMap = {
```

machi1990: Good suggestion. 

I am still thinking about this and the field information we'd want to store in this map.

My recent thoughts pushed me to think that we'll need more than field column name in the database, additional information like column type might be needed so that an appropriate query in the database is performed. I am thinking of `ObjectID` for MongoDB filtering and `JSON` types for Postgres. 

A naive map could look lik: 

```ts
/**
Field descriptor in the database
**/
export type FieldDescriptor = {
    name: string, // could come from GraphQLField.name or later on from a @column(name: 'db_name') annotation or the @db annotation
    type: string // could come from GraphQLField.type or later on from @column(type: 'db_type') or the @db annotation
}

/**
 * Maps type fields to corresponding field in the database
 */
export type ModelField = {
  [key : string]: FieldDescriptor
}
```
It needs more thinking.

machi1990:
Depends on https://github.com/aerogear/graphback/pull/1745

packages/graphback-core/src/plugin/ModelDefinition.ts
machi1990: I went with this thought. Let me know what you think.

wtrocki:
Rebase needed. It will be best to try to land this changes into #936

machi1990:
@craicoverflow will you have some time to check this out? The idea will be to get this one merged and then add on some of the stuffs we discussed here https://github.com/aerogear/graphback/issues/1023 to ignore fields that not in the database.

craicoverflow:
Sure I will check this shortly.

machi1990:
Ah forgot about this completely. Let me rebase it so that we can get this in.

Issue #1742, opened 2020-07-22
Three phased plugin engine to "createSchema" , "createResolvers" and "createResources"


machi1990:
Current the PluginEngine creates resolvers once plugin's transformation phase is done. 

See: https://github.com/aerogear/graphback/blob/8cefc6c085cdc6fe8ab40a7b0275642d4caf9c9b/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts#L65-L71

Opening this issue as a question / discussion to see if this can be changed into two phases:

1. Run the transformation of all the plugins. This is the current the `GraphbackPluginEngine#createSchema()` with the resolver part extracted into a second phase.
2. Run the resolvers creation after the transfomation phase is done. This will be  `GraphbackPluginEngine#createResolvers()` 

The advantage of this method is that we'll be sure that the model object https://github.com/aerogear/graphback/issues/1675 will have all the information: 
- those declared statically in the business models graphql schema e.g `model.graphql` and 
- those added dynamically will be there. 

We'll add a second iteration in the plugin engine (a very minimal perf cost I'd) which is okay since we'll be gaining correctness.

The two phases will be followed by the "createResources" phase, the final phase where schema and other files are saved in hard disk. 
/cc @craicoverflow @wtrocki WDYT?

wtrocki:
+1 from me but I would rely on @craicoverflow opinion more as he was designing this feature.

craicoverflow:
So SchemaCRUDPlugin might add a `getNote` query based on metadata, but PluginB might remove `getNote` based on some other custom condition or metadata. Currently the `getNote` resolver gets created anyway because it uses an earlier snapshot of the schema in the transformation cycle. But in reality it should not be there. Whereas with this approach suggested we wait till the final schema and metadata exists before examining what resolvers are needed.

+1 from me!

machi1990:
> So SchemaCRUDPlugin might add a `getNote` query based on metadata, but PluginB might remove `getNote` based on some other custom condition or metadata. Currently the `getNote` resolver gets created anyway because it uses an earlier snapshot of the schema in the transformation cycle. But in reality it should not be there. Whereas with this approach suggested we wait till the final schema and metadata exists before examining what resolvers are needed.
>

Perfectly summed up.

PR #1745, opened 2020-07-23
refactor(plugin-engine): do not create resolvers in the "GraphbackPluginEngine#createSchema" method

machi1990:
Instead have a dedicated "GraphbackPluginEngine#createResolvers" method which is called after schema
transfromation phase has finished.

Fixes https://github.com/aerogear/graphback/issues/1742

### DIFF
--- a/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
+++ b/packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
@@ -46,8 +46,20 @@ export class GraphbackPluginEngine {
     this.plugins.push(...plugins);
   }
 
+  /**
+   * Allows the transformation of schema by applying transformation logic for each plugin
+   * Creation of resolvers, which has to come after all the changes in schema have been applied
+   * Saving of the transformed schema and related files
+   */
   public createResources(): GraphbackCoreMetadata {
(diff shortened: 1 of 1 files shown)


packages/graphback-core/src/plugin/GraphbackPluginEngine.ts
```diff
@@ -57,15 +69,18 @@ export class GraphbackPluginEngine {
...
-      console.warn("GraphbackEngine: No Graphback plugins registered");
-    }
     //We need to apply all required changes to the schema we need
     //This is to ensure that every plugin can add changes to the schema
     for (const plugin of this.plugins) {
       const newSchema = plugin.transformSchema(this.metadata);
       this.metadata.setSchema(newSchema);
+    }
 
+    return this.metadata;
```
craicoverflow, line 79: `createSchema` and `createResolvers` could just be void functions, since there is no reason to return `this.metadata`, which is available as a class variable.

machi1990: Good point.

machi1990: Good point. Pushed a quick update.

Issue #1684, opened 2020-07-08
[Discussion] selected fields should only be applied to `findOne`, `findBy` and `sync` related db queries but not write/delete operations


machi1990:
https://github.com/aerogear/graphback/pull/1509 added support to retrieve only fields that the user requested. 

This was applied for all the CRUD operations. I am opening this issue for discussion as I might have ran into a potential bug where subscriptions (create, update, delete) are broken. See example scenario below:

 
```graphql
""" @model """
type Note {
  id: ID!
  title: String!
  description: String
}
```

With the above model:

I've the following subscription created:
```graphql
type Subscription {
  newNote(filter: NoteSubscriptionFilter): Note!
  updatedNote(filter: NoteSubscriptionFilter): Note!
  deletedNote(filter: NoteSubscriptionFilter): Note!
}
```

If I subscribe to the creation live updates as below:
```graphql
subscription {
  newNote {
    id,
    title,
    description
  }
}
``` 

and I issue a create mutation 
```graphql

mutation {
  createNote(
    input: {
      title: "kichwa cha habari cha noti"
      description: "maelezo zaidi ya noti"
    }
  ) {
    id
  }
}
```

The create mutation actually works as one would expect but this breaks the subscription above with the following error
```json
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Note.title.",
      "locations": [
        {
          "line": 4,
          "column": 5
        }
      ],
      "path": [
        "newNote",
        "title"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ],
  "data": null
}
```

But If the mutation is changed to 
```
mutation {
  createNote(
    input: {
      title: "kichwa cha habari cha noti"
      description: "maelezo zaidi ya noti"
    }
  ) {
    id,
    title,
    description
  }
}
```

The subscription is working as expected.

```json
{
  "data": {
    "newNote": {
      "id": "6",
      "title": "kichwa cha habari cha noti",
      "description": "maelezo zaidi ya noti"
    }
  }
}
```

This is because the [Crud Service](https://github.com/aerogear/graphback/blob/9fbe415b2e4eb93eea3ef0cbdcac1bcb74418a53/packages/graphback-core/src/runtime/CRUDService.ts#L40-L78) only sent the payload with selected fields, meaning that if the actual write operation only requested `id` in return, the subscription payload sent will only have the `id` field and missing other fields, in this case the `title` and `description`. 
 
[...]

PR #1748, opened 2020-07-24
feat: in memory subscription filtering

craicoverflow:
This PR implements simplistic in-memory filtering of subscriptions using [withFilter](https://www.apollographql.com/docs/apollo-server/data/subscriptions/#subscription-filters) from graphql-subscriptions.

### Tasks

- [x] 72498d00b40b150566289200133f0c3aebfc3c78 return all fields from mutations as discussed in #1684
- [x] add ability to filter subscriptions on create, update, delete
- [ ] tests (struggled with how to add subscription filters in tests)
- [x] documentation

### Verification

Set up a subscription with a filter:

```gql
subscription {
  newNote(filter: {title:"note1", description:"note1hello"}) {
    id
    title
    description
    comments {
      id
    }
  }
}
```

Create a mutation matching the filter params. You should get a subscription even. Now try it with different filters.

```gql
mutation {
  createNote(input: {
    title: "note1",
    description:"note1hello"
  }) {
    id
  }
}
```

### Notes 

Opening this PR up for feedback now for a conversation on the general implementation. Tried adding quick tests to help with verify but they are non-trivial tricky so can revisit this.

### DIFF
--- a/docs/crud/subscriptions.md
+++ b/docs/crud/subscriptions.md
@@ -49,7 +49,7 @@ subscription {
 }
 ```
 
-and with a filter:
+You can apply field-equality filters to the subscription:
 
 ```graphql
 subscription {
@@ -73,7 +73,7 @@ subscription {
 }
 ```
 
-and with a filter:
+You can apply field-equality filters to the subscription:
 
 ```graphql
 subscription {
@@ -97,7 +97,7 @@ subscription {
 }
 ```
 
-and with a filter:
+You can apply field-equality filters to the subscription:
 
 ```graphql
 subscription {
--- a/docs/crud/subscriptions.md
+++ b/docs/crud/subscriptions.md
@@ -35,10 +35,6 @@ type Subscription {
 // highlight-end
 ```
 
-:::caution
-Subscription filtering is not available yet and will be implemented in a future release.
-:::
-
 ### Examples
 
 Subscribing to a [`create`](./mutations#create) event on `Note`:
@@ -46,6 +42,18 @@ Subscribing to a [`create`](./mutations#create) event on `Note`:
 ```graphql
 subscription {
   newNote {
+    id
+    title
+    likes
+  }
+}
+```
+
+and with a filter:
+
+```graphql
+subscription {
+  newNote(filter: {likes: 100}) {
     id
     title
     likes
@@ -65,11 +73,35 @@ subscription {
 }
 ```
 
+and with a filter:
+
+```graphql
+subscription {
+  updatedNote(filter: {likes: 100}) {
+    id
+    title
+    likes
+  }
+}
+```
+
 Subscribing to a [`delete`](./mutations#delete) event on `Note`:
 
 ```graphql
 subscription {
   deletedNote {
+    id
+    title
+    likes
+  }
+}
+```
+
+and with a filter:
+
+```graphql
+subscription {
+  updatedNote(filter: {likes: 100}) {
     id
     title
     likes
--- a/packages/graphback-core/src/runtime/CRUDService.ts
+++ b/packages/graphback-core/src/runtime/CRUDService.ts
@@ -1,6 +1,7 @@
 import * as DataLoader from "dataloader";
 import { PubSubEngine } from 'graphql-subscriptions';
-import { GraphbackCRUDGeneratorConfig, GraphbackOperationType, upperCaseFirstChar } from '..';
+import { GraphbackCRUDGeneratorConfig, GraphbackOperationType, upperCaseFirstChar, getSubscriptionName } from '..';
+import { withSubscriptionFilter } from './withSubscriptionFilter';
 import { GraphbackCRUDService, GraphbackDataProvider, GraphbackContext, GraphbackOrderBy, GraphbackPage, ResultList, QueryFilter } from '.';
 
 /**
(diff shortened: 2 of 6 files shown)


wtrocki:
This looks good. We can have quick chat about this as it has couple side impacts and it might require more cleanup.

wtrocki:
I think we can do optional field selection for models that do not have subscription enabled. 
IF we look into code we check if the subscription is enabled. For those cases we can skip selection and get better benchmark results (they are much better in this case). 

WDYT? Trick is that all will fall apart if we would want to have relationships returned by subscriptions. For that we need to apply hasura like approach.

craicoverflow:
> I think we can do optional field selection for models that do not have subscription enabled.
> IF we look into code we check if the subscription is enabled. For those cases we can skip selection and get better benchmark results (they are much better in this case).
> 
> WDYT? Trick is that all will fall apart if we would want to have relationships returned by subscriptions. For that we need to apply hasura like approach.

Brilliant idea!

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -38,6 +39,10 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
   }
 
   public async create(data: Type, context: GraphbackContext): Promise<Type> {
+    if (this.crudOptions.subCreate) {
```
craicoverflow, line 42: @wtrocki selecting all fields if subscription for mutation type is enabled.

craicoverflow:
Added in 72498d00b40b150566289200133f0c3aebfc3c78

If we decide to have some annotations to let the user subscribe to multiple topics from a subscription, it will become more difficult or will require an alternative approach to identify that a subscription is enabled.

```
"""@subscribe(mutation: ['createNote', 'deleteNote'])
noteChange: Note
```

wtrocki:
```"""@subscribe(mutation: ['createNote', 'deleteNote'])```
Seen that before. Let's have sync next week to make sure that we did not missed anything?

**Comment by wtrocki** [comment]
Ok. Question is.. do we need that having 3 separate subscriptions seem ok.

Alternative would be to use `__deleted` and `__updated` flags on single subscriptions, but now sure about this being too much opiniated.

**Comment by wtrocki** [comment]
> If we decide to have some annotations to let the user subscribe to multiple topics from a subscription, it will become more difficult or will require an alternative approach to..

I see this unlikely to happen and if we will we will need to redesign subscriptions anyway.

File: packages/graphback-core/src/runtime/CRUDService.ts (hunk 1 above)
Comment (machi1990) [reply packages/graphback-core/src/runtime/CRUDService.ts:42]: I think this will solve https://github.com/aerogear/graphback/issues/1684 right?

**Comment by machi1990** [comment]
Had a quick look and will go over this PR once again next week, but I feel like once this is merged it will be of a blog post e.g using Kafka / AMQ to perform out of the box GraphQL subscriptions.

**Comment by wtrocki** [comment]
Yep.. this begs for separate blog post

File: packages/graphback-core/src/runtime/withSubscriptionFilter.ts [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+    return asyncIterator
+  }
+
+  return withFilter(asyncIterator, (payload: any) => {
+    const subscriptionPayload = payload[subscriptionName]
+
+    for (const key of Object.keys(filter)) {
+      if (!subscriptionPayload[key] || subscriptionPayload[key].toString() !== filter[key].toString()) {
+        return false
+      }
```
machi1990, line 14: Nice. Seems like we only support equality filtering. Do we have plans to extend this?

wtrocki: Also if you guys can come back to us with desirable GraphQL filter for subscriptions so we can adjust spec.

craicoverflow: Agreement that we can extend this, but in a follow up issue/PR.

machi1990: Sure, let's discuss these in separate issue.

craicoverflow: https://github.com/aerogear/graphback/issues/1760

machi1990: Thanks

docs/crud/subscriptions.md
```diff
@@ -53,6 +49,18 @@ subscription {
 }
 ```
 
+and with a filter:
```
machi1990, line 52: Worth adding that this is an equality filter? Pending https://github.com/aerogear/graphback/issues/1760

craicoverflow: I'll add it before Examples.

craicoverflow:
> Tested this locally and it is good.
> 
> I noticed that we are missing tests for filters, I think it should be possible to call the `crudService.subscribeToXXX(filter, context).next(.. this is where the assertion will go)` to listen for updates. E.g by calling the `create` twice and verify once with the filtering and one without filtering and verify that it was called only once etc. This is just an idea, can be explored afterwards.

Thanks, I was struggling to figure out how to unit test this and did not want do delay on it considering subscription filtering will likely be changing in future.

Issue #1625, opened 2020-07-01
Filters on "id: ID" field not working for mongo db


machi1990:
I have a use case where I want to perform a not in filter:

```
query {
  findVolunteers(
    filter: { not: { id: { in: ["5ef626c190701a3b90ebb24f", "5ef9c631da553f262c7a95dc"] } } }
  ) {
    items {
      id
    }
    count
  }
}
```

Which should return all volunteers whose id is in not equal to the one passed in the array. 

But this returns 
```
{
  "data": {
    "findVolunteers": {
      "items": [
        {
          "id": "5ef6262d90701a3b90ebb227"
        },
        {
          "id": "5ef6266890701a3b90ebb237"
        },
        {
          "id": "5ef6269490701a3b90ebb245"
        },
        {
          "id": "5ef626c190701a3b90ebb24f"
        },
        {
          "id": "5ef626e690701a3b90ebb25b"
        },
        {
          "id": "5ef6271a90701a3b90ebb26a"
        },
        {
          "id": "5ef9c631da553f262c7a95dc"
        }
      ],
      "count": 7
    }
  }
}
```

With the `5ef626c190701a3b90ebb24f` and `5ef9c631da553f262c7a95dc` present while they should have been filtered out.

/cc @craicoverflow @wtrocki

machi1990:
I guess it comes from the transformation of `id` to `_id`. 

Because even the `eq` is failing.

```graphql
query {
   all: findVolunteers {
    items {
      id
    }
    count
  }
  in: findVolunteers(
    filter: { id: {
      in:  ["5ef626c190701a3b90ebb24f", "5ef9c631da553f262c7a95dc"]
    }}
  ) {
    items {
      id
    }
    count
  }
  
  eqFirst: findVolunteers(
    filter: { id: {
      eq:  "5ef626c190701a3b90ebb24f"
    }}
  ) {
    items {
      id
    }
    count
  }
  
  eqSecond: findVolunteers(
    filter: { id: {
      eq:  "5ef9c631da553f262c7a95dc"
    }}
  ) {
    items {
      id
    }
    count
  }
}
```

gives 
```json
{
  "data": {
    "all": {
      "items": [
        {
          "id": "5ef6262d90701a3b90ebb227"
        },
        {
          "id": "5ef6266890701a3b90ebb237"
        },
        {
          "id": "5ef6269490701a3b90ebb245"
        },
        {
          "id": "5ef626c190701a3b90ebb24f"
        },
        {
[...]

PR #1749, opened 2020-07-24
feat(scalars): support scalar out of the box

machi1990:
This adds built in support the listed scalars using the graphql-scalars library.

# Table Summary

| Scalar | Description | Database Type |
| ------------- |---------------| ------ |
| `GraphbackTime` | A time string at UTC, such as 10:15:30Z, compliant with the `full-time` format outlined in section 5.6 of the RFC 3339 profile of the ISO 8601 standard for representation of dates and times using the Gregorian calendar.| `time` |
| `GraphbackDate` | A date string, such as 2007-12-03, compliant with the `full-date` format outlined in section 5.6 of the RFC 3339 profile of the ISO 8601 standard for representation of dates and times using the Gregorian calendar. | `date` |
| `GraphbackDateTime` | A date-time string at UTC, such as 2007-12-03T10:15:30Z, compliant with the `date-time` format outlined in section 5.6 of the RFC 3339 profile of the ISO 8601 standard for representation of dates and times using the Gregorian calendar.| `datetime` |
| `GraphbackObjectID` | A field whose value conforms with the standard mongodb object ID as described here: https://docs.mongodb.com/manual/reference/method/ObjectId/#ObjectId. You will need to have the [`mongodb`](https://www.npmjs.com/package/mongodb) package installed in order to use this| `varchar(24)` |
| `GraphbackJSONObject` | The `JSONObject` scalar type represents JSON objects as specified by [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf). | `json` |
| `GraphbackJSON` | The `JSON` scalar type represents JSON values as specified by [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).| `json` |
| `GraphbackTimestamp` | The javascript `Date` as integer. Type represents date and time as number of milliseconds from start of UNIX epoch.| `timestamp` |

# Progress:
- [x] modify graphql migrations to take into account the added scalars
- [x] Unit and IT tests
- [x] Directly deserialize GraphbackObjectID into Mongo ObjectID instead of a string
- [x] Documentation
- [x] Release notes


packages/graphback-core/src/scalars/index.ts
```diff
@@ -1,2 +1,70 @@
...
+export const GraphbackDateTime = new GraphQLScalarType({
+  ...extractConfig(DateTimeResolver),
+  name: "GraphbackDateTime"
+});
+
+// TODO - modify parseValue and parseLiteral so that we deserialise into Mongo ObjectID instead of a string.
+export const GraphbackObjectID = new GraphQLScalarType({
+  ...extractConfig(ObjectIDResolver),
+  name: "GraphbackObjectID"
+});
```
machi1990, line 28: I feel like `parseLiteral `and `parseValue` should directly parse the string to [ObjectID](https://docs.mongodb.com/manual/reference/method/ObjectId/) type. What graphql-scalars does is only performing the validations against the argument and returning the string. For us going on a step further would mean the the MongoDB Data Provider would not have to do stuff like:
-   https://github.com/aerogear/graphback/blob/2b77fb445ad245eec00f92999b77b126586871b4/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L66
- https://github.com/aerogear/graphback/blob/2b77fb445ad245eec00f92999b77b126586871b4/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L101
- Or modifying the [query builder](https://github.com/aerogear/graphback/blob/2b77fb445ad245eec00f92999b77b126586871b4/packages/graphback-runtime-mongodb/src/queryBuilder.ts) to make ObjectID filters work, something observed when trying the things described in https://github.com/aerogear/graphback/issues/1625#issuecomment-660083625

machi1990: What do you think? @craicoverflow @wtrocki

wtrocki: +100 from me but I feel that GraphQL-Scalars (Urigo) Might want that as well. maybe worth creating issue on their side

integration/mock.graphql
```diff
@@ -41,3 +43,6 @@ type Task {
 type Query {
   helloWorld: String
 }
+
+scalar GraphbackTimestamp
```
wtrocki, line 47: The best benefit of having embedded scalars is to not have to define them in model. What this means that our schema plugin need to add them etc. WDYT?

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -716,12 +725,14 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     schemaComposer.add(StringScalarInputType);
     schemaComposer.add(BooleanScalarInputType);
     schemaComposer.add(JSONScalarInputType);
+    schemaComposer.add(JSONObjectScalarInputType);
```
wtrocki, line 728: Do you see dedicated scalar input types as well?

packages/graphback-core/src/scalars/index.ts
```diff
@@ -1,2 +1,70 @@
-export * from 'graphql-type-json';
-export * from './scalars'
+import { GraphQLNamedType, GraphQLScalarType } from 'graphql';
+import { TimeResolver, TimestampResolver, DateResolver, DateTimeResolver, ObjectIDResolver, JSONResolver, JSONObjectResolver } from 'graphql-scalars';
+
+export const GraphbackTime = new GraphQLScalarType({
```
wtrocki, line 4: Would it be possible to have those in schema plugin. Feels like specific to this plugin to me

packages/graphql-migrations/src/abstract/getColumnTypeFromScalar.ts
```diff
@@ -93,6 +93,22 @@ export default function(
     }
   }
 
+  // graphback datetime
+  if (scalarType && scalarType.name === 'GraphbackDateTime') {
```
wtrocki, line 97: Love this. biggest benefit ever

packages/graphback-core/src/scalars/index.ts
machi1990: > +100 from me but I feel that GraphQL-Scalars (Urigo) Might want that as well. maybe worth creating issue on their side.
 
> There is easy way to introduce this without breaking GraohQL Scalars. We do not want to parse object id twice - for validation and actual object

In order to be able to parse to actual `ObjectID` we'll need to use the [mongodb](https://www.npmjs.com/package/mongodb) package. Not sure, how to approach this: peerDependency?

integration/mock.graphql
machi1990: > The best benefit of having embedded scalars is to not have to define them in model. 

+1 on this. I actually started with this in mind, till I ran through issues "unknown GraphbackTimestamp type" when running my tests. so adding them to the schema was the solution. 

Happy to chat about it if there are suggestion as that would be of huge benefit.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: These inputs are only added for filtering.

packages/graphback-core/src/scalars/index.ts
machi1990: No strong opinion on this one. I put them on the core maybe because some other plugins would want to benefit from those and act on them. Also because the file was already there. @craicoverflow what do you think?

packages/graphql-migrations/src/abstract/getColumnTypeFromScalar.ts
machi1990: Yeap, and the added resolvers.

packages/graphback-core/src/scalars/index.ts
wtrocki: Yes.. That will work

integration/mock.graphql
wtrocki: Yes.. It will mean that model files will not be valid GraphQL schemas but we can always append all scalar definitions to them so things will work

packages/graphback-core/src/scalars/index.ts
wtrocki: Changed mind. +1 on keeping them

integration/mock.graphql
craicoverflow: This would be a breaking change as the Graphback API allows a GraphQLSchema object or a string as the model. It would also mean a change to the templates to load the schema as a string and not with `loadSchemaSync`.

Should this happen in a separate PR? Nothing from current PR would need to be changed as such for this to happen, but the scope is quite different and it would have an effect in other areas of the library.

packages/graphql-migrations/src/abstract/getColumnTypeFromScalar.ts
```diff
@@ -117,6 +141,14 @@ export default function(
     }
   }
 
+  // graphback json
+  if (scalarType && (scalarType.name === 'GraphbackJSON' || scalarType.name === 'GraphbackJSONObject')) {
```
craicoverflow, line 145: Wondering if we could also check if the scalar is `JSON` or `JSONObject` here and return type json, in the event that the user is already using these scalars.

machi1990: hmm I'd say since we are not breaking anything here, if `JSON` or `JSONBObject` were there, they'd continue to work the same way. Via `@db(type)` I suppose.

integration/mock.graphql
machi1990: @craicoverflow thanks for the analysis. Maybe we get this PR in first with proper documentation and evaluate further?

wtrocki: Let's drop new issue and cover it by triaging. The way we can do it is to generate model file initially that contains those scalars and people can manage them etc.

wtrocki:
I have tested this change with frontend stack we are building. 
Amazing feature. We really need it to define what types of scalars we support (as we cannot support any scalar sadly)!

I would love dropping some alpha release once we merge this!

packages/graphback-core/src/scalars/index.ts
machi1990: I have opened https://github.com/Urigo/graphql-scalars/issues/429 upstream. While waiting for a response, I'd do it on our side.

wtrocki: Awesome!

wtrocki:
A agreed. We might probably want to have docs PR separately as it might mean a lot of work.

integration/mock.graphql
machi1990: https://github.com/aerogear/graphback/issues/1756

craicoverflow:
Really nice stuff :smile: 

My main concern is how long the embedded scalar names are, especially because the user also has to add a scalar definition to their model. This can be tricky to type out and slow. Other than creating an alias suffix like `GBJSONObject` I don't see any other way though, and the current naming is probably more clear.

machi1990:
> Really nice stuff 
> 
> My main concern is how long the embedded scalar names are, especially because the user also has to add a scalar definition to their model. This can be tricky to type out and slow. Other than creating an alias suffix like `GBJSONObject` I don't see any other way though, and the current naming is probably more clear.

The more clear the better. I am planning to work on the documentation to list them so that it will be a matter of copying them from our website and posting them in the model file. Less error prone for the user.

machi1990:
@wtrocki @craicoverflow I've added some documentations. Let me know what you think.

docs/model/scalars.md
```diff
@@ -0,0 +1,103 @@
...
+  // highlight-start
+  """
+  Usage of the Graphback DateTime scalar
+  """
+  createdAt: GraphbackDateTime
+  // highlight-end
+}
+```
+
+The highlited code shows how you can define and use one of the Graphback scalars.
```
craicoverflow, line 57:
```suggestion
The highlighted code shows how you can define and use one of the Graphback scalars.
```

craicoverflow:
Docs look good!

Filtering of `GraphbackJSONInput` does not work for Postgres. What is the long term plan here? Do you see us being able to get some sort of operator mapping on JSON fields to work? Maybe we should remove this filter until such time we have it in place. I have not tested it in Mongo, does it work there? If so this can be kept in but documented that it only works for Mongo.

```gql
query {
  findNotes(filter: {
    test: {eq: "{\"hello\": 1}"}
  }) {
    items {
      id
      test
    }
  }
}

select \"id\", \"test\" from \"note\" where \"test\" = $1 - operator does not exist: json = unknown
```

machi1990:
> Docs look good!
> 
> Filtering of `GraphbackJSONInput` does not work for Postgres. What is the long term plan here? Do you see us being able to get some sort of operator mapping on JSON fields to work? Maybe we should remove this filter until such time we have it in place. I have not tested it in Mongo, does it work there? If so this can be kept in but documented that it only works for Mongo.
> 
> ```graphql
> query {
>   findNotes(filter: {
>     test: {eq: "{\"hello\": 1}"}
>   }) {
>     items {
>       id
>       test
>     }
>   }
> }
> 
> select \"id\", \"test\" from \"note\" where \"test\" = $1 - operator does not exist: json = unknown
> ```

I think we should remove this filtering, until we are able to have this mapping in. Good catch!

machi1990:
> Docs look good!
>

Thanks!
 
> Filtering of `GraphbackJSONInput` does not work for Postgres. What is the long term plan here? Do you see us being able to get some sort of operator mapping on JSON fields to work? Maybe we should remove this filter until such time we have it in place. I have not tested it in Mongo, does it work there? If so this can be kept in but documented that it only works for Mongo.
> 
> ```graphql
> query {
>   findNotes(filter: {
>     test: {eq: "{\"hello\": 1}"}
>   }) {
>     items {
>       id
>       test
>     }
>   }
> }
> 
> select \"id\", \"test\" from \"note\" where \"test\" = $1 - operator does not exist: json = unknown
> ```

I've disabled the feature for now. I've done json queries for pg, and getting the corresponding query from the filter should be doable but that will required a bit of infrastructural work: see https://github.com/aerogear/graphback/pull/1620#discussion_r450074824 where we initially chatted about this. With that in mind, I have opened https://github.com/aerogear/graphback/issues/1761

machi1990:
@craicoverflow I just pushed new changes to disable JSON filtering. Do you mind checking? Once that is good, we should be able to merge this one.

packages/graphback-core/package.json
```diff
@@ -44,12 +44,13 @@
...
+    "graphql-scalars": "1.2.2",
     "graphql-subscriptions": "1.1.0",
-    "graphql-type-json": "0.3.2",
     "pino": "6.3.2",
     "pluralize": "8.0.0"
   },
   "peerDependencies": {
-    "graphql": "^14.0.0 || ^15.0.0"
+    "graphql": "^14.0.0 || ^15.0.0",
+    "mongodb": "3.5.9"
```
craicoverflow, line 54: This means user must use mongodb version 3.5.9? Maybe it should be less restrictive and allow a minimum version.

machi1990: This is what we have for the mongo runtime too:

https://github.com/aerogear/graphback/blob/ff8ea5b04b1414625d4d7daca203ecad58e06777/packages/graphback-runtime-mongodb/package.json#L49

wtrocki:
Verfied. Want to share this with external parties for feedback and try it out with the datastore. Amazing job @machi1990

PR #1753, opened 2020-07-27
fix(schema): allow model field on non-model types

### LINKED ISSUE #1751: Error: Non-model maps to Graphback model which is not allowed
As the error states, I have a non-model object type which has a field of a type that is a Graphback model. This was working fine in alpha versions, but now I get this error. I need this functionality, I have custom data types that have nothing to do with the access to the database, but that return types that have access to the database. I'm not sure what this restriction is for, graphback should just ignore these types like it used to, if possible.

Here's my non-model type:

```graphql
type AnonymousJWTAuthResult {
    user: Users!
    csrfToken: String!
    authJWT: String!
}
```
The `user` field is of type `Users`, which is a modle. I return the `AnonymousJWTAuthResult` in a custom mutation, not sure why this is restricted.

craicoverflow:
Fixes #1751 

The original reason we disabled model fields on non-model types was because of circular generation causing a Maximum call stack size exceeded error, in scenarios where a Model referenced a non-model which referenced the model again.

Obviously we made this too restrictive and prevented other common use cases.

### Verification

Start your Graphback server with the following model:

```gql
type AnonymousJWTAuthResult {
    user: User!
    csrfToken: String!
    authJWT: String!
}

"""@model"""
type User {
  id: ID!
  name: String
}

type Query {
  test: AnonymousJWTAuthResult
}
```

### DIFF
--- a/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
+++ b/packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
@@ -292,10 +292,6 @@ function mapObjectInputFields(schemaComposer: SchemaComposer<any>, fields: Graph
 
     let inputType
     if (isObjectType(namedType)) {
-      // if (isModelType(namedType)) {
-      //   throw new Error(`Non-model ${objectName}.${field.name} maps to Graphback model (${namedType.name}) which is not allowed.`)
-      // }
-
       typeName = getInputTypeName(typeName, GraphbackOperationType.CREATE)
       namedType = schemaComposer.getOrCreateITC(typeName).getType()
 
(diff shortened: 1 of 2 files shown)


packages/graphback-codegen-schema/src/definitions/schemaDefinitions.ts
```diff
@@ -292,9 +292,9 @@ function mapObjectInputFields(schemaComposer: SchemaComposer<any>, fields: Graph
 
     let inputType
     if (isObjectType(namedType)) {
-      if (isModelType(namedType)) {
-        throw new Error(`Non-model ${objectName}.${field.name} maps to Graphback model (${namedType.name}) which is not allowed.`)
-      }
+      // if (isModelType(namedType)) {
+      //   throw new Error(`Non-model ${objectName}.${field.name} maps to Graphback model (${namedType.name}) which is not allowed.`)
+      // }
```
machi1990, line 297: Should we remove the comment code all together?

craicoverflow: Removed, thanks.

Issue #1744, opened 2020-07-23
DeltaTypes return documents or changed fields?


ssd71:
Should `DeltaType`s always return full documents or be allowed to return only changed fields?

wtrocki:
@ssd71 Cna you provide more information. This will affect subscriptions (if we want return delta there as well). GraphQL generally has this required flag and if type has required flag it will not be able for us to not return it - thus delta might require some specific type created

craicoverflow:
It should be the full document (IMO) because the user will likely need to work with the entire document (by rendering it on their webpage as an example). 

If we start returning only changed fields the user would have to merge the objects to keep the full document, which does not seem desirable.

ssd71:
> @ssd71 Cna you provide more information. This will affect subscriptions (if we want return delta there as well). GraphQL generally has this required flag and if type has required flag it will not be able for us to not return it - thus delta might require some specific type created

@wtrocki When using delta tables, we have the option of returning only either the changed fields or entire documents, but due to GraphQL some fields are strictly, required, if we were to allow returning changed fields that would mean making all fields optional (Except maybe `id` and `_deleted` and some other metadata which will need to be returned anyway)

> It should be the full document (IMO) because the user will likely need to work with the entire document (by rendering it on their webpage as an example).

> If we start returning only changed fields the user would have to merge the objects to keep the full document, which does not seem desirable.

@craicoverflow Fair point :+1:

wtrocki:
On the other hand, it will be confusing to not get data for what you requested (or get nulls). 
Cost of adding delta types returning delta etc. History API and changes will be only available on server to be able to perform server-side conflict resolution. For client-side conflict resolution, we going to send delta in error object (where we do not have any constraints over payload format)

Can I hook up into this and ask - do we want to have subscription delta expands (fetching relationships)?

> If we start returning only changed fields the user would have to merge the objects to keep the full document, which does not seem desirable.

Total offtopic/fun fact - most of the GraphQL clients do that out of the box. We can use subscription to refill your cache and watch for cache changes to fetch full objects.

ssd71:
> It will be confusing to not get data for what you requested (or get nulls).
Cost of adding delta types returning delta etc. History API and changes will be only available on server to be able to perform server-side conflict resolution. 

Agreed

> Can I hook up into this and ask - do we want to have subscription delta expands (fetching relationships)?

Just wondering, wouldn't being able to filter subscriptions make up for this?

wtrocki:
[...]

PR #1755, opened 2020-07-27
[WIP]Datasync prototype

ssd71:
PR is in very early stages :)
User-side API will change, some feedback on that would be :fire:  

## Planned
- [x] Custom Conflicts with soft deletes
- [ ] Proper Server-side conflict resolution
- [ ] TESTS

## What works
- Custom Conflict field and resolution strategy with soft deletion, almost everything else is WIP

## How to test functionality
```typescript
// templates/ts-apollo-mongodb-datasync-backend/src/index.ts
import { createDataSyncMongoDbProvider, createDataSyncCRUDService, DataSyncPlugin, VersionConflictEngine } from '@graphback/datasync'
.
.
    dataProviderCreator: createDataSyncMongoDbProvider(db, VersionConflictEngine),
    serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
    plugins: [
      new DataSyncPlugin(VersionConflictEngine)
    ]
  });

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -1,12 +1,12 @@
 import { createVersionedInputFields,createVersionedFields } from "@graphback/codegen-schema";
-import { GraphQLNonNull, GraphQLSchema, buildSchema, GraphQLString, GraphQLResolveInfo } from 'graphql';
+import { GraphQLNonNull, GraphQLSchema, buildSchema, GraphQLString, GraphQLResolveInfo, GraphQLInt } from 'graphql';
 import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, parseRelationshipAnnotation, metadataMap, GraphbackContext, getSelectedFieldsFromResolverInfo } from '@graphback/core'
 import { parseMetadata } from "graphql-metadata";
 import { SchemaComposer, ObjectTypeComposerFieldConfig } from 'graphql-compose';
 import { IResolvers, IFieldResolver } from '@graphql-tools/utils'
 
 import { getDeltaType, getDeltaListType, getDeltaQuery } from "./deltaMappingHelper";
-import { isDataSyncService, isDataSyncModel } from "./util";
+import { isDataSyncService, isDataSyncModel, ConflictEngine } from "./util";
 
 /**
  * Configuration for Schema generator CRUD plugin
@@ -22,6 +22,14 @@ export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
  * It will generate delta queries
  */
 export class DataSyncPlugin extends GraphbackPlugin {
+  protected conflictEngine: ConflictEngine;
+
+  public constructor(conflictEngine?: ConflictEngine) {
+    super()
+    if (conflictEngine !== undefined) {
+      this.conflictEngine = conflictEngine;
+    }
+  }
 
   public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
     const schema = metadata.getSchema()
@@ -72,14 +80,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
           filter: filterTypeName
         })
 
-        // Add updatedAt arg to update and delete input types
+        // Add _version arg to update and delete input types
         // for conflict resolution
         const { fieldNames } = metadataMap;
 
         const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
-        if (inputType) {
+        if (inputType && this.conflictEngine !== undefined) {
           inputType.addFields({
-            [fieldNames.updatedAt]: {
+            [this.conflictEngine.conflictFieldName]: {
               type: GraphQLNonNull(GraphQLString)
             }
           });
@@ -158,6 +166,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
       // metadata fields needed for @versioned
 
       modelTC.addFields(metadataFields);
+
+      if (this.conflictEngine !== undefined) {
+        modelTC.addFields({
+          [this.conflictEngine.conflictFieldName]: {
+            type: GraphQLString
+          }
+        })
+      }
 
       const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.FIND))
       if (inputType) {
--- a/packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
+++ b/packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
@@ -1,16 +1,17 @@
 import { GraphQLObjectType } from 'graphql';
-import { getDatabaseArguments, metadataMap, GraphbackContext, NoDataError, TransformType, FieldTransform, GraphbackOrderBy, GraphbackPage, QueryFilter, StringInput } from '@graphback/core';
+import { getDatabaseArguments, metadataMap, GraphbackContext, NoDataError, TransformType, FieldTransform, GraphbackOrderBy, GraphbackPage, QueryFilter, StringInput, ModelTableMap } from '@graphback/core';
 import { ObjectId } from 'mongodb';
 import { MongoDBDataProvider, applyIndexes } from '@graphback/runtime-mongo';
-import { ConflictError, ConflictStateMap } from "../util";
-import { DataSyncProvider } from "./DataSyncProvider";
+import { ConflictError, ConflictStateMap, ConflictEngine} from "../util";
+import { DataSyncProvider } from "./";
 
 /**
  * Mongo provider that attains data synchronization using soft deletes
  */
 export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider<Type> implements DataSyncProvider {
+  protected conflictEngine: ConflictEngine;
 
-  public constructor(baseType: GraphQLObjectType, client: any) {
+  public constructor(baseType: GraphQLObjectType, client: any, conflictEngine?: ConflictEngine) {
     super(baseType, client);
     applyIndexes([
       {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by wtrocki** [comment]
> VersionConflictEngine

This should be default. While having ability to swap detection mechanism I think that our entire ecosystem needs to base on some solid default etc.

File: packages/graphback-datasync/src/DataSyncPlugin.ts [hunk 1]
```diff
@@ -159,6 +167,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
       modelTC.addFields(metadataFields);
 
+      if (this.conflictEngine !== undefined) {
```
wtrocki, line 170: As mentioned. Always add versioned and only override if someone provided something else.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -72,14 +80,14 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+        // Add _version arg to update and delete input types
         // for conflict resolution
         const { fieldNames } = metadataMap;
 
         const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
-        if (inputType) {
+        if (inputType && this.conflictEngine !== undefined) {
           inputType.addFields({
-            [fieldNames.updatedAt]: {
+            [this.conflictEngine.conflictFieldName]: {
```
wtrocki, line 90: It is not about field. It is method that receives object and tells you that there is conflict. 
It can be field or number of fields or hash or even call to some extra database thing etc.

packages/graphback-datasync/src/DataSyncPlugin.ts
wtrocki: Please check how current offix does it. We do not want to downgrade capabilities unless we will have reason for it.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -22,45 +23,156 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
-      throw new ConflictError(conflict);
+    const { idField } = getDatabaseArguments(this.tableMap, data);
+
+    if (!idField.value) {
+      throw new NoDataError(`Cannot update ${this.collectionName} - missing ID field`)
     }
 
-    // TODO use findOneAndUpdate to check consistency afterwards
-    return super.update(data, context);
+    const objectId = new ObjectId(idField.value);
```
wtrocki, line 47: I really struggle to understand what is happening here.

Do we want to maintain single Provider using conflicts and without conflicts? Would it make sense to extend it for better code/visiblity etc.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -22,45 +23,156 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
+      if (this.conflictEngine !== undefined) {
+        
+        const { conflictFieldName, next } = this.conflictEngine;
+        const currentState = updateSets[conflictFieldName];
+        updateFilter[conflictFieldName] = currentState;
+
+        // eslint-disable-next-line @typescript-eslint/tslint/config
+        delete updateSets[conflictFieldName]
+
+        updateSets[conflictFieldName] = next(currentState, updateSets);
```
wtrocki, line 75: nextState should have fieldName information inside IMHO.
Entire implementation is leaking some concepts.

I would recommend to:

1) Not introduce state providers yet (as this requires understanding off how the state will change in real scenarios)

2) Adapt existing implementation (obviously there are some issue there so we can chat about this)
https://offix.dev/docs/next/conflict-server

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -22,45 +23,156 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
+        $set: updateSets
+      },{
+        w: "majority",
+        returnOriginal: false
+      })
+
+      if (!updateResult.value) {
+        // There is conflict or document doesn't exist
+  
+        // Check if document exists
```
wtrocki, line 88: This call makes no sene for me. We checking if document exist after we checked for conflict?

Can you put some text how conflict is being checked.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -22,45 +23,156 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
+    
+    throw new NoDataError(`Cannot update ${this.collectionName}`);
   }
 
   public async delete(data: any, context: GraphbackContext): Promise<Type> {
-    const conflict = await this.checkForConflicts(data, context);
-    if (conflict !== undefined) {
-      throw new ConflictError(conflict);
-    }
+    // throw new Error('not implemented');
```
wtrocki, line 127: That is the big stuff - as you might notice we do not support deletes in offix as most of the business use cases we had simply did not had actual delete operation so this problem never occurred, but as generic solution goes on we want to support it. Let's chat how this could work in meta language first.

packages/graphback-datasync/src/providers/DeltaMongoDBDataProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+import { getDatabaseArguments, metadataMap, GraphbackContext, NoDataError, TransformType, FieldTransform, GraphbackOrderBy, GraphbackPage, QueryFilter, StringInput, GraphbackOperationType } from '@graphback/core';
+import { ObjectId } from 'mongodb';
+import { MongoDBDataProvider, applyIndexes } from '@graphback/runtime-mongo';
+import { ConflictError, ConflictStateMap, ConflictEngine, HashConflictEngine } from "../util";
+import { DataSyncProvider, DataSyncMongoDBDataProvider, DeltaProvider } from "./";
+
+/**
+ * Mongo provider that attains data synchronization using soft deletes
+ */
+export class DeltaMongoDBDataProvider<Type = any> extends DataSyncMongoDBDataProvider <Type> {
```
wtrocki, line 11: I will need to have map/short doc of providers we add and how they relay with each other. We can see if we can simplify things. 

For me DataSync is always Delta. Conflict can be optional but we can also leverage this - maybe we should always have versioned conflict enabled etc. - all depends on how complex conflict check will become etc.

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,203 @@
+import { applyIndexes } from "@graphback/runtime-mongo";
```
wtrocki, line 1: Name probably wrong - it suggest it is provider. 
It is more like set of the delta table helpers etc.

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,203 @@
...
+  timestamp: Date,
+  version: number,
+  sets: {
+    [fieldName: string]: any
+  }
+}
+
+
+/**
+ * Provides an interface for interacting with deltas and stuff
```
wtrocki, line 32: Would this be used for data read? Are delta served from this?

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,203 @@
...
+      version: documentVersion,
+      sets
+    }
+
+    const res = await this.db.collection(this.collectionName).insertOne(deltaDocument);
+
+    return res.insertedId;
+  }
+
+  public async sync(lastSync: number, filter?: FilterQuery<any>) {
```
wtrocki, line 72: Any reason for using delta table for reading/sync
This is:
- Very slow (Probably)
- Complex
- Hard to maintain

We have the same information in the main table (with soft deletes). We do not need to remove soft deletes IMHO.
We keep delta table for conflict resolution (with some TTL)

Using it for delta requires complex merges that can span many layers and even return incomplete data (if you will miss entire objects or you start removing objects by TTL.)

Generally with some aggressive TTL we cannot avoid having situation that client will need to handle conflict (possibly by reapplying change etc.

packages/graphback-datasync/src/providers/DeltaProvider.ts
```diff
@@ -0,0 +1,203 @@
...
+          }
+        }
+      }, {
+        $match: filter
+      }];
+
+    return this.db.collection(this.collectionName).aggregate(syncPipeline).toArray()
+  }
+
+  public async checkForConflicts(_id: string|ObjectId, version: number, updateData: any): Promise<ConflictingFieldsReport> {
```
wtrocki, line 117: Another missunderstanding.

Conflict detection doesn't need delta tables. 
Implementation bellow doesn't make sense to me. 
Let's chat.

We need delta to handle conflict but we can detect it based on o the source table.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: I was under the impression that datasync = delta, and conflict would be opt-in, so wondering if that was a misunderstanding

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: if the above findOneAndUpdate fails, updateResult.value will be `null`, that means that there was no update done,
this means that either the document does not exist in the collection or the version did not match,
next we check if the document exists in the collection using findOne, if the document exists, then the reason update failed is because version did not match

packages/graphback-datasync/src/providers/DeltaMongoDBDataProvider.ts
ssd71: I am a bit lost here: Do we want opt-in conflict or conflict all the time?

packages/graphback-datasync/src/providers/DeltaProvider.ts
ssd71: Delta being served from this is very bad from multiple standpoints, I had written this code before we had the discussion on #1744 , this is definitely going away

packages/graphback-datasync/src/providers/DeltaProvider.ts
ssd71: This is going away

Would like to chat about what would happen if delta document deleted by TTL, kinda lost there on what to do when that happens

packages/graphback-datasync/src/providers/DeltaProvider.ts
ssd71: This is only checking if the fields that are being modified by a client have changed since the client fetched the document, this happens when version is mismatched in the source table, continuing: if the fields have been modified it throws conflict error, if the fields have not changed, it accepts the update

ssd71: In other words this is conflict being resolved, conflict checking is occurring in the data provider

packages/graphback-datasync/src/providers/DeltaProvider.ts
ssd71: It will change, this whole section is work in progress, I ported this from previous PR about delta tables so it is not completely concrete yet

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: So do we always have conflicts?

ssd71: Or do we always have version but conflict is optional?

ssd71:
I will redo this PR, will summarize what I understand here that will need to change, feel free to reply to add to this:
- Conflict will be based on function that receive object and return if there is conflict
- Sync will use base table
- Version field will always stay

Besides this, I have a few other questions:
- Will we always have delta table? 
- Will we always have conflicts?
- What about deletes?
- Multiple providers or single provider? If Multiple, do we separate them on strategy(softdelete only versus softdelete plus delta table) or something else?

packages/graphback-datasync/src/DataSyncPlugin.ts
wtrocki: It all depends how we structure API. 
I was thinking that conflict stuff has different helper and uses versioned out of the box, but if you want to have one helper then we need to pass version. 

Generally reason why we want to have conflict separate is that it will more or less be less performant way to edit stuff than without it. So if you do not use conflict you do not want to be affected by conflict shenanigans. 

As mentioned all of this is pretty much open to your interpretation, but the way we have it now suggest that we getting the worse of both worlds:

- Mixed implementation
- Increased complexity
- Verbose API
- Lack of clear distinction between conflict and non conflicts.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
wtrocki: Yep.. but Initially seen that as two separate providers. 
WHY - entire delta table and tons of other complex elements are simply not needed if you just want to have delta. Soft deletes approach can be reused as it has very limited impact. This means that conflict can be addon on top of very robust API we have.

This will mean that testing will be simpler and we will avoid complexity, tons of ifs all over the provider that looks very very well done on master - and IMHO should stay this way.

So TL;DR - user API should be making decision by picking different data provider rather than dropping all into datasync provider. 

This will also make it really easy to implement schema markers. 
Just datasync is needed and you control conflicts by picking different helper. 
I might be totally wrong but when I think about using this myself - this approach will make a lot of sense. What was proposed here brings mental challenge to me to understand what will happen. 

What happens if I do not provide conflict? What is versioned conflict? Should I provide my own. tons of info and reading required upfront. 

What we really want is just.. Hey .. want conflict - use this method. If not stay with this stuff that is on master.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
wtrocki: This contradics what we shown in initial documentation that we can use upserts. 
So this approach was applied without much comments why upsert did not work etc. 

Extra queries bring performance impact, so if you write in meta language what is suggested way to handle update/delete we can consult that with mongodb team and make sure that we getting the most performant option rather than something that just works.

packages/graphback-datasync/src/providers/DeltaMongoDBDataProvider.ts
wtrocki: See explanations above.

packages/graphback-datasync/src/providers/DeltaProvider.ts
wtrocki: Cool. Sorry for commenting in early phase.

packages/graphback-datasync/src/providers/DeltaProvider.ts
wtrocki: No need to be lost. TTL is simply to minimize amount of space. If your record hit TTL you cannot perform conflict resolution on sever and return error to user. 

No matter if you are getting partial results (some of the elements in common history) - if you do not have common base we cannot realize server side conflict.

In practical means - if I made edit when offline 5 min before going on 2 weeks holiday there is no sense to automatically resolve that change on server - History will be most likely gone and objects might have moved so much that using auto strategy on server will cause havoc in this cases. Returning error to user will let him to apply their changes again.

packages/graphback-datasync/src/providers/DeltaProvider.ts
wtrocki: oh.. So this is in wrong place then. As mentioned above we should wrap anything that suggest what fields or operations affect detection. Please check current implementation.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: Apologies, this was a shortcoming I forgot to discuss, this falls on me. My problem with using upserts: If we send an updateComment mutation with invalid id like anything garbage, upsert:true would create that document and we would likely end up with garbage in the database. I would certainly like to discuss about current approach and other possible approaches.

wtrocki:
As for actual implementation I would love you to check if we can:

1) Leave current master untouched and work the same way all the times
2) Add conflict as additional data layer (with ability to configure it to support server side conflict, configure TTL and anything else you would like for conflict in future)

Basing on limited investigation I think that will be:

1) Simple to understand
2) Simple to implement
3) Easy to maintain
4) Minimize number of variations for testing.
5) Do not affect performance as conflict is addition, not entire new implementation.

If you take look at offix plugin that we wrote in 0.11 it was literally doing that - despite we did not had any means to support deletes and server side conflict at that time. If we just extend what we had then we should be good.

wtrocki:
CC @craicoverflow @machi1990 diversity of opinions.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
craicoverflow: This could be on me somewhat..

@ssd71 originally had separate providers - I saw that there was very little difference between the providers, only the extra code for conflict handling and thought it would reduce API complexity by having to mix multiple providers with very little difference. 

https://github.com/aerogear/graphback/pull/1647#discussion_r449454898

ssd71: It's still two providers though, https://github.com/aerogear/graphback/pull/1755/files/526dbcf47e7714add3a74ca55f88e8aed592f6db#diff-7cfe2dc3aac0bbf8cc840e0fab566a2f
and https://github.com/aerogear/graphback/pull/1755/files/526dbcf47e7714add3a74ca55f88e8aed592f6db#diff-be2392f5aaf3d58e02d7e9c6429f279c
Sorry about the confusion

machi1990:
Thanks for the work @ssd71 and the ping. I'll sit and review this properly early tomorrow.

wtrocki:
Ignore review for the moment. We have call tommorow to get this over the line

ssd71:
This is outdated and the work is irrelevant.

Issue #1604, opened 2020-06-29
When no relationship annotation is defined, invalid input fields are added to input type


craicoverflow:
From https://github.com/aerogear/graphback/pull/1574#pullrequestreview-438421423

Defining this model will create a non-model input type for Note:

```gql
""" @model """
type Note {
  id: ID!
  title: String!
  description: String
  comments: [Comment]!
}

""" @model """
type Comment {
  id: ID!
  text: String
  description: String
}
```

Created an input type 
```
input MutateNoteInput {
  id: ID!
  title: String
  description: String
  comments: ID
}
```

This should either:

a) throw an error because there is no relationship annotation on `Note.comments`
b) Detect an implicit relationship and only create model input types.

machi1990:
Option (a) sounds more reliable to me. The thing about (b) is it is hard to know what kind of relationship it is.

E.g when you have a model like this:
```graphql
""" @model """
type First {
  id: ID!
  second: Second
}

""" @model """
type Second {
  id: ID!
}
```  

Is it a `@oneToOne` relationship with `First` as the relationship owner or `@onToMany` relationship with `Second` as the owner of the relationship? 

Having said that, auto detection of relationship will be a killer feature and certainly worth a separate issue till everything has been figured, for now I'd say we go with (a).
  
/cc @craicoverflow @wtrocki WDYT?

craicoverflow:
For now at least, we should go with **a** I think. As you said implicit relationships with be a brilliant feature but difficult to get right without unintended consequences, and right now we are targeting stability. Another thing about **b** is it would be challenging to map multiple relationship fields between two models, because at least one of these would require an annotation. This would require extra validation. The fact that every relationship requires an annotation means the user supplies mapping information.

On these grounds, let's go with **a** :smile:

PR #1766, opened 2020-07-28
fix(core): throw an error if relationship annotation are missing

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1604

### DIFF
--- a/packages/graphback-core/src/crud/mappingHelpers.ts
+++ b/packages/graphback-core/src/crud/mappingHelpers.ts
@@ -1,8 +1,7 @@
-import { GraphQLObjectType, GraphQLSchema, GraphQLField, getNamedType, isObjectType, isScalarType, GraphQLInputType, isEnumType } from 'graphql';
+import { GraphQLObjectType, GraphQLSchema, GraphQLField, getNamedType, isObjectType, isScalarType, isEnumType } from 'graphql';
 import { parseMetadata } from 'graphql-metadata';
-import * as pluralize from 'pluralize'
+import * as pluralize from 'pluralize';
 import { getUserTypesFromSchema } from '@graphql-toolkit/common';
-import { SchemaComposer } from 'graphql-compose';
 import { parseRelationshipAnnotation, transformForeignKeyName, getPrimaryKey } from '..';
 import { GraphbackOperationType } from './GraphbackOperationType';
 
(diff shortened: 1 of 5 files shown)


packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -157,10 +157,17 @@ export function getInputFieldName(field: GraphQLField<any, any>): string {
   return relationshipAnnotation.key || transformForeignKeyName(field.name);
 }
 
-export function getInputFieldNamedType(schemaComposer: SchemaComposer<any>, field: GraphQLField<any, any>, operation: GraphbackOperationType): GraphQLInputType {
+export function getInputFieldNamedType(schemaComposer: SchemaComposer<any>, modelName: string, field: GraphQLField<any, any>, operation: GraphbackOperationType): GraphQLInputType {
```
craicoverflow, line 160: The number of parameters for this function is 4, which is a bit more than ideal. It is unavoidable to not add `modelName` here.

`schemaComposer` is passed so that this can return the input type. Another option is to return the type name from `getInputFieldNamedType` instead and let the method which called it (which would have `schemaComposer` get the input type from the name.

What you think?

machi1990: Good idea. I like it!

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -157,17 +156,24 @@ export function getInputFieldName(field: GraphQLField<any, any>): string {
   return relationshipAnnotation.key || transformForeignKeyName(field.name);
 }
 
-export function getInputFieldNamedType(schemaComposer: SchemaComposer<any>, field: GraphQLField<any, any>, operation: GraphbackOperationType): GraphQLInputType {
+export function getInputFieldNamedTypeName(modelName: string, field: GraphQLField<any, any>, operation: GraphbackOperationType): string {
```
craicoverflow, line 159: Probably no need for the NamedType part anymore.

```suggestion
export function getInputFieldTypeName(modelName: string, field: GraphQLField<any, any>, operation: GraphbackOperationType): string {
```

machi1990: Good catch!

Issue #1724, opened 2020-07-14
No pagination for delta sync


wtrocki:
Delta sync do not allow to paginate thus it can return large number of results that will overload client. Pagination should be enabled

wtrocki:
@ssd71 We will not make it for 14 so this is 15 target

ssd71:
A few questions:
- Are we going to do cursor based pagination here or page based? I feel like page based pagination will fail here suppose if we have data that is being updated between client fetching two pages, client will probably miss some updates if we only use limit and skip
- If we want to do cursor based pagination e.g something like `lastSync` and `limit`:   some documents with same timestamps will be either missed (if we use simply `timestamp:{ gt: lastSync }` ) or repeated (if we use `timestamp:{ ge: lastSync }`

machi1990:
> * If we want to do cursor based pagination e.g something like `lastSync` and `limit`:   some documents with same timestamps will be either missed (if we use simply `_lastUpdatedAt:{ gt: lastSync }` ) or repeated (if we use `_lastUpdatedAt:{ ge: lastSync }`

I believe we have this kind of pagination implemented by `findXXX` query. Maybe we can go with something similar:
https://github.com/aerogear/graphback/blob/ce7236aa84cf053d0e9090115ab5b2b95d3276e4/packages/graphback-core/src/runtime/CRUDService.ts#L96-L117

Maybe this kind of logic can be implemented for the `DataSyncCRUDService#sync` method:
https://github.com/aerogear/graphback/blob/ce7236aa84cf053d0e9090115ab5b2b95d3276e4/packages/graphback-datasync/src/services/DataSyncCRUDService.ts#L21-L28

WDYT? 

Also let's get @wtrocki @craicoverflow opinion on this.

wtrocki:
Happy with what was suggested

ssd71:
@machi1990 Sorry, but just to make sure we are on the same page: you are proposing that we use `lastSync` and `limit` as parameters to paginate a sync query (instead of `offset` and `limit`) or are you proposing that we use `offset` and `limit` themselves?

ssd71:
I found [this](https://engineering.mixmax.com/blog/api-paging-built-the-right-way/) article about paginating data from mongodb and I think it sums up what I think about this pretty well

machi1990:
[...]

Issue #1349, opened 2020-06-01
Use DateTime scalars for createdAt updatedAt fields


ssd71:
The current way we have implemented this is very hacky. Once Graphback supported date/time scalars have been added, the hack is to be removed from code and refactoring needed to be done.

Blocked by #1095

machi1990:
With https://github.com/aerogear/graphback/pull/1749, we added support for DateTime scalars i.e `GraphbackDateTime`.

PR #1771, opened 2020-07-28
[DataSync] add datasync specific _lastUpdatedAt

ssd71:
~This PR restores the `_version` field that was present in offix-plugin in previous release~
`_version` will be properly added along with conflicts in the next PR

This PR only adds `_lastUpdatedAt` and removes update conflicts from default configuration

## TODO
- [x] Integration tests
- [ ] Release notes

### DIFF
--- a/integration/tests/runtime-workflow-datasync.ts
+++ b/integration/tests/runtime-workflow-datasync.ts
@@ -37,8 +37,8 @@ const updatedTS2 = new Date(1596622363886);
 const objectId = new ObjectID("507f191e810c19729de860ea");
 
 const SYNC_NOTES = gql`
-query syncNotes($lastSync: GraphbackTimestamp!, $filter: NoteFilter) {
-    syncNotes(lastSync: $lastSync, filter: $filter) {
+query syncNotes($lastSync: GraphbackTimestamp!, $filter: NoteFilter, $limit: Int) {
+    syncNotes(lastSync: $lastSync, filter: $filter, limit: $limit) {
         items {
           _id
           title
@@ -160,6 +160,27 @@ test('Sync all notes with filtering', async () => {
         eq: 'Note A'
       }
     }
+  }});
+
+  expect(data).toBeDefined();
+  expect(data.syncNotes.items).toHaveLength(1);
+  expect(data.syncNotes.items).toEqual([
+    {
+      _id: notesId[0],
+      title: 'Note A',
+      description: 'Note A Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    }
+  ])
+  expect(typeof data.syncNotes.lastSync).toEqual('number')
+})
+
+test('Sync all notes with limit', async () => {
+
+  const { data } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    limit: 1
   }});
 
   expect(data).toBeDefined();
--- a/integration/tests/runtime-workflow-datasync.ts
+++ b/integration/tests/runtime-workflow-datasync.ts
@@ -1,0 +1,245 @@
+/* eslint-disable max-lines */
+/* eslint-disable no-null/no-null */
+/* eslint-disable no-shadow */
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable import/no-extraneous-dependencies */
+import { mkdirSync, readFileSync, rmdirSync } from 'fs';
+import * as path from 'path';
+import { ApolloServer, gql } from "apollo-server";
+import { createTestClient, ApolloServerTestClient } from 'apollo-server-testing';
+import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader'
+import { loadDocuments } from '@graphql-tools/load'
+import { buildGraphbackAPI, GraphbackAPI } from "graphback";
+import { DocumentNode } from 'graphql';
+import { MongoClient, Db, ObjectID } from 'mongodb';
+import { PubSub } from "graphql-subscriptions";
+import { advanceTo } from "jest-date-mock";
+import { createDataSyncMongoDbProvider, createDataSyncCRUDService, DataSyncPlugin, DataSyncFieldNames } from "../../packages/graphback-datasync";
+import { SchemaCRUDPlugin } from '../../packages/graphback-codegen-schema';
+import { ClientCRUDPlugin } from '../../packages/graphback-codegen-client';
+
+/** global config */
+let db: Db;
+let mongoClient: MongoClient;
+let server: ApolloServer;
+let client: ApolloServerTestClient;
+let graphbackApi: GraphbackAPI;
+
+let documents: DocumentNode;
+
+const notesId = [];
+const commentId = [];
+const metadataId = [];
+
+const modelText = readFileSync("./datasync-mongodb-model.graphql").toString();
+const startTS = new Date(1596622318448);
+const updatedTS2 = new Date(1596622363886);
+const objectId = new ObjectID("507f191e810c19729de860ea");
+
+const SYNC_NOTES = gql`
+query syncNotes($lastSync: GraphbackTimestamp!, $filter: NoteFilter) {
+    syncNotes(lastSync: $lastSync, filter: $filter) {
+        items {
+          _id
+          title
+          description
+          _lastUpdatedAt
+          _deleted
+        },
+        lastSync
+    }
+}
+`
+
+beforeAll(async () => {
+  try {
+    mkdirSync("./output-datasync-mongo");
+    mkdirSync("./output-datasync-mongo/client")
+
+    mongoClient = new MongoClient('mongodb://mongodb:mongo@localhost:27017/users?authSource=admin', { useUnifiedTopology: true });
+    await mongoClient.connect();
+    db = mongoClient.db('users');
+    graphbackApi = buildGraphbackAPI(modelText, {
+      dataProviderCreator: createDataSyncMongoDbProvider(db),
+      serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
+      plugins: [
+        new SchemaCRUDPlugin({ outputPath: "./output-datasync-mongo/schema/schema.graphql" }),
+        new ClientCRUDPlugin({ outputFile: './output-datasync-mongo/client/graphback.graphql' }),
+        new DataSyncPlugin()
+      ]
+    });
+    
+    await seedDatabase(db);
+    
+    const source = await loadDocuments(path.resolve(`./output-datasync-mongo/client/graphback.graphql`), {
+      loaders: [
+        new GraphQLFileLoader()
+      ]
+    });
+    documents = source[0].document;
+  } catch (e) {
+    throw e;
+  }
+})
+
+beforeEach(() => {
+  const { typeDefs, resolvers, contextCreator } = graphbackApi;
+  server = new ApolloServer({
+    typeDefs,
+    resolvers,
+    context: contextCreator
+  });
+  
+  client = createTestClient(server);
+})
+
+afterEach(() => server.stop())
+
+afterAll(async () => {
+  rmdirSync(path.resolve('./output-datasync-mongo'), { recursive: true });
+  await db.dropCollection('note');
+
+  return mongoClient.close();
+});
+
+async function seedDatabase(db: Db) {
+  const notes = [
+    {
+      title: 'Note A',
+      description: 'Note A Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    },
+    {
+      title: 'Note B',
+      description: 'Note B Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    }
+  ]
+
+  for (const note of notes) {
+    const { ops } = await db.collection("note").insertOne(note);
+    notesId.push(ops[0]._id);
+  }
+
+  
+}
+
+test('Sync all notes', async () => {
+
+  const { data, errors, extensions } = await client.query({ operationName:"syncNotes", query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {}
+  }});
+  expect(data).toBeDefined();
+  expect(data.syncNotes?.items).toEqual([
+    {
+      _id: notesId[0],
+      title: 'Note A',
+      description: 'Note A Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    },
+    {
+      _id: notesId[1],
+      title: 'Note B',
+      description: 'Note B Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    }
+  ])
+})
+
+test('Sync all notes with filtering', async () => {
+
+  const { data } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
+      title: {
+        eq: 'Note A'
+      }
+    }
+  }});
+
+  expect(data).toBeDefined();
+  expect(data.syncNotes.items).toHaveLength(1);
+  expect(data.syncNotes.items).toEqual([
+    {
+      _id: notesId[0],
+      title: 'Note A',
+      description: 'Note A Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false
+    }
+  ])
+  expect(typeof data.syncNotes.lastSync).toEqual('number')
+})
+
+test('Should update lastUpdatedAt field on create', async () => {
+  const createResponse = await createNote(client, { title: 'Note C'});
+
+  expect(createResponse.data).toBeDefined();
+  expect(typeof createResponse.data.createNote[DataSyncFieldNames.lastUpdatedAt]).toEqual('number')
+})
+
+test('Should update lastUpdatedAt field on update', async () => {
+  const updateResponse = await updateNote({ _id: notesId[1], description: 'Note B Description'}, client);
+
+  expect(updateResponse.data).toBeDefined();
+  expect(typeof updateResponse.data.updateNote[DataSyncFieldNames.lastUpdatedAt]).toEqual('number');
+  expect(updateResponse.data.updateNote[DataSyncFieldNames.lastUpdatedAt]).not.toEqual(startTS.valueOf());
+})
+
+test('Should update lastUpdatedAt and deleted on delete', async () => {
+  const createResponse = await createNote(client, { title: 'Note X'});
+  const noteId = createResponse.data?.createNote._id;
+
+  const deleteResponse = await deleteNote(client, noteId.toString())
+  expect(deleteResponse.data).toBeDefined();
+  expect(typeof deleteResponse.data.deleteNote[DataSyncFieldNames.lastUpdatedAt]).toEqual('number');
+  expect(deleteResponse.data.deleteNote[DataSyncFieldNames.lastUpdatedAt]).not.toEqual(createResponse.data[DataSyncFieldNames.lastUpdatedAt]);
+
+  const syncResponse = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: createResponse.data.createNote[DataSyncFieldNames.lastUpdatedAt],
+    filter: {
+      title: {
+        eq: 'Note X'
+      }
+    }
+  }});
+
+  expect(syncResponse.data).toBeDefined();
+  const { data } = syncResponse;
+  expect(data.syncNotes.items).toHaveLength(1);
+  expect(data.syncNotes.items[0][DataSyncFieldNames.deleted]).toEqual(true);
+})
+
+async function updateNote(input: any, client: ApolloServerTestClient) {
+  const response = await client.mutate({
+    operationName: "updateNote",
+    mutation: documents,
+    variables: { input }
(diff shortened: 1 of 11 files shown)


packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -37,6 +38,11 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
       throw new ConflictError(conflict);
     }
 
+    const currentVersion = data._version;
+    if (typeof (currentVersion) === "number") {
```
machi1990, line 42: Is it possible that the `_version` field will be of other type?

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -48,7 +54,11 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
     }
 
     const { idField } = getDatabaseArguments(this.tableMap, data);
+    const currentVersion = data._version;
     data = {};
+    if (typeof (currentVersion) === "number") {
```
machi1990, line 59: I am wondering if this is needed.

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: Nope, this isn't actually needed, thanks for pointing out!

machi1990: Great!

wtrocki:
Missing:

- Docs
- Release notes
- Integration tests 
- We cannot merge that without fixing datasync-starter (PR should be created first)

ssd71:
@wtrocki How do you think about doing the following in this PR as well:

- add _lastUpdatedAt and remove `updatedAt` and `createdAt`
- only use `@datasync`
- remove conflicts by default

I believe this way I could write docs better as well as write consistent tests, wdyt?

wtrocki:
> updatedAt and createdAt

-1. can we just leave it alone as non DataSync feature?

Rest is ok

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
```diff
@@ -103,14 +105,14 @@ input MutateCommentInput {
...
-  updatedAt: String!
+  _version: Int!
 }
 
 input MutateNoteInput {
   id: ID!
   title: String
   description: String
-  updatedAt: String!
+  _version: Int!
```
craicoverflow, line 115: I would imagine this should not be added to the mutation input - the user should not be able to set the value.

craicoverflow: Correct me if I am wrong @ssd71 @wtrocki

wtrocki: User needs to pass the value unchanged

machi1990: The `CreateXXXInput`, but I think it should be there for the update input, since it will be used for conflicts detection https://github.com/aerogear/graphback/pull/1771/files#diff-be2392f5aaf3d58e02d7e9c6429f279cR146.

ssd71:
Yet to rebase this on #1769 

Should I start now, or wait until it is merged?

machi1990:
> Yet to rebase this on #1769
> 
> Should I start now, or wait until it is merged?

It won't be long till it is merged, you can wait a bit.

machi1990:
> Yet to rebase this on #1769
> 
> Should I start now, or wait until it is merged?

@ssd71 #1769  has been merged. Feel free to rebase now.

ssd71:
rebased now, TODOs: https://github.com/aerogear/graphback/pull/1771#issue-457935611

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -149,21 +115,74 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
-
-      modelTC.addFields(metadataFields);
 
-      const inputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.FIND))
-      if (inputType) {
-        const metadataInputFields = createVersionedInputFields();
-        inputType.addFields(metadataInputFields);
+    modelTC.addFields({
+      [DataSyncFieldNames.lastUpdatedAt]: {
+        type: GraphbackTimestamp
```
machi1990, line 124: This is good to see and kind goes in the way of fixing https://github.com/aerogear/graphback/issues/1349. I've a few questions regarding this line:

What happens if the schema contains `scalar GraphbackTimestamp` already?  Will this work?

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -149,21 +115,74 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+      ...allowedFields,
+      [DataSyncFieldNames.deleted]: 'Boolean',
+    });
+
+    // Create and Add Delta List type to schema
+    schemaComposer.createObjectTC({
+      name: getDeltaListType(modelName),
+      fields: {
+        items: `[${getDeltaType(modelName)}]!`,
+        lastSync: GraphQLNonNull(GraphbackTimestamp)
```
machi1990, line 172: See https://github.com/aerogear/graphback/pull/1771/files#r464225435

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Thanks for the catch! I haven't fiddled around with scalars yet, hence :P, working on this

ssd71: I have updated this: Now it uses the Graphback scalar if schema already contains it, wondering if this is unnecessary or some other approach is required here:
https://github.com/aerogear/graphback/blob/c02c4b50159a1365d2fe0060435e09a76a50f507/packages/graphback-datasync/src/DataSyncPlugin.ts#L31-L33

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -22,10 +15,24 @@ export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
...
   public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
     const schema = metadata.getSchema()
     const schemaComposer = new SchemaComposer(schema);
+    if (!schemaComposer.has(GraphbackTimestamp.name)) {
+      schemaComposer.createScalarTC(GraphbackTimestamp);
+    } else {
+      const TSScalarTC = schemaComposer.getSTC(GraphbackTimestamp.name);
+      if (!isSpecifiedGraphbackScalarType(TSScalarTC.getType())) {
+        throw new Error('GraphbackTimestamp is not a Graphback scalar: DataSync requires GraphbackTimestamp to be a Graphback specified scalar.');
+      }
```
machi1990, line 34: I do not think this is necessary:

IIRC, under the hood `isSpecifiedGraphbackScalarType` checks if `GraphbackTimestamp` is among known Graphback scalars and if it is a scalar.

packages/graphback-datasync/src/DataSyncPlugin.ts
machi1990: > I have updated this: Now it uses the Graphback scalar if schema already contains it

Great, this should work. Thanks.  I was wondering if we can add a test case (snaphot test) for this?

> ..., wondering if this is unnecessary or some other approach is required here:
> https://github.com/aerogear/graphback/blob/c02c4b50159a1365d2fe0060435e09a76a50f507/packages/graphback-datasync/src/DataSyncPlugin.ts#L31-L33

I've added a comment in this regard.

ssd71: > Great, this should work. Thanks. I was wondering if we can add a test case (snapshot test) for this?

Thanks for the review! added snapshot now

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Removed 🗑️

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -97,51 +106,18 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
     }
 
-    filter._deleted = { ne: true };
+    filter[DataSyncFieldNames.deleted] = { ne: true };
 
     return super.count(filter);
   }
 
-  public sync(lastSync: string, context: GraphbackContext, filter?: any): Promise<Type[]> {
+  public sync(lastSync: number, context: GraphbackContext, filter?: any): Promise<Type[]> {
```
craicoverflow, line 114: `lastSync` is not a number, it is a `Date`. The scalar value is parsed as a Date.

```suggestion
  public sync(lastSync: Date, context: GraphbackContext, filter?: any): Promise<Type[]> {
```

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
```diff
@@ -97,51 +106,18 @@ export class DataSyncMongoDBDataProvider<Type = any> extends MongoDBDataProvider
...
 
     return super.findBy({
       ...filter,
-      updatedAt: {
-        gt: lastSync
+      [DataSyncFieldNames.lastUpdatedAt]: {
+        ge: lastSync
```
craicoverflow, line 119: Because `lastSync` is not a number, this query is not working (tested from the playground).

ssd71:
Thanks for the review @craicoverflow !

Will address all queries and update PR after sometime

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
machi1990: Good catch!

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -22,10 +15,19 @@ export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
  * It will generate delta queries
  */
 export class DataSyncPlugin extends GraphbackPlugin {
+  protected addVersion: boolean;
+
+  public constructor(addVersion: boolean = false) {
```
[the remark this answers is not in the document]
ssd71: Thanks! this is done 👍

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -149,21 +111,74 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+        }
+      });
     }
   }
+
+  protected addDeltaQuery(schemaComposer: SchemaComposer<unknown>, model: ModelDefinition) {
+    // Create Delta Type
+    const modelName = model.graphqlType.name;
+    const modelTC = schemaComposer.getOTC(modelName);
+    const allowedFieldEntries = Object.entries(modelTC.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
```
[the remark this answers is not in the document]
ssd71: I've updated this, do you think I should still be more explicit in saying this doesn't include relationship fields?
https://github.com/aerogear/graphback/blob/7402981623195f2a86203d1ed7c1b5cd1e17d341/packages/graphback-datasync/src/DataSyncPlugin.ts#L190-L199

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -149,21 +111,74 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+      ...allowedFields,
+      [DataSyncFieldNames.deleted]: 'Boolean',
+    });
+
+    // Create and Add Delta List type to schema
+    schemaComposer.createObjectTC({
+      name: getDeltaListType(modelName),
+      fields: {
+        items: `[${getDeltaType(modelName)}]!`,
+        lastSync: `${GraphbackTimestamp.name}!`
```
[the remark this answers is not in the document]
ssd71: Thanks, This is done, feel free to mark similar comments as resolved 😄

packages/graphback-datasync/src/providers/DatasyncMongoDBDataProvider.ts
ssd71: Thanks! Addressed this in last push

ssd71:
Need some pointers on writing integration test for datasync, as client plugin doesn't support datasync, maybe I should write my own queries? :P

machi1990:
> Need some pointers on writing integration test for datasync, as client plugin doesn't support datasync, maybe I should write my own queries? :P

Yes, we could do that.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -1,37 +1,39 @@
...
  *
  * Plugin is enabled by """ @datasync """ annotation
  * It will generate delta queries
  */
 export class DataSyncPlugin extends GraphbackPlugin {
-  protected addVersion: boolean;
+  protected useVersion: boolean;
 
-  public constructor(addVersion: boolean = false) {
+  public constructor(config: DataSyncPluginConfig = {
```
craicoverflow, line 24: Initializing a default value for an object directly in the function argument won't scale after you add different props because you cannot merge user defined config values with default values.

Copy the style from [SchemaCRUDPlugin](https://github.com/aerogear/graphback/blob/f8ab9f3cf4a8fe58379a0eac1655f0e9c825d82b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L43) by using spread operator of `Object.assign` to merge values in the constructor body.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -1,37 +1,39 @@
...
 
-/**
- * Configuration for Schema generator CRUD plugin
- */
-
-
 export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
 
+export interface DataSyncPluginConfig {
+  useVersion: boolean
```
craicoverflow, line 12:
```suggestion
  useVersion?: boolean
```

You can set the default in the constructor.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -118,14 +120,15 @@ export class DataSyncPlugin extends GraphbackPlugin {
   protected addDataSyncFieldsToModel(schemaComposer: SchemaComposer<any>, model: ModelDefinition) {
     const name = model.graphqlType.name;
     const modelTC = schemaComposer.getOTC(name);
+    const TimestampSTC = schemaComposer.getSTC(GraphbackTimestamp.name);
```
craicoverflow, line 123: There is no need for this, you can just refer to the scalar directly (this is getting the type from the type name, but you already have the type).

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -118,14 +120,15 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
 
     modelTC.addFields({
       [DataSyncFieldNames.lastUpdatedAt]: {
-        type: GraphbackTimestamp
+        type: TimestampSTC.getType()
```
craicoverflow, line 127: I am not 100% sure, but you can just reference the scalar type directly.

See [`7402981` (#1771)](https://github.com/aerogear/graphback/pull/1771/commits/7402981623195f2a86203d1ed7c1b5cd1e17d341#r465712478)

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -1,37 +1,39 @@
...
-  protected addVersion: boolean;
+  protected useVersion: boolean;
 
-  public constructor(addVersion: boolean = false) {
+  public constructor(config: DataSyncPluginConfig = {
+    useVersion: false
+  }) {
     super()
-    this.addVersion = addVersion
+    this.useVersion = config.useVersion
```
craicoverflow, line 28: It will be neater to keep all config in a single global variable.

```suggestion
    this.config = config
```

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -151,38 +154,47 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+    const findFilterITC = schemaComposer.getITC(getInputTypeName(modelName, GraphbackOperationType.FIND));
     schemaComposer.Query.addFieldArgs(deltaQuery, {
-      lastSync: GraphQLNonNull(GraphbackTimestamp),
-      filter: findFilterTypeName
+      lastSync: GraphQLNonNull(TimestampSTC.getType()),
+      filter: findFilterITC.getType()
+    });
+  }
+
+  private getDeltaTypeFields(modelTC: ObjectTypeComposer) {
```
craicoverflow, line 190: You have access to the GraphQL type from the calling method, and all that is being done with the type composer is getting the fields (which does not require GraphQL Compose).

```suggestion
  private getDeltaTypeFields(modelType: GraphQLObjectType) {
```

Also, in general it is easier to unit test a mock GraphQLObjectType instead of an ObjectTypeComposer.

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -151,38 +154,47 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+    const DeltaTypeFieldEntries = Object.entries(modelTC.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
```
craicoverflow, line 191: `ObjectTypeComposerFieldConfig` could probably be replaced with the GraphQL equivalent.

```suggestion
    const DeltaTypeFieldEntries = Object.entries(modelTC.getFields()).filter((e: [string, ObjectTypeComposerFieldConfig<any, unknown, any>]) => {
```

craicoverflow:
When ready for review again, use the **Re-request review** button so that it will appear in my _Pending Review_ tab.

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Hmm, I tried doing that before(and now) as well, it throws when GraphbackTimestamp is already present and used somewhere in the schema:
image
wdyt

ssd71:
CC @wtrocki

wtrocki:
Should I be:

- Reviewing changes?
- Verifying this?

integration/tests/runtime-workflow-datasync.ts
```diff
@@ -0,0 +1,245 @@
...
+      [DataSyncFieldNames.deleted]: false
+    }
+  ])
+})
+
+test('Sync all notes with filtering', async () => {
+
+  const { data } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
```
wtrocki, line 158: Are we planning to add `limit`

packages/graphback-datasync/src/util.ts
```diff
@@ -38,3 +33,9 @@ export class ConflictError extends Error {
     this.conflictInfo = stateMap;
   }
 }
+
+export const DataSyncFieldNames = {
+  version: '_version',
```
wtrocki, line 38: I though we will use double underscores same as graphql introspection format.

integration/tests/runtime-workflow-datasync.ts
ssd71: @wtrocki There was some discussion about this: https://github.com/aerogear/graphback/issues/1724

wtrocki:
What I think it is needed:

- Adding limit (obviously without offset - doesn't make sense)
- Integration test for conflict resolution.

Let's use this PR to make final version of the datasync we can play with.

packages/graphback-datasync/src/util.ts
ssd71: Sure, we can go with that

ssd71:
> Adding limit (obviously without offset - doesn't make sense)

@wtrocki Are you sure we should add it in now or we could investigate this a bit more?

> Integration test for conflict resolution.

Have to add conflicts first though :P

packages/graphback-datasync/src/DataSyncPlugin.ts
craicoverflow: My bad, thanks for checking.

You could still access the type directly with SchemaComposer:

```ts
const TimestampScalar = schemaComposer.types.get(GraphbackTimestamp.name);
```

packages/graphback-datasync/src/util.ts
ssd71: All three are using double underscores now

ssd71:
>     * Adding limit (obviously without offset - doesn't make sense)
> 
>     * Integration test for conflict resolution.

@wtrocki 
- :heavy_check_mark: limit
- :heavy_check_mark: double underscores

ssd71:
@wtrocki  integration tests failing :grimacing: 
image

machi1990:
> @wtrocki integration tests failing 
> image

This calls for the ["mapping"](https://github.com/aerogear/graphback/issues/637) capability :-P

ssd71:
> This calls for the "mapping" capability :-P

Not sure if we have that :thinking:

wtrocki:
Ok let's go back to single `_` then

wtrocki:
Good job!

machi1990:
Great work @ssd71 thanks for this and for your patient in this during the review. 

I noticed that we are still missing the release note, do you mind adding a few lines in https://github.com/aerogear/graphback/blob/master/docs/releases.md#016 in a separate PR?

PR #1773, opened 2020-07-29
chore(templates): add codegen scalar mapping for Graphback scalars

machi1990:
Follow up https://github.com/aerogear/graphback/pull/1769#discussion_r461811695


wtrocki:
can you also regenerate ./src/generated-types.tsx file. 
```
yarn codegen generate
```

craicoverflow:
I do not see the reason for regenerating `generated-types.tsx`. The model does not have any of the Graphback scalars in it, so if a user runs `yarn graphql-codegen` again upon downloading a new template the scalar types will just get removed from `generated-types.tsx`. If a user adds a Graphback scalar to their schema and generates the types file it will be added at this point.

PR #1778, opened 2020-07-29
fix(create-graphback): fix issue with generating a fullstack template

### LINKED ISSUE #1777: Choosing a fullstack template from `create-graphback` generates an invalid project
When I use `create-graphback` to generate a new project and choose either of the fullstack templates - `apollo-fullstack-react-postgres-ts` or `apollo-fullstack-react-mongo-ts` the generated project folder contains a valid `client` subdirectory and nothing else.

### Steps to reproduce

1. `npx create-graphback my-fullstack-app`
2. Choose a template to bootstrap: `apollo-fullstack-react-mongo-ts`
3. cd `my-fullstack-app`.
4. The project will only have the `client` subfolder but no other files in the project root.

machi1990:
Fixes #1777

### DIFF
--- a/packages/create-graphback/src/init/starterTemplates.ts
+++ b/packages/create-graphback/src/init/starterTemplates.ts
@@ -164,8 +164,7 @@ export async function extractTemplate(template: Template, name: string) {
       mkdirSync(output);
     }
 
-    return extractStarterFromRepository(file, tarInfo, output);
+    await extractStarterFromRepository(file, tarInfo, output);
   }
-
 
 }


machi1990:
Good catch! @craicoverflow  

I hope this should fix this issue. Milestoned this for 0.15, but wondering if it is worth a patch release?

wtrocki:
I think we might drop 0.15.0 and save us another bump on graphql-cli?

machi1990:
> I think we might drop 0.15.0 and save us another bump on graphql-cli?

Good point!

PR #1779, opened 2020-07-30
chore: auto issue labeler

machi1990:


.github/workflows/auto-issue-label.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+    autolabel:
+      runs-on: ubuntu-latest
+      name: auto label
+      steps:
+      - name: issuelabeler
+        continue-on-error: true
+        uses: docker://maxandersen/jbang-issuelabeler
+        env:
+          GITHUB_TOKEN: ${{ secrets.LABELER_TOKEN }}
```
machi1990, line 15: We'll need to create this secret. 

It is a token to access github api. When I was testing it, I used my personal token https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token

machi1990: I've added my personal token, that will get us moving once this is merged. It can be updated once we have project's account.

Issue #1767, opened 2020-07-28
Directives on Object Types are removed in generated schema


lastmjs:
I am on graphback `0.15.0-alpha1`. Object type directives in my model schema are removed by Graphback in the generated schema.

Model schema:
```graphql
type Users @directive1 @directive2 {
  id: ID!
}
```

This is what is currently happening with the generated schema:
```graphql
type Users {
  id: ID!
}
```
This is what should happen with the generated schema:
```graphql
type Users @directive1 @directive2 {
  id: ID!
}
```

craicoverflow:
Thanks for reporting this issue. It appears that a dependency of Graphback (graphql-compose) is removing the directive from the object. We reported a similar issue in the past on field definitions, which was fixed (and still works). We will follow up on this.

machi1990:
Fixed by https://github.com/aerogear/graphback/pull/1810. I'll add a test case tomorrow and hopefully do a release.

machi1990:
@lastmjs we have droppped https://github.com/aerogear/graphback/releases/tag/0.16.0-alpha1, which include a fix for this issue. Let me know if that fixes the issue on your side. Thanks!

wtrocki:
Awesome work @machi1990

machi1990:
@lastmjs we've dropped another [0.16.0-alpha3](https://github.com/aerogear/graphback/releases/tag/0.16.0-alpha3) which includes a fix for this issue. Mind checking if this release effectively fix the issue for you? Thanks!

PR #1792, opened 2020-07-31
fix: ensure field directive is mapped to schema on relationship fields

### LINKED ISSUE #1772: Directives on field definitions are removed if there is a @oneToOne annotation
I'm not sure how general the problem is here, but in my generated schema for one of my fields with the following annotation:

```graphql
    """
    @oneToOne(key: 'company_id')
    """
```
the schema directives for that field in my model schema are removed in the generated schema.

craicoverflow:
Fixes #1772 

This fixes an issue where directives on relationship fields in the model were not added to the generated schema. because Graphback modifies the relationship fields to add extra metadata/filter arguments etc. 

Graphback was removing the old field and adding a new one with the same config + some modifications but we were losing the astNode from the field. I found that the better thing to do was to use [extendField](https://graphql-compose.github.io/docs/api/InterfaceTypeComposer.html#extendfield) from GraphQL Compose.

## Verification

I have added a test case in this PR. If you want to try it yourself, using the following model:

```graphql
directive @test on FIELD_DEFINITION
  """ @model """
type Note {
  id: ID!
  title: String!
  description: String
  """
  @oneToMany(field: 'note')
  """
  comments: [Comment]! @test
  """@oneToOne"""
  comment: Comment @test
}
""" @model """
type Comment {
  id: ID!
  text: String
  description: String
  note: Note! @test
}
```


packages/graphback-codegen-schema/tests/GraphQLSchemaCreatorTest.ts
```diff
@@ -229,3 +229,50 @@ test('Creates CRUD resolvers for models', async () => {
...
+  // Note.comment field
+  expect(comment.astNode?.directives).toHaveLength(1)
+  expect(comment.astNode?.directives[0].name.value).toBe('test')
+
+  const commentType = schema.getType('Comment') as GraphQLObjectType
+  const { note } = commentType.getFields()
+
+  // Note.comments field
+  expect(note.astNode?.directives).toHaveLength(1)
+  expect(note.astNode?.directives[0].name.value).toBe('test')
```
machi1990, line 277: Thanks for adding these tests. 

Do you think we can have a snapshot test too?

machi1990:
Could the fix be applied for this issue https://github.com/aerogear/graphback/issues/1767 or it is entirely has to do with grapql-compose?

packages/graphback-codegen-schema/tests/GraphQLSchemaCreatorTest.ts
craicoverflow: Sure, pushed now :wink:

craicoverflow:
> Could the fix be applied for this issue #1767 too or it is entirely has to do with grapql-compose?

This does not fix that issue unfortunately. I think that is entirely GraphQL Compose but I will do some investigating separate from this PR.

wtrocki:
Ouch.. I merged wrong PR! Sorry @craicoverflow

machi1990:
> > Could the fix be applied for this issue #1767 too or it is entirely has to do with grapql-compose?
> 
> This does not fix that issue unfortunately. I think that is entirely GraphQL Compose but I will do some investigating separate from this PR.

Okay I see. It was more a question for my own understanding. Thanks for answering.

PR #1818, opened 2020-08-05
perf: review relationship builder to avoid parsing annotation several times

machi1990:
With this, I observed a difference in performance for schema with a lot of business model and relationships between them.

Opening as draft so that it can be tested in a comparative base on a business model generated with:
```js
function genSchema(length){
    let result = `
    type Book {
        id: ID!
        name: String!
        numPages: Int!
      }
    `;

    for(let i = 1; i <= length; i++){
        result += `
        """ @model """
        type Author${i} {
          id: ID!
          name: String!
          md5: String!
          company: String!
          books: [Book]!
        }
        
        """ @model """
        type Note${i} {
          id: ID!
          name: String!
          md5: String!
          company: String!
          """
          @oneToMany(field: 'note') 
          """
          comments: [Comment${i}]!
        }

        """ @model """
        type Comment${i} {
          id: ID!
          name: String!
          md5: String!
          """
          @oneToOne
          """
          metadata: CommentMetadata${i}
        }
        
        """ @model """
        type CommentMetadata${i} {
          id: ID!
          metadata: String!
        }
        `;
    }
    return result;
}
```

You can run two tests, one with 10 models and another one with 100 models.


packages/graphback-codegen-schema/tests/__snapshots__/GraphQLSchemaCreatorTest.ts.snap
```diff
@@ -285,9 +288,10 @@ type Fix {
   id: ID!
 
   \\"\\"\\"
-  Added as part of fixing https://github.com/aerogear/graphback/issues/1452 showing that even if filtering is disabled on a model level,
+  @oneToMany(field: 'bugFix', key: 'bugFixId') 
+   Added as part of fixing https://github.com/aerogear/graphback/issues/1452 showing that even if filtering is disabled on a model level,
   we should have filter type created as it can be used on the relationship level without throwing an error.
-  @oneToMany(field: 'bugFix', key: 'bugFixId')
+  @oneToMany(field: 'bugFix')
```
machi1990, line 294: I'll prefer not to repeat this annotation but It is safe to have this year, as graphback will recognise the relationship annotation in https://github.com/aerogear/graphback/pull/1818/files#diff-3a7edbeb794d4f75d443877f36311dd6R291

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
```diff
@@ -139,109 +149,113 @@ export class RelationshipMetadataBuilder {
...
 
     // skip - relationship with key mapping will be generated at a later stage
-    if (!metadata.key) {
+    if (!oneToOneAnnotation.key) {
       return;
     }
 
     const relationType = getNamedType(field.type) as GraphQLObjectType;
 
-    const columnField = metadata.key || transformForeignKeyName(field.name);
```
craicoverflow, line 236: This enables shorter annotations by assuming a default foreign key column name.

packages/graphback-core/src/relationships/relationshipHelpers.ts
```diff
@@ -45,59 +45,6 @@ export function isOneToManyField(field: GraphQLField<any, any>): boolean {
...
-    const parsedAnnotation = parseRelationshipAnnotation(description);
-
-    if (parsedAnnotation && parsedAnnotation.key) {
-      descriptionLines.push(relationshipDescription);
-      break;
-    }
-  }
-
-  return descriptionLines.join('\n').trim();
-}
```
craicoverflow, line 99: I think this function can be optimized for sure. It is needed to ensure the annotation is merged with the field description (that is not an annotation)

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
machi1990: Yes, except that because of the following conditions: 
 - https://github.com/aerogear/graphback/blob/6f80afa191444414d8d89d14a8198f2c3ae4d12a/packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts#L246
- https://github.com/aerogear/graphback/blob/6f80afa191444414d8d89d14a8198f2c3ae4d12a/packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts#L224
- https://github.com/aerogear/graphback/blob/6f80afa191444414d8d89d14a8198f2c3ae4d12a/packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts#L202

The second expression will never be evaluated.

packages/graphback-core/src/relationships/relationshipHelpers.ts
machi1990: Yes, a first attempt of that is:

https://github.com/aerogear/graphback/blob/6f80afa191444414d8d89d14a8198f2c3ae4d12a/packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts#L151-L196

We make sure that the annotation with the needed arguments i.e (key and field) are there. 

This comes with one problem, the original annotation expression is never stripped out: see https://github.com/aerogear/graphback/pull/1818#discussion_r465598228

craicoverflow:
`> You can run two tests, one with 10 models and another one with 100 models.`

`genSchema(100)` generates 400 types instead of 400 types, and `genSchema(10)` generates 40, so the difference is considerably greater than the original benchmark models.

machi1990:
> `> You can run two tests, one with 10 models and another one with 100 models.`
> 
> `genSchema(100)` generates 400 types instead of 400 types, and `genSchema(10)` generates 40, so the difference is considerably greater than the original benchmark models.

Thanks for checking. 

Yes, I've added relationships to these models so that we can use the generated schema to compare this change. You can generate any number of models needed.  

The idea here is to compile this change, and compare how benchmarks results with and without this change. 

1. Run the benchmark using the genSchema script shared in the description
2. Pull in this change, compile the code and replace related fils in `node_modules`
3. Run the benchmark again
4. Compare results in step 3 and 1 to see if the change has any effects for you.  

For me step (4) seems to indicate an increase in startup time.

craicoverflow:
I've compiled my results into [this gist](https://gist.github.com/craicoverflow/9af7c7c04246f0d0fcebb7217df37fb3) to make it easier to compare. Yes, it looks as though this branch has actually increased startup time but slightly increased throughput.

wtrocki:
>  generates 400 types instead of 400 types

😆

craicoverflow:
I see you re-requested my review @machi1990 - is the motive here to run the benchmark tests or is this something you are looking to get merged to master?

machi1990:
> I see you re-requested my review @machi1990 - is the motive here to run the benchmark tests or is this something you are looking to get merged to master?

I re-ran startup benchmarks again and I had positive results across several runs as compared to current master. It is something I am looking to merge into master.

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
```diff
@@ -280,39 +289,32 @@ export class RelationshipMetadataBuilder {
       throw new Error(`${modelName}.${field.name} relationship field maps to ${relationModelType.name}.${relationField.name} (${relationFieldBaseType.name} type) which should be ${modelName} type.`);
     }
 
-    const manyToOneMetadata = parseRelationshipAnnotation(relationField.description);
-
-    if (manyToOneMetadata && oneToManyMetadata.key && manyToOneMetadata.key && oneToManyMetadata.key !== manyToOneMetadata.key) {
+    if (corresspondingManyToOneMetadata && oneToManyMetadata.key && corresspondingManyToOneMetadata.key && oneToManyMetadata.key !== corresspondingManyToOneMetadata.key) {
```
craicoverflow, line 292:
```suggestion
    if (oneToManyMetadata?.key !== corresspondingManyToOneMetadata?.key) {
```

packages/graphback-core/src/relationships/relationshipHelpers.ts
craicoverflow: I can see from the snapshots that this problem still exists. Are there plans to fix it?

packages/graphback-core/src/relationships/RelationshipMetadataBuilder.ts
machi1990: Good catch. Thanks!

craicoverflow:
@machi1990 I was wondering about https://github.com/aerogear/graphback/pull/1818#discussion_r466897189 before proceeding as it looks like a bug to me.

packages/graphback-core/src/relationships/relationshipHelpers.ts
machi1990: Ah sorry, I missed this. 

I do not think this a big issue and worth the effort to be honest.

machi1990: Or `graphql-metadata` would refuse this when parsing the `annotation` since it is declared twice?

craicoverflow: I am not sure what happens if you use `graphql-metadata` to parse the annotation (which one does it choose). We want the one with `key`.

If it handles that correctly, yes I get that it is not a massive issue, however the final schema could look a bit confusing. I will do a verification again.

machi1990: `graphql-metadata` would use the first annotation hence the reason behind it (the generated annotation) coming first before the `oldDescription`. It is more of the output schema. To make it clear, we could a `xxx generated by graphback`. 
 
Also, when the annotation does contain the `key` field, this is  block will be ignored and the original description will be returned.

machi1990:
@craicoverflow thanks for the verification and review. Let's get this in.

PR #1833, opened 2020-08-07
feat(datasync): add conflicts by version

ssd71:
Rather hopeful about this one :crossed_fingers: 

```typescript
const { typeDefs, resolvers, contextCreator } = createDataSync(modelDefs, db, { Comment: { enabled: true, conflictResolution: serverSideConflictResolution }});
```

## TODOS

- [x] Agreement on if this is how it should be
- [ ] Delete Conflicts?
- [ ] Tests
- [ ] Docs

### DIFF
--- a/integration/tests/runtime-workflow-datasync-conflicts.ts
+++ b/integration/tests/runtime-workflow-datasync-conflicts.ts
@@ -40,7 +40,7 @@ beforeAll(async () => {
     mkdirSync("./output-datasync-conflict-mongo");
     mkdirSync("./output-datasync-conflict-mongo/client")
 
-    mongoClient = new MongoClient('mongodb://mongodb:mongo@localhost:27018/users?authSource=admin', { useUnifiedTopology: true });
+    mongoClient = new MongoClient('mongodb://mongodb:mongo@localhost:27017/users?authSource=admin', { useUnifiedTopology: true });
     await mongoClient.connect();
     db = mongoClient.db('users');
     graphbackApi = createDataSyncAPI(modelText, { db, dataSyncConflictMap: { Note: { enabled: true } }, graphbackAPIConfig: {
--- a/integration/tests/runtime-workflow-datasync-conflicts.ts
+++ b/integration/tests/runtime-workflow-datasync-conflicts.ts
@@ -26,6 +26,22 @@ let graphbackApi: GraphbackAPI;
 
 let documents: DocumentNode;
 
+const SYNC_NOTES = gql`
+query syncNotes($lastSync: GraphbackTimestamp!, $filter: NoteFilter, $limit: Int) {
+    syncNotes(lastSync: $lastSync, filter: $filter, limit: $limit) {
+        items {
+          _id
+          title
+          description
+          ${DataSyncFieldNames.lastUpdatedAt}
+          ${DataSyncFieldNames.deleted}
+          ${DataSyncFieldNames.version}
+        },
+        lastSync
+    }
+}
+`
+
 const notesId = [];
 const commentId = [];
 const metadataId = [];
@@ -44,10 +60,10 @@ beforeAll(async () => {
     await mongoClient.connect();
     db = mongoClient.db('users');
     graphbackApi = createDataSyncAPI(modelText, { db, dataSyncConflictMap: { Note: { enabled: true } }, graphbackAPIConfig: {
-        plugins: [
-            new SchemaCRUDPlugin({ outputPath: "./output-datasync-conflict-mongo/schema/schema.graphql" }),
-            new ClientCRUDPlugin({ outputFile: './output-datasync-conflict-mongo/client/graphback.graphql' }),
-        ]
+      plugins: [
+        new SchemaCRUDPlugin({ outputPath: "./output-datasync-conflict-mongo/schema/schema.graphql" }),
+        new ClientCRUDPlugin({ outputFile: './output-datasync-conflict-mongo/client/graphback.graphql' }),
+      ]
     } });
     
     await seedDatabase(db);
@@ -106,9 +122,9 @@ async function seedDatabase(db: Db) {
     const { ops } = await db.collection("note").insertOne(note);
     notesId.push(ops[0]._id);
     await db.collection(getDeltaTableName('note')).insert({
-        docId: ops[0]._id,
-        [DataSyncFieldNames.version]: 1,
-        document: ops[0]
+      docId: ops[0]._id,
+      [DataSyncFieldNames.version]: 1,
+      document: ops[0]
     })
   }
 }
@@ -143,6 +159,103 @@ test('Client side should win when conflict occurs', async () => {
   expect(secondUpdateResponse.data.updateNote.title).toEqual(secondUpdateTitle);
 })
 
+test('force deletes on delete conflicts', async () => {
+  const noteTitle = 'Note C';
+  const noteDescription = "Note C Description";
+  const createResponse = await createNote(client, { title: noteTitle, description: noteDescription });
+
+  expect(createResponse.data?.createNote).toBeDefined();
+  expect(createResponse.data.createNote.title).toEqual(noteTitle);
+
+  const {_id} = createResponse.data.createNote;
+  
+  const updatedTitle = "Note C v2";
+  await updateNote({ _id, title: updatedTitle, [DataSyncFieldNames.version]: 1 }, client);
+
+  await deleteNote(client, { _id, [DataSyncFieldNames.version]: 1 })
+
+  const { data, errors } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
+      title: {
+        eq: updatedTitle
+      }
+    }
+  }});
+  console.log(errors);
+  expect(data.syncNotes.items).toBeDefined();
+  expect(data.syncNotes.items[0]._id).toEqual(_id);
+  expect(data.syncNotes.items[0][DataSyncFieldNames.deleted]).toEqual(true);
+});
+
+test('deletes document when no conflicts occur', async () => {
+  const noteTitle = 'Note D';
+  const noteDescription = "Note D Description";
+  const createResponse = await createNote(client, { title: noteTitle, description: noteDescription });
+
+  expect(createResponse.data?.createNote).toBeDefined();
+  expect(createResponse.data.createNote.title).toEqual(noteTitle);
+
+  const { _id, [DataSyncFieldNames.version]: version } = createResponse.data.createNote;
+  
+  const updatedTitle = "Note D v2";
+  const updateResponse = await updateNote({ _id, title: updatedTitle, [DataSyncFieldNames.version]: version }, client);
+
+  expect(updateResponse.data?.updateNote).toBeDefined();
+
+  const {[DataSyncFieldNames.version]: updatedVersion} = updateResponse.data.updateNote;
+
+  await deleteNote(client, { _id, [DataSyncFieldNames.version]: updatedVersion })
+
+  const { data } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
+      title: {
+        eq: updatedTitle
+      }
+    }
+  }});
+
+  expect(data.syncNotes.items).toBeDefined();
+  expect(data.syncNotes.items[0]._id).toEqual(_id);
+  expect(data.syncNotes.items[0][DataSyncFieldNames.deleted]).toEqual(true);
+
+});
+
+test('restores document when server-side deleted', async () => {
+  const noteTitle = 'Note E';
+  const noteDescription = "Note E Description";
+  const createResponse = await createNote(client, { title: noteTitle, description: noteDescription });
+
+  expect(createResponse.data?.createNote).toBeDefined();
+  expect(createResponse.data.createNote.title).toEqual(noteTitle);
+
+  const { _id, [DataSyncFieldNames.version]: version } = createResponse.data.createNote;
+  
+  const deleteResponse = await deleteNote(client, { _id, [DataSyncFieldNames.version]: version })
+  expect(deleteResponse.data.deleteNote).toBeDefined();
+
+  const updatedTitle = "Note E v2";
+  const updateResponse = await updateNote({ _id, title: updatedTitle, [DataSyncFieldNames.version]: version }, client);
+
+  expect(updateResponse.data?.updateNote).toBeDefined();
+
+  const { data } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
+      title: {
+        eq: updatedTitle
+      }
+    }
+  }});
+
+  expect(data.syncNotes.items).toBeDefined();
+  expect(data.syncNotes.items[0]._id).toEqual(_id);
+  expect(data.syncNotes.items[0].title).toEqual(updatedTitle);
+  expect(data.syncNotes.items[0][DataSyncFieldNames.deleted]).toEqual(false);
+
+});
+
 
 
 async function updateNote(input: any, client: ApolloServerTestClient) {
@@ -164,11 +277,11 @@ async function createNote(client: ApolloServerTestClient, input: any) {
   return response;
 }
 
-async function deleteNote(client: ApolloServerTestClient, _id:  string | number) {
+async function deleteNote(client: ApolloServerTestClient, input: any) {
   const response = await client.mutate({
     operationName: "deleteNote",
     mutation: documents,
-    variables: { input: { _id } }
+    variables: { input }
   });
 
   return response;
--- a/integration/tests/runtime-workflow-datasync-conflicts.ts
+++ b/integration/tests/runtime-workflow-datasync-conflicts.ts
@@ -1,0 +1,175 @@
+/* eslint-disable max-lines */
+/* eslint-disable no-null/no-null */
+/* eslint-disable no-shadow */
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable import/no-extraneous-dependencies */
+import { mkdirSync, readFileSync, rmdirSync } from 'fs';
+import * as path from 'path';
+import { ApolloServer, gql } from "apollo-server";
+import { createTestClient, ApolloServerTestClient } from 'apollo-server-testing';
+import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader'
+import { loadDocuments } from '@graphql-tools/load'
+import { buildGraphbackAPI, GraphbackAPI } from "graphback";
+import { DocumentNode } from 'graphql';
+import { MongoClient, Db, ObjectID } from 'mongodb';
+import { PubSub } from "graphql-subscriptions";
+import { getDeltaTableName, DataSyncFieldNames, createDataSyncAPI } from "../../packages/graphback-datasync";
+import { SchemaCRUDPlugin } from '../../packages/graphback-codegen-schema';
+import { ClientCRUDPlugin } from '../../packages/graphback-codegen-client';
+
+/** global config */
+let db: Db;
+let mongoClient: MongoClient;
+let server: ApolloServer;
+let client: ApolloServerTestClient;
+let graphbackApi: GraphbackAPI;
+
+let documents: DocumentNode;
+
+const notesId = [];
+const commentId = [];
+const metadataId = [];
+
+const modelText = readFileSync("./datasync-mongodb-model.graphql").toString();
+const startTS = new Date(1596622318448);
+const updatedTS2 = new Date(1596622363886);
+const objectId = new ObjectID("507f191e810c19729de860ea");
+
+beforeAll(async () => {
+  try {
+    mkdirSync("./output-datasync-conflict-mongo");
+    mkdirSync("./output-datasync-conflict-mongo/client")
+
+    mongoClient = new MongoClient('mongodb://mongodb:mongo@localhost:27018/users?authSource=admin', { useUnifiedTopology: true });
+    await mongoClient.connect();
+    db = mongoClient.db('users');
+    graphbackApi = createDataSyncAPI(modelText, { db, dataSyncConflictMap: { Note: { enabled: true } }, graphbackAPIConfig: {
+        plugins: [
+            new SchemaCRUDPlugin({ outputPath: "./output-datasync-conflict-mongo/schema/schema.graphql" }),
+            new ClientCRUDPlugin({ outputFile: './output-datasync-conflict-mongo/client/graphback.graphql' }),
+        ]
+    } });
+    
+    await seedDatabase(db);
+    
+    const source = await loadDocuments(path.resolve(`./output-datasync-conflict-mongo/client/graphback.graphql`), {
+      loaders: [
+        new GraphQLFileLoader()
+      ]
+    });
+    documents = source[0].document;
+  } catch (e) {
+    throw e;
+  }
+})
+
+beforeEach(() => {
+  const { typeDefs, resolvers, contextCreator } = graphbackApi;
+  server = new ApolloServer({
+    typeDefs,
+    resolvers,
+    context: contextCreator
+  });
+  
+  client = createTestClient(server);
+})
+
+afterEach(() => server.stop())
+
+afterAll(async () => {
+  rmdirSync(path.resolve('./output-datasync-conflict-mongo'), { recursive: true });
+  const dropCollections = ["note", getDeltaTableName("note")].map((name: string) => db.dropCollection(name));
+  await Promise.all(dropCollections);
+
+  return mongoClient.close();
+});
+
+async function seedDatabase(db: Db) {
+  const notes = [
+    {
+      title: 'Note A',
+      description: 'Note A Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false,
+      [DataSyncFieldNames.version]: 1
+    },
+    {
+      title: 'Note B',
+      description: 'Note B Description',
+      [DataSyncFieldNames.lastUpdatedAt]: startTS.valueOf(),
+      [DataSyncFieldNames.deleted]: false,
+      [DataSyncFieldNames.version]: 1
+    }
+  ]
+
+  for (const note of notes) {
+    const { ops } = await db.collection("note").insertOne(note);
+    notesId.push(ops[0]._id);
+    await db.collection(getDeltaTableName('note')).insert({
+        docId: ops[0]._id,
+        [DataSyncFieldNames.version]: 1,
+        document: ops[0]
+    })
+  }
+}
+
+
+test('Should update successfully when no conflict', async () => {
+  const updatedTitle = 'Note A v2';
+  const firstUpdateResponse = await updateNote({ _id: notesId[0], title: updatedTitle, [DataSyncFieldNames.version]: 1 }, client);
+
+  expect(firstUpdateResponse.data?.updateNote).toBeDefined();
+  expect(firstUpdateResponse.data.updateNote.title).toEqual(updatedTitle);
+
+  const updatedDescription = "Note A Description v2";
+  const secondUpdateResponse = await updateNote({ _id: notesId[0], description: updatedDescription, [DataSyncFieldNames.version]: 1 }, client);
+
+  expect(secondUpdateResponse.data?.updateNote).toBeDefined();
+  expect(secondUpdateResponse.data.updateNote.description).toEqual(updatedDescription);
+})
+
+test('Client side should win when conflict occurs', async () => {
+  const updatedTitle = 'Note B v2';
+  const firstUpdateResponse = await updateNote({ _id: notesId[0], title: updatedTitle, [DataSyncFieldNames.version]: 1 }, client);
+
+  expect(firstUpdateResponse.data?.updateNote).toBeDefined();
+  expect(firstUpdateResponse.data.updateNote.title).toEqual(updatedTitle);
+
+  
+  const secondUpdateTitle = "Note B v3";
+  const secondUpdateResponse = await updateNote({ _id: notesId[0], title: secondUpdateTitle, [DataSyncFieldNames.version]: 1 }, client);
+
+  expect(secondUpdateResponse.data?.updateNote).toBeDefined();
+  expect(secondUpdateResponse.data.updateNote.title).toEqual(secondUpdateTitle);
+})
+
+
+
+async function updateNote(input: any, client: ApolloServerTestClient) {
+  const response = await client.mutate({
+    operationName: "updateNote",
+    mutation: documents,
+    variables: { input }
+  });
+
+  return response;
+}
+
+async function createNote(client: ApolloServerTestClient, input: any) {
+  const response = await client.mutate({
+    operationName: "createNote",
+    mutation: documents, variables: { input }
+  });
+
+  return response;
+}
+
+async function deleteNote(client: ApolloServerTestClient, _id:  string | number) {
+  const response = await client.mutate({
+    operationName: "deleteNote",
+    mutation: documents,
+    variables: { input: { _id } }
+  });
+
+  return response;
+}
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -4,12 +4,14 @@ import { SchemaComposer, ObjectTypeComposerFieldConfig, ObjectTypeComposer } fro
 import { IResolvers, IFieldResolver } from '@graphql-tools/utils'
 
 import { getDeltaType, getDeltaListType, getDeltaQuery } from "./deltaMappingHelper";
-import { isDataSyncService, isDataSyncModel, DataSyncFieldNames } from "./util";
+import { isDataSyncService, isDataSyncModel, DataSyncFieldNames, DataSyncModelConflictConfig } from "./util";
 
 export const DATASYNC_PLUGIN_NAME = "DataSyncPlugin";
 
 export interface DataSyncPluginConfig {
-  useVersion?: boolean
+  modelConfigMap: {
+    [modelName: string]: DataSyncModelConflictConfig
+  }
 }
 
 /**
@@ -21,14 +23,12 @@ export interface DataSyncPluginConfig {
 export class DataSyncPlugin extends GraphbackPlugin {
   protected config: DataSyncPluginConfig;
 
-  public constructor(config: DataSyncPluginConfig = {
-    useVersion: false
-  }) {
+  public constructor(config?: DataSyncPluginConfig) {
     super()
-    this.config = {
-      useVersion: false,
+    this.config = { 
+      modelConfigMap: {},
       ...config
-    }
+    };
   }
 
   public transformSchema(metadata: GraphbackCoreMetadata): GraphQLSchema {
@@ -132,7 +132,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
       }
     });
 
-    if (this.config.useVersion) {
+    const modelUsesVersion = !!this.config.modelConfigMap[model.graphqlType.name]?.enabled;
+    if (modelUsesVersion) {
       modelTC.addFields({
         [DataSyncFieldNames.version]: {
           type: GraphQLInt
@@ -145,7 +146,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
     // Add _version argument to UpdateInputType
     const updateInputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
-    if (this.config.useVersion && updateInputType) {
+    const modelUsesVersion = this.config.modelConfigMap[model.graphqlType.name]?.enabled;
+    if ((modelUsesVersion) && updateInputType) {
       updateInputType.addFields({
         [DataSyncFieldNames.version]: {
           type: GraphQLNonNull(GraphQLInt)
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -132,7 +132,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
       }
     });
 
-    const modelUsesVersion = !!this.config.modelConfigMap[model.graphqlType.name]?.enabled;
+    const modelUsesVersion = this.config.modelConfigMap[model.graphqlType.name]?.enabled;
     if (modelUsesVersion) {
       modelTC.addFields({
         [DataSyncFieldNames.version]: {
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -147,7 +147,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
     // Add _version argument to UpdateInputType
     const updateInputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
     const modelUsesVersion = this.config.modelConfigMap[model.graphqlType.name]?.enabled;
-    if ((modelUsesVersion) && updateInputType) {
+    if (modelUsesVersion && updateInputType) {
       updateInputType.addFields({
         [DataSyncFieldNames.version]: {
           type: GraphQLNonNull(GraphQLInt)
--- a/packages/graphback-datasync/src/deltaHelper.ts
+++ b/packages/graphback-datasync/src/deltaHelper.ts
@@ -1,0 +1,46 @@
+import { ModelDefinition, getTableName } from '@graphback/core';
+import { Db, Collection, ObjectId } from "mongodb";
+import { DataSyncFieldNames } from './util';
+
+export function getDeltaTableName(tableName: string) {
+  return `${tableName}Delta`
+}
+
+/**
+ * M
+ */
+export class MongoDeltaHelper {
+  protected db: Db;
+  protected collectionName: string;
+  public constructor(model: ModelDefinition, db: Db) {
+    this.db = db;
+    this.collectionName = getDeltaTableName(getTableName(model.graphqlType));
+  }
+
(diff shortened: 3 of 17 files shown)


packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -131,7 +128,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
       }
     });
 
-    if (this.config.useVersion) {
+    if (this.config?.modelUsesVersion[model.graphqlType.name] || false) {
```
wtrocki, line 131: move to variable

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+  protected collection: Collection;
+  public constructor(model: ModelDefinition, db: Db) {
+    this.collection = db.collection(getDeltaTableName(getTableName(model.graphqlType)));
+  }
+
+  public async insertDiff(updateDocument: any) {
+    const { _id,  ...delta } = updateDocument;
+    const version = delta[DataSyncFieldNames.version];
+    // eslint-disable-next-line @typescript-eslint/tslint/config
+    delete delta[DataSyncFieldNames.version];delete delta[DataSyncFieldNames.lastUpdatedAt];delete delta[DataSyncFieldNames.deleted];
```
wtrocki, line 22: Why deleting those fields... and why in single line :)

ssd71: Deleting so as to ensure only the fields changed by the user are kept in the diff

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+      docId: _id,
+      [DataSyncFieldNames.version]: version,
+      delta
+    }
+    console.log(diff);
+    await this.collection.insertOne(diff);
+  }
+
+  // returns final resolved update document
+  public async resolveConflictFromDelta(updateDocument: any): Promise<any> {
```
wtrocki, line 34: Resolving conflict algorithm for dummies.
```
1. Accept client side version form client and validate it.
2. Look for the delta table entry with id and version.
3. This entry will be our `base` for conflict resolution.
4. Fetch server side entry 
5. Calculate diff for server side entry
5. Perform conflict check (check if fields that were changed are mergeable based on the diffs)
6. if mergable perform merge
7. If not mergable (the same fields changed - perform predefined conflict resolution.
8. Call out resolution algorithm (user provided function or default ones we provide)
We can use `MostRecentWins strategy` - that is why do not remove last updated :)
9.  try to save result to database - conditial check if version did not changed since then
7. Return data to user.

Another condition - if items reached their time (TTL ;) we have no way to kick offix compatible error to user.

```

All of this logic is already implemented in:
https://github.com/aerogear/offix/tree/master/packages/offix-conflicts-client

See also offix conflict server. Obviously you are free to invent your own algorithm but based on what I see I will probably prefer us to copy paste the code that was written and used in production rather than try to invent new way :)

packages/graphback-datasync/src/helpers/createDataSync.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function createDataSync(model: string | GraphQLSchema, db: Db, dataSyncConfig: { [modelName: string]: DataSyncModelConflictConfig} = {}, graphbackConfig: DataSyncGraphbackConfig = {}): GraphbackAPI {
+
+  return buildGraphbackAPI(model, {
+    serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
+    ...graphbackConfig,
+    dataProviderCreator: createDataSyncConflictProviderCreator(db, dataSyncConfig),
+    plugins:[
+      ...graphbackConfig.plugins || [],
+      new DataSyncPlugin({ modelUsesVersion: Object.keys(dataSyncConfig).reduce((versionMap: any, modelName: string) => {
```
wtrocki, line 20: Separate variable?

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+  return `${tableName}Delta`
+}
+
+/**
+ * M
+ */
+export class MongoDeltaHelper {
+  protected collection: Collection;
+  public constructor(model: ModelDefinition, db: Db) {
+    this.collection = db.collection(getDeltaTableName(getTableName(model.graphqlType)));
```
wtrocki, line 15: Extract variable

packages/graphback-datasync/src/helpers/createDataSync.ts
wtrocki: Actually why we do not do this inside plugin. It feels wrong to do it here.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,96 @@
...
+    await this.deltaHelper?.insertDiff(result).catch((e: any) => { console.error(`Error in inserting delta: ${e}`) });
+
+
+    return result;
+  }
+
+  public async update(updateDocument: any, context: GraphbackContext): Promise<Type> {
+    const { _id } = updateDocument;
+
+    const MAX_RETRIES = 3;
```
wtrocki, line 50: Probably not the best place to setup this value.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,96 @@
...
+    const updateFilter = { _id, [DataSyncFieldNames.deleted]: false };
+
+    for(let i = 0; i < MAX_RETRIES; i++) {
+      const serverStateData = await this.db.collection(this.collectionName).findOne({ _id, [DataSyncFieldNames.deleted]: false });
+
+      if (serverStateData){
+
+        
+        if (updateDocument[DataSyncFieldNames.version] !== serverStateData[DataSyncFieldNames.version]) {
+          const conflict: ConflictStateMap = {
```
wtrocki, line 60: This doesn't mean conflict. 
It means only that versions are different. 
We need diffs to know if conflict occured.

Server data might have different fields than client as well.

packages/graphback-datasync/src/DataSyncPlugin.ts
craicoverflow, line 131:
```suggestion
    if (this.config?.modelUsesVersion[model.graphqlType.name]) {
```

Not sure if that is needed. Is the `|| false` in case the key does not exist in `modelUsesVersion`? In that case the condition would be `if (false)`.

packages/graphback-datasync/src/helpers/createDataSync.ts
```diff
@@ -0,0 +1,27 @@
...
+import { GraphbackAPIConfig, GraphbackAPI, buildGraphbackAPI, GraphbackConfig } from "graphback";
+import { PubSub } from 'graphql-subscriptions';
+import { createDataSyncMongoDbProvider, createDataSyncConflictProviderCreator } from '../providers';
+import { ConflictError, ConflictStateMap, DataSyncModelConflictConfig } from "../util";
+import { createDataSyncCRUDService } from '../services';
+import { DataSyncPlugin } from '../DataSyncPlugin';
+
+type DataSyncGraphbackConfig = Partial<GraphbackConfig>
+
+export function createDataSync(model: string | GraphQLSchema, db: Db, dataSyncConfig: { [modelName: string]: DataSyncModelConflictConfig} = {}, graphbackConfig: DataSyncGraphbackConfig = {}): GraphbackAPI {
```
craicoverflow, line 12:
```suggestion
export function createDataSyncAPI(model: string | GraphQLSchema, db: Db, dataSyncConfig: { [modelName: string]: DataSyncModelConflictConfig} = {}, graphbackConfig: DataSyncGraphbackConfig = {}): GraphbackAPI {
```

packages/graphback-datasync/src/helpers/createDataSync.ts
craicoverflow: +1 it should work the same if a user wants to use `buildGraphbackAPI` directly instead of `createDataSync`

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+      docId: _id,
+      [DataSyncFieldNames.version]: version,
+      document: updatedDocument
+    }
+    await this.db.collection(this.collectionName).insertOne(diff);
+  }
+
+  public async findBaseForConflicts(updateDocument: any) {
+    if (updateDocument._id === undefined || updateDocument[DataSyncFieldNames.version] === undefined) {
+      throw new Error('Both _id and version field are needed for finding base');
```
machi1990, line 34:
```suggestion
      throw new Error(`Both _id and ${DataSyncFieldNames.version} field are needed for finding base`);
```

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Db, Collection, ObjectId } from "mongodb";
+import { DataSyncFieldNames } from './util';
+
+export function getDeltaTableName(tableName: string) {
+  return `${tableName}Delta`
+}
+
+/**
+ * M
+ */
```
machi1990, line 11: comments looks incomplete here.. can we extend this a little bit. 

It is useful since we use `typedoc` to generate API reference (even though this is not part of something we'll want to expose but let's get the habit running).

packages/graphback-datasync/src/helpers/createDataSync.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ConflictError, ConflictMetadata, DataSyncModelConflictConfig } from "../util";
+import { createDataSyncCRUDService } from '../services';
+import { DataSyncPlugin } from '../DataSyncPlugin';
+
+type DataSyncGraphbackConfig = Partial<GraphbackConfig>
+
+export function createDataSyncAPI(model: string | GraphQLSchema, db: Db, dataSyncConfigMap: { [modelName: string]: DataSyncModelConflictConfig} = {}, graphbackConfig: DataSyncGraphbackConfig = {}): GraphbackAPI {
+
+  return buildGraphbackAPI(model, {
+    serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
```
machi1990, line 15: Are we going to work with a in memory message queue for now? Looks limiting.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
+import { ModelDefinition, GraphbackContext, TransformType, FieldTransform, NoDataError, GraphbackOperationType } from '@graphback/core';
+import { DataSyncModelConflictConfig, DataSyncFieldNames, ConflictMetadata, ConflictError } from '../util';
+import { MongoDeltaHelper } from '../deltaHelper';
+import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
+
+export const MAX_RETRIES = 3;
+
+/**
+ * Da
```
machi1990, line 9: Comments needs some meat.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
+import { ModelDefinition, GraphbackContext, TransformType, FieldTransform, NoDataError, GraphbackOperationType } from '@graphback/core';
+import { DataSyncModelConflictConfig, DataSyncFieldNames, ConflictMetadata, ConflictError } from '../util';
+import { MongoDeltaHelper } from '../deltaHelper';
+import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
+
+export const MAX_RETRIES = 3;
```
machi1990, line 6:
```suggestion
export const MAX_RETRIES = 3;
```

It will be useful to be able to configure this. 

@wtrocki what do you think?

It does not have to be done in this PR. We can open an issue to chat about it.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+    this.deltaHelper = new MongoDeltaHelper(model, client);
+
+  }
+
+  public async create(data: any, context: GraphbackContext): Promise<Type> {
+    const contextWithoutSelectedFields: GraphbackContext = { ...context }
+    contextWithoutSelectedFields.graphback.options.selectedFields = [];
```
machi1990, line 33: Why is this needed?

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+    throw new Error(`Cannot update ${this.collectionName}`);
+  }
+
+  public checkForConflict(clientData: any, base: any, serverData: any, operation: GraphbackOperationType): ConflictMetadata|undefined {
+    const ignoredKeys = [DataSyncFieldNames.lastUpdatedAt, DataSyncFieldNames.version, DataSyncFieldNames.deleted];
+
+    const clientDiff: any = {};
+    const serverDiff: any = {};
+
+    let isConflict = false;
```
machi1990, line 91:
```suggestion
    let conflictFound = false;
```

Just a nitpick, as it looks like we are checking if we have a conflict.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+    for (const key of Object.keys(clientData)) {
+      if (base[key] !== serverData[key]) {
+        if (!ignoredKeys.includes(key)) {
+          serverDiff[key] = serverData[key];
+          if (clientDiff[key]) {
+            isConflict = true;
+          }
+        }
+      }
+    }
```
machi1990, line 111: can this be done in one loop?

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -131,7 +131,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
       }
     });
 
-    if (this.config.useVersion) {
+    const modelUsesVersion = !!this.config.modelConfigMap[model.graphqlType.name]?.enabled;
```
craicoverflow, line 134: I don't think a double-negation is required here.

```suggestion
    const modelUsesVersion = !this.config.modelConfigMap[model.graphqlType.name]?.enabled;
```

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -144,7 +145,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
     // Add _version argument to UpdateInputType
     const updateInputType = schemaComposer.getITC(getInputTypeName(model.graphqlType.name, GraphbackOperationType.UPDATE));
-    if (this.config.useVersion && updateInputType) {
+    const modelUsesVersion = this.config.modelConfigMap[model.graphqlType.name]?.enabled;
+    if ((modelUsesVersion) && updateInputType) {
```
craicoverflow, line 149:
```suggestion
    if (modelUsesVersion && updateInputType) {
```

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
+import { ModelDefinition, getTableName } from '@graphback/core';
+import { Db, Collection, ObjectId } from "mongodb";
+import { DataSyncFieldNames } from './util';
+
+export function getDeltaTableName(tableName: string) {
+  return `${tableName}Delta`
+}
+
+/**
+ * M
```
craicoverflow, line 10: Missing/incomplete comments

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export class MongoDeltaHelper {
```
craicoverflow, line 12: > a helper class is used to assist in providing some functionality, which isn't the main goal of the application or class in which it is used. 

I think the name `MongoDeltaHelper` is too generic/non-specific. It is in fact a very integral class providing a very specific purpose. I see a helper class as being a collection of methods which do not necessarily work together or have common purpose.

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+  protected db: Db;
+  protected collectionName: string;
+  public constructor(model: ModelDefinition, db: Db) {
+    this.db = db;
+    this.collectionName = getDeltaTableName(getTableName(model.graphqlType));
+  }
+
+  public async insertDiff(updatedDocument: any) {
+    const version = updatedDocument[DataSyncFieldNames.version];
+    const { _id } = updatedDocument;
```
craicoverflow, line 22: I believe this could be a one-liner:

```suggestion
const { _id, [DataSyncFieldNames.version]: version } = updatedDocument;
```

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+    const diff: any = {
+      docId: _id,
+      [DataSyncFieldNames.version]: version,
+      document: updatedDocument
+    }
+    await this.db.collection(this.collectionName).insertOne(diff);
+  }
+
+  public async findBaseForConflicts(updateDocument: any) {
+    if (updateDocument._id === undefined || updateDocument[DataSyncFieldNames.version] === undefined) {
```
craicoverflow, line 33: Should have same effect:

```suggestion
    if (!updateDocument._id || !updateDocument[DataSyncFieldNames.version]) {
```

packages/graphback-datasync/src/deltaHelper.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export class MongoDeltaHelper {
+  protected db: Db;
+  protected collectionName: string;
+  public constructor(model: ModelDefinition, db: Db) {
+    this.db = db;
+    this.collectionName = getDeltaTableName(getTableName(model.graphqlType));
+  }
+
+  public async insertDiff(updatedDocument: any) {
```
craicoverflow, line 20: We could bring better type support to this class with generics:

```ts
export class MongoDeltaHelper<T = any>
...
public async insertDiff(updatedDocument: T) {
```

packages/graphback-datasync/src/helpers/createDataSync.ts
```diff
@@ -0,0 +1,23 @@
...
+import { GraphbackAPIConfig, GraphbackAPI, buildGraphbackAPI, GraphbackConfig } from "graphback";
+import { PubSub } from 'graphql-subscriptions';
+import { createDataSyncMongoDbProvider, createDataSyncConflictProviderCreator } from '../providers';
+import { ConflictError, ConflictMetadata, DataSyncModelConflictConfig } from "../util";
+import { createDataSyncCRUDService } from '../services';
+import { DataSyncPlugin } from '../DataSyncPlugin';
+
+type DataSyncGraphbackConfig = Partial<GraphbackConfig>
+
+export function createDataSyncAPI(model: string | GraphQLSchema, db: Db, dataSyncConfigMap: { [modelName: string]: DataSyncModelConflictConfig} = {}, graphbackConfig: DataSyncGraphbackConfig = {}): GraphbackAPI {
```
craicoverflow, line 12: I think 5 arguments is too complicated for the user API, what do you think about reducing it to max 2? 

`DataSyncGraphbackConfig` can copy the required config props (crud, plugins) from [`GraphbackAPIConfig`](https://github.com/aerogear/graphback/blob/ce7236aa84cf053d0e9090115ab5b2b95d3276e4/packages/graphback/src/buildGraphbackAPI.ts#L7). This means very minimal repetition, but greatly simplifies user experience, which is paramount IMO.

```ts
export function createDataSyncAPI(model: string | GraphQLSchema, config: DataSyncGraphbackConfig): GraphbackAPI {
```

Thoughts? Do you see any complications that might arise from this? cc @machi1990

packages/graphback-datasync/src/helpers/createDataSync.ts
```diff
@@ -0,0 +1,23 @@
+import { GraphQLSchema } from 'graphql';
+import { Db } from 'mongodb';
+import { GraphbackAPIConfig, GraphbackAPI, buildGraphbackAPI, GraphbackConfig } from "graphback";
+import { PubSub } from 'graphql-subscriptions';
+import { createDataSyncMongoDbProvider, createDataSyncConflictProviderCreator } from '../providers';
+import { ConflictError, ConflictMetadata, DataSyncModelConflictConfig } from "../util";
+import { createDataSyncCRUDService } from '../services';
+import { DataSyncPlugin } from '../DataSyncPlugin';
+
+type DataSyncGraphbackConfig = Partial<GraphbackConfig>
```
craicoverflow, line 10: `GraphbackConfig` is not used in runtime API (it is used for the `generate` command. I think you are looking for `GraphbackAPIConfig`?

packages/graphback-datasync/src/helpers/createDataSync.ts
craicoverflow, line 15: What if user wants to use their own PubSub library?

craicoverflow: `serviceCreator` should still be an (optional) option in the API config, as the user may want to customise/extend it.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
craicoverflow, line 9: Requires API docs

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+          throw new ConflictError(conflict)
+        }
+      }
+    }
+
+    this.deltaHelper = new MongoDeltaHelper(model, client);
+
+  }
+
+  public async create(data: any, context: GraphbackContext): Promise<Type> {
```
craicoverflow, line 31:
```suggestion
  public async create(data: Partial<Type>, context: GraphbackContext): Promise<Type> {
```

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const result = await super.create(data, contextWithoutSelectedFields);
+
+    await this.deltaHelper.insertDiff(result).catch((e: any) => { console.error(`Error in inserting delta: ${e}`) });
+
+
+    return result;
+  }
+
+  public async update(updateDocument: any, context: GraphbackContext): Promise<Type> {
```
craicoverflow, line 45:
```suggestion
  public async update(updateDocument: Partial<Type>, context: GraphbackContext): Promise<Type> {
```

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
craicoverflow, line 33: What is this for?

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+    const { _id } = updateDocument;
+
+    for(let i = 0; i < MAX_RETRIES; i++) {
+      
+      const serverData = await this.db.collection(this.collectionName).findOne({ _id, [DataSyncFieldNames.deleted]: false});
+      const base = await this.deltaHelper.findBaseForConflicts(updateDocument);
+      const clientData = updateDocument;
```
craicoverflow, line 52: Why use this separate variable which references `updateDocument`? I don't see `updateDocument` being modified after this (unless I missed it)

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      let resolvedUpdate = Object.assign({}, updateDocument);
+
+      const updateFilter: any = { _id, [DataSyncFieldNames.version]: serverData[DataSyncFieldNames.version] }
```
craicoverflow, line 56: Explicit type declarations are great, but when it is `any` there is no need to do it.

```suggestion
      const updateFilter = { _id, [DataSyncFieldNames.version]: serverData[DataSyncFieldNames.version] }
```

packages/graphback-datasync/src/providers/createDataSyncMongoDbProvider.ts
```diff
@@ -18,3 +19,18 @@ export function createDataSyncMongoDbProvider(db: Db): (...args: any[]) => Graph
     }
   }
 }
+
+export function createDataSyncConflictProviderCreator(db: Db, datasyncConfigMap: { [modelName: string]: DataSyncModelConflictConfig } = {}) : (...args: any[]) => GraphbackDataProvider {
```
craicoverflow, line 23: Needs API documentation with JSDoc

packages/graphback-datasync/src/deltaHelper.ts
wtrocki, line 20: Is it diff or snapshot?

wtrocki, line 20: How this will work if I:

Update just single field (userName) in the user object that has 10 fields?

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
wtrocki: Yep.. Not sure why my comments are gone.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
wtrocki:
```suggestion
  public async create(data: Omit<Type, '_id'>, context: GraphbackContext): Promise<Type> {
```

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
wtrocki, line 33: What is the purpose of this?

wtrocki:
This looks like way to go but it seems incomplete to me. We need to have this runnable with the example app itself.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
craicoverflow: Or `Partial<Omit<Type, '_id'>>`?

wtrocki: Now sure. Do you mean Create will not push every field ?

craicoverflow: Actually I was thinking about optional fields but this is already handled since the fields are optional it is still technically a complete Type.

craicoverflow: //ignoreme

packages/graphback-datasync/src/DataSyncPlugin.ts
ssd71: Hmm wouldn't that make the expression true when the rest of the statement is undefined?

packages/graphback-datasync/src/deltaHelper.ts
ssd71: it is snapshot

ssd71: I am using the updated document to add the diff:
https://github.com/aerogear/graphback/blob/601fe9f40cadf9ee5617073f491d05052d6ed9e5/packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts#L74-L76
Should I change that?

wtrocki: No. just  my bad. All good here

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
ssd71: @machi1990 @craicoverflow @wtrocki I am using it to get the complete document instead of just the selected fields. This is to ensure that the complete document is stored in the delta table:
https://github.com/aerogear/graphback/blob/601fe9f40cadf9ee5617073f491d05052d6ed9e5/packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts#L37-L39

Perhaps we do not need a delta table entry for the `create` operation, wdyt :thinking:

ssd71:
Thanks for the detailed review guys! :D

packages/graphback-datasync/src/DataSyncPlugin.ts
craicoverflow: My bad, what I mean to say I don't think there is a need for **any** negation here.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
craicoverflow: Yeah you are right to do this - I was thinking that all fields are fetched anyway, but only if subscriptions are enabled for that mutation type.

However, I think you can just modify the context directly, instead of cloning it to remove the selectedFields. There should be one context per request, and it has a short lifetime so modifying it directly should be fine.

craicoverflow: See https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/runtime/CRUDService.ts#L43

ssd71:
> This looks like way to go but it seems incomplete to me. We need to have this runnable with the example app itself.

@wtrocki By example app do you mean the template or  datasync-starter

wtrocki:
Yes

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
machi1990: Opened https://github.com/aerogear/graphback/issues/1856 for this.

ssd71:
Quick question: Should I do tests/docs in this PR as well or another one?

craicoverflow:
Can this be taken out of draft?

craicoverflow:
> Quick question: Should I do tests/docs in this PR as well or another one?

Tests - yes
Docs - preferable but not mandatory since it depends on when we release.

wtrocki:
Couple questions:
1. Do sample app (in templates) work?
2. What will be the default conflict resolution mechanism and would it be possible to use it for customization
3. Do we support deletes?

Ping me if answer on those is yes.

ssd71:
@wtrocki 

1. Template works and is configured to use Server-side conflict resolution
2. Default conflict resolution strategy when conflict is enabled is to throw an offix compatible conflict error when a conflict is detected, it is not available for customisation but custom conflict resolution can be implemented by using the `ConflictResolutionStrategy` interface
3. No, as it wasn't discussed, should I open issue about it?

wtrocki:
> Template works and is configured to use Server-side conflict resolution
> Default conflict resolution strategy when conflict is enabled is to throw an offix compatible conflict error when a conflict is detected, it is not available for customisation but custom conflict resolution can be implemented by using the ConflictResolutionStrategy interface

This makes no sense to me - we have enabled server side conflict resolution but we are throwing error to user to do client side resolution. 

> No, as it wasn't discussed, should I open issue about it?

Do you have any questions? We kinda need this in offix.

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,35 @@ export const DataSyncFieldNames = {
...
+
+export interface DataSyncModelConfigMap {
+  [modelName: string]: DataSyncModelConflictConfig
+}
+
+export interface ConflictResolutionStrategy {
+  resolveUpdate(conflict: ConflictMetadata): any
+}
+
+export const ServerSideConflictResolution: ConflictResolutionStrategy = {
```
wtrocki, line 63: I would recomend to provide names for strategies. ServerFirst, LastUpdatedFirst etc.

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,35 @@ export const DataSyncFieldNames = {
...
+  resolveUpdate(conflict: ConflictMetadata): any {
+    const { serverDiff, clientDiff } = conflict;
+
+    const resolved: any = {};
+
+    Object.keys(clientDiff).forEach((fieldName: string) => {
```
wtrocki, line 69: I cannot confirm if this algorithm works. Would be possible to explain motivation here.

Conflict detected by version doesn't mean actual conflict. Our implementation should know if we have conflict or mergability. Interface should clearly represent that.

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,35 @@ export const DataSyncFieldNames = {
...
+export interface DataSyncModelConfigMap {
+  [modelName: string]: DataSyncModelConflictConfig
+}
+
+export interface ConflictResolutionStrategy {
+  resolveUpdate(conflict: ConflictMetadata): any
+}
+
+export const ServerSideConflictResolution: ConflictResolutionStrategy = {
+  resolveUpdate(conflict: ConflictMetadata): any {
```
wtrocki, line 64: resolve would be more appropriate. Resolution for deletion and update doesn't matter at all - we simply getting different diffs and.. in deletion conflicts is obviously guaranteed but that is not something we need to be bothered.

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,35 @@ export const DataSyncFieldNames = {
...
+    const { serverDiff, clientDiff } = conflict;
+
+    const resolved: any = {};
+
+    Object.keys(clientDiff).forEach((fieldName: string) => {
+      if (!serverDiff[fieldName]) {
```
wtrocki, line 70: We really need to have tests for this part. It is clearly from running this code that with this strategy we going to lose some data. 

Maybe you wanted to use full server snapshot and then apply client side diff on top of it? 
This would be something like client side first (without checking if conflict actually occurred. 

Second approach will be to merge automatically what is mergable and pass to conflictResolutionStrategy only fields that are conflicted - I have done that before but do not like as it limits flexibility of having full view of the doc.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+  protected deltaSource: MongoDeltaSource
+
+  public constructor(model: ModelDefinition, client: any, dataSyncConflictConfig: DataSyncModelConflictConfig) {
+    super(model, client);
+    this.conflictConfig = dataSyncConflictConfig;
+
+    if (!this.conflictConfig.conflictResolution?.resolveUpdate) {
+      this.conflictConfig.conflictResolution = {
+        resolveUpdate(conflict: ConflictMetadata) {
+          throw new ConflictError(conflict)
```
wtrocki, line 23: Let's disscuss. We talking here about our users holding GB'S of data in diff table with still client side resolution by default. 

IMHO server side should be default, but would love to get your opinion as well

packages/graphback-datasync/src/util.ts
ssd71: > I cannot confirm if this algorithm works. Would be possible to explain motivation here.

This is merging the fields that weren't changed

> Conflict detected by version doesn't mean actual conflict. Our implementation should know if we have conflict or mergability. Interface should clearly represent that.

I though this was doing it:https://github.com/aerogear/graphback/blob/94c0db84ebb75ff2bce18350da8680bd144fb27b/packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts#L84-L121

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
ssd71: > IMHO server side should be default, but would love to get your opinion as well

Onboard with this

packages/graphback-datasync/src/util.ts
ssd71: > Maybe you wanted to use full server snapshot and then apply client side diff on top of it?

To me:
serverDiff: What fields have changed on the server since client passed version _version
clientDiff: What fields the client wants to update
resolved: The fields that are in clientDiff but not in serverDiff

But I see that that was wrong so I will change it

wtrocki:
I think to summarize what I want on this PR is:

1. Use the default server-side strategy. (Most recent wins or client wins depending on capabilities)
2. and add unit test for:

- Delete: Client side changed single field, server side deleted everything
- Merge: Client side changed single field, server side changed other field. 
- Conflict: Client and server changed the same field 

3. Implement another strategy (give people choice - like serverWins and export it.

packages/graphback-datasync/src/util.ts
wtrocki: > resolved: The fields that are in clientDiff but not in serverDiff

Ok.. So we have problem with the interface. Maybe quick chat about this. 
I thought resolved is actual final object that will be saved. I see it now when debugging. 

Generally let's have quick chat about this approach - I never seen it but this actually might make a lot of sense.

ssd71:
Added default strategy: clientWins:
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts#L20-L22

Added some tests as best as I could manage, please confirm if these are valid or if something needs to change:
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L83
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L97
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L112
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L128
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L145

There is one more strategy: serverWins:
https://github.com/aerogear/graphback/blob/582df32e1e784a0f999ed9df3763ba58eb56e139/packages/graphback-datasync/src/util.ts#L64
Will write tests for it now

ssd71:
Added tests for serverWins strategy:
https://github.com/aerogear/graphback/blob/ece5c2436e0ee3601143284107802a810ea6fd6b/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L177
https://github.com/aerogear/graphback/blob/ece5c2436e0ee3601143284107802a810ea6fd6b/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L191
https://github.com/aerogear/graphback/blob/ece5c2436e0ee3601143284107802a810ea6fd6b/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L207
https://github.com/aerogear/graphback/blob/ece5c2436e0ee3601143284107802a810ea6fd6b/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L220
https://github.com/aerogear/graphback/blob/ece5c2436e0ee3601143284107802a810ea6fd6b/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts#L237

ssd71:
Starting to add integration test now.

Quick question: Should I change existing DataSync integration test to use createDataSyncAPI? It would save me the trouble of having to write another integration test specifically for createDataSyncAPI but It would also mean that using buildGraphbackAPI to do datasync will be untested. wdyt

ssd71:
Integration test is WIP, will be back later today to:
- add more tests

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
wtrocki: Should be done

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,59 @@ export const DataSyncFieldNames = {
...
+      throw new ConflictError(conflict);
+    }
+
+    const resolved = Object.assign(serverData, clientDiff, serverDiff);
+
+    return resolved
+  },
+  resolveDelete(conflict: ConflictMetadata): any {
+
+    throw new ConflictError(conflict);
```
wtrocki, line 78: We might need to know what operation thrown conflict (in case someone will start batching requests on client)

Separate issue probably?

packages/graphback-datasync/src/util.ts
```diff
@@ -39,3 +44,59 @@ export const DataSyncFieldNames = {
...
+    if (serverData[DataSyncFieldNames.deleted] === true) {
+      resolved[DataSyncFieldNames.deleted] = false;
+    }
+
+    return resolved;
+  },
+  resolveDelete(conflict: ConflictMetadata): any {
+    const { serverData, clientData } = conflict;
+
+    if (serverData[DataSyncFieldNames.deleted] === true || clientData[DataSyncFieldNames.deleted] === true) {
```
wtrocki, line 97: We are missing deletion integration tests.

[Question] want to delete from client (on server this changed) - client wins should force delete. With current handlers it looks like deletion support makes no sense at all (we always return error to client - in both strategies.

wtrocki:
This looks very very good. 
I would merge this in order to start playing with this in other project but we need more integration tests.

ssd71:
Fixed clientWins strategy to force deletes as well as added some more integration tests for deletion

integration/tests/runtime-workflow-datasync-conflicts.ts
```diff
@@ -143,6 +159,103 @@ test('Client side should win when conflict occurs', async () => {
...
+
+  const { data, errors } = await client.query({ query: SYNC_NOTES , variables: {
+    lastSync: 0,
+    filter: {
+      title: {
+        eq: updatedTitle
+      }
+    }
+  }});
+  console.log(errors);
```
wtrocki, line 185: Probably worth to assert this to be empty. :)

Issue #1304, opened 2020-05-25
use dataloader to resolve oneToOne relationships


machi1990:
I was wondering how many db queries we make with a gql query like 

```
query findAllProducts {
  findAllProducts {
    id,
    description,
    label,
    distributionCentre {
      id,
      name
    }
  }
}
```

Looking at the generated `resolvers.ts` file, for a one product I’d say at least two db queries - https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/src/resolvers/resolvers.ts#L27 

for N products, I’d say N+1 queries?

It will be interesting to perform just minimum db round trips.

This is with graphback `0.12.0`

wtrocki:
Dataloader will agregate them by ID. We can enable logging in mongo to prove it or simply debug batch method on Graphbacks DataProvider

machi1990:
Interesting, and seems like a non issue then. 

These logs got my curiosity and did not know we have `dataloader` underhood aggregating the queries: 
```
querying object Product
querying object DistributionCentre with filter {"id":"5ea305da65c3575d036ae57b"}
querying object DistributionCentre with filter {"id":"5ea9a23a134a4aa669e306d4"}
querying object DistributionCentre with filter {"id":"5ea9a8b9134a4aa669e3086f"}
querying object DistributionCentre with filter {"id":"5ea9a23a134a4aa669e306d4"}
```
> Always 2 queries

Much better. Is getting it to perform `one query` a possibility?

wtrocki:
One query will go against GraphQL execution plan that uses relationships. We actually tried that before - see motivation in graphql-query-mapper library for more info.

wtrocki:
This was done 1 year ago. Since then there are couple good data mapping libs out there - sqlmancer etc

craicoverflow:
Closing for now cc @machi1990

machi1990:
If you have the following model
```
""" @model """
type Note {
  id: ID!
  title: String!
  """
  @oneToMany field: 'note'
  """
  comments: [Comment!]!
}

""" @model """
type Comment {
  id: ID!
  text: String
  note: Note!
}
```

If I issue a graphql query to retrieves *Notes* like below
```graphql
query {
  findNotes {
    items {
      id
      comments {
        id
      }
    }
  }
}

```

I am seeing only two db queries below thanks to dataloader.
```log
querying object Note with filter undefined
  knex:query select * from "note" undefined +21s
[...]

PR #1834, opened 2020-08-07
perf: use dataloader to resolve oneToOne or manyToOne relationships

machi1990:
This removes the number of queries issued to the database

Fixes https://github.com/aerogear/graphback/issues/1304

## Reproducer.
Go to current [master](https://github.com/aerogear/graphback/commit/9e71c786e429984046a7f6dba022b0effd760ef6) and use the PostgresSQL template with the following model:

```graphql
""" @model """
type Note {
  id: ID!
  title: String!
  """
  @oneToMany(field: 'note')
  """
  comments: [Comment!]!
}

""" @model """
type Comment {
  id: ID!
  text: String
  note: Note!
}
```

## start the server

```shell
DEBUG=knex:query yarn develop
```
create a note which has three comments linked to it. 

If I issue a graphql query to retrieves *Notes* like below
```graphql
query {
  findNotes {
    items {
      id
      comments {
        id
      }
    }
  }
}

```

I am seeing only two db queries below thanks to dataloader, this is fine.

```log
querying object Note with filter undefined
  knex:query select * from "note" undefined +21s
  knex:query select * from "comment" where "noteId" in (?, ?) undefined +5ms
```

However, when I query for *Comments* with a query like 
```graphql
query {
  findComments {
    items {
      id,
        note {
          id
        }
    }
  }
}

```

I am seeing the following logs, notice the number of sql query to retrieve note. 
```log
knex:query select * from "comment" undefined +1m
  knex:query select * from "note" where "id" = ? limit ? undefined +13ms
  knex:query select * from "note" where "id" = ? limit ? undefined +5ms
  knex:query select * from "note" where "id" = ? limit ? undefined +3ms
  knex:query select * from "note" where "id" = ? limit ? undefined +0ms
....
```
I think this should be only one query, and not several ones. 

Apply this patch and you should see, and verify again to see that the number of queries has been reduced.

### DIFF
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -1,6 +1,7 @@
 /* eslint-disable max-lines */
 import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { resolve, dirname, join } from 'path';
+import * as DataLoader from "dataloader";
 import { getFieldName, metadataMap, printSchemaWithDirectives, getSubscriptionName, GraphbackCoreMetadata, GraphbackOperationType, GraphbackPlugin, ModelDefinition, addRelationshipFields, extendRelationshipFields, extendOneToManyFieldArguments, getInputTypeName, FieldRelationshipMetadata, GraphbackContext, getSelectedFieldsFromResolverInfo, isModelType, getPrimaryKey, isSpecifiedGraphbackJSONScalarType, graphbackScalarsTypes } from '@graphback/core'
 import { GraphQLNonNull, GraphQLObjectType, GraphQLSchema, GraphQLInt, GraphQLFloat, isScalarType, isSpecifiedScalarType, GraphQLResolveInfo, isObjectType, GraphQLField } from 'graphql';
 import { SchemaComposer, NamedTypeComposer } from 'graphql-compose';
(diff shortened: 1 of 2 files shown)


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -701,18 +702,27 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     const relationOwner = relationship.ownerField.name;
     const model = modelNameToModelDefinition[modelName];
 
+    // construct a unique key to identify the dataloader
+    const dataLoaderName = `${modelName}-${relationship.kind}-${relationIdField.name}-${relationship.relationForeignKey}-${Math.random().toString(16).slice(2)}-DataLoader`;
```
wtrocki, line 706: Out of curiosity - why we need random here?

machi1990: Just to generate a unique key

machi1990: But I think we'll be fine without it as the segment before it are unique enough.

wtrocki: I think that this dynamic dataloader was hack done at the time where we did not provided our contexCreator. This was done before we even had services mounted to context. Since we control now context this pattern look very very strange and we can have our dataloaders created inside contextCreator. 

Not saying that we should do it, but as guy that committed this (ugly?) hack I wanted to say why we were doing it and why this is no longer relevant.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -701,18 +702,27 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
...
       }
 
-      const selectedFields = getSelectedFieldsFromResolverInfo(info, model);
-      const graphback = {
-        services: context.graphback.services,
-        options: { selectedFields }
-      };
+      if (!context[dataLoaderName]) {
+        context[dataLoaderName] = new DataLoader<string, any>(async (keys: string[]) => {
+          const selectedFields = getSelectedFieldsFromResolverInfo(info, model);
```
wtrocki, line 714: Love it! Thanak you for adding this. Not sure if we can test it somehow?
From user/response point if view result will be the same.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: Thank you for the context. I went with this approach to create them lazily and not having a global one inside contextCreator as think that the dataloader should be:
 - created only when we need them and 
- ensure its uniqueness (for each model's relationship) to avoid loading entity on a different dataloader hence service (not related to the entity). 

PS: Still discovering this (had to went through the `dataloader` docs to understand somethings) so happy to be directed in the right path here.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: I think once I make the integration tests pass, we should be good https://app.circleci.com/pipelines/github/aerogear/graphback/6791/workflows/c990cf6b-8a62-4513-a09f-c6961a505386/jobs/6925

I also have a TODO item about `relationships` in the benchmark folder.

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
wtrocki: TBH.. this works pretty well (and it was reducing number of queries for OneToMany side), but some people might say this is hack :D

machi1990: > TBH.. this works pretty well (and it was reducing number of queries for OneToMany side), but some people might say this is hack :D

Yes, it is possible to do all in our query. 
No, we cannot really do it because of the way we resolve things. 

Agreeing with you here, two queries (1. to retrieve parent object, and 2. retrieve related children objects) is better than 1 + N db queries. I guess I can say there is room for improvement in the future ;-)

PR #1835, opened 2020-08-07
feat: subscription filtering v2

### LINKED ISSUE #1760: Investigate expansion of SubscriptionFilter to allow non-simplistic filtering
Currently the SubscriptionFilter input allows direct value comparisons. Since we are going to route of in-memory subscription filtering this can be expanded to be more aligned with the model filter input in the `findBy` query.

**NoteFilter**

```gql
input NoteFilter {
  id: IDInput
  title: StringInput
  opened: BooleanInput
  likes: IntInput
  completedPercentage: FloatInput
  and: [NoteFilter!]
  or: [NoteFilter!]
  not: NoteFilter
}
```

**NoteSubscriptionFilter**

```graphql
input NoteFilter {
  id: ID
  title: String
  opened: Boolean
  likes: Int
  completedPercentage: Float
}
```

1. Will it be straightforward to implement this advanced filtering on subscriptions?
2. What operators should be used? All the same as in `findBy`?
1. Can we reuse the same filter input, or should the <T>SubscriptionFilter? Depends on answer to 2.

craicoverflow:
Closes #1760 

This PR has increased the filtering capabilities for subscriptions by using the same filter inputs as the `findBy` query.

```gql
subscription {
  updatedNote(filter: {
    title: {
      eq: "hello"
    }
  }) {
    id
    title
  }
}
```

### Tasks

- [x] update filter input in schema 
- [x] create filter predicate 
- [x] tests
- [x] docs
- [x] Add support for AND/OR/NOT operators


packages/graphback-core/src/runtime/subscriptionFilter.ts
```diff
@@ -0,0 +1,40 @@
+import { QueryFilter } from './QueryFilter';
+
+type PredicateFn = (input: any) => boolean;
+
+interface IPredicate {
+  eq(filterValue: any): PredicateFn
+  gt(filterValue: any): PredicateFn
```
craicoverflow, line 7: I will add more expressions later.

machi1990:
This looks good, what I see is us being able to construct the final `Predicate` from a `QueryFilter` and we can possibly use them in [InMemoryDataProvider](https://github.com/aerogear/graphback/blob/9e71c786e429984046a7f6dba022b0effd760ef6/performance/libs/inmemory-data-provider.js) that's somehow fully feature (Off topic in this PR) ;-) 

> Filtering on the id: ID field does not work because we are passing a number usually, but the scalar parses this as a string so a comparison between the two values does not work.

Maybe we can use `.toString()` as in: 
- https://github.com/aerogear/graphback/blob/9e71c786e429984046a7f6dba022b0effd760ef6/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts#L109
- https://github.com/aerogear/graphback/blob/9e71c786e429984046a7f6dba022b0effd760ef6/packages/graphback-runtime-mongodb/src/MongoDBDataProvider.ts#L135

lastmjs:
It should be possible to filter by updated field. For example, if I have a user's id, I should be able to subscribe to changes to a specific field of the user, for example an `available` field. In this case I don't want an equality filter except for the user's id. I only want to receive updates when the user's `available` field is update. Will this be possible?

craicoverflow:
Under the current subscription filtering we don't support receiving updates when a user's `available` field is updated.

```ts
const filter = {
  available: ['available', 'unavailable', 'back soon']
}
```

Would you see something like this as being satisfactory, where you can provide multiple values, or do you think the subscription should be triggered regardless of the value of `available`?

lastmjs:
For this particular use case, I could probably get away with providing
values, since if is a Boolean. But for other use cases this will not be
sufficient. For example, what if I am waiting for a field to update that is
an arbitrary string or a date? I believe I have one use case for this
already, where I want to know when a field is updated from null to a date
string. That can't be known before runtime

craicoverflow:
> Maybe we can use `.toString()` as in:

Yeah, that would probably work fine. I have not done it yet because at the time I wanted to check out the different data types as I was concerned converting all of them to string might not work for some expressions (that are not `eq`).

craicoverflow:
Ah yes, converting everything to a string will not work. Take the following example:

`10 < 9` is false
but
`"10" < "9"` is true

One option is to identify the data type of the filter field and convert the payload field type to compare in the same type.

craicoverflow:
@lastmjs that sounds like a valid use case, would you mind creating an issue and we can triage it as a team?

machi1990:
> Ah yes, converting everything to a string will not work. Take the following example:
> 
> `10 < 9` is false
> but
> `"10" < "9"` is true
> 

```js

// ID comparison
let value = "1" < 9;
console.log(value); // output true
value = "1" > 9;
console.log(value); // output false   

value = "1" == 1;
console.log(value); // output true   

// ObjectID assuming that its converted to string

value = "1" < "9";
console.log(value) // outputs true
value = "10" < "9"
console.log(value) // output true
value = "10" > "9"
console.log(value) // output false
```

With this in mind, If the payload is a number, no conversion needed. Otherwise convert payload and filter to string and do appropriate comparison.

> One option is to identify the data type of the filter field and convert the payload field type to compare in the same type.

Yes, this could help out somehow. But I am wondering if we can achieve this for `ObjectID`, as far as I know it only has `.equals()` method.

craicoverflow:
This is ready for review now @machi1990 and @wtrocki 

Question - should we add AND/OR/NOT filter capabilities to subscription filtering?

craicoverflow:
Ignore my request for review - missing a case.

wtrocki:
AND seems like must have. I can live without OR

Competition has same filters for query and subscribtions
https://hasura.io/docs/1.0/graphql/manual/subscriptions/index.html

packages/graphback-core/src/runtime/subscriptionFilter.ts
```diff
@@ -0,0 +1,172 @@
+import { convertType, isDateObject } from '../utils/convertType';
```
machi1990, line 1: I'll remove `subscription` from the content of this file and from the file name to something like `PredicateBuilder`, `InMemoryFilter` or something along the line. What I see is that we have really nice in memory filter capabilities that can be used for anything other than subscription (I am thinking of in memory datastores ;-)). What's your thoughts on this?

packages/graphback-core/src/runtime/subscriptionFilter.ts
```diff
@@ -0,0 +1,172 @@
...
+ * Dynamically creates a subscription filter predicate using the filter object values
+ * 
+ * @param {QueryFilter} filter - subscription filter input object
+ */
+export function createSubscriptionFilterPredicate<T = any>(filter: QueryFilter): (input: Partial<T>) => boolean {
+  const andFilter = filter.and;
+  const orFilter = filter.or;
+  const notFilter = filter.not;
+
+  const filterFields = Object.keys(filter).filter((key: string) => !['and', 'or', 'not'].includes(key));
```
machi1990, line 95: not needed we can skip "boolean operators" in the loop, https://github.com/aerogear/graphback/pull/1835/files#diff-5cab2776f685ff7a7ce3e24f71aeea51R99

packages/graphback-core/src/runtime/subscriptionFilter.ts
```diff
@@ -0,0 +1,172 @@
...
+  return (payload: Partial<T>): boolean => {
+    let predicateResult = true;
+    for (const fieldName of filterFields) {
+
+      const fieldFilter = filter[fieldName];
+
+      for (const [expr, exprVal] of Object.entries(fieldFilter)) {
+        const predicateFn: PredicateFn = predicateMap[expr](exprVal)
+
+        predicateResult = predicateFn(payload[fieldName])
```
machi1990, line 106:
```suggestion
        predicateResult = predicateResult && predicateFn(payload[fieldName])
```

Are we missing an `and` operator? 

Also looking at the code, I think it should be possible to short circuit the loop 

```
if (!predicateFn(payload[fieldName])) {
  predicateResult = false;
  break;
}
```

I'll propose treating `or` operator first, since this will enable for further short circuiting e.g instead of a `break;` statement in my suggestion we could `return false` when the `orPredicate` evaluation is false.

packages/graphback-core/src/utils/convertType.ts
```diff
@@ -0,0 +1,37 @@
...
+      return String(value);
+  }
+}
+
+/**
+ * Check if value is a Date object
+ * 
+ * @param {any} value 
+ */
+export const isDateObject = (value: any): boolean => Object.prototype.toString.call(value) === '[object Date]';
```
machi1990, line 37: Do we need to export this? seems like an inside thingy! 

Also could this be simplified to ? 
```suggestion
const isDateObject = (value: any): boolean => value instanceof Date';
```

packages/graphback-core/tests/subscriptionFilterTest.ts
```diff
@@ -0,0 +1,736 @@
+/* eslint-disable max-lines */
+import { Scalars, Maybe, QueryFilter, StringInput, IdInput, BooleanInput, IntInput, FloatInput, DateTimeInput } from '../src/runtime/QueryFilter';
+import { createSubscriptionFilterPredicate } from '../src/runtime/subscriptionFilter';
+
```
machi1990, line 4: General comment: I love these tests! Thank you for taking the time to add them. 

Can we remove `subscription`? Same comment as the one I made above for the `file under test`.

packages/graphback-core/tests/subscriptionFilterTest.ts
```diff
@@ -0,0 +1,736 @@
...
+type User = {
+  __typename?: 'User';
+  id?: Scalars['ID'];
+  name?: Scalars['String'];
+  verified?: Maybe<Scalars['Boolean']>;
+  age?: Scalars['Int'];
+  score?: Scalars['Float'];
+  createdAt?: Scalars['DateTime'];
+};
```
machi1990, line 13: Are these enough to cover the `ObjectID` case?

packages/graphback-core/src/runtime/subscriptionFilter.ts
craicoverflow: For sure, makes total sense!

packages/graphback-core/src/utils/convertType.ts
craicoverflow: https://stackoverflow.com/a/643827

`new Date("some random text")` is an instance of Date, so this could lead to false positives. This way we will know for certain that this is a valid date object.

craicoverflow: No need to export it (I was at one time using it from another file).

machi1990: > https://stackoverflow.com/a/643827
> 
> `new Date("some random text")` is an instance of Date, so this could lead to false positives. This way we will know for certain that this is a valid date object.

Yes, the Date constructor is a weird one but I do not think that this can happen though as some scalars library will perform the validation for us. e.g https://github.com/Urigo/graphql-scalars/blob/dc0c2bdfdbf6742b79c0a223024fe2fee1329457/src/scalars/iso-date/DateTime.ts#L36

packages/graphback-core/src/runtime/subscriptionFilter.ts
craicoverflow: Yep, `or` is missing now. But will apply your second proposal.

packages/graphback-core/src/utils/convertType.ts
craicoverflow: Correct, but as you mentioned this filter can be used anywhere in theory, and could be used without going through a GraphQL scalar, avoiding validation. Would it be better to cater both cases?

By the way, this is used externally actually: https://github.com/aerogear/graphback/blob/fc8814c0156b70003156bb6838762bf351fed126/packages/graphback-core/src/runtime/subscriptionFilter.ts#L62

packages/graphback-core/tests/subscriptionFilterTest.ts
craicoverflow: I used all the typed filters were available in `QueryFilter.ts`:

https://github.com/aerogear/graphback/blob/c2a718aa9da5d00ab74877996f2a72bcc65c414d/packages/graphback-core/src/runtime/QueryFilter.ts#L17-L83

I'm guessing a PR will need to be created to update QueryFilter with ObjectID.

packages/graphback-core/src/utils/convertType.ts
machi1990: Okay, we can leave this as it is now.

packages/graphback-core/tests/subscriptionFilterTest.ts
machi1990: Sure, we'll quickly need it bu happy to having this done in followup PR.

craicoverflow:
Sorry if I missed something @machi1990 - applied your feedback in https://github.com/aerogear/graphback/pull/1835/commits/3b0bbb7e007b42643980ef3b951077d487b02e8f. Thanks for the review!

craicoverflow:
Thanks @machi1990 ! Follow up will be in https://github.com/aerogear/graphback/issues/1864

machi1990:
> Thanks @machi1990 ! Follow up will be in #1864

Thanks! Let's merge this ;-)

Issue #1837, opened 2020-08-08
invalid input syntax for type timestamp: "null"


lastmjs:
I have this mutation:

```graphql
            mutation (
                $createdAt: DateTime!
                $updatedAt: DateTime!
                $userCallSid: String!
                $podId: ID!
                $callId: ID!
                $incomingCallType: INCOMING_CALL_TYPE!
            ) {
                createIncoming_queue_items(input: {
                    created_at: $createdAt
                    updated_at: $updatedAt
                    user_call_sid: $userCallSid
                    pod_id: $podId
                    call_id: $callId
                    incoming_call_type: $incomingCallType
                    handled_at: null
                }) {
                    id
                }
            }
```
`handled_at` is a custom scalar called `DateTime`. It's nullable. I keep getting this error with this mutation: 

```
{"errors":[{"message":"insert into \"incoming_queue_items\" (\"call_id\", \"created_at\", \"handled_at\", \"incoming_call_type\", \"pod_id\", \"updated_at\", \"user_call_sid\") values ($1, $2, $3, $4, $5, $6, $7) returning * - invalid input syntax for type timestamp: \"null\"","locations":[{"line":10,"column":17}],"path":["createIncoming_queue_items"]}],"data":{"createIncoming_queue_items":null}}
```
I haven't setup any custom code for my DateTime scalar, do I need to? Seems like a null value should work as is

machi1990:
Automatically generated comment to notify maintainers
/cc @craicoverflow, @machi1990, @wtrocki

wtrocki:
https://graphback.dev/docs/next/model/scalars

Scalars we support

machi1990:
@lastmjs we've dropped another [0.16.0-alpha3](https://github.com/aerogear/graphback/releases/tag/0.16.0-alpha3) which includes a fix for this issue. Mind checking if this release effectively fix the issue for you? Thanks!

lastmjs:
Just tested and seems to work! Thanks!

PR #1849, opened 2020-08-10
fix(core): avoid stringfying "null" values as `"null"`

machi1990:
A null value type is an object, this was the transformed to "null" by
the fixed logic resulting in errors for non string columns in database.

Fixes https://github.com/aerogear/graphback/issues/1837

### DIFF
--- a/packages/graphback-core/src/db/dataMapper.ts
+++ b/packages/graphback-core/src/db/dataMapper.ts
@@ -35,7 +35,7 @@ export const getDatabaseArguments = (modelMap: ModelTableMap, data?: any): Table
     transFormedData = {};
     for (const key of keys) {
       const value: any = data[key];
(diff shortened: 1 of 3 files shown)


craicoverflow:
Verified manually. Maybe we could have a test case with a null value to verify this?

https://github.com/machi1990/graphback/blob/d311eb9bae2b2f1fa5fbbca45a0ffeb3ad5b252b/packages/graphback-core/tests/dataMapperTest.ts#L56

machi1990:
> Verified manually. Maybe we could have a test case with a null value to verify this?
> 
> https://github.com/machi1990/graphback/blob/d311eb9bae2b2f1fa5fbbca45a0ffeb3ad5b252b/packages/graphback-core/tests/dataMapperTest.ts#L56

Sure, good suggestion.

Issue #1859, opened 2020-08-12
The loader.load() function must be called with a value,but got: null.


lastmjs:
It seems that the dataloader, if a relation does not exist, throws this error: `The loader.load() function must be called with a value,but got: null.` This happens even if the relation is nullable.

Here's my example query:

```graphql
        query ($twilioParentCallSid: String!) {
            findCalls(filter: {
                twilio_parent_call_sid: {
                    eq: $twilioParentCallSid
                }
            }) {
                items {
                    id
                    duration
                    billable_duration
                    ignore_during_audit
                    direction
                    origin_phone
                    marchex_recommendation
                    audit_result
                    twilio_parent_call_sid
                    destination_phone
                    action_item {
                        id
                    }
                    user {
                        id
                    }
                    company {
                        id
                        pod {
                            id
                        }
                        voicemail_audio_file_name
                        voicemail_text
                        queue_audio_file_name
                        callbacks_allowed
                    }
                    contact {
                        id
                        status
                        phone_number
                        sales
                    }
                    duration_events(filter: {
                        valid: {
                            eq: true
                        }
                    }) {
                        id
                        type
                        timestamp
                        description
                        user {
                            id
                        }
                    }
                }
            }
        }  
```
Here's what is returned:

```
  "errors": [
    {
      "message": "The loader.load() function must be called with a value,but got: null.",
      "locations": [
        {
          "line": 19,
          "column": 21
        }
      ],
      "path": [
        "findCalls",
        "items",
        0,
        "action_item"
      ]
    },
    {
      "message": "The loader.load() function must be called with a value,but got: null.",
      "locations": [
        {
          "line": 22,
          "column": 21
        }
      ],
      "path": [
        "findCalls",
        "items",
        0,
        "user"
      ]
    }
  ],
  "data": {
    "findCalls": {
      "items": [
        {
          "id": "147",
          "duration": 0,
          "billable_duration": 0,
          "ignore_during_audit": false,
          "direction": 0,
          "origin_phone": "REDACTED",
          "marchex_recommendation": "PASS",
          "audit_result": "NOT_AUDITED",
          "twilio_parent_call_sid": "REDACTED",
          "destination_phone": "REDACTED",
          "action_item": null,
          "user": null,
          "company": {
            "id": "1",
            "pod": {
              "id": "1"
            },
            "voicemail_audio_file_name": null,
            "voicemail_text": null,
            "queue_audio_file_name": null,
            "callbacks_allowed": true
          },
[...]

PR #1861, opened 2020-08-12
fix: add a null / undefined check before loading relationship with dataloader and make sure that relationship field is loaded in these cases

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1859


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -726,7 +728,15 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
         });
       }
 
-      return context[dataLoaderName].load(parent[relationship.relationForeignKey])
+      const relationForeignKey = parent[relationship.relationForeignKey];
+
+      // eslint-disable-next-line no-null/no-null
+      if (relationForeignKey === undefined || relationForeignKey === null) {
```
craicoverflow, line 734: Catches both:

```suggestion
      if (!relationForeignKey) {
```

machi1990: I wanted to cater for cases where `relationForeignKey` is either `0` or `''`.

PR #1862, opened 2020-08-12
feat(schema-plugin): use GraphbackDateTime scalar for generated createdAt updatedAt fields

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1349


wtrocki:
Since we moved to our own fields in datasync this brings zero conflict. Amazing change. 
Solid merge for me. @ssd71 you need to rebase your branch after.

Issue #1825, opened 2020-08-06
[DataSync] automatic indexing of "lastUpdatedAt" field for better query performance


ssd71:
I believe this would make for much faster sync queries, and remove the step of in-memory sorting when ordering by lastUpdatedAt, if one chooses to do it in a findBy query

machi1990:
Automatically generated comment to notify maintainers
/cc @craicoverflow, @machi1990, @wtrocki

machi1990:
+1, makes sense to me. This will mean adding the `@index` annotation by default right? @ssd71

wtrocki:
Aren't we doing it now 🙈

ssd71:
> +1, makes sense to me. This will mean adding the `@index` annotation by default right? @ssd71

Yep

> Aren't we doing it now :see_no_evil:

Nope :smile:

wtrocki:
🤦

machi1990:
This is low hanging fruit. I kind of have my eyes in the codebase that does this, I am grabbing this issue in order to leave @ssd71 focused on interesting issues. WDYT?

wtrocki:
@machi1990 This is not longer up to date. We have separate _lastUpdated field now for datasync.

machi1990:
> @machi1990 This is not longer up to date. We have separate _lastUpdated field now for datasync.

Yes, that's the one that's indexed automatically in the patch I opened.

Issue #1597, opened 2020-06-29
DataSync silently override createdAt/updatedAt custom fields


machi1990:
If I have a model with custom fields 

```
"""
Represents action that is assigned to volunteer

@model
@versioned
"""
type Entity {
  id: ID!
  createdAt: DateTime
  completedAt: DateTime
  updatedAt: DateTime
}
```

When I start the server with the datasync plugin the two fields `createdAt`, `updatedAt` are silently overwritten and their type is changed to `String`. 

```
"""
Represents action that is assigned to volunteer

@model
@versioned
"""
type Entity {
  id: ID!
  createdAt: String
  completedAt: DateTime
  updatedAt: String
}
```

/cc @ssd71 @wtrocki @craicoverflow

ssd71:
This is a sort of workaround that we decided until we add Graphback native DateTime scalars, current implementation is quite unstable(hacky) and is bound to change, it's generally not recommended to explicitly specify those fields as of now :smile:, I think we should decide something about this soon, before something else breaks

Practically speaking as far as I am aware of, it's not recommended to use any type other than string (or Double) for these fields because of the way this hack works.

craicoverflow:
I see three approach options:

1. Detect if these fields exist and don't overwrite the fields. The issue with this is how does Graphback decide whether the scalar added by the user is a valid one. 

If a field `createdAt: Date` is specified by the user it is probably invalid and will not work for DataSync, since it does not have enough information about the time.

2. Overwrite the field as it is doing, because this is one of the main functions of the `@versioned` field and a caveat of adding it is Graphback will need to control this field.

3. Throw an error and don't start the server until the user removes their custom field.

My personal preference is approach 3, since a rule to follow is to not overwrite user specified code/configuration. I see approach 1 as causing issues as Graphback does not know until query execution if the scalar is valid or not.

craicoverflow:
^^ Posted hours ago, GitHub would not let me comment.

machi1990:
> I see three approach options:
> 
> 1. Detect if these fields exist and don't overwrite the fields. The issue with this is how does Graphback decide whether the scalar added by the user is a valid one.
> 
> If a field `createdAt: Date` is specified by the user it is probably invalid and will not work for DataSync, since it does not have enough information about the time.
> 
> 1. Overwrite the field as it is doing, because this is one of the main functions of the `@versioned` field and a caveat of adding it is Graphback will need to control this field.
[...]

PR #1863, opened 2020-08-12
feat: create index for the lastUpdateAt field

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1825

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -127,7 +127,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
     modelTC.addFields({
       [DataSyncFieldNames.lastUpdatedAt]: {
-        type: TimestampSTC.getType()
+        type: TimestampSTC.getType(),
+        description: "@index(name: 'Datasync_lastUpdatedAt')"
       }
     });
 
@@ -160,7 +161,7 @@ export class DataSyncPlugin extends GraphbackPlugin {
     const TimestampSTC = schemaComposer.getSTC(GraphbackTimestamp.name);
 
     const DeltaTypeFieldNames = this.getDeltaTypeFieldNames(modelTC.getType());
-    
+
     // Add Delta Type to schema
     const DeltaOTC = modelTC.clone(getDeltaType(modelName));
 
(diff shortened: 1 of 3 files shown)


wtrocki:
TL;DR - generic lastUpdated at is no longer used due to the fact that people might have clashes on types. We really need to control that type name etc. Not sure why @ssd71 approved it :)

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -127,7 +127,8 @@ export class DataSyncPlugin extends GraphbackPlugin {
 
     modelTC.addFields({
       [DataSyncFieldNames.lastUpdatedAt]: {
```
machi1990, line 129: @wtrocki this field corresponds to https://github.com/aerogear/graphback/blob/e450454bac3d9d05cca0e5478e447c9e09395791/packages/graphback-datasync/src/util.ts#L39 which is `_lastUpdatedAt`. So it is not the generic one that's got indexed.

machi1990: We can also see this from the index https://github.com/aerogear/graphback/pull/1863/files#diff-a7d16cb5768bfd64feb81ba0b1a8921bR185

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
```diff
@@ -173,6 +173,23 @@ describe('Delta Queries', () => {
...
+  it('creates index for the lastUpdateAt field', async () => {
+
+    context = await createTestingContext(postSchema);
+
+    const indexes = await context.db.collection('post').indexes();
+
+    expect(indexes[1]).toMatchObject(
+      {
+        "key": {
+          "_lastUpdatedAt": 1
```
ssd71, line 185:
```suggestion
          [DataSyncFieldNames.lastUpdatedAt]: 1
```
Maybe this would be better

wtrocki:
Ah.. - Was this PR merged? Sorry I was the one who got confused. DId not know that this changes are on master.

packages/graphback-datasync/tests/DataSyncMongoDBTests.ts
machi1990: Thanks, applied.

machi1990:
> Ah.. - Was this PR merged? Sorry I was the one who got confused. DId not know that this changes are on master.

There were two PRs that are slightly not related but somehow touches the dates. 

The first one was https://github.com/aerogear/graphback/pull/1862 which only touched the `createdAt` and `updatedAt` fields by changing their type to timestamp instead of a string. I believe with https://github.com/aerogear/graphback/issues/1597 we should be able to control how these fields are added in order to not override custom added ones; 

This PR only adds indexing capabilities onto the `_lastUpdatedAt` field.

PR #1870, opened 2020-08-13
Update QueryFilter type with Graphback scalars and disable generation of filters for unknown custom scalars.

### LINKED ISSUE #1864: Add custom scalars to QueryFilter
[`QueryFilter`](https://github.com/aerogear/graphback/blob/c2a718aa9da5d00ab74877996f2a72bcc65c414d/packages/graphback-core/src/runtime/QueryFilter.ts#L81) only deals with default scalars. We could add [Graphback scalars](https://graphback.dev/docs/next/model/scalars#table-summary) to provide more strongly typed query-ability.

Let's also disable generating filters for unknown scalars since we have provided most scalars by now.

craicoverflow:
Closes #1864 

This PR does:

- Add Graphback scalars to the `QueryFilter` and the and/or/not operators
- Removes generation of filter inputs for unknown custom scalars.

### DIFF
--- a/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
+++ b/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
@@ -1,4 +1,4 @@
-import { ObjectID } from 'mongodb';
+import { ObjectID } from 'bson';
 import { convertType, isDateObject } from '../utils/convertType';
 import { QueryFilter } from './QueryFilter';
 
--- a/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
+++ b/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
@@ -1,3 +1,4 @@
+import { ObjectID } from 'mongodb';
 import { convertType, isDateObject } from '../utils/convertType';
 import { QueryFilter } from './QueryFilter';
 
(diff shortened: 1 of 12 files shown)


craicoverflow:
@machi1990 I had implemented the special checks for ObjectID but the commit got lost and I had not noticed - I got it back with `git reflog` and it should work for ObjectID now :smile: 

https://github.com/aerogear/graphback/pull/1870/commits/d528664f1f2f08cdcd12200d1a31c1e858bfc32b

packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
```diff
@@ -1,3 +1,4 @@
+import { ObjectID } from 'mongodb';
```
machi1990, line 1: This is not related to this PR but wondering weather we should trim this to something like https://github.com/aerogear/graphback/compare/master...machi1990:chore/depends-on-bson-directly?expand=1 by having `bson` as a dependency and not mongodb as peerDependency. The idea will be that his can be used directly without requiring a user to have `mongodb` as depedency.

machi1990: You can cherry-pick the commit from my fork. Or merge this and I'll do it afterwards. Let me know what works for you.

craicoverflow: I will cherry pick yours, thanks for that.

machi1990: Cool.

craicoverflow: https://app.circleci.com/pipelines/github/aerogear/graphback/7175/workflows/35ccf632-8913-461e-baa0-837fc059a9ab/jobs/7315

I have cherry picked your commit but the build is failing - this is down to there being no ObjectId type declarations available from `bson-ext`, which are needed to creating the Scalar type [here](https://github.com/craicoverflow/graphback/blob/22a2be49ee8225adb88ce2c1ce55cdbe045197ae/packages/graphback-core/src/runtime/QueryFilter.ts#L14)

packages/graphback-core/src/scalars/index.ts
```diff
@@ -27,12 +28,18 @@ const { parseLiteral, parseValue, ...objectIDConfig } = extractConfig(ObjectIDRe
...
+    return new ObjectIDFinal(parseLiteral(ast, variables));
   },
   parseValue: (value: any) => {
-    const { ObjectID } = require('mongodb');
+    let ObjectIDFinal = ObjectID;
+    try {
+      ObjectIDFinal = require('bson-ext').ObjectID;
+    } catch {}
 
     return new ObjectID(parseValue(value));
```
machi1990, line 44: Aah this is my mistake, it should be `new ObjectIDFinal(...)`. 

The idea is to always prefer [bson-ext](https://github.com/mongodb-js/bson-ext) a C++ native implementation which is faster than [js-bson](https://github.com/mongodb/js-bson)

I'll create a followup PR for this.

PR #1883, opened 2020-08-17
[WIP]feat(datasync): add TTL

ssd71:
This PR adds TTL for datasync

## TODO
- [ ] Docs

### DIFF
--- a/integration/datasync-mongodb-model.graphql
+++ b/integration/datasync-mongodb-model.graphql
@@ -1,6 +1,8 @@
 """
 @model
-@datasync
+@datasync(
+  ttl: 5184000
+)
 """
 type Note {
   _id: GraphbackObjectID!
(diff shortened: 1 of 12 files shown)


integration/datasync-mongodb-model.graphql
```diff
@@ -1,6 +1,8 @@
 """
 @model
-@datasync
+@datasync(
+  ttl: 5184000
```
wtrocki, line 4: Does it mean TTL for soft deletes or TTL for the conflict. 
Would they be handled by the same value now? 

Any explanation what TTL means here will be good.

wtrocki:
Looks good so far. Let's get docs fixed first. Then we can updated docs in this PR. 
I do not want to loose this change (being undocumented etc.)

integration/datasync-mongodb-model.graphql
ssd71: This is the ttl for the base table, both for conflicts as well as soft deletes

ssd71: TTL for base table would be setup in schema and TTL for delta table would be given in createDataSyncAPI, I could not find a better way to put this because createDataSyncAPI has to do both i.e. just delta queries AND delta queries with conflict resolution, and we need base table TTL for both of these cases, Only alternative is to pass a map of model name to base table TTL separately to createDataSyncAPI

That maybe sounds like it might be problematic and as such I might need some time to fix it up

wtrocki: This is good. If you think about it those values need to be connected

wtrocki:
Can we get that in. I would love new alpha for testing

Issue #1876, opened 2020-08-14
Cannot return null for non-nullable field Users.email.


lastmjs:
This is a strange one. Here is my query:

```graphql
        query ($callId: ID!) {
            getCalls(id: $callId) {
                duration_events {
                    user {
                        id
                    }
                }
                company {
                    users {
                        user {
                            email
                        }
                    }
                }
            }
        }
```
Executing this query, I get the following result:

```graphql
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Users.email.",
      "locations": [
        {
          "line": 11,
          "column": 11
        }
      ],
      "path": [
        "getCalls",
        "company",
        "users",
        0,
        "user",
        "email"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "Error: Cannot return null for non-nullable field Users.email.",
            "    at completeValue (/var/task/node_modules/graphql/execution/execute.js:560:13)",
            "    at completeValueCatchingError (/var/task/node_modules/graphql/execution/execute.js:495:19)",
            "    at resolveField (/var/task/node_modules/graphql/execution/execute.js:435:10)",
            "    at executeFields (/var/task/node_modules/graphql/execution/execute.js:275:18)",
            "    at collectAndExecuteSubfields (/var/task/node_modules/graphql/execution/execute.js:713:10)",
            "    at completeObjectValue (/var/task/node_modules/graphql/execution/execute.js:703:10)",
            "    at completeValue (/var/task/node_modules/graphql/execution/execute.js:591:12)",
            "    at completeValue (/var/task/node_modules/graphql/execution/execute.js:557:21)",
            "    at /var/task/node_modules/graphql/execution/execute.js:492:16",
            "    at processTicksAndRejections (internal/process/task_queues.js:97:5)"
          ]
        }
      }
    }
  ],
  "data": {
    "getCalls": {
      "duration_events": [
        {
          "user": {
            "id": "21"
          }
        },
        {
          "user": {
            "id": "21"
          }
        },
        {
          "user": {
            "id": "21"
          }
        }
      ],
      "company": null
    }
  }
}
```
Now, when I add the `email` field to the user selected in `duration_events`, the query works as expected. Here is the query that works:

```graphql
        query ($callId: ID!) {
            getCalls(id: $callId) {
                duration_events {
                    user {
                        id
                      	email # adding this makes the query work
                    }
                }
                company {
                    users {
                        user {
[...]

PR #1888, opened 2020-08-17
fix: make sure dataloader name is unique enough

machi1990:
This makes sure that relationships are loaded within unique context thus avoid resolving fields in
different context results than the one it was requested

Fixes https://github.com/aerogear/graphback/issues/1876

# To verify the patch: 

- Run the added integration test without the change in the codebase should err. (simply pulling these changes and running the test should be enough). 
- Then build the patch and run the tests again  `yarn build && yarn test:integration -u`

### DIFF
--- a/packages/graphback-core/src/runtime/CRUDService.ts
+++ b/packages/graphback-core/src/runtime/CRUDService.ts
@@ -166,7 +166,9 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
 
   public batchLoadData(relationField: string, id: string | number, filter: any, context: GraphbackContext) {
     // TODO use relationfield mapping
-    const keyName = `${this.modelName}${upperCaseFirstChar(relationField)}DataLoader`;
+    const selectedFields = context.graphback.options?.selectedFields || [];
+    const fetchedKeys = selectedFields.join('-');
+    const keyName = `${this.modelName}-${upperCaseFirstChar(relationField)}-${fetchedKeys}-${JSON.stringify(filter)}-DataLoader`;
     if (!context[keyName]) {
       context[keyName] = new DataLoader<string, any>((keys: string[]) => {
         return this.db.batchRead(relationField, keys, filter, context);
--- a/integration/postgres-model.graphql
+++ b/integration/postgres-model.graphql
@@ -18,7 +18,7 @@ type Note {
 """ @model """
 type Comment {
   id: ID!
-  text: String
+  text: String!
   createdAt: GraphbackTimestamp
   description: String
   """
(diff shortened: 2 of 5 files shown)


packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -166,7 +166,9 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
 
   public batchLoadData(relationField: string, id: string | number, filter: any, context: GraphbackContext) {
     // TODO use relationfield mapping
-    const keyName = `${this.modelName}${upperCaseFirstChar(relationField)}DataLoader`;
+    const selectedFields = context.graphback.options?.selectedFields || [];
+    const fetchedKeys = selectedFields.join('-');
+    const keyName = `${this.modelName}-${upperCaseFirstChar(relationField)}-${fetchedKeys}-DataLoader`;
```
machi1990, line 171: I think filters should be part of the key too. 

What do you think? @craicoverflow @wtrocki

wtrocki:
Is the problem here happening when someone executes multiple queries for the same resolver?
While this works thinking if there is another technique we can use like context splitting etc.

machi1990:
> Is the problem here happening when someone executes multiple queries for the same resolver?

Yes, I believe the problem can occur in any case as long we touch same resolver. 
 
> While this works thinking if there is another technique we can use like context splitting etc.

Interesting. Never heard of this technique before, wondering how it works?. I'd be also happy to try this out.

wtrocki:
OK. Context splitting is just to override context for each root resolver call.
So generally this will be working only for specific providers like graphql-tools as execution logic differs between those. 

Looks like what you have provided is working and literally the only thing I can think of right now. 
Generally batching with generic providers will be kinda hard as you can have countless cases of filters that you cannot predict (and loaders need to be precreated for those). 

I think we shown that in generic providers dataloaders do not work (extra queries and code complexity is just unjustifiable huge. Using some ORM like solution (that will deal with joins) will be much easier here

craicoverflow:
> I think we shown that in generic providers dataloaders do not work (extra queries and code complexity is just unjustifiable huge. Using some ORM like solution (that will deal with joins) will be much easier here

Does this solution also imply only having root resolvers and no relationship resolvers?

wtrocki:
> Does this solution also imply only having root resolvers and no relationship resolvers?

Yep. Query mapper idea. Generally not on the scope for 1.0 but just saying that we cannot really overcome limitations of dataloader.

machi1990:
> I think we shown that in generic providers dataloaders do not work (extra queries and code complexity is just unjustifiable huge. Using some ORM like solution (that will deal with joins) will be much easier here

I was thinking of this too but worth a separate discussion/issue. Also we have to be careful with MongoDB, currently we work with normalized table, not sure how those the joins will work.

wtrocki:
I have tried another alternative would be to cleanup all loaders on before entering root resolvers by using something like: https://github.com/prisma-labs/graphql-middleware

However, this is also dependent on the execution plan (most of the implementations execute things per root resolver)

Your implementation is by no means the best generic take I can think off! Amazing job

PR #1889, opened 2020-08-18
fix: always prefer bson-ext if it is installed over bson

machi1990:
Follows https://github.com/aerogear/graphback/pull/1870/files#r472129139

### DIFF
--- a/packages/graphback-core/src/scalars/objectId.ts
+++ b/packages/graphback-core/src/scalars/objectId.ts
@@ -1,0 +1,27 @@
+import { ObjectID as BsonObjectID } from "bson";
+
+/* eslint-disable */
+export function isObjectID(value: any) {
+  if (value instanceof BsonObjectID) {
+    return true;
+  }
+  try {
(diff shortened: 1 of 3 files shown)


packages/graphback-core/src/scalars/objectId.ts
```diff
@@ -0,0 +1,27 @@
+import { ObjectID as BsonObjectID } from "bson";
+
+/* eslint-disable */
```
craicoverflow, line 3: Could a function scoped blanket disablement of ESLint mask other warnings in future?

machi1990: Theoretically yes, but I honestly do not see us adding anything else in these two functions.

craicoverflow: We probably won't yes, but there is a possibility that if our ESLint rules change and this function is affected it won't be noticed. Unlikely scenario anyway.

PR #1899, opened 2020-08-19
refactor(codegen-schema): case to IObjectTypeResolver instead if IFieldResolver

### LINKED ISSUE #1898: Use `IObjectTypeResolver` instead of `IFieldResolver` in SchemaCRUDPlugin
The SchemaCRUDPlugin is casting the resolver query, mutation and subscription objects to `IFieldResolver` whereas they should be `IObjectResolver`.

Everything is working fine, but let's correct this in case it causes problems in the future.

craicoverflow:
Closes #1898 

When writing the example in [Create your own Graphback plugin](https://graphback.dev/docs/next/plugins/create) example I noticed that this was done incorrectly in the SchemaCRUDPlugin. Everything still works the exact same as before.


machi1990 (approved):
Good catch! Thanks

PR #1901, opened 2020-08-19
feat(graphqlserve): add conflicts to --datasync

ssd71:
This PR adds conflicts to --datasync option of graphql serve with default strategy of clientsidewins which can be overridden by the --conflict=client or --conflict=server option

### DIFF
--- a/docs/graphqlserve/graphqlserve.md
+++ b/docs/graphqlserve/graphqlserve.md
@@ -107,6 +107,21 @@ Once we have a model with datasync capabilities, we can run our GraphQL server b
 ```bash
 gqlserve serve Note.graphql --datasync
 ```
+
+Conflict resolution strategies for datasync enabled models can be specified via the --conflict option:
+
+```bash
+gqlserve serve Note.graphql --datasync --conflict=clientSideWins
+```
+
+This defaults to ClientSideWins, if unset.
+
+The TTL for delta tables, can also be set using the --deltaTTL option:
+```bash
+gqlserve serve Note.graphql --datasync --deltaTTL=172800
+```
+
+This value defaults to `172800` when unused
  
 ### Printing your GraphQL schema
 
--- a/packages/graphql-serve/src/commands/serve.ts
+++ b/packages/graphql-serve/src/commands/serve.ts
@@ -1,7 +1,5 @@
 import yargs from 'yargs';
-import { serveHandler } from '../components/serveHandler';
-
-type Params = { model?: string, port?: number, datasync: boolean };
+import { serveHandler, GraphQLServeParams } from '../components/serveHandler';
 
 export const command = 'serve [modelDir] [options]';
 
@@ -25,9 +23,20 @@ export const builder = (args: yargs.Argv): void => {
     type: 'boolean',
     alias: 'ds'
   })
+
+  args.option('conflict', {
+    describe:"Specify a conflict resolution strategy with --datasync",
+    type: 'string',
+    choices: ["clientSideWins", "serverSideWins"]
+  })
+
+  args.option('deltaTTL', {
+    describe: "Specify a TTL for delta tables with --datasync",
+    type: "number"
+  })
   args.example('$0 serve . -p 8080', 'generate schema from data model files in current directory and start GraphQL server on port 8080')
 }
 
-export async function handler(args: Params): Promise<void> {
+export async function handler(args: GraphQLServeParams): Promise<void> {
   await serveHandler(args);
 }
--- a/packages/graphql-serve/src/components/serveHandler.ts
+++ b/packages/graphql-serve/src/components/serveHandler.ts
@@ -1,7 +1,11 @@
 import { buildGraphbackServer, GraphbackServer } from "../GraphbackServer";
+import { ConflictResolutionStrategyName, DataSyncServeConfig } from "../runtime";
 
-export const serveHandler = async (argv: { model?: string, port?: number, datasync }): Promise<GraphbackServer> => {
-  const server = await buildGraphbackServer(argv.model, !!argv.datasync);
+export type GraphQLServeParams = { model?: string, port?: number, datasync: boolean, conflict?: ConflictResolutionStrategyName, deltaTTL?: number };
+
+export const serveHandler = async (argv: GraphQLServeParams): Promise<GraphbackServer> => {
+  const datasyncServeConfig: DataSyncServeConfig = { datasync: !!argv.datasync, conflict: argv.conflict, deltaTTL: argv.deltaTTL}
+  const server = await buildGraphbackServer(argv.model, datasyncServeConfig);
 
   if ('port' in argv) {
     const portNumber = argv.port;
--- a/packages/graphql-serve/src/runtime.ts
+++ b/packages/graphql-serve/src/runtime.ts
@@ -1,12 +1,20 @@
 
-import { GraphbackAPI, buildGraphbackAPI } from 'graphback'
+import { GraphbackAPI, buildGraphbackAPI, filterModelTypes } from 'graphback'
 import { createMongoDbProvider } from '@graphback/runtime-mongo'
 import { MongoMemoryServer } from 'mongodb-memory-server';
 import { MongoClient, Db } from 'mongodb';
 import { loadModel } from './loadModel';
-import { GraphQLSchema } from 'graphql';
-import { DataSyncPlugin, createDataSyncMongoDbProvider, createDataSyncCRUDService } from "@graphback/datasync"
-import { PubSub } from 'graphql-subscriptions';
+import { GraphQLSchema, GraphQLObjectType } from 'graphql';
+import { createDataSyncAPI, DataSyncModelConfigMap, ClientSideWins, ServerSideWins, ConflictResolutionStrategy, isDataSyncType } from "@graphback/datasync"
+
+
+export type ConflictResolutionStrategyName = "clientSideWins"|"serverSideWins";
+
+export interface DataSyncServeConfig {
+  datasync: boolean,
+  conflict?: ConflictResolutionStrategyName,
+  deltaTTL?: number
+}
 
 export interface Runtime {
   schema: GraphQLSchema;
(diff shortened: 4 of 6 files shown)


packages/graphql-serve/src/runtime.ts
```diff
@@ -25,19 +28,45 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
...
+    return !!parseMetadata('datasync', type);
+  })
+
+  dataSyncModelTypes.forEach((type: GraphQLObjectType) => {
+    map[type.name] = {
+      enabled: true,
+      deltaTTL: 172800,
+      conflictResolution: conflictResolutionStrategy
+    }
+  })
```
machi1990, line 44: I suppose we need this because we do not have a way to supply default config?

packages/graphql-serve/src/runtime.ts
```diff
@@ -25,19 +28,45 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
...
+  const map: DataSyncModelConfigMap = {};
+
+  const dataSyncModelTypes = filterModelTypes(model).filter((type: GraphQLObjectType) => {
+    return !!parseMetadata('datasync', type);
+  })
+
+  dataSyncModelTypes.forEach((type: GraphQLObjectType) => {
+    map[type.name] = {
+      enabled: true,
+      deltaTTL: 172800,
```
machi1990, line 41: Maybe this could be an option too?

ssd71: Agreed :+1: 

@wtrocki wdyt

packages/graphql-serve/src/runtime.ts
machi1990: Not saying we should introduce it but more a question.

packages/graphql-serve/src/runtime.ts
```diff
@@ -25,19 +28,45 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
   return client
 }
 
+export function getConflictMapForModels(model: GraphQLSchema, conflictResolutionStrategy: ConflictResolutionStrategy): DataSyncModelConfigMap {
+  const map: DataSyncModelConfigMap = {};
+
+  const dataSyncModelTypes = filterModelTypes(model).filter((type: GraphQLObjectType) => {
+    return !!parseMetadata('datasync', type);
+  })
```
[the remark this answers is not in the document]
ssd71: It does actually, but it uses ModelDefinition instead of GraphQL Type, hence a bit troublesome to use it here:
https://github.com/aerogear/graphback/blob/bb774b287d6f387964b2cd740775e33b5984dc8e/packages/graphback-datasync/src/util.ts#L5-L7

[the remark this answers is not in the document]
machi1990: Great. We can add the below function along side this 
- `isDataSyncType(graphqlType)` 

Then the `isDataSyncModel(model: ModelDefinition)` would simply call this function with `model.graphqlType`. 

I guess what I am hoping to see is having this knowledge in one place where it belongs, and not scattered.

packages/graphql-serve/src/runtime.ts
```diff
@@ -25,19 +33,50 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
   return client
 }
 
+export function getConflictMapForModels(model: GraphQLSchema, conflictResolutionStrategy: ConflictResolutionStrategy, deltaTTL: number): DataSyncModelConfigMap {
+  const map: DataSyncModelConfigMap = {};
+
+  const dataSyncModelTypes = filterModelTypes(model).filter((type: GraphQLObjectType) => {
+    return !!isDataSyncType(type);
+  })
```
machi1990, line 41: Much better. 

I think it could be also written as a oneliner.

```ts
 const dataSyncModelTypes = filterModelTypes(model).filter(isDataSyncType)
```

Bu most important we do not need the `!!` since the function returns a boolean already.

packages/graphql-serve/src/runtime.ts
wtrocki: So we do nto have any way to supply default config? 
How this even works when I provide empty object?

This is must have!

packages/graphql-serve/src/runtime.ts
ssd71: Ahh, thanks! :D

packages/graphql-serve/src/runtime.ts
wtrocki: Good call!

packages/graphql-serve/src/runtime.ts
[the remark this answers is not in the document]
wtrocki: This entire thing looks like terrible hack for the missing feature (global config) that needs to be added.

packages/graphql-serve/src/runtime.ts
wtrocki: I think that method should not exist. We need global config in the plugin itself.

packages/graphql-serve/src/runtime.ts
```diff
@@ -25,19 +33,48 @@ export const createMongoDBClient = async (): Promise<MongoClient> => {
...
-      serviceCreator: createDataSyncCRUDService({ pubSub: new PubSub() }),
-      dataProviderCreator: createDataSyncMongoDbProvider(db),
-      plugins: [new DataSyncPlugin()]
+  if (datasyncServeConfig.datasync) {
+    let conflictResolutionStrategy = ClientSideWins;
+    let deltaTTL = 172800;
+    if (datasyncServeConfig.conflict === "client") {
+      conflictResolutionStrategy = ClientSideWins;
+    }
+    if (datasyncServeConfig.conflict === "server") {
```
wtrocki, line 66: Undocumented `server` value

packages/graphql-serve/src/runtime.ts
machi1990: :-) Yeap. I did not want to make call for having a global default config.

That should solve it. I am sorry @ssd71 for making you do rounds on this. :-(

packages/graphql-serve/src/runtime.ts
ssd71: server is supposed to mean serverSideWins, should I expand "server" to "serverSideWins"? Not sure I understand

ssd71: Do you mean that I should describe both options:
https://github.com/aerogear/graphback/blob/f5f02b58740dcdb6d7e13943b80dc3c6b6c3a472/packages/graphql-serve/src/commands/serve.ts#L30

packages/graphql-serve/src/runtime.ts
ssd71: > So we do nto have any way to supply default config?
How this even works when I provide empty object?

Default value for conflictResolution? Built into plugin

Do you mean default value for deltaTTL?

packages/graphql-serve/src/commands/serve.ts
```diff
@@ -25,6 +25,17 @@ export const builder = (args: yargs.Argv): void => {
     type: 'boolean',
     alias: 'ds'
   })
+
+  args.option('conflict', {
+    describe:"Specify a conflict resolution strategy with --datasync",
+    type: 'string',
+    choices: ["client", "server"]
```
craicoverflow, line 32: In my opinion having just "client" and "server" is not obvious enough as to what they do. If the choices were `serverSideWins` or `clientSideWins` it would become much clearer, and self-documenting.

packages/graphql-serve/src/components/serveHandler.ts
```diff
@@ -1,7 +1,8 @@
 import { buildGraphbackServer, GraphbackServer } from "../GraphbackServer";
+import { ConflictResolutionStrategyName } from "../runtime";
 
-export const serveHandler = async (argv: { model?: string, port?: number, datasync }): Promise<GraphbackServer> => {
-  const server = await buildGraphbackServer(argv.model, !!argv.datasync);
+export const serveHandler = async (argv: { model?: string, port?: number, datasync: boolean, conflict?: ConflictResolutionStrategyName, deltaTTL?: number }): Promise<GraphbackServer> => {
```
craicoverflow, line 4: The number of properties of `argv` is growing rather long. It might be time to create an interface for this?

```suggestion
export const serveHandler = async (args: GraphQLServeArgs): Promise<GraphbackServer> => {
```

packages/graphql-serve/src/components/serveHandler.ts
```diff
@@ -1,7 +1,8 @@
...
+  const server = await buildGraphbackServer(argv.model, { datasync: !!argv.datasync, conflict: argv.conflict, deltaTTL: argv.deltaTTL});
```
craicoverflow, line 5: It would be neater to create an interface before passing the argument:

```suggestion
  const server = await buildGraphbackServer(argv.model, dataSyncServeConfig);
```

Easier to read, easier to modify, easier to scale.

packages/graphql-serve/src/runtime.ts
ssd71: > I suppose we need this because we do not have a way to supply default config?

@machi1990 Do you mean default conflict config itself or default deltaTTL?

Conflict config is used to `enable` conflicts as well, so I think maybe putting a default conflict config in the plugin would mean enabling conflicts for all models, do you mean something like a global config?

ssd71:
@machi1990 @wtrocki 
I think all that's left to address in this PR is the lack of a global conflict config, I would like it if I could have a bit more time discussing about that and creating another PR about it instead of hastily adding it in this PR, as it does not appear like a trivial thing to me, what do you guys think? Sorry if I left something out

packages/graphql-serve/src/components/serveHandler.ts
ssd71: Thanks for the review!

docs/graphqlserve/graphqlserve.md
```diff
@@ -107,6 +107,21 @@ Once we have a model with datasync capabilities, we can run our GraphQL server b
 ```bash
 gqlserve serve Note.graphql --datasync
 ```
+
+Conflict resolution strategies for datasync enabled models can be specified via the --conflict option:
+
+```bash
+gqlserve serve Note.graphql --datasync --conflict=client
```
machi1990, line 114: I think it is `clientSideWins` now instead of `client`

machi1990:
> @machi1990 @wtrocki
> I think all that's left to address in this PR is the lack of a global conflict config, I would like it if I could have a bit more time discussing about that and creating another PR about it instead of hastily adding it in this PR, as it does not appear like a trivial thing to me, what do you guys think? Sorry if I left something out

Hey, +1 on dealing with that on a separate issue / PR.

docs/graphqlserve/graphqlserve.md
ssd71: Done this now. Thanks!

wtrocki:
Happy to merge this!

packages/graphql-serve/src/runtime.ts
machi1990: Let's discuss here https://github.com/aerogear/graphback/issues/1902

machi1990 (approved):

PR #1903, opened 2020-08-19
Install knex as a peer dependency

### LINKED ISSUE #1640: Should Knex be a peerDependency on `@graphback/runtime-knex`?
Knex is required to be installed into the template to create the database. Knex is also installed as a dependency of `@graphback/runtime-knex`.

This might make it difficult for users to add Graphback to their existing project if they already use a different version of Knex to our libraries. 

Would there be a benefit of letting this be acquired as a peerDependency from the template?

craicoverflow:
Resolves #1640

### Verification

To verify, you will want a template outside of this monorepo.

Install Graphback to the project like so:

```
"dependencies": {
  "@graphback/runtime-knex": "/home/username/code/aerogear/graphback/packages/graphback-runtime-knex"
}
```

Now, you can change your Knex version to anything in 0.20.x || 0.21.x and your application will run and you can perform database calls.

### DIFF
--- a/packages/graphback-runtime-knex/package.json
+++ b/packages/graphback-runtime-knex/package.json
@@ -39,10 +39,10 @@
   "dependencies": {
     "@graphback/core": "0.15.1",
     "dataloader": "2.0.0",
-    "graphql-subscriptions": "1.1.0",
-    "knex": "0.21.4"
+    "graphql-subscriptions": "1.1.0"
   },
   "peerDependencies": {
+    "knex": "0.20.x || 0.21.x",
     "graphql": "^14.0.0 || ^15.0.0"
   }
 }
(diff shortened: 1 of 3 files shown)


wtrocki (approved):

packages/graphback-runtime-knex/package.json
```diff
@@ -39,10 +39,10 @@
   "dependencies": {
     "@graphback/core": "0.15.1",
     "dataloader": "2.0.0",
-    "graphql-subscriptions": "1.1.0",
-    "knex": "0.21.4"
+    "graphql-subscriptions": "1.1.0"
   },
   "peerDependencies": {
+    "knex": "0.20.x || 0.21.x",
```
machi1990, line 45: Will this work if `0.22.0` is released?

craicoverflow: It will give a warning, but it will still work yes. We could add 0.22.x || 0.23.x

machi1990: Thanks

PR #1906, opened 2020-08-20
fix(schema): missing Date filters

### LINKED ISSUE #1904: Custom DateTime filter no longer available when upgrading from 0.16.0-alpha5 to 0.16.0-alph6
Here's an example filter in `0.16.0-alpha5`:

```graphql
type Incoming_queue_itemsFilter {
id: IDInput
call_id: IDInput
created_at: DateTimeInput
handled_at: DateTimeInput
incoming_call_type: StringInput
pod_id: IDInput
updated_at: DateTimeInput
user_call_sid: StringInput
and: [Incoming_queue_itemsFilter!]
or: [Incoming_queue_itemsFilter!]
not: Incoming_queue_itemsFilter
}
```
When I upgrade to `0.16.0-alpha6` the filter looks like this:

```graphql
type Incoming_queue_itemsFilter {
id: IDInput
call_id: IDInput
incoming_call_type: StringInput
pod_id: IDInput
user_call_sid: StringInput
and: [Incoming_queue_itemsFilter!]
or: [Incoming_queue_itemsFilter!]
not: Incoming_queue_itemsFilter
}
```
I have been using a simple `scalar DateTime` just fine until `0.16.0-alph6`, and there is no mention of this change in the release notes. I hope it's just a bug and that this functionality hasn't been removed.

craicoverflow:
Fixes #1904

### DIFF
--- a/docs/crud/queries.md
+++ b/docs/crud/queries.md
@@ -203,7 +203,7 @@ input BooleanInput {
 }
 ```
 
-Graphback generates a filter input for the [Graphback scalars](../model/) too. For convenience, a filter input for `Date`, `DateTime`, `Time`, `Timestamp` scalars is also generated; however we do not officially support these scalars and recommend you to use the Graphback scalars.
+Graphback generates a filter input for the [Graphback scalars](../model/scalars.md) too. For convenience, a filter input for `Date`, `DateTime`, `Time`, `Timestamp` scalars is also generated; however we do not officially support these scalars and recommend you to use the Graphback scalars.
 
 ### Pagination
 
--- a/docs/crud/queries.md
+++ b/docs/crud/queries.md
@@ -203,6 +203,8 @@ input BooleanInput {
 }
 ```
 
+Graphback generates a filter input for the [Graphback scalars](../model/) too. For convenience, a filter input for `Date`, `DateTime`, `Time`, `Timestamp` scalars is also generated; however we do not officially support these scalars and recommend you to use the Graphback scalars.
+
 ### Pagination
 
 You can perform pagination with the `page` argument.
--- a/packages/graphback-core/src/runtime/QueryFilter.ts
+++ b/packages/graphback-core/src/runtime/QueryFilter.ts
@@ -29,7 +29,11 @@ export const FILTER_SUPPORTED_SCALARS = [
   'GraphbackTimestamp',
   'GraphbackTime',
   'GraphbackDate',
-  'GraphbackDateTime'
+  'GraphbackDateTime',
+  'Timestamp',
+  'Time',
+  'Date',
+  'DateTime'
 ];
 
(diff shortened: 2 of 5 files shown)


packages/graphback-core/src/runtime/QueryFilter.ts
```diff
@@ -29,7 +29,11 @@ export const FILTER_SUPPORTED_SCALARS = [
   'GraphbackTimestamp',
   'GraphbackTime',
   'GraphbackDate',
-  'GraphbackDateTime'
+  'GraphbackDateTime',
+  'Timestamp',
+  'Time',
+  'Date',
+  'DateTime'
```
machi1990, line 36: Good quick fix. 

It looks like we do not generate filters for all the custom scalars, but known ones (as in 0.15 release). Maybe we should add it as breaking change and mention it in the filters docs as a `caution`. What do you think?

craicoverflow: Yeah, I am on board with that!

machi1990: Thank you :-)

craicoverflow: How is this? https://github.com/aerogear/graphback/pull/1906/commits/fff7a323c681f04d72b6c689a7fc436474f46ac3

docs/crud/queries.md
```diff
@@ -203,6 +203,8 @@ input BooleanInput {
 }
 ```
 
+Graphback generates a filter input for the [Graphback scalars](../model/) too. For convenience, a filter input for `Date`, `DateTime`, `Time`, `Timestamp` scalars is also generated; however we do not officially support these scalars and recommend you to use the Graphback scalars.
```
machi1990, line 206:
```suggestion
Graphback generates a filter input for the [Graphback scalars](../model/scalars.md) too. For convenience, a filter input for `Date`, `DateTime`, `Time`, `Timestamp` scalars is also generated; however we do not officially support these scalars and recommend you to use the Graphback scalars.
```

packages/graphback-core/src/runtime/QueryFilter.ts
machi1990: Great. I added a small suggestion. 

Also with https://github.com/aerogear/graphback/pull/1870 we disable support for unknown scalars. What do you think about adding it to the release note?

craicoverflow: Thanks! Added now in https://github.com/aerogear/graphback/pull/1906/commits/2358b6ad0d522a1e575c98dedc5fa8d0069c3188

machi1990 (approved):
Thanks for the quick fix and docs update. All good!

PR #1912, opened 2020-08-20
chore: update auto labelling rules

machi1990:


machi1990:
I got the regex wrong thus some labels were not working. This should fix it.

/cc @wtrocki @craicoverflow

craicoverflow (approved):

Issue #1915, opened 2020-08-20
Types missing for 'bson' in @graphback/core


Eunovo:
* **Module**: @graphback/core
* **Version**: 0.16.0-beta1
* **yarn version:** 0.1.2
* Screenshot of console log

![image](https://user-images.githubusercontent.com/37949128/90776044-808de500-e2f1-11ea-9456-0864ba9bf985.png)

I get above error when I try to compile my typescript project

machi1990:
Automatically generated comment to notify maintainers
/cc @craicoverflow, @machi1990, @wtrocki

machi1990:
Is it the Graphback project that you are compiling?

Eunovo:
No. I'm using the graphback-core in offix

machi1990:
Ah strange that it tries to compile an external dependency (might be how the project is configured).

wtrocki:
Yep... This is due to using it within CLI. This can be replicated by using generate from cli.

PR #1916, opened 2020-08-20
chore(deps): add bson types

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1915

/cc @Eunovo maybe this will fix the issue for you.

### DIFF
--- a/packages/graphback-core/package.json
+++ b/packages/graphback-core/package.json
@@ -28,6 +28,7 @@
     "node": ">=10.15.3"
   },
   "devDependencies": {
+    "@types/bson": "4.0.2",
     "@types/jest": "26.0.9",
     "@types/node": "12.12.54",
     "@types/pino": "6.3.0",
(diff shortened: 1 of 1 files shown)


packages/graphback-core/package.json
```diff
@@ -58,7 +59,7 @@
     "pluralize": "8.0.0"
   },
   "peerDependencies": {
-    "graphql": "^14.0.0 || ^15.0.0",
-    "bson-ext": "2.0.5"
+    "bson-ext": "2.0.5",
```
craicoverflow, line 62: Should this be fixed, or should we allow a range of versions, like `2.x.x`?

craicoverflow (approved):

packages/graphback-core/package.json
machi1990: Done.

craicoverflow (approved):

PR #1923, opened 2020-08-21
Transient field

craicoverflow:
This PR adds a `@transient` field annotation. This will:

- Omit the field from all generated input types
- Skip column creation in graphql-migrations

This PR also replaces `@db(skip: true)` field annotation (see https://github.com/aerogear/graphback/commit/85d50f3c332ee35c46fbe8e6c3e81d97ae60db7b), as there is no need to maintain two (this creates confusion).

### Verification

Use the `@transient` annotation on a field:

```graphql
"""
@model
"""
type User {
  id: ID!
  """@transient"""
  computedField: String!
}
```

The CreateUserInput, UserFilter, UserSubscriptionFilter and MutateUserFilter will not have this field.


craicoverflow:
Thanks, yes I found `@ignore` to be very ambiguous.

PR #1935, opened 2020-08-24
chore: make the benchmark package private to avoid publishing it

machi1990:
The package was published as https://www.npmjs.com/package/graphback-benchmarks, let's make it
private to avoid such cases.

### DIFF
--- a/performance/package.json
+++ b/performance/package.json
@@ -1,6 +1,7 @@
 {
   "name": "graphback-benchmarks",
   "version": "0.15.1",
+  "private": true,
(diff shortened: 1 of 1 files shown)


wtrocki (approved):

machi1990:
@wtrocki thanks. 

I think we should un-publish the package too. Thoughts? 

/cc @craicoverflow

craicoverflow:
> I think we should un-publish the package too. Thoughts?

Yup for sure.

Issue #1925, opened 2020-08-21
Avoid querying relationship field from GraphbackService (database) when relationship owner contains full dataset


machi1990:
Consider the business model below:

```graphql
""" @model """
type Comment {
  id: ID!
  text: String
  createdAt: GraphbackDate
  """
  @oneToOne
  """
  metadata: CommentMetadata
}

""" @model """
type CommentMetadata {
  id: ID!
  metadata: GraphbackJSON
  """
  @oneToOne
  """
  comment: Comment
}

type Query {
  todaysCommentWithMetadata: [Comment]
}
scalar GraphbackJSON
scalar GraphbackDate
```

**Is your feature request related to a problem? Please describe.**

The code of my custom resolver `todaysCommentWithMetadata`, returns an array of `Comment` each with their associated `Metadata` all in a single database query. However, when I enabled database query logging, I noticed two queries being sent to the database 

1. My custom database query
2. The query to resolve relationship. **This should not happen since the first query has all the data**

**Describe the solution you'd like**

Just one query sent to the database as my custom resolver contain the full dataset and there is no need to re-fetch the "metadata" field from database .

We should only retrieve the relationship field from database if it is not in the parent object. 
 
With something like 
```ts
if (parent[relationOwner]) {
   return parent[relationOwner];
}
```

This should happen before the following lines:
- https://github.com/aerogear/graphback/blob/7d4f80c47cb5ff0c58da9dd493c424ffd0015fcb/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L697
- https://github.com/aerogear/graphback/blob/7d4f80c47cb5ff0c58da9dd493c424ffd0015fcb/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L732

**Describe alternatives you've considered**
N/A

craicoverflow:
Seems very sensible, +1.

PR #1937, opened 2020-08-24
fix(schema-plugin): make sure that "null"/"undefined" relationship values are considered resolved

machi1990:
Improve the condition that checked if parent object has the relationship resolved. This also take
care of when the relationship field is "undefined" or "null". With the previous check, we'll query
the database again which seemed wrong. I've added a some data values in the previous tests.

Related to https://github.com/aerogear/graphback/pull/1933 and
https://github.com/aerogear/graphback/pull/1925

### DIFF
--- a/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
+++ b/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
@@ -694,7 +694,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     const model = modelNameToModelDefinition[modelName];
 
     resolverObj[relationOwner] = (parent: any, args: any, context: GraphbackContext, info: GraphQLResolveInfo) => {
-      if (parent[relationOwner]) {
+      if (Object.keys(parent).includes(relationOwner)) {
         return parent[relationOwner];
       }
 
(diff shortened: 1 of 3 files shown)


packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
```diff
@@ -694,7 +694,7 @@ export class SchemaCRUDPlugin extends GraphbackPlugin {
     const model = modelNameToModelDefinition[modelName];
 
     resolverObj[relationOwner] = (parent: any, args: any, context: GraphbackContext, info: GraphQLResolveInfo) => {
-      if (parent[relationOwner]) {
+      if (relationOwner in parent) {
```
craicoverflow, line 697: The `in` key can lead to false positives or negatives:

```ts
'constructor' in obj; // true
'__proto__' in obj; // true
'hasOwnProperty' in obj; // true

obj.hasOwnProperty('constructor'); // false
obj.hasOwnProperty('__proto__'); // false
obj.hasOwnProperty('hasOwnProperty'); // false
```

I can't see this being a problem for the `parent` arg, but wanted to highlight it in case.

https://masteringjs.io/tutorials/fundamentals/hasownproperty#:~:text=The%20key%20difference%20is%20that,property%2C%20and%20a%20hasOwnProperty%20function.

However, `hasOwnProperty` is not available in TS yet, another solution is to make our own: https://fettblog.eu/typescript-hasownproperty/

craicoverflow (approved):

packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts
machi1990: Really good point. 

`hasOwnProperty in obj` is an oversight from me. 

Let me fix that.

PR #1938, opened 2020-08-24
Small improvements around CI and tests

### LINKED ISSUE #1921: ability to ignore CI runs when patch does not affect source files
We should not run ci when we modify docs etc
Example
- https://github.com/aerogear/graphback/pull/1854
- https://github.com/aerogear/graphback/pull/1786

and couple of other doc change related PRs.

### LINKED ISSUE #1936: createIndexesTest is unstable
I've seen the test failing in separate random occasions and had to re-start CI to make it pass. 

The failure is: 
```
FAIL  packages/graphback-runtime-mongodb/tests/createIndexesTest.ts
  ● MongoDB indexing › can create relation indexes

    MongoError: ns does not exist: test.comment

      at MessageStream.messageHandler (node_modules/mongodb/lib/cmap/connection.js:266:20)
      at processIncomingData (node_modules/mongodb/lib/cmap/message_stream.js:144:12)
      at MessageStream._write (node_modules/mongodb/lib/cmap/message_stream.js:42:5)
```

See https://app.circleci.com/pipelines/github/aerogear/graphback/7359/workflows/259093cd-6792-4e23-bb39-b7632159eb02/jobs/7506 for example

machi1990:
Fixes #1936 
Fixes #1921 

### Tasks
- [x] Ignore deps update via renovate for the performance benchmark folder
- [x] Brings in GH actions
- [x] Improve issue template
- [x] Fixes failing index tests for MongoDB
- [x] Investigate failing website build
- [x] Cache dependencies
- [x] IT tests against two major releases of each database we support e.g PG 9 and 12, Mongo 3 and 4 
- [x] Cancel Previous runs
 
Hope to bring the above changes in and I'll look into the below items next: 

### As separate manual task and required before merging this PR
- [x] Disable Circle CI
- [x] Generate an NPM token and add it in GH secrets

### On Follow up PR
- [ ] Add CI to deploy website automatically when new release is created.


machi1990:
@ssd71 I asked you for a review, since you might have a better idea around flasky indexes tests.

ssd71:
While testing for indexes themselves is a good idea, and shouldn't take an unreasonable amount of time because the test databases should be empty, do you think we're better off mocking or `jest.spy`ing the createIndexes function? We could test the function itself separately perhaps.

Feel free to ignore if that sounds like a crazy idea in this context :P

machi1990:
> While testing for indexes themselves is a good idea, and shouldn't take an unreasonable amount of time because the test databases should be empty, do you think we're better off mocking or `jest.spy`ing the createIndexes function? We could test the function itself separately perhaps.
> 
> Feel free to ignore if that sounds like a crazy idea in this context :P

It is a good idea, but we can stick with the current tests. From my understanding, the reason they are failing is that the createIndexes function is async which probably means that in some random occasion the function has not been full resolved once the assertion is done.

.github/workflows/build-website.yml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    branches:
+      - master
+    paths:
+      - 'docs'
+      - 'website'
+      - 'package.json'
+      - 'packages/**/src'
+
+  pull_request:
```
craicoverflow, line 13: So this builds the website when a PR is opened? Is this to validate that it _can_ build?

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+env:
+  CI: true
+  NODE_OPTIONS: --max_old_space_size=4096
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:13
```
craicoverflow, line 16: We could run this on multiple versions (9,10,11,12) with a [matrix](https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#configuring-a-build-matrix).

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_USER: postgresql
+          POSTGRES_DB: users
+        ports:
+          - 5432:5432
+        # needed because the postgres container does not provide a healthcheck
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+      mongodb:
+        image: mongo:4.2.9
```
craicoverflow, line 27: Could we test multiple versions with a Matrix?

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,82 @@
...
+
+env:
+  NODE_OPTIONS: --max_old_space_size=4096
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:13
```
craicoverflow, line 47: We could run this on multiple versions (9,10,11,12) with a [matrix](https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#configuring-a-build-matrix).

craicoverflow (dismissed):
Looks great! Any way of verifying the release will work?

I've left some comments around further improvements that could be made through testing multiple versions of Postgres, Mongo.

.github/workflows/release.yml
wtrocki: I think github actions have fair usage so we need to be careful to not hit the quota ;D

.github/workflows/lint-and-test.yml
wtrocki: But this will mean much longer builds. Longer builds === productivity loss

.github/workflows/build-website.yml
machi1990: Yes, we are just verifying that it can build and we have not regressed in any way in this regard. Planning to add auto-deploy on each release at least. But that will be on separate PR.

.github/workflows/lint-and-test.yml
wtrocki: If we would be something like knex that would make sense but since we use wrapper library we rely on their capabilities there.

.github/workflows/build-website.yml
```diff
@@ -0,0 +1,42 @@
...
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - master
+    paths:
+      - 'docs'
+      - 'website'
+      - 'package.json'
+      - 'packages/**/src'
```
wtrocki, line 21: Probably not needed.

.github/workflows/release.yml
machi1990: Yes, very good point. I initially started with this but the build was slow and decided to drop it. 

Let me add it to things that can be improved in followup issue.

.github/workflows/release.yml
machi1990: Same reason as https://github.com/aerogear/graphback/pull/1938#discussion_r476282885. But testing with at least two major release makes a lot of sense.

.github/workflows/build-website.yml
machi1990: Needed because we are building API docs. We want to make sure that the generated links have not broken.

.github/workflows/release.yml
craicoverflow: Okay, we could settle on only having it on release action, we would unlikely hit the quota then

.github/workflows/lint-and-test.yml
craicoverflow: Okay, we could settle on only having it on release action, we would unlikely hit the quota then

.github/workflows/build-website.yml
craicoverflow: Excellent.

.github/workflows/release.yml
machi1990: I'll do it in this PR. I'll narrow the matrix to two major releases per database.
PG: v9 and v13
Mongo: v3 and v4

What do you think?

.github/workflows/release.yml
machi1990: :+1: Good point.

.github/workflows/release.yml
craicoverflow: I see v13 of PG is still in beta so I am not sure if we should bother with v13?

machi1990:
@craicoverflow @wtrocki I've pushed some minor improvements here and there. Can you review please? 

PS: CircleCI build is failing because I have removed the config file. The plan is to deactivate it completely once we are happy with this workflow.

craicoverflow (approved):
Amazing!

wtrocki:
I would disable circleCI before merge so it will not confuse our branch protection rules

machi1990:
I have created https://github.com/aerogear/graphback/issues/1945 to automate website deployment via https://v2.docusaurus.io/docs/deployment/#triggering-deployment-with-github-actions

Issue #1941, opened 2020-08-24
Cannot read property 'and' of undefined


lastmjs:
graphback version: 0.16.0-alpha5

All of my subscriptions without filter now return this error:

```json
{
  "errors": [
    {
      "message": "Cannot read property 'and' of undefined",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "newIncoming_queue_items"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ]
}
```
The error above came from this subscription query:

```graphql
subscription {
  newIncoming_queue_items {
    id
  }
}
```
Any new subscription without a filter is doing this.

This is another example of something very core breaking when I upgrade. My subscriptions were all working very reliably before this change. Seems regressions are easily added. I'm trying to take graphback into production, but now I fear any updates

machi1990:
Automatically generated comment to notify maintainers
/cc @craicoverflow, @machi1990, @wtrocki

wtrocki:
Easy fix. Workaround is to pass empty object as filter

machi1990:
Hi @lastmjs thank for reporting this issue. It should be an easy fix to make sure that it's retro-compatible with previous releases. 
I'll take care of this and drop a beta release. 
 
> This is another example of something very core breaking when I upgrade. My subscriptions were all working very reliably before this change. Seems regressions are easily added. I'm trying to take graphback into production, but now I fear any updates

Our target is to make Graphback stable enough and we appreciate your inputs. Sorry for this issues you are running into, they are top priority for us. Testing is an area that we are giving much focus, it may arrive that some test cases are missing as of this case we are working on covering this missing pieces.

machi1990:
[0.16.0-beta2](https://github.com/aerogear/graphback/releases/tag/0.16.0-beta2) has been released and should effectively fix this issue.

PR #1942, opened 2020-08-25
fix(subscription): cannot read property 'and' of "undefined" subscription filter

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1941

### DIFF
--- a/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
+++ b/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
@@ -114,6 +114,7 @@ const predicateMap: IPredicate = {
  * @param {QueryFilter} filter - subscription filter input object
  */
 export function createInMemoryFilterPredicate<T = any>(filter: QueryFilter): (input: Partial<T>) => boolean {
+  filter = filter || {};
   const andFilter = filter.and;
   const orFilter = filter.or;
   const notFilter = filter.not;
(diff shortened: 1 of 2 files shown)


packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
```diff
@@ -966,4 +966,24 @@ describe('createInMemoryFilterPredicate', () => {
     expect(filterSubscription({ name: 'Homer Thompson' })).toEqual(true)
     expect(filterSubscription({ name: 'Bart Simpson' })).toEqual(false)
   });
-});
\ No newline at end of file
+
+  describe('empty or undefined filter', () => {
+    describe('empty filter', () => {
```
wtrocki, line 971: Doesn't seem to be empty to me :D

machi1990: I am passing an empty object. 

```ts
const filterSubscription = createInMemoryFilterPredicate<User>({})
```

packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
```diff
@@ -114,6 +114,7 @@ const predicateMap: IPredicate = {
  * @param {QueryFilter} filter - subscription filter input object
  */
 export function createInMemoryFilterPredicate<T = any>(filter: QueryFilter): (input: Partial<T>) => boolean {
```
craicoverflow, line 116:
```suggestion
export function createInMemoryFilterPredicate<T = any>(filter: QueryFilter = {}): (input: Partial<T>) => boolean {
```

craicoverflow (approved):

packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts
machi1990: Yeap, this will work when filter is `undefined` but not cases where filter is `null`, probably won't happen but keeping in case users write their own filter and pass `null`.

Issue #1627, opened 2020-07-01
All templates should have custom resolver example 


wtrocki:
All templates should have custom resolver example in order to show how graphback can be used with custom logic

machi1990:
Fixed by https://github.com/aerogear/graphback/pull/1956

Issue #1637, opened 2020-07-02
Graphback context do not have typings for services


wtrocki:

wtrocki:
See https://github.com/aerogear/graphback/pull/1643
Challenge here is that we still need to generate types from somewhere. 
More deeper integration with GraphQL-CLI codegen is probably needed to enable better local development regime :D

craicoverflow:
The only thing I can think of is by letting the SchemaCRUDPlugin generate the context file:

```ts
interface GeneratedContext {
  graphback: {
    services: {
      Note: GraphbackCRUDService
    }
  }
}
```

wtrocki:
No need for generation. We will not be able to get service names but.. connecting CRUDservices is trivial:

`[service:string: GraphbackCRUDService]`

If we start using generics and drop graphback namespace this will be:

`resolveUser(_, _, { services:{ User: GraphbackCRUDService<User>}})`

AKA type it yourself - user will come from codegen then.

craicoverflow:
The decomposition of the context wil be a little bit bulkier than what you proposed, as the typing needs to be decomposed separate to the context:

```ts
resolverFn: (_: any, __: any, { graphback: { services: Note } }: { graphback: { services: GraphbackCRUDService<Note> } })
```

But this is still the best solution.

craicoverflow:
The problem with this approach is that you no longer have the complete context object as it is decomposed as a parameter.

```ts
const customResolvers = {
  Query: {
    getDraftNotes: async (_: any, __: any, { graphback: Note }: { graphback: GraphbackCRUDService<Note> }) => {
      const results = await Note.findBy({
        filter: {
          title: {
            startsWith: '[DRAFT]'
          }
        } as QueryFilter<NoteFilter>
      }, context, info); // there is no context object anymore!!

      return results.items;
    }
  }
}
```

While this is an optional parameter, and in fact currently the only method that uses it in the CRUDService is `batchLoadData`, it should still be kept.

So the user has three choices:

1. Sacrifice passing context to CRUDService for typing in the resolver function.
2. Sacrifice typings for a full context.
3. Add boilerplate to the function by doing some decomposition and casting in the function body.

Neither way is perfect and has its downfalls.

craicoverflow:
**FOURTH APPROACH**

Another approach is that the user can define their own custom context interface to use in custom resolvers:

```ts
interface AppContext {
  graphback: {
    Note: GraphbackCRUDService<NoteFilter>
  }
}

[...]

PR #1944, opened 2020-08-25
Pass context and info to CRUD service, which will handle selected fields, aggregations etc.

craicoverflow:
See **Release notes** below for what has changed in this PR..

This PR improves the user API by decreasing reliance on user modification of the context in resolvers. Now you can create a custom resolver without the context if you wish:

```ts
const customResolvers = {
  Query: {
    getDraftNotes: async (_: any, __: any, context: GraphbackContext, info: GraphQLResolveInfo) => {
      const noteService = context.graphback.services.Note;

      const results = await noteService.findBy({
        filter: {
          title: {
            startsWith: '[WIP]'
          }
        }
      })

      return results.items;
    }
  }
}
```

It is a matter of passing the optional context and info to the CRUD layer, where Graphback will handle field selection, aggregrations etc.

This PR also adds other fixes in the same area (like usng generics instead of any)

#### Changed GraphbackCRUDService `findBy` method signature. This also applies to all services that implement this interface.

```patch
- findBy(filter: QueryFilter<Type>, context: GraphbackContext, page?: GraphbackPage, orderBy?: any): Promise<ResultList<Type>>;
+ findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo): Promise<ResultList<Type>>;
```

**args**

`findBy` now accepts a new interface, `FindByArgs`, which wraps the `filter`, `page` and `orderBy` optional arguments.

```ts
await noteService.findBy({
  filter: {
    id: {
      gt: 100
    }
  },
  page: {
    offset: 0,
    limit: 10
  },
  orderBy: {
    field: 'id'
  }
})
```

**context**

The context parameter is now optional.

**info**

You can now optionally pass the `GraphQLResolveInfo` info object to the CRUD service, to perform extra optimizations like retrieving only the selected fields from the query.

```ts
await noteService.findBy(filter, context, info);
```

#### `context` parameter removed from `subscribeToCreate`, `subscribeToDelete`, `subscribeToUpdate` methods in GraphbackCRUDService.

This method was unused.

#### Removed `context` parameter from all GraphbackDataProvider methods. This also applies to all providers that implement this interface.

All CRUD methods in `GraphbackDataProvider` had a `context` parameter used to get the selected fields. 
These have been replaced with (optional) `selectedFields` - you can pass a string array of the fields you want to select from the database.

```ts
await noteProvider.findBy(filter, ['id', 'name']);
```

#### Changed GraphbackDataProvider `findBy` method signature. This also applies to all providers that implement this interface.

**args**

`findBy` now accepts a new interface, `FindByArgs`, which wraps the `filter`, `page` and `orderBy` optional arguments.

```ts
await noteService.findBy({
  filter: {
    id: {
      gt: 100
    }
  },
  page: {
    offset: 0,
    limit: 10
  },
  orderBy: {
    field: 'id'
  }
})
```

#### Remove resolver options from GraphbackContext

Resolver options was removed from the context because the `count` aggregation and `selectedFields` extraction logic was moved to the CRUDService method.

#### CRUDService, DataSyncCRUDService now accepts a `ModelDefinition` as the first constructor parameter.

To instantiate a CRUDService you must pass the full `ModelDefinition` instead of the model name.

```ts
const myService = new CRUDService(modelDefinition, ...);
```

#### KeycloakCrudService now accepts a `ModelDefinition` as the first constructor parameter.

To instantiate a CRUDService you must pass the full `ModelDefinition` instead of the model name.

```ts
const myService = new KeycloakCrudService(modelDefinition, ...);
```

#### DataSyncProvider `sync` method signature has been changed

```patch
- sync(lastSync: Date, context: GraphbackContext, filter?: any, limit?: number): Promise<Type[]>;
+ sync(lastSync: Date, selectedFields?: string[], filter?: QueryFilter, limit?: number): Promise<Type[]>;
```

The `context` argument has been replaced with (optional) `selectedFields`.

#### `context` parameter is removed from the `create` method in `MongoDBDataProvider` and `DatasyncMongoDBDataProvider` providers.

This parameter did not do anything.

## Follow up

- This enables https://github.com/aerogear/graphback/issues/1627 to be completed.

### DIFF
--- a/integration/tests/runtime-workflow-postgres.ts
+++ b/integration/tests/runtime-workflow-postgres.ts
@@ -13,11 +13,11 @@ import { loadDocuments } from '@graphql-tools/load';
 import * as Knex from 'knex';
 import { buildGraphbackAPI, GraphbackAPI } from "graphback/src";
 import { DocumentNode } from 'graphql';
+import { ObjectID } from 'mongodb';
 import { migrateDB } from '../../packages/graphql-migrations/src';
 import { createKnexDbProvider } from "../../packages/graphback-runtime-knex";
 import { SchemaCRUDPlugin } from '../../packages/graphback-codegen-schema';
 import { ClientCRUDPlugin } from '../../packages/graphback-codegen-client';
-import { ObjectID } from 'mongodb';
 
 /** global config */
 let db: Knex;
@@ -113,6 +113,25 @@ query getNoteWithFullDataSet {
         ratings
       }
     }
+}`;
+
+const FIND_NOTES_WITH_COUNT = gql`
+query findNotes {
+  findNotes {
+    items {
+      id
+    }
+    count
+  }
+}`;
+
+const FIND_NOTES_WITHOUT_COUNT = gql`
+query findNotes {
+  findNotes {
+    items {
+      id
+    }
+  }
 }`;
 
 beforeAll(async () => {
@@ -283,6 +302,26 @@ test('Find all notes', async () => {
   })
 })
 
+test('Find all notes with count', async () => {
+  const { data, errors } = await client.query({
+    operationName: 'findNotes',
+    query: FIND_NOTES_WITH_COUNT
+  })
+
+  expect(data).toBeDefined()
+  expect(data.findNotes.count).toBeDefined()
+})
+
+test('Find all notes without count', async () => {
+  const { data } = await client.query({
+    operationName: 'findNotes',
+    query: FIND_NOTES_WITHOUT_COUNT
+  })
+
+  expect(data).toBeDefined()
+  expect(data.findNotes.count).toBeUndefined()
+})
+
 test('Find all notes except the first', async () => {
   const { data } = await client.query({
     operationName: 'findNotes',
@@ -407,15 +446,16 @@ test('Find all comments', async () => {
 })
 
 test('filter comments using built in scalars', async () => {
-  const { data } = await client.query({ operationName: "findComments", query: documents,
-  variables: {
-    filter: {
-      createdAt: {
-        eq: createdAt.getTime()
+  const { data } = await client.query({
+    operationName: "findComments", query: documents,
+    variables: {
+      filter: {
+        createdAt: {
+          eq: createdAt.getTime()
+        }
       }
     }
-  }
-});
+  });
 
   expect(data).toBeDefined();
   expect(data.findComments).toEqual({
@@ -586,44 +626,46 @@ test('Delete Note 1', async () => {
 });
 
 test('getComment in simultaneous with different resolvers pinfo', async () => {
-  const { data } = await client.query({ operationName:"getCommentDetailsAtOnce", query: GET_COMMENTS , variables: {
-    id: 1
-  }});
+  const { data } = await client.query({
+    operationName: "getCommentDetailsAtOnce", query: GET_COMMENTS, variables: {
+      id: 1
+    }
+  });
 
   expect(data).toEqual({
-       firstQuery:{
-          id: "1",
-          note:{
-             id:"1",
-             comments:[
-                {
-                   text:"Note A Comment"
-                },
-                {
-                   text:"Note A Comment 2"
-                }
-             ]
+    firstQuery: {
+      id: "1",
+      note: {
+        id: "1",
+        comments: [
+          {
+            text: "Note A Comment"
+          },
+          {
+            text: "Note A Comment 2"
           }
-       },
-       secondQuery:{
-          id: "1",
-          note:{
-             title:"Note 1 New Title",
-             comments:[
-                {
-                   id:"1"
-                },
-                {
-                   id:"2"
-                }
-             ]
+        ]
+      }
+    },
+    secondQuery: {
+      id: "1",
+      note: {
+        title: "Note 1 New Title",
+        comments: [
+          {
+            id: "1"
+          },
+          {
+            id: "2"
           }
-       }
- });
+        ]
+      }
+    }
+  });
 })
 
 test('get full dataset from custom query without querying the database again', async () => {
-  const { data } = await client.query({ operationName:"getNoteWithFullDataSet", query: GET_NOTE_WITH_CUSTOM_QUERY });
+  const { data } = await client.query({ operationName: "getNoteWithFullDataSet", query: GET_NOTE_WITH_CUSTOM_QUERY });
   customNoteWithFullDataSet.comments[2].metadata = null;
   expect(data.getNoteWithFullDataSet).toEqual(customNoteWithFullDataSet);
 })
--- a/packages/graphback-core/src/runtime/CRUDService.ts
+++ b/packages/graphback-core/src/runtime/CRUDService.ts
@@ -105,12 +105,12 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
     return this.db.findOne(args, selectedFields);
   }
 
-  public async findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo): Promise<ResultList<Type>> {
+  public async findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo, path?: string): Promise<ResultList<Type>> {
     let selectedFields: string[];
     let requestedCount: boolean = false;
     if (info) {
-      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model, 'items');
-      requestedCount = getResolverInfoFieldsList(info).some((field: string) => field === "count");
+      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model, path);
+      requestedCount = path === 'items' && getResolverInfoFieldsList(info).some((field: string) => field === "count");
     }
 
     const items: Type[] = await this.db.findBy(args, selectedFields);
--- a/packages/graphback-core/src/runtime/CRUDService.ts
+++ b/packages/graphback-core/src/runtime/CRUDService.ts
@@ -1,8 +1,13 @@
 import * as DataLoader from "dataloader";
 import { PubSubEngine, withFilter } from 'graphql-subscriptions';
+import { GraphQLResolveInfo } from 'graphql';
 import { GraphbackCRUDGeneratorConfig, GraphbackOperationType, upperCaseFirstChar, getSubscriptionName } from '..';
+import { ModelDefinition } from '../plugin/ModelDefinition';
+import { getSelectedFieldsFromResolverInfo, getResolverInfoFieldsList } from '../plugin/getSelectedFieldsFromResolverInfo';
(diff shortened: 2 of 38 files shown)


wtrocki:
This ranks as the most awesome PR of this year! Adding to my queue to check even further.
Amazing work!

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -72,40 +79,51 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
...
 
-    if (context.graphback.options?.aggregations?.count) {
-      count = await this.db.count(filter);
+  public async findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo): Promise<ResultList<Type>> {
+    let selectedFields: string[];
+    let requestedCount: boolean = false;
+    if (info) {
+      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model, 'items');
+      requestedCount = getResolverInfoFieldsList(info).some((field: string) => field === "count");
     }
```
machi1990, line 114: What  will be the resulting query when a `oneToOne` or `manyToOne` relationship is being resolved? 
- https://github.com/aerogear/graphback/blob/6a6deec2c5189456ab08c2fd10f6315ec2dd26fe/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L718

?

packages/graphback-datasync/src/services/DataSyncCRUDService.ts
```diff
@@ -12,16 +13,16 @@ export interface SyncList<T> {
...
   }
   /**
    * sync
    * For delta queries
    */
-  public async sync(lastSync: Date, context: GraphbackContext, filter?: any, limit?: number): Promise<SyncList<T>> {
-
-    const res = await (this.db as DataSyncProvider).sync(lastSync, context, filter, limit);
+  public async sync(lastSync: Date, info?: GraphQLResolveInfo, filter?: any, limit?: number): Promise<SyncList<T>> {
+    const selectedFields = getSelectedFieldsFromResolverInfo(info, this.model)
```
machi1990, line 24: Should we check if `info` is defined?

machi1990 (commented):
I'd have loved to actually not pass the `GraphQLResolverInfo` down to service layer but if we consider that layer to be like a `Controller` or `Router` layer then I can live with that. 

Added two questions, otherwise this looks good from first look. I'll have a second detailed look later on.

packages/graphback-runtime-knex/tests/service/DefaultCRUDServiceTest.ts
```diff
@@ -262,94 +244,11 @@ test('find users where name starts with "John"', async () => {
...
+  const result = await services.User.findBy({ filter: { name: { startsWith: 'John' } } })
 
   expect(result.items).toHaveLength(2)
 })
 
-test('find and count all users', async () => {
-  const { services } = await setup({
-    schemaSDL: `
-    """
-    @model
```
machi1990, line 274: I suppose this block is deleted because it was hard to fake an `info` object?

packages/graphback-datasync/src/services/DataSyncCRUDService.ts
craicoverflow: Yes, good spot.

packages/graphback-runtime-knex/tests/service/DefaultCRUDServiceTest.ts
craicoverflow: Yes, that is an unfortunate compromise. I tried for some time yesterday to create a mock info object but to no avail, and to my knowledge there is no library to do it.

To make up for the loss of this tested I added an integration test: https://github.com/aerogear/graphback/blob/6a6deec2c5189456ab08c2fd10f6315ec2dd26fe/integration/tests/runtime-workflow-postgres.ts#L305

machi1990: Perfect, I noticed those. We can live with that. Thanks for checking it out.

packages/graphback-core/src/runtime/CRUDService.ts
craicoverflow: Damn, that looks wrong. Need to check it out further.

craicoverflow: Yeah this is something I overlooked. This assumes the path will always fetch fields at the items node.

craicoverflow: The only reliable way I can think of doing this is by adding a `path` argument to the findBy method:

```ts
findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo, path?: string): Promise<ResultList<Type>>;
```

I think there is no way to be 100% sure that the path is correct if we try to do this dynamically in the CRUD method. path would be opt in, and in most cases would be undefined, except for the `find<T>` resolver.

machi1990: That's what I was thinking of also.

machi1990: And I think the only place that will need changing (to pass the `path` arg) is for `findXXXs` resolver.

craicoverflow: Yep you are correct I found this while testing it out.

machi1990: Cool!

craicoverflow: Fixed in https://github.com/aerogear/graphback/pull/1944/commits/d425e44b73a58ecc768435ddee28e79e364a800b

wtrocki:
> I'd have loved to actually not pass the GraphQLResolverInfo down to service layer but if we consider that layer to be like a Controller or Router layer then I can live with that.

Done review. This is really what we needed to make things right (apart from typesafety), how would you see this done?

craicoverflow:
> Done review. This is really what we needed to make things right (apart from typesafety), how would you see this done?

Apart from typesafety of what? The services?

machi1990:
> > Done review. This is really what we needed to make things right (apart from typesafety), how would you see this done?
> 
> Apart from typesafety of what? The services?

I guess he's talking about https://github.com/aerogear/graphback/issues/1637

craicoverflow:
Thought so but was not entirely sure. I am in the middle of investigating that (feels like follow up PR).

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -72,40 +79,51 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
...
+  }
 
-    if (context.graphback.options?.aggregations?.count) {
-      count = await this.db.count(filter);
+  public async findBy(args?: FindByArgs, context?: GraphbackContext, info?: GraphQLResolveInfo, path?: string): Promise<ResultList<Type>> {
+    let selectedFields: string[];
+    let requestedCount: boolean = false;
+    if (info) {
+      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model, path);
+      requestedCount = path === 'items' && getResolverInfoFieldsList(info).some((field: string) => field === "count");
```
machi1990, line 113: We should only perform the count aggregation query if path is `items` i.e we are querying for the list of "raw" items with pagination.

e.g if we issue a query like 
```graphql
query {
  getXXX(id: 1) {
   id, 
  yyy {
    id 
    count
  }
 }
}

``

This will perform the count query, while it should not. 

yyy and xxx have a one to one relationship between them.

Comment (craicoverflow) [reply packages/graphback-core/src/runtime/CRUDService.ts:113]: Yep, you are absolutely right. Great eye!

Comment (craicoverflow) [reply packages/graphback-core/src/runtime/CRUDService.ts:113]: I've pushed that fix now:

```ts
 requestedCount = path === 'items' && getResolverInfoFieldsList(info).some((field: string) => field === "count");
```

machi1990: Thanks.

integration/tests/runtime-workflow-postgres.ts
```diff
@@ -283,6 +302,26 @@ test('Find all notes', async () => {
...
 })
 
+test('Find all notes with count', async () => {
+  const { data, errors } = await client.query({
+    operationName: 'findNotes',
+    query: FIND_NOTES_WITH_COUNT
+  })
+
+  expect(data).toBeDefined()
+  expect(data.findNotes.count).toBeDefined()
```
machi1990, line 312: If `count==null` this will be `true`, while in reality the value is not sent. Maybe we should check for the actual value.

integration/tests/runtime-workflow-postgres.ts
```diff
@@ -586,44 +626,46 @@ test('Delete Note 1', async () => {
...
 
 test('getComment in simultaneous with different resolvers pinfo', async () => {
-  const { data } = await client.query({ operationName:"getCommentDetailsAtOnce", query: GET_COMMENTS , variables: {
-    id: 1
-  }});
+  const { data } = await client.query({
+    operationName: "getCommentDetailsAtOnce", query: GET_COMMENTS, variables: {
+      id: 1
+    }
+  });
```
machi1990, line 633: This is not related, but I am curios what editor settings do you have. Maybe we can share them across. I usually do not care about this sort of things and run the  `"linter"` hoping it will take care of it.

packages/graphback-datasync/src/services/DataSyncCRUDService.ts
```diff
@@ -12,16 +13,19 @@ export interface SyncList<T> {
...
    * For delta queries
    */
-  public async sync(lastSync: Date, context: GraphbackContext, filter?: any, limit?: number): Promise<SyncList<T>> {
-
-    const res = await (this.db as DataSyncProvider).sync(lastSync, context, filter, limit);
+  public async sync(lastSync: Date, info?: GraphQLResolveInfo, filter?: any, limit?: number): Promise<SyncList<T>> {
+    let selectedFields: string[];
+    if (info) {
+      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model)
+    }
```
machi1990, line 27: Should we pass "items" in this too? Cause a delta list will be sent.

See https://graphback.dev/docs/datasync/soft-delete

machi1990: /cc @ssd71

craicoverflow: Yes!!

integration/tests/runtime-workflow-postgres.ts
craicoverflow: Definitely.

Issue #1945, opened 2020-08-25
Add action to deploy website automatically when new release is created. 


machi1990:

machi1990:
/cc @craicoverflow, @wtrocki
Automatically generated comment to notify maintainers

machi1990:
See https://v2.docusaurus.io/docs/deployment/#triggering-deployment-with-github-actions

PR #1947, opened 2020-08-26
ci: auto-deploy website on push to master

machi1990:
This is based on https://v2.docusaurus.io/docs/deployment/#triggering-deployment-with-github-actions

Fixes https://github.com/aerogear/graphback/issues/1945

### DIFF
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -5,23 +5,25 @@ on:
     branches:
       - master
     paths:
-      - 'docs'
-      - 'website'
+      - 'docs/**'
+      - 'website/**'
       - 'package.json'
-      - 'packages/**/src'
+      - 'packages/**/src/**'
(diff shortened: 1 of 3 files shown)


craicoverflow (approved):

craicoverflow:
I see the entire CI suite runs for this PR even though all that is changed is .md and .yaml

It would be amazing if what runs is based on the files changed - obviously not in this PR it is a future improvement that could prevent us hitting API limits and speed up PRs.

machi1990:
> I see the entire CI suite runs for this PR even though all that is changed is .md and .yaml
> 
> It would be amazing if what runs is based on the files changed - obviously not in this PR it is a future improvement that could prevent us hitting API limits and speed up PRs.

So the entire CI is ran because I modified the two workflows. Normally, I've added "paths filtering rules" to prevent running all workflows but some of them based on changes.

PR #1950, opened 2020-08-27
follow up to #1944

craicoverflow:
This follows up some feedback in https://github.com/aerogear/graphback/pull/1944

### DIFF
--- a/packages/graphback-datasync/src/services/DataSyncCRUDService.ts
+++ b/packages/graphback-datasync/src/services/DataSyncCRUDService.ts
@@ -1,6 +1,7 @@
 import { CRUDService, CRUDServiceConfig, ModelDefinition, getSelectedFieldsFromResolverInfo } from '@graphback/core';
 import { GraphQLResolveInfo } from 'graphql';
 import { DataSyncProvider } from "../providers";
+import { DataSyncFieldNames } from '../util';
 
 export interface SyncList<T> {
   items: T[],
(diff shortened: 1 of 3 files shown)


machi1990 (approved):
Thanks for the followup.

packages/graphback-datasync/src/services/DataSyncCRUDService.ts
```diff
@@ -23,7 +23,7 @@ export class DataSyncCRUDService<T = any> extends CRUDService<T> {
   public async sync(lastSync: Date, info?: GraphQLResolveInfo, filter?: any, limit?: number): Promise<SyncList<T>> {
     let selectedFields: string[];
     if (info) {
-      selectedFields = getSelectedFieldsFromResolverInfo(info, this.model)
```
craicoverflow, line 26: I see that the tests are failing with this change, on further investigation I can see that the selected fields were never extracted and passed to this method.

https://github.com/aerogear/graphback/blob/c8d747fb47e73713b1f806488e1ce56bf3f41c5e/packages/graphback-datasync/src/DataSyncPlugin.ts#L119

machi1990: Might be worth checking if the model is a Delta Table or the real
model. IIRC, the sync method operates on the delta table which has different fields than the original model. 

If my memory is right, we might as well drop this selection here. 

/cc @ssd71 to confirm this.

craicoverflow: Yep, this is because of the Delta Table which **has** the _deleted field, but the model does not have it.

craicoverflow:
@wtrocki this is merged you are free to release now.

PR #1951, opened 2020-08-27
chore(templates): add codegen to backend templates

craicoverflow:
This PR adds a `codegen.yml` file to each backend template and I have added a generated schema and types file to the templates.


wtrocki (approved):

wtrocki:
Obviously we do not have way to use them yet, but it is best to have separate PR for that in future that will showcase sample resolvers.

craicoverflow:
> Obviously we do not have way to use them yet, but it is best to have separate PR for that in future that will showcase sample resolvers.

Yes that is the reason for this PR. I will be doing that next.

machi1990 (approved):

machi1990:
Should we completely ignore CI runs for `templates`?

PR #1952, opened 2020-08-27
feat(context): move service map to context top-level

craicoverflow:
The `graphback` property has been removed from `GraphbackContext`.

```patch
export interface GraphbackContext {
-  graphback: {
-    services: GraphbackServiceConfigMap
-  }
+  graphback: GraphbackServiceConfigMap
}
```

Now you can reach the Graphback services like so: 

```patch
- context.graphback.services.Note.findBy(...);
+ context.graphback.Note.findBy(...);
```

This is another enabler PR for #1637

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -153,6 +153,21 @@ await noteService.findBy({
 
 Resolver options (`context.graphback.options`) was removed from the context because the `count` aggregation and `selectedFields` extraction logic was moved to the CRUDService method.
 
+### Removed graphback key from GraphbackContext
+
+The `graphback` property has been removed from `GraphbackContext`.
+
+```patch
+export interface GraphbackContext {
+-  graphback: {
+-    services: GraphbackServiceConfigMap
+-  }
++  services: GraphbackServiceConfigMap
+}
+```
+
+Now you can reach the Graphback services by calling `context.services.Note.findBy(...)`.
+
 #### CRUDService, DataSyncCRUDService now accepts a `ModelDefinition` as the first constructor parameter.
 
 To instantiate a CRUDService you must pass the full `ModelDefinition` instead of the model name.
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -155,18 +155,18 @@ Resolver options (`context.graphback.options`) was removed from the context beca
 
 ### Removed graphback key from GraphbackContext
 
-The `graphback` property has been removed from `GraphbackContext`.
+The `graphback.services` property has been removed from `GraphbackContext`, and `graphback` is now the service map property.
 
 ```patch
 export interface GraphbackContext {
 -  graphback: {
--    services: GraphbackServiceConfigMap
+-    graphback: GraphbackServiceConfigMap
 -  }
-+  services: GraphbackServiceConfigMap
++  graphback: GraphbackServiceConfigMap
 }
 ```
 
-Now you can reach the Graphback services by calling `context.services.Note.findBy(...)`.
+Now you can reach the Graphback service map by calling `context.graphback.Note.findBy(...)`.
 
 #### CRUDService, DataSyncCRUDService now accepts a `ModelDefinition` as the first constructor parameter.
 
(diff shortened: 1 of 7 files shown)


packages/graphback-core/src/runtime/interfaces.ts
```diff
@@ -14,9 +14,7 @@ export interface GraphbackServiceConfigMap {
  * GraphQL context interface according to Graphback runtime layer format
  */
 export interface GraphbackContext {
-  graphback: {
-    services: GraphbackServiceConfigMap
-  }
+  services: GraphbackServiceConfigMap
```
machi1990, line 17: `services` is too general and it is likely that the `context` could contain a key with this name. What do you think of `graphbackServices`?

docs/releases.md
```diff
@@ -153,6 +153,21 @@ await noteService.findBy({
...
+```patch
+export interface GraphbackContext {
+-  graphback: {
+-    services: GraphbackServiceConfigMap
+-  }
++  services: GraphbackServiceConfigMap
+}
+```
+
+Now you can reach the Graphback services by calling `context.services.Note.findBy(...)`.
```
machi1990, line 169: I love these notes. Thanks!

packages/graphback-core/src/runtime/interfaces.ts
craicoverflow: I was thinking the same, and would like to use `models`, but then it would also need to be changed from `services` to `models` in `GraphbackAPI` (see API reference)[https://graphback.dev/docs/next/api/graphback/interfaces/_buildgraphbackapi_.graphbackapi#services] for consistency, but here it might gert confusing as the user might think that models means actual model GraphQL files or something.

I kept it as `services` then because we originally tried the approach of being ultra flexible and fitting into any use case without interfering with some user defined code - but this lead us down the road of having really unfriendly APIs which we are rolling back on now.

So in essence, a prerequisite of using Graphback could be that our services will be called services in the context, you will need to update your code accordingly.

Alternatively, we could just use `graphback` and have no `services` key:

```ts
context.graphback.Note.findBy(...);
```

I like conciseness in a user API, don't mind long variables internally, but love to make things simple for the user.

machi1990: > I was thinking the same, and would like to use `models`, but then it would also need to be changed from `services` to `models` in `GraphbackAPI` (see API reference)[https://graphback.dev/docs/next/api/graphback/interfaces/_buildgraphbackapi_.graphbackapi#services] for consistency, but here it might gert confusing as the user might think that models means actual model GraphQL files or something.
> 
> I kept it as `services` then because we originally tried the approach of being ultra flexible and fitting into any use case without interfering with some user defined code - but this lead us down the road of having really unfriendly APIs which we are rolling back on now.
> 
> So in essence, a prerequisite of using Graphback could be that our services will be called services in the context, you will need to update your code accordingly.

This may work fine if user looks set to define custom services, but it maybe not play well if the user has another framework that plays with the context and uses `services` as a key. 

> Alternatively, we could just use `graphback` and have no `services` key:
> 
> ```ts
> context.graphback.Note.findBy(...);
> ```
> 
> I like conciseness in a user API, don't mind long variables internally, but love to make things simple for the user.

I like the alternative suggestion.

wtrocki:
Services as key can be easily changed in createContext method. Prisma do not even care to namespace services and mount every obj to context

packages/graphback-core/src/runtime/interfaces.ts
craicoverflow: Same, I will go with that.

machi1990: Thanks.

wtrocki (approved):

wtrocki:
Let's merge. I will push templates and play with the changes

craicoverflow:
Not ready for a merge

craicoverflow:
I have adjusted the code and updated the description. Now you can call Graphback services like this:

`context.graphback.Note.findBy(...)`

craicoverflow:
> Services as key can be easily changed in createContext method.

Not true, the `SchemaCRUDPlugin` relies on the existence of `services` to load the services.

machi1990 (approved):

craicoverflow:
Thanks @machi1990

PR #1956, opened 2020-08-27
chore(templates): add example custom resolver

craicoverflow:
**DRAFT**

This PR adds a custom resolver to the Postgres Backend template.

Opening this as a draft so you can see if this is a good approach. It is pretty much as was discussed in  https://github.com/aerogear/graphback/issues/1637

Once we have agreement I can follow up and add this to the other templates.

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -17,7 +17,6 @@
   },
   "devDependencies": {
     "@graphql-tools/graphql-file-loader": "6.1.0",
-    "apollo-boost": "0.4.9",
     "apollo-server": "2.17.0",
     "apollo-server-testing": "2.17.0",
     "jest": "26.4.2",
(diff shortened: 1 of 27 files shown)


machi1990 (approved):
Pretty much what we discussed on the issue. I am okay with this approach. 

Once @wtrocki is happy with this too, I am hoping to see the docs updated: they can be based on this example and the codegen trick.

wtrocki:
We need:

- Documentation
- Add this to mongodb as well 
- Use custom method in the client

wtrocki (approved):

machi1990 (commented):
Code LGTM thanks @craicoverflow. I'll need to verify this early tomorrow when I'm infront of my workstation.

integration/package.json
```diff
@@ -17,7 +17,6 @@
   },
   "devDependencies": {
     "@graphql-tools/graphql-file-loader": "6.1.0",
-    "apollo-boost": "0.4.9",
```
machi1990, line 20: Good catch!

machi1990 (approved):

PR #1974, opened 2020-09-01
review github actions job

machi1990:


wtrocki:
Screenshot 2020-09-01 at 12 41 22

LOL :D

machi1990:
> Screenshot 2020-09-01 at 12 41 22
> 
> LOL :D

Looks like the cache trick did not work.

wtrocki:
Question Why we cannot use different approach. Use single item that does build, lint and unit tests. I never seen such granular approach. Usually this is really fast to get feedback.

First you get build failing, then linting then tests then integration for mongo and postgress (just latest version. This is also not wasting github resources. 
Also going and clicking on many jobs to get info I need sounds pointless. 

Ideally we can have single job for those + separate for website builds.

Also strange that we use release one here. If we have single build step we can make it as prerequisite for release and actual release having only extra build step + probably more automation to push to branch etc.

machi1990:
> Question Why we cannot use different approach. Use single item that does build, lint and unit tests. I never seen such granular approach. Usually this is really fast to get feedback.

This approach was to make use of database matrix and also an attempt to run the build in parallel and not sequentially. 

> First you get build failing, then linting then tests then integration for mongo and postgress (just latest version. This is also not wasting github resources.
> Also going and clicking on many jobs to get info I need sounds pointless.

If we are fine with testing against one database I do not have a problem with it. We can go the sequential way. 

> Ideally we can have single job for those + separate for website builds.
> 
> Also strange that we use release one here. If we have single build step we can make it as prerequisite for release and actual release having only extra build step + probably more automation to push to branch etc.

I opted to put the release job in the same file  instead of a separate workflow to avoid duplication. I did not know if it's possible for one one workflow to depend on the other.

machi1990:
> Also going and clicking on many jobs to get info I need sounds pointless.

Hmm I would not say it's pointless. Instead of having to scroll through a list of sequential logs, you can click on a specific job that failed and get the details of the failure. 


> > Question Why we cannot use different approach. Use single item that does build, lint and unit tests. I never seen such granular approach. Usually this is really fast to get feedback.
> 
> This approach was to make use of database matrix and also an attempt to run the build in parallel and not sequentially.
> 
> > First you get build failing, then linting then tests then integration for mongo and postgress (just latest version. This is also not wasting github resources.
> > Also going and clicking on many jobs to get info I need sounds pointless.
> 
> If we are fine with testing against one database I do not have a problem with it. We can go the sequential way.

Seems like this is a way we'd want to go. I'd update the PR in this direction. 
 
> > Ideally we can have single job for those + separate for website builds.
> > Also strange that we use release one here. If we have single build step we can make it as prerequisite for release and actual release having only extra build step + probably more automation to push to branch etc.
> 
> I opted to put the release job in the same file instead of a separate workflow to avoid duplication. I did not know if it's possible for one one workflow to depend on the other.

Regarding this, I am happy to take any pointer on how to re-use a workflow on another workflow.

wtrocki:
>  I did not know if it's possible for one one workflow to depend on the other.

Sorry I meant actions not workflows.
https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#about-github-actions

wtrocki:
Quick check proven that considfering how much hassle it would be to build action we can have simple `ci` script in the package.json that will run all. but generally release will be much different and the only thing that will be similar is build so there is no need to reuse workflow IMHO

.github/workflows/build-website.yml
```diff
@@ -28,10 +28,6 @@ jobs:
     # Allow failure when website build fails. This won't block merging but the status will still be red allowing for investigations
     continue-on-error: true
     steps:
-    - name: Cancel Previous Runs
-      uses: n1hility/cancel-previous-runs@v2
-      with:
-        token: ${{ secrets.GITHUB_TOKEN }}
```
craicoverflow, line 34: Would this not improve CI completion because old runs would be cancelled instead of queued?

machi1990: This was supposed to have an effect of cancelling previous running jobs. 

When I inspected some CI runs, it was not actually kicking in and I preferred to remove it for now.

craicoverflow (approved):

PR #1988, opened 2020-09-03
fix: moving to graphql-cli instead of codegen

wtrocki:
Very small change to templates to use graphql-cli instead of codegen (legacy). 
Removed codegen files and added graphql-cli dependency. 

Dependencies will be swapped at runtime but only for root (@machi1990 client one has no dependency swapped. Orignal template was yarn workspace. Here we use two separate projects so that is why installation on root is not working. However, if we drop dependency from the client we will be able to use the root one. @machi1990 WDYT? We also need better documentation showing how to regenerate typings on server.

Let's make sure that the templates branch is updated after the merge.

## Verification

Go to any template and:

1. yarn codegen 
2. cd client && yarn codegen

You should see successful execution and no changes.


wtrocki:
@machi1990 left some open question on this. Happy to help you with review etc.

templates/ts-react-apollo-client/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.16.1",
   "private": true,
   "devDependencies": {
+    "@graphql-cli/codegen": "1.17.8",
```
wtrocki, line 6: Actually we might try to not have cli here but all those packages in root. This way graphql-cli will be able to update them

machi1990:
> Very small change to templates to use graphql-cli instead of codegen (legacy).
> Removed codegen files and added graphql-cli dependency.

+1 

> Dependencies will be swapped at runtime but only for root (@machi1990 client one has no dependency swapped. Orignal template was yarn workspace. Here we use two separate projects so that is why installation on root is not working. However, if we drop dependency from the client we will be able to use the root one. @machi1990 WDYT? 

+1 this is the correct observation of what was happening. This should fix it for the client side project too. 

> We also need better documentation showing how to regenerate typings on server.
 
+1. I think it will be same as the one we have on client side. I can add a commit in this PR if that's okay with you. 

> Let's make sure that the templates branch is updated after the merge.

+100. 

> ## Verification
> Go to any template and:
> 
> 1. yarn codegen
> 2. cd client yarn codegen

wtrocki:
@machi1990 feel free to make small calibrations

machi1990:
> @machi1990 feel free to make changes that are necessary

Done https://github.com/aerogear/graphback/pull/1988/commits/3b341cd326db040d5bb729542a3546a9c3bfc3ea. I've also fixed the `codegen` command. 

From my side all is good. @craicoverflow can you also have a look? If it's okay with you too, we can merge this and update the templates branch.

machi1990 (approved):

templates/ts-react-apollo-client/package.json
craicoverflow: `graphql-codegen` works with graphqlrc files, so you could stick with that?

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
```diff
@@ -2,6 +2,21 @@
 ## Please review configuration and adjust it for your own project
 schema: ./src/schema/schema.graphql
 extensions:
+  graphback:
+    model: ./model/**/*.graphql
```
craicoverflow, line 6: We should load the model from this in the template, otherwise config needs to be maintained in multiple places

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
```diff
@@ -2,6 +2,21 @@
...
+    crud:
+      create: true
+      update: true
+      findOne: true
+      find: true
+      delete: true
+      subCreate: true
+      subUpdate: true
+      subDelete: true
```
craicoverflow, line 15: We should load the CRUD config from here into `buildGraphbackAPI`, otherwise CRUD config needs to be maintained in multiple places

templates/ts-react-apollo-client/.graphqlrc.yml
```diff
@@ -1,7 +1,7 @@
 ## GraphQL Config Generated by Graphback
 ## Please review configuration and adjust it for your own project
-schema: ./schema.graphql
-documents: ./src/graphql/**/*.graphql
+schema: ../src/schema/schema.graphql
+documents: ./src/graphql/**/**.graphql
```
craicoverflow, line 4:
```suggestion
documents: ./src/graphql/**/*.graphql
```

Should this be kept as it was? Not sure if that would make a difference or work.

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
machi1990: This is only needed for the `graphback-cli`

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
machi1990: I can get rid of these to remove the bloat. Only added for the `graphback-cli` command

templates/ts-react-apollo-client/.graphqlrc.yml
machi1990: good catch.

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
craicoverflow: Yep, but this still means that path is declared in multiple places - in code and in this

machi1990:
@craicoverflow @wtrocki  regarding https://github.com/aerogear/graphback/pull/1988#discussion_r483039530, the idea was to consume everything using `@graphql-cli`.

wtrocki (commented):
LGTM

templates/ts-apollo-mongodb-backend/.graphqlrc.yml
machi1990: Sorry I completely misunderstood your initial comment. You are right. That could improve things. 

Let me push something for you to review.

machi1990:
@craicoverflow I've added https://github.com/aerogear/graphback/pull/1988/commits/e3ace03bd24bb7291f754d67c2e081899cf3b1db to avoid duplication as suggested in https://github.com/aerogear/graphback/pull/1988#discussion_r483040329.

PR #1997, opened 2020-09-04
Templates fix

wtrocki:
Fixing couple very small bugs in templates;

- [x] Added ability to run and play with the templates without linking graphback versions. If we really would need non production versions then we can disable package.json by renaming it to something else.
I think this might be controversial change but it is really painful to work with templates now without it.

- [x] Hidden implementation details of the codegen and graphback commands. 
- [x] Moved all code to use _id to reuse the same app
- [x] [GraphQL-Migrations]: Fixes an issue where "Primary key with `ID` ScalarType and field name different to `id` isn't auto-incrementable"

## Verification
0. yarn install
1. Run all templates together with the client app.
2. Create Notes and comments. 

After that we can push that to 0.16.0 and verify it in production 🤣


templates/ts-apollo-postgres-backend/README.md
```diff
@@ -38,15 +38,7 @@ The project has been created using `graphback`. Run the project using the follow
...
-- Define your schema in the `model/runtime.graphql` file. Or use the default:
-
-```
-type User {
-  id: ID!
-  name: String
-}
-```
-
+- Inspect your schema in the `model/runtime.graphql` file.
```
craicoverflow, line 41:
```suggestion
- Inspect your schema in the `model/datamodel.graphql` file.
```

templates/ts-apollo-mongodb-backend/package.json
```diff
@@ -9,7 +9,7 @@
     "develop": "ts-node src/index.ts",
     "start": "node dist/src/index.js",
     "watch": "tsc -w",
-    "codegen": "graphql codegen"
+    "generate": "graphback generate && graphql codegen"
```
machi1990, line 12: +1

templates/ts-react-apollo-client/.env
```diff
@@ -0,0 +1 @@
+SKIP_PREFLIGHT_CHECK=true
```
craicoverflow, line 1: This won't be needed once the templates are deployed to a user's machine through our CLI, it is only there for working around the workspaces issue right? I did not add it for this reason, as the user is getting environment variables which they don't need.

`SKIP_PREFLIGHT_CHECK=true yarn start`

machi1990, line 1: This will mean users having this when a template is downloaded. Do we really want that?

wtrocki: I see this actually happening when initializing the project using graphql-cli.
Not sure about our CLI.

machi1990: i.e are the problems we were seeing e.g `eslint` crashing with react-scripts' `eslint` only related to how the templates directory is setup in our repository?

wtrocki: I do not know. It is just that released version of the graphql-cli doesn't work with instruction in the readme. I have tested it twice with different template options. Always needed to skip_preflight

wtrocki: If you guys testing graphql-cli and **do not see this** then maybe I have react installed globally or something

templates/ts-react-apollo-client/.graphqlrc.yml
```diff
@@ -1,10 +1,10 @@
 ## GraphQL Config Generated by Graphback
 ## Please review configuration and adjust it for your own project
-schema: ../src/schema/schema.graphql
+schema: ../ts-apollo-postgres-backend/src/schema/schema.graphql
```
machi1990, line 3: I understand it was there to generate the file but should this be reverted?

templates/ts-react-apollo-client/schema.graphql
```diff
@@ -0,0 +1,173 @@
...
+  _id: ID!
+  text: String
+  description: String
+
+  """@manyToOne(field: 'comments', key: 'noteId')"""
+  note: Note
+}
+
+input CommentFilter {
+  _id: IDInput
```
machi1990, line 21: So this may not solve the problem for mongo's ObjectID?

wtrocki: It actually works with the mongo and postgress. Despite being not the proper type application works on the runtime and compiles. Types here are mostly informative and they all map to strings anyway.

wtrocki:
Ready for review

templates/ts-react-apollo-client/schema.graphql
machi1990: I can't test this for now (away from my machine) so I trust you on this.

templates/ts-react-apollo-client/.env
wtrocki: I have feeling that this might be issue on my side - you can actually check that during verification since we have separate workspace now

templates/ts-react-apollo-client/schema.graphql
wtrocki: Never trust me 🤣

templates/ts-react-apollo-client/.env
craicoverflow: I created a new project with graphql-cli fresh (outside repos) and the client starts fine without this env - are you seeing it outside repos like this?

machi1990: I do not remember having the issue when the template was downloaded outside graphback repo. In my opinion, we can remove this file and do @craicoverflow suggestion in https://github.com/aerogear/graphback/pull/1997#discussion_r483677834. It's easier to forget that so let's add it to https://github.com/aerogear/graphback/blob/master/CONTRIBUTING.md

wtrocki: Ok.. I tracked the problem. I have created sample project in the folder that had number of package.json files above and some of those had react. After moving out all works.

wtrocki:
Something strange happening. I do not see github actions job triggered. Checking

wtrocki:
Looks like some github issue. See builds happenening for the commits in this PR.

craicoverflow:
> Something strange happening. I do not see github actions job triggered. Checking

I think GH Actions do not trigger when the changes are only in templates.

craicoverflow:
I'm seeing a couple of issues during verification.

**graphql-migrations**

```
Foreign field id on type Note not found on field note.
Foreign field id on type Note not found on field note.
Foreign field id on type Note not found on field note.
```

And in the playground, I can't perform a mutation without passing the ID field.

```sh
insert into \"note\" (\"description\", \"title\") values ($1, $2) returning * - null value in column \"_id\" violates not-null constraint
```

This is likely because only fields matching `id: ID` exactly will be auto-generated. A fix for this (and a possible improvement is to assume fields with scalar `ID` are auto-generated, and the field name does not matter.

machi1990:
> I'm seeing a couple of issues during verification.
> 
> **graphql-migrations**
> 
> ```
> Foreign field id on type Note not found on field note.
> Foreign field id on type Note not found on field note.
> Foreign field id on type Note not found on field note.
> ```
> 
> And in the playground, I can't perform a mutation without passing the ID field.
> 
> ```shell
> insert into \"note\" (\"description\", \"title\") values ($1, $2) returning * - null value in column \"_id\" violates not-null constraint
> ```
> 
> This is likely because only fields matching `id: ID` exactly will be auto-generated. 

This was going to be an issue. 

> A fix for this (and a possible improvement is  to assume fields with scalar `ID` are auto-generated, and the field name does not matter.

I'd say a scalar `ID` that's a primary key i.e either annotated with `@id` or the automatic inferred one `id: ID`. 

Because we'd want to stamp auto-increment on any field with  scalar type `ID`. 
e.g 
```graphql
counter: ID
```

machi1990:
> I'm seeing a couple of issues during verification.
> 
> **graphql-migrations**
> 
> ```
> Foreign field id on type Note not found on field note.
> Foreign field id on type Note not found on field note.
> Foreign field id on type Note not found on field note.
> ```
> 
> And in the playground, I can't perform a mutation without passing the ID field.
> 
> ```shell
> insert into \"note\" (\"description\", \"title\") values ($1, $2) returning * - null value in column \"_id\" violates not-null constraint
> ```
> 
> This is likely because only fields matching `id: ID` exactly will be auto-generated. A fix for this (and a possible improvement) is to assume fields with scalar `ID` are auto-generated, and the field name does not matter.

What do you think about fixing this for a 0.16.2 release? @craicoverflow @wtrocki

craicoverflow:
+1 from me @machi1990

machi1990:
> +1 from me @machi1990

I'll drive this PR home.

wtrocki:
> This is likely because only fields matching id: ID exactly will be auto-generated. A fix for this (and a possible improvement) is to assume fields with scalar ID are auto-generated, and the field name does not matter.

Ouch. This is tough one. Is problem in the DataProvider or Database migrations?

machi1990:
> > This is likely because only fields matching id: ID exactly will be auto-generated. A fix for this (and a possible improvement) is to assume fields with scalar ID are auto-generated, and the field name does not matter.
> 
> Ouch. This is tough one. Is problem in the DataProvider or Database migrations?

It's a data migrations issue.

wtrocki:
Closed by mistake. @machi1990 Thank you so much. Maybe it makes sense to have separate PR with the fix that later can enable this PR to work.

machi1990:
> Closed by mistake. @machi1990 Thank you so much. Maybe it makes sense to have separate PR with the fix that later can enable this PR to work.

@craicoverflow @wtrocki I pushed two commits.
- 1. That fixes the issue
- 2. That bumps the version to 0.16.2 so that we can release straight after this is merged. 

I tested the fix locally and it worked. Can you also check please?

docs/releases.md
```diff
@@ -8,6 +8,12 @@ title: Release Notes
 This file contains changes and migration steps for the Graphback project. 
 Please follow individual releases for more information.
 
+# 0.16.2
+
+### Bug Fixes
+
+* Primary key with `ID` ScalarType and field name different to `id` isn't auto-incrementable. Fixed in [#1997](https://github.com/aerogear/graphback/pull/1997)
```
craicoverflow, line 15: Bug or feature? It was intended behaviour until now

craicoverflow (approved):
Checked and verified. One comment inline

craicoverflow:
Would the permanent solution for this be two split the client template into two:

- ts-react-postgres-apollo-client
- ts-react-mongo-apollo-client

Each template would just need to symlink to the corresponding frontend template.

machi1990:
> Would the permanent solution for this be two split the client template into two:
> 
> * ts-react-postgres-apollo-client
> * ts-react-mongo-apollo-client
> 
> Each template would just need to symlink to the corresponding frontend template.

Very good point. This will mean having to main two similar client templates with the difference being the `_id`?

docs/releases.md
machi1990: I honestly was not sure: A bug in specification maybe? :-)

machi1990 (approved):

machi1990:
[0.16.2](https://github.com/aerogear/graphback/releases/tag/0.16.2) released with these changes.

PR #2018, opened 2020-09-09
refactor(core): no need to await for publish as it can be done asynchronously

machi1990:


wtrocki:
**So the current situation wasn't good.** 
If there is a problem with the network connection client would got error (despite proper save).

**After the change we get:**
Save will happen and user will get a positive result but subscriptions might not get it. 

I think this is good behavior/improvement. Verifying to be sure

machi1990:
> **Current situation**
> If there is a problem with the network connection to queue client would got an error (despite proper save to db) client cannot react or handle this error (as it is related to server side problem).
> 

Exactly. Any publish error would have resulted in failure of the whole request even if the save / update / delete operations went well. 

> **After the change we get:**
> Save will happen and user will get a positive result but subscriptions might not get it.
> Logging should be improved (but that is actually an template side code.

+1. I think a `.then()` and `.catch()` as we can have appropriate logs for each case but went with `.finally()` and we can add a log message here. I can create a "good first issue" for that. 


> I think this is good behavior/improvement. Verifying to be sure

Thanks.

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -55,7 +55,8 @@ export class CRUDService<Type = any> implements GraphbackCRUDService<Type>  {
       const topic = this.subscriptionTopicMapping(GraphbackOperationType.CREATE, this.model.graphqlType.name);
       //TODO use subscription name mapping
       const payload = this.buildEventPayload('new', result);
-      await this.pubSub.publish(topic, payload);
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      this.pubSub.publish(topic, payload).finally(() => {});
```
wtrocki, line 59: ok. Havent's seen that. Why finnally is empty? Makes no sense

wtrocki:
Approved and then seen your comment.

> +1. I think a .then() and .catch() as we can have appropriate logs for each case but went with .finally() and we can add a log message here. I can create a "good first issue" for that.

Is there any problem with logging that prevents us from this

packages/graphback-core/src/runtime/CRUDService.ts
machi1990: We can surely add logs, I was just too lazy to add them.

machi1990: I've added a console.error on this, so that users get's to see what went wrong.

wtrocki:
All good for me

Issue #2019, opened 2020-09-09
Validation for relationships fails on valid case


wtrocki:
Problem with OpenVolunteerModel

```
Error: Volunteer.actions and VolunteerAction.volunteer 'key' annotations are different. Ensure both are the same or remove one so that it can be generated.
```

https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/server/model/main.graphql#L38-L41

The action doesn't have a key specified and ManyToOne side has (this was just copied from the generated schema for visibility of the field in the schema)

Error is confusing as key is not there. Removing key from manyToOne will fix it.

machi1990:
/cc @craicoverflow, @machi1990
Automatically generated comment to notify maintainers

wtrocki:
I'm not sure if there is some fix here.

wtrocki:
This is a bug. Little bit debugging showed me that result schema has some extra annotations generated and ends up with 2 annotations under single object:

<img width="807" alt="Screenshot 2020-09-09 at 11 52 59" src="https://user-images.githubusercontent.com/981838/92589824-0ec30e80-f293-11ea-89f1-bcb4676f0fd4.png">

wtrocki:
So trick that I have done is:

- Generate schema
- When to schema to take generated fields and include them into model (they should be ok to be in the model.
- Generate code:

```
<b>Generation failed: Error: Volunteer.actions and VolunteerAction.volunteer 'key' annotations are different. Ensure both are the same, or remove one so that it can be generated.</b>
```

wtrocki:
See also #2020

craicoverflow:
What happens when you remove all manyToOne annotations from the data model? It should succeed

wtrocki:
It will fail validation (unless this change is applied)

craicoverflow:
Can you show me the output?

craicoverflow:
Can you reproduce this error using the models from the templates? They are smaller and more focused so it is easier to see why the error occurs - as in OVP the model is large and there could be other schema errors.

machi1990:
> So trick that I have done is:
> 
> * Generate schema
> * When to schema to take generated fields and include them into model (they should be ok to be in the model.
> * Generate code:
> 
> ```
> <b>Generation failed: Error: Volunteer.actions and VolunteerAction.volunteer 'key' annotations are different. Ensure both are the same, or remove one so that it can be generated.</b>
> ```

[...]

PR #2020, opened 2020-09-09
fix: do not require manyToOne to be defined on the relationship field

wtrocki:
With the current shape of relationships, graphback is really confusing with validation failing with errors that should not happen.

1) ManyToOne is not documented yet it is **required** and **validated** to be present

2) Using ManyToOne in model is causing validation issues (despite usage being valid)

3) Using only OneToMany will not work with model - client-side generation plugin is not seeing changes and breaks when generating client side queries. (that can be a separate issue but is also a sign of problem) 

3) Validation for relationships is clearly broken (Specifying the same `key` value on both sides still ends up with the validation error.

4) Model itself is a little bit confusing as we are **forced** to remove fields that should stay.

### Solution

I have tested this only for mongo by I think we should have short and long term discussions on this. 
For short I think we need to disable validation and require manyToOne annotation for the fields that are already covered with OneToMany. This is the fix in this PR.

### Long term plan

I think we need some small refactor on relationships to be able to get them aligned. 
We can use target schema in the client side plugin as well - that will also fix the problem I see with not being able to generate queries when field is missing, but other problems with persist.

### DIFF
--- a/packages/graphback-core/src/crud/mappingHelpers.ts
+++ b/packages/graphback-core/src/crud/mappingHelpers.ts
@@ -163,7 +163,7 @@ export function getInputFieldTypeName(modelName: string, field: GraphQLField<any
 
     const relationshipAnnotation = parseRelationshipAnnotation(field.description);
 
-    if (!relationshipAnnotation) {
+    if (!relationshipAnnotation && relationshipAnnotation.kind !== 'manyToOne') {
       throw new Error(`Missing relationship definition on: "${modelName}.${field.name}". Visit https://graphback.dev/docs/model/datamodel#relationships to see how you can define relationship in your business model.`);
     }
 


wtrocki:
@machi1990 [Offtopic] Website build is triggered for the change that did not affected website

machi1990:
> @machi1990 [Offtopic] Website build is triggered for the change that did not affected website

I think we already covered this many times. The change affects the API (source code) hence potentially affecting the generated API code - e.g https://graphback.dev/docs/api/graphback/modules/_buildgraphbackapi_, thus we need to run the website build to make sure that the links are not broken.

machi1990 (commented):
@wtrocki thanks for the detailed explanation. I'd like to test this properly in OVP. Is there a branch I can use there?

wtrocki:
Website Problem fixed in separate PR with explanation:
https://github.com/aerogear/graphback/pull/2021

OVP branch to replicate issue
https://github.com/aerogear/OpenVolunteerPlatform/pull/153

machi1990:
> OVP branch to replicate issue
> [aerogear/OpenVolunteerPlatform#153](https://github.com/aerogear/OpenVolunteerPlatform/pull/153)

Thanks. Checking.

wtrocki:
Before

```graphql
  """@manyToOne(field: 'actions', key: 'volunteerId')"""
  volunteer: Volunteer

  """@manyToOne(field: 'actions', key: 'recipientId')"""
  recipient: Recipient
```

After this fix:
```graphql
type VolunteerAction {
  volunteer: Volunteer
  recipient: Recipient
}
```

Generation successful!

packages/graphback-core/src/crud/mappingHelpers.ts
```diff
@@ -163,7 +163,7 @@ export function getInputFieldTypeName(modelName: string, field: GraphQLField<any
 
     const relationshipAnnotation = parseRelationshipAnnotation(field.description);
 
-    if (!relationshipAnnotation) {
+    if (!relationshipAnnotation && relationshipAnnotation.kind !== 'manyToOne') {
```
machi1990, line 166: I think this will always fail with `cannot read property kind of undefined` since the relationship is always undefined by the `!relationshipAnnotation` check as seem to be indicated by CI: https://github.com/aerogear/graphback/pull/2020/checks?check_run_id=1090647585

This is too specific for `manyToOne` relationships and I am wondering what happens for other relationships?

Maybe https://github.com/aerogear/graphback/issues/2019#issuecomment-689514459 is what we need here /cc @craicoverflow and generally I think we need an issue that highlights relationships enhancement.

wtrocki: It should have ? Added

PR #2021, opened 2020-09-09
fix: do not build/publish website on code changes

wtrocki:


wtrocki:
I have done checkout today and noticed increased size of the repo. I think the way that publish happens is that every time gh pages will generate completely different website (Every file will be changed).

Over time this will significantly increase the size of the repository as every commit on every PR deploys website! This also counts renovate bump PR's that we leave for long time (and they are build numerous times a day)

.github/workflows/build-website.yml
```diff
@@ -7,10 +7,7 @@ on:
     paths:
       - 'docs/**'
       - 'website/**'
-      - 'package.json'
-      - 'packages/**/src/**'
-      - '.github/workflows/build-website.yml'
-
```
machi1990, line 13: My opinion would be to build the website once the code changes, see https://github.com/aerogear/graphback/pull/2020#issuecomment-689479859. 

As for the number of commits, it's not an issue in my opinion:
- The commits are not in the main branch, they are in gh-pages branch, so the main history remains clean.
- We are not limited by the number of commits to make in this branch
- This is fully automated, I'll not need to checkout the gh-pages branch locally (which could take some time since history / the commits are growing linearly). 

I am plus +1 for keeping this so that any regression on the built API is caught in the instant the change is made and not "days" later a change in `docs` or `website` is made (as in this PR).

Happy to get @craicoverflow opinion on this too.

wtrocki:
Fully support your decision, but then I think we should deploy on merge to master. We have PR's with 20 commits. Each publishing website.. to production...

craicoverflow:
>  but then I think we should deploy on merge to master

This is what is happening I believe?

> Each publishing website.. to production..

They are not publishing until merge to master I believe?

.github/workflows/build-website.yml
craicoverflow: Overall I am happy with the current CI trigger for the website.

We chatted about this today and there are no adverse affects of having this - a PR might take a minute extra to be ready but we are saving lots of time keeping the website up to date.

machi1990:
> > Each publishing website.. to production..
> 
> They are not publishing until merge to master I believe?

Exactly.

wtrocki:
My bad. Website build works but commits disabling publishing. Closing

Issue #1758, opened 2020-07-27
Analyze Performance suite build for Graphback


wtrocki:
Check results and identify areas for improvements:

Tasks:
- [x] Analyse startup time
- [x] Analyse throughput
- [x] Analyse database query performance. A first round of analysis on https://github.com/aerogear/graphback/pull/2024. 
- [x] Memory usage (RSS column Added)
- [x] Bring in the tests suite in our repository
- [x] A script to run tests in docker container so we always have the same resources/memory limits for better results by isolating any hardware specificity (See https://github.com/aerogear/graphback/pull/2028)
- [x] Display 99.999 percentiles instead of mean values in comperative table of benchmark results. See https://github.com/aerogear/graphback/pull/2056
- [ ] Add a script to wipe out old perfs results e.g those of two previous major releases. See https://github.com/aerogear/graphback/pull/2028#discussion_r486334034 /cc @craicoverflow 
- [ ] Try out https://github.com/TechEmpower/FrameworkBenchmarks. See [JavaScript Frameworks](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/JavaScript)

wtrocki:
CC @martinhesko @machi1990 @craicoverflow

machi1990:
It will be interesting to add a database and see how performance fares in https://www.techempower.com/benchmarks especially against other graphql related frameworks or [Javascript libraries](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/JavaScript).

machi1990:
I'll look into dockerising the benchmark tests and provide a onetime script to launch it.

craicoverflow:
Looks like this has can be closed. @machi1990 has done great work in analysing performance and integrating it into the workflow

PR #2024, opened 2020-09-09
Review data providers

machi1990:
Round of database perf analysis as par https://github.com/aerogear/graphback/issues/1758

### DIFF
--- a/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
+++ b/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
@@ -39,14 +39,11 @@ export class KnexDBDataProvider<Type = any> implements GraphbackDataProvider<Typ
     const { idField, data: updateData } = getDatabaseArguments(this.tableMap, data);
 
     //tslint:disable-next-line: await-promise
-    const updateResult = await this.db(this.tableName).update(updateData).where(idField.name, '=', idField.value);
-    if (updateResult === 1) {
-      //tslint:disable-next-line: await-promise
-      const dbResult = await this.db.select(this.getSelectedFields(selectedFields)).from(this.tableName).where(idField.name, '=', idField.value);
-      if (dbResult && dbResult[0]) {
-        return dbResult[0]
-      }
+    const updateResult = await this.db(this.tableName).update(updateData, this.getSelectedFields(selectedFields)).where(idField.name, '=', idField.value);
+    if (updateResult && updateResult[0]) {
+      return updateResult[0];
     }
+
     throw new NoDataError(`Cannot update ${this.tableName}`);
   }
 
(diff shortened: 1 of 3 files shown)


packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -39,26 +39,21 @@ export class KnexDBDataProvider<Type = any> implements GraphbackDataProvider<Typ
...
-    if (updateResult === 1) {
-      //tslint:disable-next-line: await-promise
-      const dbResult = await this.db.select(this.getSelectedFields(selectedFields)).from(this.tableName).where(idField.name, '=', idField.value);
-      if (dbResult && dbResult[0]) {
-        return dbResult[0]
-      }
+    const updateResult = await this.db(this.tableName).update(updateData, this.getSelectedFields(selectedFields)).where(idField.name, '=', idField.value);
+    if (updateResult && updateResult[0]) {
+      return updateResult[0];
     }
```
craicoverflow, line 45: @wtrocki I'd love to know the original reason for two separate queries in the first place - I recall there being a reason so don't want to approve this without input on this

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
```diff
@@ -39,26 +39,21 @@ export class KnexDBDataProvider<Type = any> implements GraphbackDataProvider<Typ
...
+
     throw new NoDataError(`Cannot update ${this.tableName}`);
   }
 
   public async delete(data: Partial<Type>, selectedFields?: string[]): Promise<Type> {
     const { idField } = getDatabaseArguments(this.tableMap, data);
 
     //tslint:disable-next-line: await-promise
-    const beforeDelete = await this.db.select(this.getSelectedFields(selectedFields)).from(this.tableName).where(idField.name, '=', idField.value);
```
craicoverflow, line 57: @wtrocki I'd love to know the original reason for two separate queries in the first place - I recall there being a reason so don't want to approve this without input on this

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
wtrocki: Support for MYSQL, Postgres. Some dbs return different data. For PG this is good

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
wtrocki: This provider history originates from times where we were supporting sqllite etc.

packages/graphback-runtime-knex/src/KnexDBDataProvider.ts
machi1990: Thanks for the context. But what this patch brings is no different from https://github.com/aerogear/graphback/blob/9615ffb50453b9772b39eda22d645f16dd91569a/packages/graphback-runtime-knex/src/KnexDBDataProvider.ts#L31

To have a more general purpose Relational database data provider, eventually, I'd see us renaming the  [SQLiteKnexDBDataProvider](https://github.com/aerogear/graphback/blob/9615ffb50453b9772b39eda22d645f16dd91569a/packages/graphback-runtime-knex/src/SQLiteKnexDBDataProvider.ts) to something more general and guide people on how to integrate it in their application. Much like what is done here https://graphback.dev/docs/databases/sqlite#using-knex-provider

wtrocki: Yep. I did not meant to say your change is somehow wrong. We should merge it. eventually check docs for sqlite

craicoverflow (approved):

PR #2026, opened 2020-09-10
fix(graphql-serve): print-schema command was not working

machi1990:
The command was broken cause of the wrong import statement.

This caused an `undefined` error.

To verify:

Build the current master and launch the print schema command:

```bash
cd packages/graphql-serve/dist/
node index.js print-schema templates/ts-apollo-mongodb-datasync-backend/model/datamodel.graphql
```

Apply the patch, build and re-launch the command.


craicoverflow (approved):

wtrocki:
Is this requirement for release?

PR #2028, opened 2020-09-10
dockerise benchmarks

machi1990:
Addresses the docker script part of https://github.com/aerogear/graphback/issues/1758

The script is still interactive, we can live with that for now. Maybe it's something we'll need to change in the future if we'll like to automate this completely and be able to run it in VMs with limited resources etc (I did not focus on this part)

Opening as draft to get me know what you think.


CONTRIBUTING.md
```diff
@@ -30,18 +30,33 @@ The SDK is made by a group of modules, each one implemented in a different NPM p
...
+```shell
 npm uninstall -g graphback-cli
 ```
 
 ## Debugging command line tool
 
 When using Visual Studio Code developers can use debugger configurations for command line tool.
 Each individual command can be executed and it will stop on breakpoints.
+
+## Run benchmark inside docker with limited resources
```
craicoverflow, line 49:
```suggestion
## Run benchmark inside Docker with limited resources 
```

machi1990: Thanks!

package.json
```diff
@@ -32,6 +32,8 @@
     "typescript": "4.0.2"
   },
   "scripts": {
+    "bench": "cd performance && yarn start",
+    "bench:docker": "docker run -it --rm -v \"$PWD\":/usr/src/bench -w /usr/src/bench -m=\"256m\" --cpus=\"2\" node:14.10 npm run bench",
```
craicoverflow, line 36: Could the executed Docker command be moved to a docker-compose file?

machi1990: What will be the benefits?

performance/results/0.16-apollo-express.json
```diff
@@ -0,0 +1 @@
...8259509999305,"rss":49.344,"cpuPercentage":0,"memoryPercentage":0.1,"url":"http://localhost:29128/graphql","requests":{"average":471.2,"mean":471.2,"stddev":3.87,"min":466,"max":477,"total":2356,"p0_001":466,"p0_01":466,"p0_1":466,"p1":466,"p2_5":466,"p10":466,"p25":468,"p50":472,"p75":473,"p90":477,"p97_5":477,"p99":477,"p99_9":477,"p99_99":477,"p99_999":477,"sent":2361},"latency":{"average":10.09,"mean":10.09,"stddev":3.74,"min":4,"max":31.248632,"p0_001":4,"p0_01":4,"p0_1":7,"p1":7,"p2_5":7,"p10":7,"p25":8,"p50":8,"p75":11,"p90":16,"p97_5":18,"p99":23,"p99_9":29,"p99_99":31,"p99_999":31,"totalCount":2356},"throughput":{"average":6919372.8,"mean":6919372.8,"stddev":57659.12,"min":6843210,"max":7004745,"total":34597860,"p0_001":6844415,"p0_01":6844415,"p0_1":6844415,"p1":6844415,"p2_5":6844415,"p10":6844415,"p25":6873087,"p50":6934527,"p75":6946815,"p90":7008255,"p97_5":7008255,"p99":7008255,"p99_9":7008255,"p99_99":7008255,"p99_999":7008255},"errors":0,"timeouts":0,"mismatches":0,"duration":5.02,"start":"2020-09-10T12:37:53.067Z","finish":"2020-09-10T12:37:58.083Z","connections":5,"pipelining":1,"non2xx":0,"1xx":0,"2xx":2356,"3xx":0,"4xx":0,"5xx":0,"server":"0.16-apollo-express"}
```
craicoverflow, line 1: As a pattern going forward, should we always specify the patch version in our results? So `performance/results/0.16-apollo-fastify.json` would be `performance/results/0.16.0-apollo-fastify.json`.

it helps us measure the difference to the most granular level possible.

machi1990: :-) I just changed this to only measure the minor and major release. In an ideal world, patch releases should not introduce performance issues this is what I had in mind. But it does not cost much to measure this. Happy to revert this change it.

package.json
craicoverflow: It can be overridden easily by combining multiple compose files. So a Windows user could override it with a Windows path mapping. Probably no need though unless someone comes asking :)

machi1990: Okay. Sure, it can be done in followups.

craicoverflow: Other reasons:

- Generally easier to control the flags passed like `-d`, `--verbose` etc - Since `npm run bench` runs after the Docker command here I can't access those Docker flags.

performance/results/0.16-apollo-express.json
craicoverflow: Yeah in an ideal world they shouldn't, I guess sometimes they do - for example, a refactor is often a patch release since there is no feature added - but this ideally should improve performance.

machi1990: Exactly. Thanks for spotting this. Also, we can later on we can add a script to remove old files e.g files from previous three minor / major releases so that this folder does not grown incessantly.

machi1990:
> This looks good! I am seeing the results between runs are closer together.

Great. Thanks for the review. Let me promote the PR as ready for review then ;-)

craicoverflow (approved):

Issue #2009, opened 2020-09-07
Semantics of how the `or` operator should be used


craicoverflow:
* **Module**: `@graphback/runtime-mongodb`, `@graphback/runtime-knex`
* **Version**: *
* **Node.js / npm versions:** NA

When combined with the `root` query operators, the `or` operator evaluates to `find X where <root query conditions> OR <or conditions>`. Take the example below. This outputs the following pseudo-query in both Postgres and MongoDB: `find Blogs where name is "blog3" or name is "blog3"` or condition3....

```graphql
{
  findBlogs(filter: {name: {eq: "blog3"}, or: [{name: {eq: "blog1"}}]}) {
    items {
      id
      name
    }
  }
}
```

Upon initial raising of this as being potentially incorrect by @machi1990 and then checking out alternative libraries (namely AWS AppSync), the query above should output: `find Blogs where name is "blog3" and (name is "blog3" || condition2 || condition 3)...`

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

craicoverflow:
Thanks for highlighting this one @machi1990 - this is in fact a bug and should be actioned ASAP.

craicoverflow:
Going to start this.

machi1990:
Thanks @craicoverflow

CoreyKovalik:
This should be re-opened, as I believe there are still bugs in the knex query builder.  We are experiencing that this simply _**does not work as expected**_ for Postgres + Knex for any and/or query on the same field.

Example:

```
// returns no results
query {
  findLocations (filter: { country_code: { eq: "US" }, or: { country_code: { eq: "CN" } } })
  
// the following 2 queries seem to work as expected, and return different results....
query {
  findLocations (filter: { country_code: { eq: "US" }, or: { city: { contains: "a" } } })

query {
  findLocations (filter: { country_code: { eq: "CN" }, or: { city: { contains: "a" } } })
  
// However, `or` seems to be operate no different than `and`.
// Expected: locations return with Either `country_code` === "US" or countries starting with `C`, i.e. `China`
// Actual: No results, assuming because "United States" doesn't have "C"
findLocations (filter: { country_code: { eq: "US" }, or: { country_name: { contains: "C" } } }
```

Issue #1967, opened 2020-08-31
Search case sensitivity


evanshortiss:
**Is your feature request related to a problem? Please describe.**

Yes. I'm trying to query a string field, but my query fails since it's case sensitive. For example the DB contains "GITHUB", but my search contains "github".

**Describe the solution you'd like**

The ability to toggle case sensitivity for a query using a Boolean.

**Describe alternatives you've considered**

Could regular expressions be an option? So instead of `findJunctions (filter: { name: { contains: "BREA" } })` a developer could pass `findJunctions (filter: { name: { regex "/brea/ig" } })`.

machi1990:
Automatically generated comment to notify maintainers
/cc @craicoverflow, @machi1990, @wtrocki

craicoverflow:
> The ability to toggle case sensitivity for a query using a Boolean.

I've thought about this and though it would be easy enough to add, the filter input fields are purely to map to an underlying database table/collection - adding case sensitivity to the filter would also mix configuration into it this. We don't want to make the inputs too complex.

In your particular use case, do you see the need to combine `{name: { contains: "BREA" }}` with other conditions, or is this solely as a search query? 
If the latter, this would be a great additional plugin:

- https://www.postgresql.org/docs/9.5/textsearch.html
- https://docs.mongodb.com/manual/text-search/

wtrocki:
Yes. Text search is one of the proposed goals on roadmap post 1.0. 
Doing any search, filtering on general entities now requires building complex filtering.

wtrocki:
Actually we can see if we can do something like Hasura done: 
https://hasura.io/docs/1.0/graphql/core/queries/query-filters.html#text-search-or-pattern-matching-operators-like-similar-etc

evanshortiss:
@craicoverflow exactly, this is a search query on a string field and supporting the DB-level text search would be ideal since it removes most ambiguity. 
> great additional plugin

I'm not incredibly famiilar with how Graphback internals operate; are the various filters implemented as plugins? Once a plugin is authored it can be exposed as a filter in the schema, similar to what @wtrocki linked?

craicoverflow:
Yes, the various filters that you see are applied to the schema through the [`SchemaCRUDPlugin`](https://graphback.dev/docs/plugins/crud-schema), this is a default plugin that is automatically executed on startup. 

What this plugin does:

[...]

PR #2034, opened 2020-09-11
refactor(knex): refactor knex query builder and fixes an issue where or queries were not mapped correctly

craicoverflow:
This PR completely rewrites the Knex query builder, which was only ever intended as a prototype version originally and was very hard to maintain.

## Tasks

- [x] Fixes an issue where the or operator does not map to SQL as it should #2009  
- [x] Refactor query builder

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,6 +8,17 @@ title: Release Notes
 This file contains changes and migration steps for the Graphback project. 
 Please follow individual releases for more information.
 
+# 0.17.0
+
+### Bug Fixes
+
+* Logical `or` filter selector was not mapped correctly in both `graphback-runtime-mongo` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [1ebe7e9](https://github.com/aerogear/graphback/pull/2034/commits/1ebe7e9bc8d3a61f0b3ef65b588881d16b7ae63f))
+* Logical `or` filter selector was not mapped correctly in both `graphback-runtime-knex` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [6d43f28](https://github.com/aerogear/graphback/commit/6d43f288865a2c8c0d441e486a156301ca6cc42a))
+
+### Breaking Changes
+
+* Refactored the Knex Query Mapper ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [6d43f28](https://github.com/aerogear/graphback/commit/6d43f288865a2c8c0d441e486a156301ca6cc42a))
+
 # 0.16.2
 
 ### Bug Fixes
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -8,12 +8,13 @@ title: Release Notes
 This file contains changes and migration steps for the Graphback project. 
 Please follow individual releases for more information.
 
-# 0.17.0
+# 1.0.0
 
 ### Bug Fixes
 
-* Logical `or` filter selector was not mapped correctly in both `graphback-runtime-mongo` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [1ebe7e9](https://github.com/aerogear/graphback/pull/2034/commits/1ebe7e9bc8d3a61f0b3ef65b588881d16b7ae63f))
-* Logical `or` filter selector was not mapped correctly in both `graphback-runtime-knex` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [6d43f28](https://github.com/aerogear/graphback/commit/6d43f288865a2c8c0d441e486a156301ca6cc42a))
+* Logical `or` filter selector was not mapped correctly in `graphback-runtime-mongo` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [1ebe7e9](https://github.com/aerogear/graphback/pull/2034/commits/1ebe7e9bc8d3a61f0b3ef65b588881d16b7ae63f))
+* Logical `or` filter selector was not mapped correctly in `graphback-runtime-knex` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [6d43f28](https://github.com/aerogear/graphback/commit/6d43f288865a2c8c0d441e486a156301ca6cc42a))
+* Logical `or` predicate was not applied correctly in `createInMemoryPredicateFilter` ([#2034](https://github.com/aerogear/graphback/pull/2034), fixed by [01f9912](https://github.com/aerogear/graphback/commit/01f99121a9462e5a277657359094ab131e6f809c))
 
 ### Breaking Changes
 
--- a/packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
+++ b/packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
@@ -871,14 +871,14 @@ describe('createInMemoryFilterPredicate', () => {
       name: {
         endsWith: 'Simpson'
       },
-      and: {
+      and: [{
         age: {
           gt: 10
         },
         verified: {
           eq: true
         }
-      }
+      }]
     }
 
     const filterSubscription = createInMemoryFilterPredicate<User>(filter)
@@ -929,11 +929,11 @@ describe('createInMemoryFilterPredicate', () => {
       name: {
         eq: 'Homer Simpson'
       },
-      or: {
+      or: [{
         name: {
           eq: 'Homer Thompson'
         }
-      }
+      }]
     }
 
     const filterSubscription = createInMemoryFilterPredicate<User>(filter)
--- a/packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
+++ b/packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
@@ -929,42 +929,47 @@ describe('createInMemoryFilterPredicate', () => {
       name: {
         eq: 'Homer Simpson'
       },
-      or: [{
-        name: {
-          eq: 'Homer Thompson'
-        }
-      }]
-    }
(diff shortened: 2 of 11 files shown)


wtrocki:
Verifying

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
```diff
@@ -736,3 +732,379 @@ type Todo {
...
+      },
+      {
+        not: {
+          a: {
+            eq: 2
+          }
+        }
+      }
+    ]
+  }
```
machi1990, line 1105: If I read this correctly, this equals: `either "a" in [1, 11] or "a" not equal 2`, right? If so, should the expectation be 
`expect(items).toHaveLength(3)` since `{a: 3}` satisfies the condition too?

packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
```diff
@@ -871,14 +871,14 @@ describe('createInMemoryFilterPredicate', () => {
       name: {
         endsWith: 'Simpson'
       },
-      and: {
+      and: [{
```
machi1990, line 874: General comment about in memory filter: 

Good to see this file modified, should we also update this condition:  

https://github.com/aerogear/graphback/blob/c0d25d556583a3df7fd85e1fbaf0a43a245b2586/packages/graphback-core/src/runtime/createInMemoryFilterPredicate.ts#L146
?

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
craicoverflow: Yes - I read it wrong and it confused me, looks like this one is wrong!

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
```diff
@@ -736,3 +732,379 @@ type Todo {
...
+              }
+            }, {
+              b: {
+                eq: 5
+              }
+            }
+          ]
+        }
+      ]
+    }
```
machi1990, line 858: This reads `either "a" equal 1 or ("c" equal 6 or "b" equal 5)`, right? 

Looking at the sample 
```ts
 [
          {
            a: 1,
            b: 5,
            c: 8
          },
          {
            a: 1,
            b: 2,
            c: 10
          },
          {
            a: 1,
            b: 5,
            c: 3
          },
          {
            a: 6,
            b: 6,
            c: 3
          }
        ]
```

I expected the query to return `3` items but the assertion `expect(items).toHaveLength(2);` seem to indicate two.

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
machi1990: I suppose it's the same for Mongo? (I've not gotten to that part yet)

craicoverflow: The tests are different so it may not be the case, Mongo might work fine. I'm hitting difficulty with this one, and it is really to do with the arrangement of the SQL operators with Knex - so I'd say the issue is not present in Mongo.

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
craicoverflow: This reads `a = 1 and (c equal 6 or b equal 5)`. Notice that there is only a single item in the first `or` array.

machi1990: > Notice that there is only a single item in the first or array.

I right. I missed this part, good catch! Thanks.

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
machi1990: Tried this out for Mongo and indeed it's not the case.  So looks like a Knex only issue.

packages/graphback-runtime-knex/src/knexQueryMapper.ts
```diff
@@ -1,140 +1,168 @@
...
+const mapQueryFilterOperatorToKnexWhereCondition = (operator: QueryFilterOperator, value: any): KnexWhereConditionMap => {
+  const operatorToWhereConditionMap: { [key in QueryFilterOperator]: KnexWhereConditionMap } = {
+    eq: ['=', value],
+    ne: ['<>', value],
+    lt: ['<', value],
+    le: ['<=', value],
+    ge: ['>=', value],
+    gt: ['>', value],
+    contains: ['like', `%${value}%`],
+    startsWith: ['like', `${value}%`],
```
machi1990, line 39: Adding new operator ( e.g the one described in https://github.com/aerogear/graphback/issues/1967#issuecomment-685824275) would be as simple as adding the mapping here? 

Same question for Mongo query builder.

craicoverflow: Yes!

machi1990 (commented):
Finished my first round of review and verification. I added a couple of comments as I was getting along the way. 

I also think we should have a section in our documentation where we explain the semantics of the filters using pseudo code as the one used in this https://github.com/aerogear/graphback/pull/2034#discussion_r487798631

packages/graphback-runtime-knex/src/knexQueryMapper.ts
machi1990: It's much better already. Thank you.

When I was looking at the Mongo part it seemed doable also, not quite similar but it is simple enough.

docs/releases.md
```diff
@@ -8,6 +8,17 @@ title: Release Notes
 This file contains changes and migration steps for the Graphback project. 
 Please follow individual releases for more information.
 
+# 0.17.0
```
machi1990, line 11: `1.0.0`  ?

craicoverflow: Hmm, we could be on v1 beta for a while and this fixes an issue which should really be fixed ASAP if someone wants to use Graphback on a stable release, that is why I did not do v1 for it.

machi1990: > Hmm, we could be on v1 beta for a while and this fixes an issue which should really be fixed ASAP if someone wants to use Graphback on a stable release, that is why I did not do v1 for it.

Okay, makes sense. I thought the idea of dropping `1.0.0` soon is after estimating that feature wise we are complete, and only bug fixes and possible performance fixes will go in. So I did not see us adding anything that could delay that release.

machi1990: /cc @wtrocki

packages/graphback-core/tests/createInMemoryFilterPredicateTest.ts
craicoverflow: Yes, I'll check that out.

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
craicoverflow: Fixed now in https://github.com/aerogear/graphback/pull/2034/commits/9cbb4f0cdb61aac4f27a59c04ea89b7c9b63cb9b

packages/graphback-runtime-knex/tests/data/KnexDbDataProviderTest.ts
```diff
@@ -4,19 +4,26 @@
...
+  if (existsSync(dbPath)) {
+    unlinkSync(dbPath)
+  }
+})
+
 afterEach(() => {
   if (existsSync(dbPath)) {
     unlinkSync(dbPath)
   }
 })
```
machi1990, line 23: I suppose this is needed because we are not using the in `memory` connection? Was there any preference of using `file system` over an in memory one?

craicoverflow: Actually I put the second one in as a temporary thing because I was running the tests in parallel in two places (watch mode and through VS Code and they were conflicting), I can take this out again.

And there probably was some reason at the time, but I feel like it has been obsoleted and can change it now.

craicoverflow: Running into issues: `SQLITE_ERROR: no such table: todo` on ever test case. I'll leave it as is for now.

machi1990: Okay, I just spotted it. As you prefer, but if we can use an in memory store maybe the test would be somehow faster and can be ran in parrallel?

machi1990 (approved):
All my comments have been addressed. Thanks Enda for a huge followup PR.

/cc @wtrocki you might want to have a look too.

docs/releases.md
wtrocki: IMHO we have been in beta stage for a while with major changes happening. 
Instead of beta it might make sense to use 1.0.0-rc.1 (but for release notes that will be 1.0.0)

No strong opinions but I feel that 1.0.0 beta is going to send mixed messages

craicoverflow: We have been breaking changes in minor releases - so in this sense it was better not to have it on rc1 - breaking changes between rc* versions sounds like a nightmare. I will set it to v1 now.

craicoverflow: See https://github.com/aerogear/graphback/pull/2049/files

It would have been hard to maintain the templates release branch with multiple breaking changes on `templates-release-1.0.0`

Issue #1954, opened 2020-08-27
Graphback CLI missing docs


machi1990:
The graphback-cli doc page is empty at the moment - https://graphback.dev/docs/cli/graphback-cli 

Opening this issue to add a paragraph or two.

machi1990:
/cc @craicoverflow, @wtrocki
Automatically generated comment to notify maintainers

craicoverflow:
https://graphback.dev/docs/cli/graphback-cli

PR #2053, opened 2020-09-15
docs(graphback-cli): add generate command documentation

machi1990:
Fixes https://github.com/aerogear/graphback/issues/1954

PS: The issue was a good first issue and tagged for Hacktoberfest but I thought we could bring this early on since https://graphback.dev/docs/cli/graphback-cli looks weird.

### DIFF
--- a/docs/cli/graphback-cli.md
+++ b/docs/cli/graphback-cli.md
@@ -3,3 +3,80 @@ id: graphback-cli
 title: Graphback CLI
 sidebar_label: Graphback CLI
 ---
+
+Graphback helps you to kickstart your experience with any existing GraphQL implementation by generating backend and client side CRUD layer using your [GraphQL data model](../model/datamodel.md). The CLI exposes a single `generate` command.  
+
+## Installation
+
+You can add `Graphback CLI` on your existing GraphQL project using either of the following commands. 
+
+With npm:
+
+```bash
+npm install --save-dev graphback-cli
+```
+
+Or with yarn: 
+
+```bash
+yarn add --dev graphback-cli
+```
+
+## Usage
+
+### Prerequisite
+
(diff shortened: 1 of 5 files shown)


docs/cli/graphback-cli.md
```diff
@@ -2,4 +2,80 @@
...
+      graphback-schema:
+        format: graphql
+        outputPath: ./src/schema/schema.graphql
+      graphback-client:
+        format: 'graphql'
+        outputFile: ./src/graphql/graphback.graphql   
+```
+
+:::note
+If you used [graphql-cli](https://github.com/Urigo/graphql-cli) to initialise the project, this file will be automatic setup for you. Otherwise, you can still use the `init` command from the cli to initialise this configuration setup.
```
craicoverflow, line 51: I know that UK English uses `s` instead of `z`, but since graphback.dev targets the world, I wonder which we should use? https://stackoverflow.com/a/157837

```suggestion
If you used [graphql-cli](https://github.com/Urigo/graphql-cli) to initialise the project, this file will be automatic setup for you. Otherwise, you can still use the `init` command from the CLI to initialize this configuration setup.
```

docs/cli/graphback-cli.md
```diff
@@ -2,4 +2,80 @@
...
+:::
+
+
+### Using the command
+
+To use Graphback generate command. In your `package.json` file, edit the `scripts` parts and add the following line:
+
+```json
+"generate": "graphback generate"
+```
```
craicoverflow, line 61: This seems like an unnecessary, opinionated way of running it. We can show the user the steps of how to run it directly with `yarn graphback generate` and they can choose to alias this if they wish?

craicoverflow (approved):

docs/cli/graphback-cli.md
machi1990: I did not want to leave npm users aside and went this way.

PR #2056, opened 2020-09-16
perf: display 99.999 percentile which is a better representation than average

machi1990:
From https://developer.nvidia.com/content/analysing-stutter-%E2%80%93-mining-more-percentiles-0:

> Percentiles are not the only way to measure stutter and debate has raged many times, here at NVIDIA, about the best approach. But I consistently find that percentiles are easy to measure; they have an intuitive interpretation; they can be used to quantitatively compare different results; and – perhaps most importantly – they always match my subjective game-play experience.

> What does a percentile mean when applied to frame times? The Tech Report has plenty of great discussion and examples. But briefly, for a given percentage, say 99%, a percentile of 35.6ms means that 99% of the frame times fall below 35.6 ms. Long frames matter for stutter analysis and so I often think of the percentile meaning the other way around: 1% of values are 35.6 ms or above.

I think display this value makes is more representative of the benchmark than the "mean" value which was previously displayed as average does not say much about data distribution.


wtrocki:
Nice!

wtrocki (approved):

PR #2057, opened 2020-09-16
refactor(core): decrease number of times model definitions are built

craicoverflow:
This splits the `models` metadata variable into a getter/setter to prevent building all models every time the `GraphbackCoreMetadata#getModelDefinitions` is called.

I ran benchmarks a few times and there is an improvement in startup time:

```
│ Server                                        │ Requests/s │ Latency │ Throughput/Mb │ Startup time (s) │ Resident Set Size (MB) │ CPU Percentage │
│ 1.0.0-graphback-apollo-express-large-schema   │ 256.4      │ 18.97   │ 3.59          │ 5.71             │ 134.94                 │ 1.51           │
│ 0.16.2-graphback-apollo-express-large-schema  │ 252.2      │ 19.27   │ 3.53          │ 6.08             │ 135.13                 │ 1.58           │
```

Issue #2035, opened 2020-09-11
Graphback templates improvments


wtrocki:
## Templates should have separation of concerns

When using Graphback templates we geting index.ts file that has multiple elements and concerns that developers need to read. It will be much better to have graphback.ts file 

See example PR that has this:
https://github.com/RedHat-Middleware-Workshops/rhtr-2020-api-mgmt-kafka-workshop/pull/3

Additionally, we could extract migrations and have this also visible for users as well. 

## Templates should use different subscription mechanisms 

In memory, subscriptions are too low

## Enterprise template is needed

We need to provide enterprise template. See #2016 

## Template can be consumable as docker images (along with graphback code)

Building all templates into quay will be desirable to be able to test stuff in container land.

machi1990:
/cc @craicoverflow, @machi1990
Automatically generated comment to notify maintainers

machi1990:
Very good issue. Thank you for opening it. 

I think in this template we can push for a "certain" architecture / project structure. E.g instead of having a `db.ts` file, maybe we can have a `create-data-provider.ts`, something self explanatory, with clear separations of folders to indicate a structure / organisation.     

/!\ Thinking out loud.

> ## Enterprise template is needed
> We need to provide an enterprise template. See #2016

Do you see this issue superseding #2016?

craicoverflow:
In my view, the current templates are "starting points" and do need need to push a certain architecture. From experience I dislike when templates have a certain architecture which I don't plan on keeping, it then creates more work for me. I would like to see an opinionated, production/enterprise template though, this can have more stuff and be more organised.

- `ts-apollo-mongodb-basic-backend`
- `ts-apollo-postgres-basic-backend`
- `ts-apollo-mongodb-enterprise-backend`
- etc...

machi1990:
> In my view, the current templates are "starting points" and do need need to push a certain architecture. From experience I dislike when templates have a certain architecture which I don't plan on keeping, it then creates more work for me. I would like to see an opinionated, production/enterprise template though, this can have more stuff and be more organised.
> 
> * `ts-apollo-mongodb-basic-backend`
> * `ts-apollo-postgres-basic-backend`
> * `ts-apollo-mongodb-enterprise-backend`
> * etc...

[...]

PR #2060, opened 2020-09-16
perf: add a raw graphback using fastify graphql benchmark server

machi1990:
This follows up my reading on
https://www.nearform.com/blog/implementing-a-graphql-gateway-with-fastify/

Checkout the numbers from the first two rows of the stats table:
bench

When I was writing this I also had ideas: 
- of a blog post for writing a graphback template / app with fastify.
- using fastify for graphql-serve instead 

WDYT?


wtrocki:
@machi1990 I have no words how awesome this work is! I is really deserving short tweet about how we improved performance.

machi1990:
> @machi1990 I have no words how awesome this work is! It is really deserving short blog post or tweet about how we improved performance. 

Let's wait for https://github.com/aerogear/graphback/pull/2057 and once that is merged, we'll see myself and @craicoverflow how we can make this happen. 

I was also thinking of a simple blog post on how to write a graphback template, but this can also be section in our docs (I'll need to think more about this idea). For now, I am leaning towards: 
- creating an issue to offer a plain fastify template (this won't have anything to do with https://github.com/aerogear/graphback/issues/2035), this template can go with the tweet / blog post about perf...
- creating an issue to update graphql-serve to use fastify. If @wtrocki and @craicoverflow agree with me on this.   

> I will also check how we can get Apollo to work the same - what is causing 200% decrease.

I do not think I'll be looking into that for now. If you'll be up for it, I'll be looking forward to your findings. 

> I will also apply those patterns to datasync starter.

Yeap.

wtrocki (approved):

craicoverflow (approved):
@machi1990 these numbers look great - to be honest I don't have a strong understanding of their meanings other than startup time, so I am happy for you to use fastify with GraphQL Serve!

machi1990:
> @machi1990 these numbers look great - to be honest I don't have a strong understanding of their meanings other than startup time, so I am happy for you to use fastify with GraphQL Serve!

Sure, here is the lowdown. 

1) request count - the higher the better (total number of requests sent within a second).
2) Latency (in milliseconds) - the lower the better, time it takes to process the request and receive a response
3) Throughput in megabyes, total load transfer, the higher the better.  
4) Startup, the time it took to receive the first `OK` response, the lower the better

The rest are just noise in terms of resource consumptions e.g memory and cpu (their measurement is not that precise but gives us an indication of how we much resource a graphback can consume during startup). The lower the better, especially in constrained container environments where you'd spin several instances (scale horizontally) of your graphback application.

wtrocki:
Makes no sense for me - graphql serve it is not going to be used in production due to in memory db that cannot be swapped. Work that I'm doing now - using container with schema provided will greatly benefit from it.

machi1990:
> Makes no sense for me - graphql serve it is not going to be used in production due to in memory db that cannot be swapped. Work that I'm doing now - using container with schema provided will greatly benefit from it.

Okay, I hear the argument but in my case, be it in dev or prod environment, i'd prefer a performant version of my app especially if the investment cost is the same. But this is not priority for now, we can revisit it later.

wtrocki:
> creating an issue to offer a plain fastify template (this won't have anything to do with #2035), this template can go with the tweet / blog post about perf...

That will be cool - will do that as well for container/datasync-starter

machi1990:
> > creating an issue to offer a plain fastify template (this won't have anything to do with #2035), this template can go with the tweet / blog post about perf...
> 
> That will be cool - will do that as well for container/datasync-starter

I've opened https://github.com/aerogear/graphback/issues/2065 to track this. It is a good first issue and a great way for someone to get familiar with various Graphback's concepts

PR #2068, opened 2020-09-17
docs: update GraphQLResolveInfo usage

machi1990:


docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,54 @@ await context.Note.findBy(args, context);
...
 You can optionally pass this to the `GraphbackCRUDService` method. A common use for this parameter is mapping the selected fields from the GraphQL query to the database query. 
 
 ```ts
 await context.Note.findBy(args, context, info);
-```
\ No newline at end of file
+```
+
+:::info
+If you want to map nested the selected fields from the GraphQL query to the database query, you can optionally pass the `nested path` as an argument to the `.findBy(..)` method.
```
craicoverflow, line 111:
```suggestion
If you want to nested map the selected fields from the GraphQL query to the database query, you can optionally pass the `nested path` as an argument to the `.findBy(..)` method. 
```

craicoverflow: I think this makes more sense

docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,54 @@ await context.Note.findBy(args, context);
...
+
+// retrieves comments owned by notes selected above. Specifying the "comments" path so start selection
+const { items: comments } = await context.Comment.findBy({noteId: {in: noteIds}}, context, info, 'comments');
+
+// ... do the rest
+```
+:::
+
+:::caution
+When using the built in `GraphbackCRUDService` such as [Crud Service](../api/graphback-core/classes/_runtime_crudservice_.crudservice.md), Graphback will use the `info` object to map the selected fields from the GraphQL query to the database query. If the requested fields from the GraphQL query are not contained by the `model` held by the service in question, then an empty selection of fields will be sent to the database. For example, if the GraphQL query contains the `Note` model informations, passing the `info` into a
```
craicoverflow, line 146:
```suggestion
When using the built in `GraphbackCRUDService` such as [CRUDService](../api/graphback-core/classes/_runtime_crudservice_.crudservice.md), Graphback will use the `info` object to map the selected fields from the GraphQL query to the database query. If the requested fields from the GraphQL query are not contained by the `model` held by the service in question, then an empty selection of fields will be sent to the database. For example, if the GraphQL query contains the `Note` model informations, passing the `info` into a
```

docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,54 @@ await context.Note.findBy(args, context);
...
+The optional `GraphQLResolveInfo` object is available to every resolver and contains information about the GraphQL schema and the current operation.
 You can optionally pass this to the `GraphbackCRUDService` method. A common use for this parameter is mapping the selected fields from the GraphQL query to the database query. 
 
 ```ts
 await context.Note.findBy(args, context, info);
-```
\ No newline at end of file
+```
+
+:::info
```
craicoverflow, line 110: It is strange to see such a large `info` block like this, these are usually reserved for small tidbits of information

image

machi1990: I agree with you. I can make it a sub section. WDYT?

docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,54 @@ await context.Note.findBy(args, context);
...
+const noteIds = allNotes.map(({ id }) => id);
+
+// retrieves comments owned by notes selected above. Specifying the "comments" path so start selection
+const { items: comments } = await context.Comment.findBy({noteId: {in: noteIds}}, context, info, 'comments');
+
+// ... do the rest
+```
+:::
+
+:::caution
```
craicoverflow, line 145: It is strange to see such a large `info` block like this, these are usually reserved for small tidbits of information

image

The yellow is hard to read in a large block like that.

machi1990: Good point. Let's make it a sub-section then.

docs/resolvers/custom-resolvers.md
craicoverflow: Yep that would be best.

craicoverflow (changes requested):
See inline comments

machi1990:
> See inline comments

Thanks for the review @craicoverflow I've pushed an update.

docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,53 @@ await context.Note.findBy(args, context);
...
-The `GraphQLResolveInfo` object is available to every resolver and contains information about the GraphQL schema and the current operation.
+The optional `GraphQLResolveInfo` object is available to every resolver and contains information about the GraphQL schema and the current operation.
 You can optionally pass this to the `GraphbackCRUDService` method. A common use for this parameter is mapping the selected fields from the GraphQL query to the database query. 
 
 ```ts
 await context.Note.findBy(args, context, info);
+```
+
+### Usage Of `GraphQLResolveInfo`
+When using the built in `GraphbackCRUDService` such as [CRUDService](../api/graphback-core/classes/_runtime_crudservice_.crudservice.md), Graphback will use the `info` object to map the selected fields from the GraphQL query to the database query. If the requested fields from the GraphQL query are not contained by the `model` held by the service in question, then an empty selection of fields will be sent to the database. For example, if the GraphQL query contains the `Note` model informations, passing the `info` into a
```
craicoverflow, line 111:
```suggestion
When using the built an implementation of the `GraphbackCRUDService` interface, such as [CRUDService](../api/graphback-core/classes/_runtime_crudservice_.crudservice.md), Graphback will use the `info` object to map the selected fields from the GraphQL query to the database query. If the requested fields from the GraphQL query are not contained by the `model` held by the service in question, then an empty selection of fields will be sent to the database. For example, if the GraphQL query contains the `Note` model informations, passing the `info` into a
```

docs/resolvers/custom-resolvers.md
```diff
@@ -100,9 +100,53 @@ await context.Note.findBy(args, context);
...
+ ```ts
+ await context.Comment.findBy(args, context, info)
+ ``` 
+ will likely return false results or crash in a relational database. 
+
+In order to avoid such cases, we recommend to pass the `info` object only when the fields requested by the incoming GraphQL query are contained by the model. 
+
+
+### Retrieving Nested Fields From `GraphQLResolveInfo`
+If you want to retrieve nested selected fields from the GraphQL query to the database query, you can optionally pass the `nested path` as an argument to the `.findBy(..)` method.
```
craicoverflow, line 121:
```suggestion
If you want to retrieve nested selected fields from the GraphQL query to the database query, you can optionally pass the `nested path` as an argument to the [`findBy`](../api/core/classes/_runtime_crudservice_.crudservice.md#findby) method. 
```

You could link to the API reference here (not sure if that link works)

craicoverflow (approved):
Left some comments, but they are mere suggestions! Approved

machi1990:
> Left some comments, but they are mere suggestions! Approved

Thanks, let me re-apply them for 0.16.x version too.

PR #2071, opened 2020-09-17
fix(graphql-migrations): migration failed for auto increments column

### LINKED ISSUE #2047: Second database migration throws an error
* **Module**: `graphql-migrations`
* **Version**: 0.16

This occurs when the primary key field is NOT "id" - it fails on `uuid: ID`, `_id: ID` etc.

```shell
> yarn develop
yarn run v1.22.5
$ ts-node src/index.ts
🚀  Server ready at http://localhost:4000/graphql
Migrated database
^C
> yarn develop
yarn run v1.22.5
$ ts-node src/index.ts
🚀  Server ready at http://localhost:4000/graphql
(node:71304) UnhandledPromiseRejectionWarning: error: alter table "public"."note" alter column "_id" drop not null - column "_id" is in a primary key
```

craicoverflow:
Fixes #2047 

Could not add a unit test for this because it is only a Postgres issue - we do need to think about integration tests for this scenario but due to it being a blocker we should get it merged and release first.

### DIFF
--- a/integration/tests/graphql-migrations-postgres.ts
+++ b/integration/tests/graphql-migrations-postgres.ts
@@ -1,0 +1,47 @@
+import * as Knex from "knex";
+
+import { buildSchema } from 'graphql';
+import { migrateDB } from "../../packages/graphql-migrations/src";
+import { Operation } from "../../packages/graphql-migrations/src/diff/Operation";
+
+
+const dbConfig: Knex.Config = {
+  client: "pg",
+  connection: {
+    user: "postgresql",
+    password: "postgres",
+    database: "users",
+    host: "localhost",
+    port: parseInt(process.env.POSTGRES_PORT, 10) || 5432
+  }
+}
+
+const db = Knex(dbConfig);
+
+beforeEach(async () => {
+  await db.raw(`DROP SCHEMA public CASCADE;
+CREATE SCHEMA public;
+GRANT ALL ON SCHEMA public TO postgresql;
+GRANT ALL ON SCHEMA public TO public;`);
+})
+
+afterAll(() => db.destroy());
+
+test('consecutive migrations', async () => {
+  const model = buildSchema(`
+  """@model"""
+  type User {
+    """@id"""
+    _id: ID
+    name: String
+  }
+`)
+
+  let ops = await migrateDB(dbConfig, model)
+  expect(ops.results).toHaveLength(3);
+  const opsResults = ops.results.map((op: Operation) => op.type);
+  expect(opsResults).toEqual(['table.create', 'table.primary.set', 'column.create'])
+
+  ops = await migrateDB(dbConfig, model)
+  expect(ops.results).toHaveLength(0);
+})
--- a/packages/graphql-migrations/src/connector/read.ts
+++ b/packages/graphql-migrations/src/connector/read.ts
@@ -11,6 +11,7 @@ import getTypeAlias from '../util/getTypeAlias'
 import getUniques from '../util/getUniques'
 import listTables from '../util/listTables'
 import transformDefaultValue from '../util/transformDefaultValue'
+import { isAutoIncrementableColumn } from '../util/isAutoIncrementableColumn'
 
 /**
  * @param {Config} config Knex configuration
@@ -92,13 +93,13 @@ class Reader {
           const columnName = this.getColumnName(key)
           if (!columnName) { continue }
           const info = columnInfo[key];
-          const {args, type} = getTypeAlias(info.type, info.maxLength);
+          const { args, type } = getTypeAlias(info.type, info.maxLength);
           const foreign = foreignKeys.find((k: any) => k.column === key);
-          const isPrimaryKey  = primaries.some((primary:  {column: string, indexName: string }) => primary.column === columnName);
+          const isPrimaryKey = primaries.some((primary: { column: string, indexName: string }) => primary.column === columnName);
           const column: TableColumn = {
             name: columnName,
             isPrimaryKey,
-            autoIncrementable: type === "increments" || type === "bigIncrements", // TODO
+            autoIncrementable: isPrimaryKey && isAutoIncrementableColumn(this.config.client.toString(), info),
             comment: this.getComment(columnComments, key),
             annotations: {},
             args,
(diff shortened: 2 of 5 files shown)


packages/graphql-migrations/src/connector/read.ts
```diff
@@ -92,13 +93,13 @@ class Reader {
...
-          const {args, type} = getTypeAlias(info.type, info.maxLength);
+          const { args, type } = getTypeAlias(info.type, info.maxLength);
           const foreign = foreignKeys.find((k: any) => k.column === key);
-          const isPrimaryKey  = primaries.some((primary:  {column: string, indexName: string }) => primary.column === columnName);
+          const isPrimaryKey = primaries.some((primary: { column: string, indexName: string }) => primary.column === columnName);
           const column: TableColumn = {
             name: columnName,
             isPrimaryKey,
-            autoIncrementable: type === "increments" || type === "bigIncrements", // TODO
+            autoIncrementable: isAutoIncrementableColumn(this.config.client.toString(), info),
```
machi1990, line 102:
```suggestion
            autoIncrementable: isPrimaryKey && isAutoIncrementableColumn(this.config.client.toString(), info),
```

machi1990:
I've added one comment. I'll properly test this tomorrow.

packages/graphql-migrations/src/connector/read.ts
craicoverflow: What additional benefit will this bring? Non-primary key columns can also be auto-incrementing, so adding this check would ignore that.

packages/graphql-migrations/src/util/isAutoIncrementableColumn.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Detect if the column is incrementable based on column information
+ * 
+ * @param {string} client - database client 
+ * @param {Knex.ColumnInfo} columnInfo - metadata for column
+ */
+export function isAutoIncrementableColumn (client: string, columnInfo: Knex.ColumnInfo): boolean {
+  switch (client) {
+    case 'pg':
+      return columnInfo.defaultValue?.toString().startsWith('nextval(');
```
machi1990, line 12: Should we check if the type is `integer` or `bigInteger` too? Because a `varchar` with a default value of `'nextval(dshdjhsjdh)'` whould be auto-incrementable too with this logic.

machi1990 (changes requested):
> I've added one comment. I'll properly test this tomorrow.

I am still seeing the error

```log
alter table "public"."note" alter column "_id" drop not null - column "_id" is in a primary key
```

I used the postgres templates.

I guess it might be the fact that the the column in database has a default value while the parsed column from the schema does not. This does not go well with the diff computation.

packages/graphql-migrations/src/connector/read.ts
machi1990: You are right, but in our case I thought the auto-increment-ability was only handled for primary keys (worth checking) hence my suggestion. My worry is "if we allow other columns (from dbs) to be auto-incrementable while they are not when parsed from schema" could open another set of bugs with the diff algorithm.

packages/graphql-migrations/src/util/isAutoIncrementableColumn.ts
craicoverflow: My reasoning for not checking it is because this is what is being read from the remote database, and if the default value is `nextval('whatever')`  that is an error on the remote database and it should not handed by Postgres:

```sql
create table IF NOT EXISTS testtable (
	column1 varchar default nextval('xxxxx'),
	secondcol text
);
```

This throws:

```
ERROR:  relation "dshdjhsjdh" does not exist
SQL state: 42P01
```

I have also been able to successfully create a sequence and auto-incrementing logic on a varchar.

You can try it out here and see the auto-incrementing in action:

```sql
CREATE SEQUENCE if not exists testtable_id_seq;

create table IF NOT EXISTS testtable (
	column1 varchar default nextval("'testtable_id_seq'"),
	secondcol text
);

insert into testtable (secondcol) values ('hello');
```

The `testtable_id_seq` sequence is decoupled from the datatypes which use it, and "1", "2"... are also valid varchars so it works just fine.

craicoverflow:
> I am still seeing the error

Weird, you  have run `yarn build`?

packages/graphql-migrations/src/connector/read.ts
craicoverflow: Yes, you are right about this. Thanks :+1:

packages/graphql-migrations/src/util/isAutoIncrementableColumn.ts
machi1990: Okay, thanks for the findings and great explanation. My reason was highly biased on `ID`as `integer`.  Sorry for the noise.

machi1990:
> > I am still seeing the error
> 
> Weird, you have run `yarn build`?

Yes.

integration/tests/graphql-migrations-postgres.ts
```diff
@@ -0,0 +1,47 @@
...
+
+test('consecutive migrations', async () => {
+  const model = buildSchema(`
+  """@model"""
+  type User {
+    """@id"""
+    _id: ID
+    name: String
+  }
+`)
```
machi1990, line 38: <3 Thanks!

craicoverflow: Not sure why you are still getting the error - your old database could have data in it that is preventing the column from being dropped.

machi1990:
> > I've added one comment. I'll properly test this tomorrow.
> 
> I am still seeing the error
> 
> ```
> alter table "public"."note" alter column "_id" drop not null - column "_id" is in a primary key
> ```
> 
> I used the postgres templates.
> 
> I guess it might be the fact that the the column in database has a default value while the parsed column from the schema does not. This does not go well with the diff computation.

I was testing the first patch and forgot to rebase over the latest change you've made today. This seems to be no longer the issue and the PR fixes the problem. Thanks for driving this baby home :-)

machi1990 (approved):

Issue #2076, opened 2020-09-18
Datasync cannot be used as plugin for generating schema to file.


wtrocki:
Due to different name loading datasync plugin from core is not possible. 
Using package name is also not possible due to `@` sign.

Error:
Cannot find module '@graphback/codegen-datasync'

er/.graphqlrc.yml:
Plain value cannot start with reserved character @ at line 26, column 7:

machi1990:
/cc @ssd71
/cc @craicoverflow, @machi1990
Automatically generated comment to notify maintainers

craicoverflow:
This format would work well 

```yaml
plugins:
  - name: '@graphback/codegen-schema'
    format: 'graphql'
  - name: '@graphback/datasync'
    format: 'graphql'
```

but since the plugin config is strongly associated with GraphQL CLI, this would be a major change, so the backward compatible suggestion proposed in #2077 is the best solution!

machi1990:
Was this fixed by https://github.com/aerogear/graphback/pull/2077? 

/cc @craicoverflow @wtrocki

craicoverflow:
Yes :+1:

PR #2077, opened 2020-09-18
feat: add ability to perform package override in config

wtrocki:
I have noticed that https://github.com/aerogear/graphback/blob/package-override/docs/plugins/creating-your-own-plugin.md no longer has information about config as well.

### DIFF
--- a/packages/graphback/src/loadPlugins.ts
+++ b/packages/graphback/src/loadPlugins.ts
@@ -1,23 +1,41 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 import { GraphbackPlugin } from "@graphback/core";
 
-export function loadPlugins(pluginConfig: any): GraphbackPlugin[] {
-  if (!pluginConfig) {
+interface PluginDefaultProps {
+  packageName?: string
+}
+
+interface PluginConfigMap {
+  [pluginName: string]: PluginConfig
+} 
+
+interface PluginConfig extends PluginDefaultProps {
+  [key: string]: any
+}
+
+export function loadPlugins(pluginConfigMap: PluginConfigMap): GraphbackPlugin[] {
+  if (!pluginConfigMap) {
     return [];
   }
 
   const pluginInstances = [];
-  for (const pluginLabel of Object.keys(pluginConfig)) {
+  for (const pluginLabel of Object.keys(pluginConfigMap)) {
     let pluginName = pluginLabel;
     if (pluginLabel.startsWith('graphback-')) {
       // Internal graphback plugins needs rename
       pluginName = pluginLabel.replace('graphback-', '@graphback/codegen-');
     }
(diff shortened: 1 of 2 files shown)


wtrocki:
#2076

craicoverflow (dismissed):
```
src/loadPlugins.ts(16,8): error TS2304: Cannot find name 'config'.
src/loadPlugins.ts(17,20): error TS2304: Cannot find name 'config'.
```

Did you run this? Want me to take it over?

craicoverflow:
Pushed some changes to fix this, will add documentation now also.

wtrocki:
I'm looking if that is something you would accept. I see couple issues:

- It looks like we completely removed the graphql config side of things from the docs
- Currently config is used only for client side generation. 
- WOuld this approach work for you.

I have made change in node_modules and ported it (but improperly).

craicoverflow:
> Would this approach work for you?

Yes, although I changed the property name from `packageOverride` to `packageNameOverride` to be more declarative.

I will look to add docs change in another PR and we can get this merged and released.

wtrocki:
@craicoverflow - I found way out. If I define output file in runtime (schema plugin) I will get schema updated so this is not priority but when we land this I would be able to do proper setup in datasync starter. 

Thank you so much!

craicoverflow:
```yaml
plugins:
  - name: '@graphback/codegen-schema'
    format: 'graphql'
  - name: '@graphback/datasync'
    format: 'graphql'
```

wtrocki (commented):
Cannot approve but amazing work

packages/graphback/src/loadPlugins.ts
```diff
@@ -1,23 +1,28 @@
...
-  for (const pluginLabel of Object.keys(pluginConfig)) {
+  for (const pluginLabel of Object.keys(pluginConfigMap)) {
     let pluginName = pluginLabel;
     if (pluginLabel.startsWith('graphback-')) {
       // Internal graphback plugins needs rename
       pluginName = pluginLabel.replace('graphback-', '@graphback/codegen-');
     }
+    // override package name
+    if(pluginConfigMap[pluginLabel].packageNameOverride){
+      pluginName = pluginConfigMap[pluginLabel].packageNameOverride;
```
wtrocki, line 18:
```suggestion
      pluginName = pluginConfigMap[pluginLabel].packageName
```

craicoverflow: Updated now in https://github.com/aerogear/graphback/pull/2077/commits/0c2dcd6dea81a50ecb5cf339cf535a1999c15680

craicoverflow:
> Cannot approve but amazing work

I can approve but should not approve myself, @machi1990 mind taking a look?

machi1990 (approved):

PR #2081, opened 2020-09-18
docs(plugins): add information on plugin configuration with graphql-config

craicoverflow:


docs/plugins/overview.md
```diff
@@ -8,4 +8,36 @@ With Graphback plugins you can customise the generation of the GraphQL schema, a
...
+  dataProviderCreator: createKnexDbProvider(db),
+  // highlight-start
+  plugins: [
+    new SchemaCRUDPlugin()
+  ]
+  // highlight-end
+});
+```
+
+Invoking `DataSyncPlugin` like this will update the in-memory schema and resolvers and return them as `typeDefs` and `resolvers`.
```
machi1990, line 38:
```suggestion
Invoking `SchemaCRUDPlugin` like this will update the in-memory schema and resolvers and return them as `typeDefs` and `resolvers`. 
```

machi1990: This was marked as resolved but `DataSyncPlugin` is still there and I think it should be `SchemaCRUDPlugin`. Right?

craicoverflow: Does it get marked resolved automatically when a new commit is pushed? Anyway, if not it was a mistake.

machi1990 (approved):
LGTM

PR #2091, opened 2020-09-21
chore(deps): set Graphback version to `next` in templates

craicoverflow:
This PR sets Graphback dependency version to `next` in each of the templates. `next` is an alias for the latest beta version and we need the templates to use this. This saves a manual bump of the version.

@wtrocki maybe there is a better way to do this with renovate?


wtrocki (approved):

wtrocki:
I think there is better way do do this with yarn workspaces, but this is something that solves current problem and renovate/workspaces is something we tried and it might require research so.. Good to have it this way

PR #2101, opened 2020-09-22
fix(keycloak-authz): make relationship field auth configurable at GraphQL field level

craicoverflow:
Relationship fields needed to be configured with the database field name. This made it confusing to set auth rules since often the user does not know what the database column name is (it usually maps to `{fieldName}Id`, but this knowledge is not apparent from the schema and is hard to track.

This fixes that so it can be configured with the model field name instead.

This PR also updates the docs, which showed configuration happening from the 1:M side, whereas it should be from the M:1 side.

Before and after:

```patch
Comment: {
  relations: {
-    noteId: {
+    note: {
       roles: ['admin']
    }
  }
}
```

### DIFF
--- a/docs/authentication/keycloak.md
+++ b/docs/authentication/keycloak.md
@@ -57,24 +57,23 @@ With this configuration the following rules are in place.
 
 ## Relationships Autorization
 
-Developers can also add authorization rules on sepecific relationships for data fetching purposes.
+Developers can also add authorization rules on specific one-to-many relationships for data fetching purposes.
 Relationship rules will be added on top of the existing rules that are defined for the individual objects.
+
+To apply a relationship field rule to the one-to-many field `Task.users`, you must configure it on the `User.task` configuration object, the inverse field which `Task.users` maps to.
 
 ```ts
 const authConfig = {
-  Task: {
+  User: {
     relations: {
-        taskUsers: { roles: ['admin'] }
-        allTasksComments: { roles: ['commenter'] }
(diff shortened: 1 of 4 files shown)


machi1990 (commented):
Makes sense to me.

Can we add a test?

craicoverflow:
A test exists already at https://github.com/craicoverflow/graphback/blob/keycloak-fix/packages/graphback-keycloak-authz/tests/KeycloakCrudService.test.ts#L358 .. but I think I can improve this.

machi1990:
> A test exists already at https://github.com/craicoverflow/graphback/blob/keycloak-fix/packages/graphback-keycloak-authz/tests/KeycloakCrudService.test.ts#L358 .. but I think I can improve this as it can be configured to give a false positive (as it was doing already). I will work on that now.

Thanks

wtrocki (approved):

craicoverflow:
> A test exists already at [https://github.com/craicoverflow/graphback/blob/keycloak-fix/packages/graphback-keycloak-authz/tests/KeycloakCrudService.test.ts#L358](https://github.com/craicoverflow/graphback/blob/keycloak-fix/packages/graphback-keycloak-authz/tests/KeycloakCrudService.test.ts?rgh-link-date=2020-09-22T13%3A46%3A10Z#L358) .. but I think I can improve this as it can be configured to give a false positive (as it was doing already). I will work on that now.

Actually...this test validates is correctly, no need for changes.

Issue #2103, opened 2020-09-22
Could not find a declaration file for module 'bson'


craicoverflow:
* **Module**: `@graphback/core`
* **Version**: *
* **Node.js / npm versions:**
* Example/snippet of the model that is causing issue

This error occurs when I run `yarn build` from the Postgres template

```shell
> yarn build
yarn run v1.22.5
$ tsc
node_modules/@graphback/core/types/runtime/QueryFilter.d.ts:1:26 - error TS7016: Could not find a declaration file for module 'bson'. '/home/ephelan/code/playground/graphback/testing/next/postgres-backend/node_modules/bson/lib/bson.js' implicitly has an 'any' type.
  Try `npm install @types/bson` if it exists or add a new declaration (.d.ts) file containing `declare module 'bson';`

1 import { ObjectID } from 'bson';
                           ~~~~~~


Found 1 error.

error Command failed with exit code 2.

```

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

machi1990:
Does the error occur too when the template is outside the Graphback repo?

craicoverflow:
I only experienced this when outside the repo - I am doing all testing of templates in isolation.

machi1990:
Okay, thanks for clarifications.

PR #2104, opened 2020-09-22
Fix script path to compiled `index.js` in templates and move `@types/bson` from devDependencies in `@graphback/core`

craicoverflow:
- [`81f9612` (#2104)](https://github.com/aerogear/graphback/pull/2104/commits/81f96128e3daa27a2345070ab961b0df01c1d102) fixes an issue in the templates where path to `dist/index.js` was wrong in the `start` script.

- [`e9f6460` (#2104)](https://github.com/aerogear/graphback/pull/2104/commits/e9f6460035f676e03445a99a4d36500984a9e083) fixes https://github.com/aerogear/graphback/issues/2103.

### DIFF
--- a/packages/graphback-core/package.json
+++ b/packages/graphback-core/package.json
@@ -28,7 +28,6 @@
     "node": ">=10.15.3"
   },
   "devDependencies": {
-    "@types/bson": "4.0.2",
     "@types/jest": "26.0.14",
     "@types/node": "12.12.62",
     "@types/pino": "6.3.0",
@@ -47,6 +46,7 @@
   "dependencies": {
     "@graphql-tools/merge": "6.2.3",
     "@graphql-tools/utils": "6.2.3",
+    "@types/bson": "4.0.2",
     "bson": "4.1.0",
     "dataloader": "2.0.0",
     "glob": "7.1.6",
--- a/templates/ts-apollo-mongodb-backend/package.json
+++ b/templates/ts-apollo-mongodb-backend/package.json
@@ -3,11 +3,11 @@
   "version": "1.0.0",
   "description": "Graphback runtime template with Apollo Server and MongoDb",
   "private": true,
-  "main": "dist/src/index.js",
+  "main": "dist/index.js",
   "scripts": {
     "build": "tsc",
     "develop": "ts-node src/index.ts",
-    "start": "node dist/src/index.js",
+    "start": "node dist/index.js",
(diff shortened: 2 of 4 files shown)


packages/graphback-core/package.json
```diff
@@ -47,6 +46,7 @@
   "dependencies": {
     "@graphql-tools/merge": "6.2.3",
     "@graphql-tools/utils": "6.2.3",
+    "@types/bson": "4.0.2",
```
machi1990, line 49: Hmm is this needed?

craicoverflow: See https://github.com/aerogear/graphback/issues/2103

machi1990: Really surprising that this has to be a "prod" dependency. I wonder if the error in #2103 occurs if the template is outside Graphback repo. If not, we probably should let this as devDependencies.

craicoverflow: I only experienced this when outside the repo - I am doing all testing of templates in isolation.

wtrocki: I have seen this logged by some community member on graphql-cli as well before. Good change!

wtrocki: Basically this should be reexported in d.ts files generated.

craicoverflow: Apparently it is not always a given that `@types` should be installed as a devDependency, it is not very clear to me, but it can happen. See https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338

wtrocki: This happens when type is being used internally but it is reexpored indirectly. All those db types etc. When dealing with compiled versions you do not need to deal with that - it is just that d.ts files require it on API level. Some people do reexport them when compiling (that will add them to d.ts) but @craicoverflow fix is amazing for me

machi1990: Okay. I'd not expect this to be an issue for an already compiled / transpiled dependency. This is like typescript trying to recompile graphback or something.

wtrocki (approved):
Installed it on latest beta. Works fine for me! 
Amazing work!

packages/graphback-core/package.json
wtrocki: More like if you trying to reference graphback in your code typescript will inspect d.ts files etc..

Issue #1965, opened 2020-08-31
The closing block of a generated Graphback query is not inline


craicoverflow:
* **Module**: graphback-codegen-client
* **Version**: *
* **Node.js / npm versions:** *

This is a minor styling bug I noticed in the queries that are generated by the `@graphback/codegen-client` plugin.

This does not affect functionality in any way, but would be nice to get fixed.

```patch
query findNotes($filter: NoteFilter, $page: PageRequest, $orderBy: OrderByInput) {
    findNotes(filter: $filter, page: $page, orderBy: $orderBy) {
      items {
        ...NoteExpandedFields
      }
      offset
      limit
      count
    }
- }
+}
```

### Testing your changes

From the `templates/ts-react-apollo-client` directory, run `yarn graphback generate`.

Note, that you need to change the path of `extensions.graphback.model` in the `.graphqlrc` file to point to a valid model file ([this one](https://github.com/aerogear/graphback/blob/f262f4cd5d36dc1ef13e9d7188ce9ff7389898ab/templates/ts-apollo-postgres-backend/model/datamodel.graphql#L21)).

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

shivambhatia5432:
Hi, I would like to take this issue can you guide me in right direction.

craicoverflow:
Hi @shivambhatia5432 - the issue description should hopefully have enough information for you to get started. The best thing to do is understand what Graphback does by looking at [our docs](https://graphback.dev/docs/next/introduction).

This issue specifically is located in the [ClientCRUD Plugin](https://graphback.dev/docs/next/plugins/crud-client). You can find this in the `graphback-codegen-client` package in this repo.

shivambhatia5432:
Hi, can you please elaborate the issue I am on exact same line and file you are talking about, I cannot understand meaning of " to be inline with `q`." as I am not able to generate Graphback query.
Thanks for assigning me this issue.

craicoverflow:
Are you in the `graphback.graphql` file? This file is generated when you run `yarn graphback generate` from the template. The ClientCRUDPlugin is a plugin which generates this code and writes it to the `graphback.graphql` file.

> I cannot understand meaning of " to be inline with q.

Sorry, I should be more clear. The closing bracket should be at the same indentation level as the `q` from the start of `query findNotes`.

[...]

PR #2107, opened 2020-09-22
#1965 Solved (The closing block of a generated Graphback query is not inline)

shivambhatia5432:
Indentation of end bracket of query corrected. I hope it is correct according to the requirements.

### DIFF
--- a/templates/ts-react-apollo-client/src/graphql/graphback.graphql
+++ b/templates/ts-react-apollo-client/src/graphql/graphback.graphql
@@ -43,13 +43,13 @@ query findNotes($filter: NoteFilter, $page: PageRequest, $orderBy: OrderByInput)
       limit
       count
     }
-  }
+ }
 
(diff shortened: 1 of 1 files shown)


machi1990 (changes requested):
Hi @shivambhatia5432 thanks for opening this PR. It's all good except that the file modified is a generated file. The piece of code we should be looking at modifying is this:

https://github.com/aerogear/graphback/blob/5bdb505ff9ad80344af26ab29055432f56985306/packages/graphback-codegen-client/src/templates/gqlTemplates.ts#L6-L94. 

You can modify the templates there:

- Do a `yarn build` in the project root 
- and then `yarn generate` in the https://github.com/aerogear/graphback/blob/3e6633fe6433ff96206f3450ccf2def1a4cf809b/templates/ts-react-apollo-client folder to check if all is good. 

I am available for any questions

Issue #1856, opened 2020-08-11
Expose MAX_RETRIES as an environment variable to templates


machi1990:
Context: https://github.com/aerogear/graphback/pull/1833#discussion_r467787854

The `MAX_RETRIES` value for conflict resolution is hard-coded into Graphback. It would be very useful if this was configurable from the templates through an environment variable.

If the user does not set this value, it should fall back on a default (the current value).

https://github.com/aerogear/graphback/blob/8763612f240a59cd15c011333c7d6fab530050ec/packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts#L6

machi1990:
/cc @craicoverflow, @wtrocki
Automatically generated comment to notify maintainers

craicoverflow:
Could you add more to the description of this issue? This is a nice first issue for someone but requires reading of old PRs for context.

craicoverflow:
I've updated the issue description.

RinkiyaKeDad:
Hey I'd like to work on this! How would I have access to the value the user has entered for ```MAX_RETRIES```?

craicoverflow:
Thanks! By exposing it as an environment variable.

RinkiyaKeDad:
Thanks for responding! I'm not sure if I get what you mean. I don't see a .env file in the providers folder so how will I use something like ```process.env._userInput``` in ```DataSyncConflictProvider.ts```.

craicoverflow:
You are welcome!

The template is where the .env file exists - it is already there athttps://github.com/aerogear/graphback/blob/master/templates/ts-apollo-mongodb-datasync-backend/.env

This is where the user would configure the value

There does not need to be a .env file in the providers folder, you need to export it from here.

machi1990:
+1 on this variable to be configurable via environment variables. 

As for the naming we can go with `CONFLICT_RESOLUTION_MAX_RETRIES`, but we are open to suggestions /cc @ssd71 @wtrocki 

We we'll need to do here is read the variable from `process.env`, if it's set use that value, otherwise default to `3`. 

As for the `.env` file, there is no need for it in the `DataSyncConflictProvider.ts` or the datasync package but we'll need to add this inside our datasync template, in this file https://github.com/aerogear/graphback/blob/8763612f240a59cd15c011333c7d6fab530050ec/templates/ts-apollo-mongodb-datasync-backend/.env to be precise.

[...]

PR #2108, opened 2020-09-22
Solving Issue : #1965 

shivambhatia5432:
Indentation is corrected of closing bracket with respect to 'q' of query.

### DIFF
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -25,26 +25,26 @@ export const findOneQuery = (t: ModelDefinition) => {
   const fieldName = getFieldName(t.graphqlType.name, GraphbackOperationType.FIND_ONE)
 
   return `query ${fieldName}($id: ${t.primaryKey.type}!) {
-    ${fieldName}(id: $id) {
-      ...${t.graphqlType.name}ExpandedFields
-    }
-  }`
-}
+            ${fieldName}(id: $id) {
+              ...${t.graphqlType.name}ExpandedFields
+            }
+          }`
+  }
 
 export const findQuery = (t: GraphQLObjectType) => {
   const fieldName = getFieldName(t.name, GraphbackOperationType.FIND)
   const inputTypeField = getInputTypeName(t.name, GraphbackOperationType.FIND)
 
   return `query ${fieldName}($filter: ${inputTypeField}, $page: PageRequest, $orderBy: OrderByInput) {
-    ${fieldName}(filter: $filter, page: $page, orderBy: $orderBy) {
-      items {
-        ...${ t.name}ExpandedFields
-      }
-      offset
-      limit
-      count
-    }
-  }`
+            ${fieldName}(filter: $filter, page: $page, orderBy: $orderBy) {
+              items {
+                ...${ t.name}ExpandedFields
+              }
+              offset
+              limit
+              count
+            }
+          }`
 }
 
 
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -29,7 +29,7 @@ export const findOneQuery = (t: ModelDefinition) => {
               ...${t.graphqlType.name}ExpandedFields
             }
           }`
-  }
+}
 
 export const findQuery = (t: GraphQLObjectType) => {
   const fieldName = getFieldName(t.name, GraphbackOperationType.FIND)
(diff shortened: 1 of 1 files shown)


wtrocki:
Looks like tests are failing due to not updated snapshots. Run tests with -updateSnapshot flag

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -25,26 +25,26 @@ export const findOneQuery = (t: ModelDefinition) => {
   const fieldName = getFieldName(t.graphqlType.name, GraphbackOperationType.FIND_ONE)
 
   return `query ${fieldName}($id: ${t.primaryKey.type}!) {
-    ${fieldName}(id: $id) {
-      ...${t.graphqlType.name}ExpandedFields
-    }
-  }`
+            ${fieldName}(id: $id) {
```
wtrocki, line 28: Why indentation was changed

shivambhatia5432: Okay should I only change the closing bracket with respect to 'q' of query?

shivambhatia5432 (commented):
Changes done

shivambhatia5432:
Test failed again how can I run tests with -updateSnapshot flag?

wtrocki:
yarn test. Please check jest doc for reference

craicoverflow (changes requested):
Thanks for the PR @shivambhatia5432!

Unfortunately, this did not work. When I verified you can see that the indentation is still not in line with the `q`

```graphql
query findNotes($filter: NoteFilter, $page: PageRequest, $orderBy: OrderByInput) {
    findNotes(filter: $filter, page: $page, orderBy: $orderBy) {
      items {
        ...NoteExpandedFields
      }
      offset
      limit
      count
    }
  }

query getNote($id: ID!) {
    getNote(id: $id) {
      ...NoteExpandedFields
    }
  }
```

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -28,23 +28,23 @@ export const findOneQuery = (t: ModelDefinition) => {
     ${fieldName}(id: $id) {
       ...${t.graphqlType.name}ExpandedFields
     }
-  }`
+         }`
```
craicoverflow, line 31: This is in line with the `q` in this function, but that does not necessarily mean they will line up the same way in the output file.

Read this article to help you understand how to fix the problem:

https://2ality.com/2016/05/template-literal-whitespace.html

As always, if you are still stuck we are here to help!

craicoverflow:
As mentioned in comments; this article to help you understand how to fix the problem:

https://2ality.com/2016/05/template-literal-whitespace.html

As always, if you are still stuck we are here to help!

craicoverflow:
hey @shivambhatia5432 - need any help?

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
shivambhatia5432: I am still not able to solve the issue after yarn test 17 test are failing even if I have undone all my changes. Solution in this article https://2ality.com/2016/05/template-literal-whitespace.html is to return query like this image.
Screenshot 2020-09-24 at 5 32 40 PM

craicoverflow: You don't need to use any special helpers or anything other than what is already there - you are on the right track in that you just need to position the closing bracket `}` in the right place. You are putting it on the wrong place - if you can figure out where it will go then it should work.

shivambhatia5432: I am getting 17 errors while doing yarn test, how can I check if output is correct or not, after doing the changes?

craicoverflow: From the `templates/ts-react-apollo-client` directory, run `yarn graphback generate`.

Note, that you need to change the path of `extensions.graphback.model` in the `.graphqlrc` file to point to a valid model file ([this one](https://github.com/aerogear/graphback/blob/f262f4cd5d36dc1ef13e9d7188ce9ff7389898ab/templates/ts-apollo-postgres-backend/model/datamodel.graphql#L21)).

shivambhatia5432: Gneration Successful! what should I do next? Sorry I have no past experience of open-source so I will need some help in the starting.

craicoverflow: Cool! So the generated file is `graphback.graphql` - can you check if the `}` are indented or at the start of line (check the issue for how it should look).

No problem, we will get through this!

craicoverflow: I have updated the original issue description to make it clearer (see the code block).

craicoverflow:
Sorry @shivambhatia5432 - this has been fixed in #2130. Thanks so much for your PR and collboration. We have more issues if you want to try them out such as https://github.com/aerogear/graphback/issues/1856

PR #2115, opened 2020-09-23
chore(templates): exclude directories from build

craicoverflow:
Running `yarn build` failed on fullstack templates as the `tsconfig.json` is not configured to compile React and does not have the required dependencies.

Since the client template comes with react-scripts, we can let them be built individually.


wtrocki (approved):
Verfied! good to go

Issue #2117, opened 2020-09-23
[Error: "Mutation" defined in resolvers, but not in schema]


craicoverflow:
* **Module**: `@graphback/datasync`
* **Version**: 1.0.0-beta*
* **Node.js / npm versions:**
* Example/snippet of the model that is causing issue

This error is occurring when the DataSync plugin is used but mutations are disabled.

https://github.com/aerogear/graphback/blob/8763612f240a59cd15c011333c7d6fab530050ec/packages/graphback-datasync/src/DataSyncPlugin.ts#L79-L84

This should be fixed in the DataSyncPlugin, but I will also see if we can prevent this my checking and removing empty resolver objects in `GraphbackCoreMetadata`

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

PR #2122, opened 2020-09-24
remove empty resolver creation in plugin

craicoverflow:
Fixes https://github.com/aerogear/graphback/issues/2117

The DataSyncPlugin was creating empty root Query, Mutation and Subscription resolver objects.

### DIFF
--- a/packages/graphql-serve/src/commands/printSchema.ts
+++ b/packages/graphql-serve/src/commands/printSchema.ts
@@ -22,5 +22,6 @@ export function handler(args: Params): void {
   const schemaSDL = printSchemaHandler(args);
 
   console.log("Generated schema:\n");
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   console.log(chalk.cyan(schemaSDL));
 }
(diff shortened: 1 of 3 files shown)


packages/graphql-serve/src/commands/printSchema.ts
```diff
@@ -1,5 +1,5 @@
 import yargs from 'yargs';
-import * as chalk from 'chalk';
```
craicoverflow, line 2: @machi1990 I know that you changed this back, but I was getting a build error:

`Unsafe call of an any typed value`. Any reason the import alias is preferred here?

machi1990: This was changed in https://github.com/aerogear/graphback/pull/2026 since the `printSchema` command was not working.

machi1990 (approved):

packages/graphql-serve/src/commands/printSchema.ts
machi1990: I've accepted the change without re-verifying this part as it seems unrelated. I trust you on that and the command should work.

craicoverflow: Okay it is not working for me now - what I will do is disable that line with eslint

PR #2130, opened 2020-09-25
fix: remove indent from generated query

RinkiyaKeDad:
I think I have found a fix for issue #1965. The indentation issue would have also come when using subscriptions so I've tried to fix that too. 

I could not test my changes. I followed the written instructions but had the following problem:
1. First I changed the path of extensions.graphback.model in the .graphqlrc file,
ss1
2. Then I ran the command in the templates folder and received this error,
ss2

While I wasn't able to test it I think the changes I've made should fix the issue.

### DIFF
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -28,7 +28,8 @@ export const findOneQuery = (t: ModelDefinition) => {
     ${fieldName}(id: $id) {
       ...${t.graphqlType.name}ExpandedFields
     }
-  }`
+}
+`
 }
 
 export const findQuery = (t: GraphQLObjectType) => {
@@ -44,7 +45,8 @@ export const findQuery = (t: GraphQLObjectType) => {
       limit
       count
     }
-  }`
+}
+`
 }
 
 
@@ -90,7 +92,8 @@ export const subscription = (t: GraphQLObjectType, fieldName: string, inputTypeF
   ${fieldName}(filter: $filter) {
       ...${t.name}Fields
   }
-} `
+} 
+`
 }
 
 export const createFragments = (types: ModelDefinition[]) => {
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -9,7 +9,7 @@ export const fragment = (t: GraphQLObjectType) => {
 
   return `fragment ${t.name}Fields on ${t.name} {
 ${returnFieldsString}
-} `
+}`
 }
 
 export const expandedFragment = (t: GraphQLObjectType) => {
@@ -18,7 +18,7 @@ export const expandedFragment = (t: GraphQLObjectType) => {
 
   return `fragment ${t.name}ExpandedFields on ${t.name} {
 ${returnFieldsString}
-} `
+}`
 }
 
 export const findOneQuery = (t: ModelDefinition) => {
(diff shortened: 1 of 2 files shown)


machi1990:
@RinkiyaKeDad amazing change and nice PR description, thanks. @craicoverflow mind having a look at this one?

machi1990:
@RinkiyaKeDad the build: https://github.com/aerogear/graphback/runs/1165819855?check_suite_focus=true if failing and it is related to the change you made. 

Do you mind running `yarn test -u` in the [packages/graphback-codegen-client/](https://github.com/aerogear/graphback/blob/b16aa6011c4ee00d00be3120ce27d28a01fad99d/packages/graphback-codegen-client/) directory to update the snapshot? Thanks.

wtrocki:
Looks good. Please update snapshots

RinkiyaKeDad:
Hey @machi1990 I ran the command you told in order to update the snapshots. Recevied this error:
ss3
Please help me. Thanks.

wtrocki:
> Please help me. Thanks

Every repo has contributimg.md file that everyone should follow. it is not perfect but it has all info and it will resolve problems. It is ok to ask maintainers to solve problems with code, but we cannot solve problems with your environment etc.

RinkiyaKeDad:
Really sorry @wtrocki. Didn't know this :sweat_smile:. Will take care from next time.

wtrocki:
File in question: https://github.com/aerogear/graphback/blob/master/CONTRIBUTING.md
If you follow steps your problem will be resolved.

RinkiyaKeDad:
Can you please just tell me if this is a  problem with my environment? 
ss4

I see that the package.json of graphback-codegen-client mentions ```@graphback/core``` as a dependency and just be sure I even ran ```yarn install``` in the graphback-codegen-client directory where it gave me a success message saying "Already up-to-date". This means that ```@graphback/core``` should be present in node_modules but then why is it having trouble importing it?

machi1990:
> Can you please just tell me if this is a problem with my environment?
> ss4
> 
> I see that the package.json of graphback-codegen-client mentions `@graphback/core` as a dependency and just be sure I even ran `yarn install` in the graphback-codegen-client directory where it gave me a success message saying "Already up-to-date". This means that `@graphback/core` should be present in node_modules but then why is it having trouble importing it?

Thanks for following up on this one. The error you are seeing is because the project was not built. Can you run the below command in the project root? 

```sh
yarn
yarn build
```

Once that is done, you can run the `yarn test -u` command in the `graphback-codegen-client` package. 

Let me know if that work.

wtrocki:
https://github.com/aerogear/graphback/blob/master/CONTRIBUTING.md#building

RinkiyaKeDad:
Sorry, I only ran ```yarn install``` after reading the CONTRIBUTING.md since I was under the assumption that building is something you only did before deployment :sweat_smile: Thanks a lot for helping me out :)

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
```diff
@@ -28,7 +28,8 @@ export const findOneQuery = (t: ModelDefinition) => {
     ${fieldName}(id: $id) {
       ...${t.graphqlType.name}ExpandedFields
     }
-  }`
+}
+`
```
craicoverflow, line 32: This fixes the original problem, so thank you. This also introduces an extra blank line too though:

image

craicoverflow (changes requested):
This fixes the original problem, so thank you. This also introduces an extra blank line too though:

image

packages/graphback-codegen-client/src/templates/gqlTemplates.ts
RinkiyaKeDad: @craicoverflow thanks for reviewing. This is happening for the generated mutations too, so should I fix this everywhere?
ssToday

craicoverflow: Sure! If you see other area that are wrong you can fix those too and we can review the code.

craicoverflow (approved):
Thank

PR #2131, opened 2020-09-28
chore: add perf data from 1.0.1 release

machi1990:


machi1990:
After several runs in my computer, the results are relatively the same.

machi1990:
I am going to automatically merge this as its generated anyway. 

/cc @craicoverflow @wtrocki

wtrocki:
Usually, it was also generating photo right?

machi1990:
> Usually, it was also generating photo right?

Nope. We do not have that option. It's just raw data that you can display in a table with `cd performance && yarn compare:stats`

wtrocki:
Ah perfect!

Issue #2135, opened 2020-09-30
ClientSide plugin unable to generate subscription query unless otherwise mutation operation in model  is set to `true`


machi1990:
* **Module**:  graphback/client-codegen
* **Version**: 1.0.1
* **Node.js / npm versions:**
* Example/snippet of the model that is causing issue

```graphql
"""
@model(create: false, update: false, delete: false, find: false, subCreate: true, subUpdate: true, subDelete: true)
"""
type Test {
  _id: GraphbackObjectID!
}
```

This will not generate any subscription. 

while:

```graphql
"""
@model(create: true, update: false, delete: false, find: false, subCreate: true, subUpdate: true, subDelete: true)
"""
type Test {
  _id: GraphbackObjectID!
}
``` 

Will only create the `newTest` subscription since the `@model.create` is set to true.

I think the generation should depend only on `subXXX` knobs.

machi1990:
/cc @craicoverflow, @wtrocki
Automatically generated comment to notify maintainers

wtrocki:
> Version: 1.0.1

I would see this as candidate for 1.0.2 :)

machi1990:
We should be looking at these lines:

- https://github.com/aerogear/graphback/blob/2f8bb076b524c062e00e1cf1494d9af495bf3f12/packages/graphback-codegen-client/src/templates/gqlTemplates.ts#L168
- https://github.com/aerogear/graphback/blob/2f8bb076b524c062e00e1cf1494d9af495bf3f12/packages/graphback-codegen-client/src/templates/gqlTemplates.ts#L177
- https://github.com/aerogear/graphback/blob/2f8bb076b524c062e00e1cf1494d9af495bf3f12/packages/graphback-codegen-client/src/templates/gqlTemplates.ts#L186

and modify the three condition to only check for `t.crudOptions.subXXX`.


/cc @RinkiyaKeDad this is a good issue too if you are looking for something.

PR #2139, opened 2020-10-01
fix: unable to generate subscription query unless mutation operation in model is set to `true`

RinkiyaKeDad:
I've updated the ```if``` conditions and it now depends only on the ```subXXX``` knobs. This should fix issue #2135. Please review the code. Thanks.

### DIFF
--- a/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
+++ b/packages/graphback-codegen-client/src/templates/gqlTemplates.ts
@@ -165,7 +165,7 @@ const createSubscriptions = (types: ModelDefinition[]) => {
   types.forEach((t: ModelDefinition) => {
     const name = t.graphqlType.name;
 
-    if (t.crudOptions.create && t.crudOptions.subCreate) {
+    if (t.crudOptions.subCreate) {
       const operation = getSubscriptionName(name, GraphbackOperationType.CREATE);
       const inputTypeField = getInputTypeName(t.graphqlType.name, GraphbackOperationType.SUBSCRIPTION_CREATE)
       subscriptions.push({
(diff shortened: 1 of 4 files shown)


machi1990 (changes requested):
@RinkiyaKeDad thanks you for the PR. It should be okay for `graphql` generated files. I think we should do the same for `.ts` generated files:
See
- https://github.com/aerogear/graphback/blob/e0a9f05025a06997a319aea8f8fe8ff070589cb8/packages/graphback-codegen-client/src/templates/tsTemplates.ts#L156-L191

Once that is done, can you add a snapshot test: one for `ts` and `gql`?

You can base them on 
https://github.com/aerogear/graphback/blob/e0a9f05025a06997a319aea8f8fe8ff070589cb8/packages/graphback-codegen-client/tests/GraphQLClientCreatorTest.ts#L9-L35. 

For these added tests, you can use 
```ts
 const crudMethods = {
    "create": false,
    "update": false,
    "findOne": true,
    "find": true,
    "delete": false,
    "subCreate": true,
    "subUpdate": true,
    "subDelete": true
  }
```

as a configuration.

Once that is done, we should be good to go. 

Thanks.

RinkiyaKeDad:
@machi1990 Thanks for reviewing my PR. I have never added tests before so there might be some errors. Can you please point me where I could read more about these tests and how to add them properly? Nevertheless I've tried adding them based on the ones already present. I've named the two tests and their output files arbitrarily, please tell me what would be an appropriate name for them. Thanks a lot.

machi1990:
> @machi1990 Thanks for reviewing my PR. I have never added tests before so there might be some errors. Can you please point me where I could read more about these tests and how to add them properly? Nevertheless I've tried adding them based on the ones already present. I've named the two tests and their output files arbitrarily, please tell me what would be an appropriate name for them. Thanks a lot.

Thanks for the followup. The added tests and naming looks okay to me. This is good to go.

machi1990 (approved):

PR #2155, opened 2020-10-03
Enable TypeScript declaration maps

Alexendoo:
This generates a corresponding `.d.ts.map` file for each declaration file. It means that when someone uses "go to definition" on anything imported from a graphback package they will end up at the original definition in `src/foo.ts` rather than in a type definition file `types/foo.d.ts`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "alwaysStrict": false,
     "sourceMap": true,
     "declaration": true,
+    "declarationMap": true,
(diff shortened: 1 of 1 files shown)


craicoverflow (approved):
I did not know about this config option - this is something I always wanted. Thank you so much for improving developer experience :smile:

wtrocki:
@Alexendoo God bless you for this change! It's been quite hard to work even inside repo without this.

PR #2160, opened 2020-10-06
feat(datasync): add ability to configure MAX_RETRIES 

RinkiyaKeDad:
I tried solving issue #1856 based on the discussion we had. I'm still not sure I understood how to read the variable from ```process.env```. Nevertheless I've tried fixing this. If the code is incorrect (which I think it is 😅 ) please point me towards how I can fix it.


packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -3,7 +3,7 @@ import { DataSyncModelConflictConfig, DataSyncFieldNames, ConflictMetadata, Conf
 import { MongoDeltaSource } from '../deltaSource';
 import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
 
-export const MAX_RETRIES = 3;
+export const MAX_RETRIES = process.env.CONFLICT_RESOLUTION_MAX_RETRIES ? process.env.CONFLICT_RESOLUTION_MAX_RETRIES : 3;
```
craicoverflow, line 6: What you had was correct but this was a shorter way to do it:

```suggestion
export const MAX_RETRIES = process.env.CONFLICT_RESOLUTION_MAX_RETRIES || 3;
```

craicoverflow (changes requested):
What you have is correct! Good job.

However what is missing:

- Unit tests covering this case. You can add the test cases in [this file](https://github.com/aerogear/graphback/blob/master/packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts). All they need to do is test the value of MAX_RETRIES with and without an environment variable.

- Documentation. This needs to be documented as an option in the template README so users understand what the environment variable is for https://github.com/aerogear/graphback/blob/06ed8820a87ec0c03c936416c27be19009f23f4f/templates/ts-apollo-mongodb-datasync-backend/README.md

craicoverflow:
Also, how are you doing your commits? We use conventional commits and have a hook in our scripts to enfoce you to use this. Can you make sure to change your commit message to follow this? Thanks!

RinkiyaKeDad:
Thanks for reviewing @craicoverflow! I have gone through the format of writing commits, will use it from now on. Sorry for not sticking to the convention in earlier commits.

1. I haven't written tests before but will try to write one for this.


> * test the value of MAX_RETRIES with and without an environment variable.

Can you please explain what "testing a value" means and how I would go about doing that?

2. Will update the docs once done with writing the tests so that I have better clarity :)

Much thanks :)

craicoverflow:
> Can you please explain what "testing a value" means and how I would go about doing that?

https://jestjs.io/docs/en/using-matchers

Tests are written in Jest. They have good docs, this link should point you in the right direction (with examples).

> P.S. I followed the commiting convention in a recent commit. Can you please tell if that is fine? Or what more can I do to improve my commits.

yep that is better, thank you!

RinkiyaKeDad:
Thanks for the help @craicoverflow! I've added a test for ```MAX_RETRIES``` please review it. Will proceed to write docs after that.

craicoverflow:
> Thanks for the help @craicoverflow! I've added a test for `MAX_RETRIES` please review it. Will proceed to write docs after that.

Thanks for adding the test @RinkiyaKeDad - this tests if the value exists, it should check if the value is 3.

You should also add another test case where you pass an environment variable and check if the MAX_RETRIES matches that value. You can use https://github.com/motdotla/dotenv for this.

craicoverflow:
In fact you might not even need dotenv, and might be able to use `process.env`, which would be better. See https://stackoverflow.com/a/48042799 as an example.

packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts
```diff
@@ -23,6 +23,10 @@ describe('DataSyncConflictMongoDBDataProvider', () => {
   scalar GraphbackObjectID
   `;
 
+  test('MAX_RETRIES always has a valid value', async () => {
```
craicoverflow, line 26: I think if you move this to outside the `describe` method it should work. the describe method creates a context which this test does not use, so tests are failing

RinkiyaKeDad:
Thanks for reviewing!


> it should check if the value is 3

Why are we checking the value to be 3 particularly? Should we not just check if the value is simply present or not because otherwise if the user edits the value to be something apart from 3 wouldn't the test fail?


> where you pass an environment variable

I don't think I understood this clearly. Should I just pass any environment variable? And if I do, wouldn't that test simply fail always cause the environment variable wouldn't have been set to anything?

craicoverflow:
> Why are we checking the value to be 3 particularly? Should we not just check if the value is simply present or not because otherwise if the user edits the value to be something apart from 3 wouldn't the test fail?

Because the default value is 3 when there is no environment variable present. It is not good enough to know that the value exists, but we need to be sure that the value has not changed by accident.

If I decide to do some work in this file and accidentally changed MAX_RETRIES to 3 and did not notice, them your test case will still pass, but 3 max retries would not be desired and this would introduce a bug.

> Should I just pass any environment variable? And if I do, wouldn't that test simply fail always cause the environment variable wouldn't have been set to anything?

You need to set the `CONFLICT_RESOLUTION_MAX_RETRIES` environment variable and then test if `MAX_RETRIES` matches the value. We need to know that MAX_RETRIES can be configured with environment variables and this is how you can test that.

packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts
```diff
@@ -4,6 +4,31 @@ import { ConflictError, DataSyncFieldNames, ServerSideWins, ConflictResolutionSt
...
+    process.env = { ...OLD_ENV };
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV; 
+  });
+
+  test('will receive process.env variables', () => {
+  
+    process.env.CONFLICT_RESOLUTION_MAX_RETRIES = 3;
```
craicoverflow, line 25: Use a number other than 3, since this is the default value anyway it can lead to a false positive.

packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts
```diff
@@ -4,6 +4,31 @@ import { ConflictError, DataSyncFieldNames, ServerSideWins, ConflictResolutionSt
...
+
+    //const testedModule = require('../../config/env').default --> is this relevant to this case??
```
craicoverflow, line 27: No, you can remove that.

packages/graphback-datasync/tests/DataSyncConflictProviderTest.ts
RinkiyaKeDad: Had a doubt here. TS shows a problem saying: ```Type '4' is not assignable to type 'string'.``` How do I make ```process.env.CONFLICT_RESOLUTION_MAX_RETRIES``` to not be of type string in the first place?

craicoverflow: Try it out with string:

```
process.env.CONFLICT_RESOLUTION_MAX_RETRIES = '4';
```

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -3,7 +3,7 @@ import { DataSyncModelConflictConfig, DataSyncFieldNames, ConflictMetadata, Conf
 import { MongoDeltaSource } from '../deltaSource';
 import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
 
-export const MAX_RETRIES = 3;
+export const MAX_RETRIES = process.env.CONFLICT_RESOLUTION_MAX_RETRIES || 3;
```
craicoverflow, line 6: Since `process.env` values must be a string, you will need to make a helper to do a conversion.

You can use [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) to convert from string to int, and is `isNan` set it to 3 as the default.

RinkiyaKeDad:
I've added type conversion while exporting ```MAX_RETRIES``` and assigned a value as string to ```process.env.CONFLICT_RESOLUTION_MAX_RETRIES ```.

packages/graphback-datasync/src/providers/DataSyncConflictProvider.ts
```diff
@@ -3,7 +3,7 @@ import { DataSyncModelConflictConfig, DataSyncFieldNames, ConflictMetadata, Conf
 import { MongoDeltaSource } from '../deltaSource';
 import { DataSyncMongoDBDataProvider } from './DatasyncMongoDBDataProvider';
 
-export const MAX_RETRIES = process.env.CONFLICT_RESOLUTION_MAX_RETRIES || 3;
+export const MAX_RETRIES = parseInt(process.env.CONFLICT_RESOLUTION_MAX_RETRIES, 2) || 3;
```
machi1990, line 6:
```suggestion
export const MAX_RETRIES = parseInt(process.env.CONFLICT_RESOLUTION_MAX_RETRIES, 10) || 3;
```

Why the base is 2 when parsing?

RinkiyaKeDad: Sorry I got confused :sweat_smile:

craicoverflow (approved):
Thanks for your PR! I've added some additional fixes to get this one merged.

PR #2164, opened 2020-10-06
fix: Remove performance from workspace

wtrocki:
Something that is used occasionally should not be part of the CI/CD build or local dev install.
It also list packages in the root making impression that we use them somewhere

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,8 +75,7 @@
     "templates/ts-apollo-mongodb-datasync-backend",
     "templates/ts-apollo-postgres-backend",
     "templates/ts-react-apollo-client",
-    "integration",
-    "performance"
+    "integration"
   ],
   "husky": {
     "hooks": {


machi1990:
What does this PR fix? 

I see we are removing the performance package from workspace but I do not understand what underlying problem is being fixed. 

I also have questions around:
- How do you easily produce benchmark or the "current" graphback version that's not released? 
- does this mean one has to to perform `yarn install` in the performance folder too? 

> Something that is used occasionally should not be part of the CI/CD build or local dev install.

I am more in the camp of: *even if it is done occasionally, let's make it easier and approachable enough so that it is no hassle when it is done.*

In the long ran, and preferably "perf benchmark or load testing generally" should be part of the CI/CD.

wtrocki:
@machi1990 Performance has number of packages that will be hoisted and I do not think that they should not be hoisted (same way as template) to avoid issues. I got some express issues and it was due to performance folder. Can close it as this is just my own take.

wtrocki:
Basically, express modules clashed and I could not run the template.

Issue #2163, opened 2020-10-06
Relationships are not synced for the datasync (Offix Datastore relies on them to get data)


wtrocki:
ObjectDelta missing relationship fields.

machi1990:
/cc @ssd71
/cc @craicoverflow, @machi1990
Automatically generated comment to notify maintainers

wtrocki:
FYI @kingsleyzissou - I need to fix that first to get over the required data in order to filter it.

craicoverflow:
This is down as a `bug` - but this was intended behaviour and a known limitation until now. Should it be a `feat`?

wtrocki:
Delta should contain exactly the state of the database. I see it as bug but not problem to make it feature :D

craicoverflow:
That's fine, let us keep it as a bug.

craicoverflow:
Docs will also need to be updated to include relationships in delta object.

RinkiyaKeDad:
Hey could I work on this issue? Can you please provide some more details about what needs to be done?

wtrocki:
I think @kingsleyzissou worked on this issue and we could technically close it due to solution available outside graphback. @kingsleyzissou Please close if you are happy with this

kingsleyzissou:
Yep, I'll close this issue

PR #2165, opened 2020-10-06
fix: adding relationships to datasync plugin

wtrocki:
It is not possible to simply answer relationships status for single model - relationships are build per entire schema which will make it hard to use this info in data sync that typically works with subset of types.

Creating to see how we can get that done for the demo and fix datasync.
See #2163

### DIFF
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -1,7 +1,7 @@
 import { SchemaComposer } from 'graphql-compose';
 import { IResolvers, IFieldResolver } from '@graphql-tools/utils';
-import { GraphQLNonNull, GraphQLSchema, buildSchema, GraphQLResolveInfo, GraphQLInt, GraphQLBoolean, GraphQLList, GraphQLObjectType, GraphQLField } from 'graphql';
-import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType, parseRelationshipAnnotation, GraphbackContext, GraphbackTimestamp, transformForeignKeyName } from '@graphback/core';
+import { GraphQLNonNull, GraphQLSchema, buildSchema, GraphQLResolveInfo, GraphQLInt, GraphQLBoolean, GraphQLList } from 'graphql';
+import { GraphbackCoreMetadata, GraphbackPlugin, ModelDefinition, getInputTypeName, GraphbackOperationType,  GraphbackContext, GraphbackTimestamp } from '@graphback/core';
 import { getDeltaType, getDeltaListType, getDeltaQuery } from "./deltaMappingHelper";
 import { isDataSyncService, isDataSyncModel, DataSyncFieldNames, GlobalConflictConfig, getModelConfigFromGlobal } from "./util";
 
@@ -205,55 +205,5 @@ export class DataSyncPlugin extends GraphbackPlugin {
       filter: findFilterITC.getType(),
       limit: GraphQLInt
     });
-  }
-
-  private getDeltaTypeFieldNames(modelTC: GraphQLObjectType): string[] {
-
-    const entries = Object.entries(modelTC.getFields()).map(([fieldName, field]: [string, GraphQLField<any, any>]) => {
-      const relationship = parseRelationshipAnnotation(field.description);
-      if (relationship) {
-        console.log("RELATIONSHIP", relationship)
-        if (relationship.kind === "oneToOne") {
-          return relationship.key || transformForeignKeyName(field.name);
-        }
-        if (relationship.kind === "manyToOne") {
-          return relationship.key || transformForeignKeyName(field.name);
-        }
-      }
-
-      return fieldName;
-    }).filter((entry: any) => {
-      // console.log("filter", entry)
-
-      return !!entry
-    });
-    // console.log(entries);
-
-    return entries;
-  }
-
-  private addDBKeyFields(modelTC: GraphQLObjectType): string[] {
-
-    const entries = Object.entries(modelTC.getFields()).map(([fieldName, field]: [string, GraphQLField<any, any>]) => {
-      const relationship = parseRelationshipAnnotation(field.description);
-      if (relationship) {
-        console.log("RELATIONSHIP", relationship)
-        if (relationship.kind === "oneToOne") {
-          return relationship.key || transformForeignKeyName(field.name);
-        }
-        if (relationship.kind === "manyToOne") {
-          return relationship.key || transformForeignKeyName(field.name);
-        }
-      }
-
-      return fieldName;
-    }).filter((entry: any) => {
-      // console.log("filter", entry)
-
-      return !!entry
-    });
-    // console.log(entries);
-
-    return entries;
   }
 }
--- a/packages/graphback-datasync/src/DataSyncPlugin.ts
+++ b/packages/graphback-datasync/src/DataSyncPlugin.ts
@@ -179,16 +179,12 @@ export class DataSyncPlugin extends GraphbackPlugin {
       DeltaOTC.addFields({ [field.name]: field.type.toString() })
     }
 
-    modelTC.addFields({
+    DeltaOTC.addFields({
       [DataSyncFieldNames.lastUpdatedAt]: {
         type: TimestampSTC.getType(),
         description: "@index(name: 'Datasync_lastUpdatedAt')"
       }
     });
-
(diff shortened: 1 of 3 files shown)


packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -206,13 +206,23 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
-    });
 
-    return DeltaTypeFieldEntries.map(([fieldName, _]: [string, GraphQLField<any, any>]) => {
-      return fieldName
-    });
+    return Object.entries(modelTC.getFields()).map(([fieldName, field]: [string, GraphQLField<any, any>]) => {
+      const relationship = parseRelationshipAnnotation(field.description);
+      if (relationship) {
+        if (relationship.kind === "oneToOne") {
+          return relationship.key;
```
wtrocki, line 214: Not sure what to do with `oneToOne`

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -206,13 +206,23 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+        }
+        if (relationship.kind === "manyToOne") {
+          // This seems to be always set for manyToOne
```
craicoverflow, line 217: Probably because manyToOne is always a generated field, and we always set "key".

packages/graphback-datasync/src/DataSyncPlugin.ts
```diff
@@ -206,13 +206,23 @@ export class DataSyncPlugin extends GraphbackPlugin {
...
+          return relationship.key;
+        }
+        if (relationship.kind === "oneToMany") {
+          return "";
```
craicoverflow, line 221: This means the final field list will have an empty string value:

```shell
['_id', 'title', 'description', '', '_lastUpdatedAt']
```

craicoverflow (changes requested):
Mostly good to go, left a comment inline about empty items in the field list.

Also, would it be possible to have a test case capturing the inclusion of relationship in the delta object?

This is down as a `fix` - but this was intended behaviour and a known limitation until now. Should it be a `feat`?

wtrocki:
> This is down as a fix - but this was intended behaviour and a known limitation until now. Should it be a feat?

It's a bug. 

I could not make sense for this for OneToOne - it looks like OneToOne is always unidirectional

craicoverflow:
> I could not make sense for this for OneToOne - it looks like OneToOne is always unidirectional

Yes, there is a longstanding issue for this (https://github.com/aerogear/graphback/issues/930) - it was never seen as a priority.

This PR fixed it but approach at the time seemed convoluted https://github.com/aerogear/graphback/pull/964 but actually, changing the `virtual` prop to `transient` would have made it make sense, I just did not have the lexicon at the time.

wtrocki:
@craicoverflow After an entire day of hacking, I realized we can solve the problem a better way. 

Update input type represents what delta should look like (it will have all the fields that required and even if another plugin will add fields to it they should still be part of the delta type.

This seems to be reasonable, however because of time constraints I will need to leave that fix for the moment and apply some of the workarounds.

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
```diff
@@ -19,11 +19,9 @@ type Comment {
...
 type CommentDelta {
   id: ID!
-  title: String!
-  description: String!
-
-  \\"\\"\\"@index(name: 'Datasync_lastUpdatedAt')\\"\\"\\"
-  _lastUpdatedAt: GraphbackTimestamp
+  title: String
+  description: String
+  noteId: ID
```
wtrocki, line 24: That is what was required. Question - would we be able to return that from the database?

craicoverflow: Return the column info?

craicoverflow:
> Update input type represents what delta should look like (it will have all the fields that required and even if another plugin will add fields to it they should still be part of the delta type.

True, but it will not match the required level of the fields.

wtrocki:
> True, but it will not match the required level of the fields.

Required level?

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
```diff
@@ -128,11 +131,9 @@ type Note {
...
 type NoteDelta {
   id: ID!
-  title: String!
-  description: String!
-
-  \\"\\"\\"@index(name: 'Datasync_lastUpdatedAt')\\"\\"\\"
-  _lastUpdatedAt: GraphbackTimestamp
+  title: String
+  description: String
+  mainCommentId: ID
```
wtrocki, line 136: Same for OneToOne - it works quite nicely.

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
wtrocki: What this means

craicoverflow:
> > True, but it will not match the required level of the fields.
> 
> Required level?

Nullability.

wtrocki:
> Nullability.

Still not getting it. Delta should never have null enforced. It is in very core of Delta. 
In this particular implementation, we ignoring completely data efficiency and send entire record, but.. for delta to be clear and compatible in future we need to have non nullable fields.

craicoverflow:
Thanks for explanation.

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
craicoverflow: > would we be able to return that from the database?

Can you explain what this means? I was asking if you want to return column information from database.

wtrocki: I wasn't sure if knext will return it but I see that this works.

craicoverflow: http://knexjs.org/#Builder-columnInfo

wtrocki:
I found a way to workaround this issue. Will be back to this after other deadlines.

wtrocki:
CC @kingsleyzissou

packages/graphback-datasync/tests/__snapshots__/DataSyncPluginTest.ts.snap
craicoverflow: graphql-migrations uses it:L https://github.com/aerogear/graphback/blob/83a80cdbb1104da7b36acdfa54b37a871c3ff1a0/packages/graphql-migrations/src/connector/read.ts#L84

wtrocki:
Let's discuss - I think I'm missing context for this changes and as we see integration tests failing because they causing issues/impacting overal codebase.

craicoverflow:
> Let's discuss - I think I'm missing context for this changes and as we see integration tests failing because they causing issues/impacting overal codebase.

I debugged integration test and get this error:

`Cannot query field "_lastUpdatedAt" on type "NoteDelta".`

wtrocki:
yep. Fixed that problem. That should be ready to go in terms of the graphback but for the offix we will still need information about relationships as whole. Here we getting delta with `someID` but we do not know if that is relationship or not (and offix needs to have that info)

wtrocki:
@craicoverflow I put some thoughts into this and...

This PR resolves number of issues:

- Relationships are not included
- Non nulls are included

However, it do not solve the problem for us in datastore to know which field comes from relationship.
This problem can be investigated by @kingsleyzissou and we can see how we can get that information properly.

From my point we should merge that and release it so we can test it properly with the offix and come back with future improvements. Opinions welcomed

packages/graphback-datasync/tests/__snapshots__/DataSyncPlugin.test.ts.snap
```diff
@@ -235,12 +244,12 @@ type Comment {
...
-  title: String!
-  description: String!
+  title: String
+  description: String
+  _version: Int!
 
   \\"\\"\\"@index(name: 'Datasync_lastUpdatedAt')\\"\\"\\"
   _lastUpdatedAt: GraphbackTimestamp
-  _version: Int
   _deleted: Boolean
```
wtrocki, line 253: deleted should be required.

craicoverflow (approved):

PR #2152, opened 2020-10-03
Rename test files from *Test.ts to *.test.ts

Odin94:
Solves #2136

Having some trouble getting the tests to run properly with regards to databases, eg. `Starting the instance failed, please enable debug for more infomation` for MongoDB tests and `EBUSY: resource busy or locked, unlink '...\graphback\packages\graphback-runtime-knex\tests\data/db.sqlite'` for anything that touches db.sqlite. 

Maybe an issue because I'm on Windows.

PR #2166, opened 2020-10-07
chore: rename test files

craicoverflow:
Co-PR of https://github.com/aerogear/graphback/pull/2152

I could not use the same PR as the author used their master, which I could not check out to.


machi1990:
I think we should be consistent and apply the naming to all the test files e.g the ones in https://github.com/aerogear/graphback/tree/35768bca52f4a2709e6ccda5541636935216421b/packages/graphql-migrations/tests/specs etc

craicoverflow:
Yeah, those were missed, thanks.

machi1990 (approved):

PR #2171, opened 2020-10-09
fix(codegen-schema): remove auto primary key from create input

### LINKED ISSUE #1900: MongoDB create method should ignore client side provided _id
Regression. Both datasync and schema plugins add _id to create. 
This is not only invalid but causes clientSide id to break server:

```
message: "Variable "$input" got invalid value "storeclient.6f0bee74-25b0-48b4-a471-96bb9e15843a" at "input._id"; Expected type "GraphbackObjectID". Value is not a valid mongodb object id of form: storeclient.6f0bee74-25b0-48b4-a471-96bb9e15843a
```

Model seems invalid as well:
```
type Todo {
  _id: GraphbackObjectID!
  title: String
  description: String
  completed: Boolean
  _lastUpdatedAt: GraphbackTimestamp
}

input CreateTodoInput {
  ## This should be gone!
  _id: GraphbackObjectID
  title: String
  description: String
  completed: Boolean
}
```

craicoverflow:
Fixes #1900. 

The `Create<T>Input` should exclude fields which are auto-incrementing. The fields that are excluded are `id: ID` and `_id: GraphbackObjectID`.

## Types of changes

- [ ] Other (please specify)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aerogear/graphback/blob/master/CONTRIBUTING.md) doc
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

## Verification

Start a Graphback server with a model where the primary key is `id: ID` or `_id: GraphbackObjectID`.

Example:

```graphql
""" @model """
type Note {
  id: ID!
  title: String!
  description: String
}
```

In the GraphQL Playground, check if the `CreateNoteInput` has the `id` field. It should not.

Expected output:

```graphql
input CreateNoteInput {
  id: ID
  title: String!
  description: String
}
```


craicoverflow:
@machi1990 if you could have a look at this today sometime? :smile:

packages/graphback-codegen-schema/tests/GraphQLSchemaCreator.test.ts
```diff
@@ -405,14 +405,58 @@ test('Transient field is excluded from input types', () => {
...
+
+test('Auto generated primary key fields are not included in the Create<T>Mutation', () => {
+  const modelAST = `
+    """
+    @model
+    """
+    type User {
+      id: ID!
+      name: String!
+    }
```
machi1990, line 428: What happens for a model

```graphql
 """
    @model
    """
    type User {
      id: ID!
      name: String!
      _id: GraphbackObjectID
    }
```

?

craicoverflow: It would probably just not include both - but then that is the fault of the user and should not even be done. These are documented as auto-incrementing so the user should not do this.

craicoverflow: Actually upon checking, this is validated earlier in the execution:

`Error: UserFour type should not have two potential primary keys: "_id" and "id". Use '@id' annotations to indicate which one is to be used.`

Comes from: https://github.com/aerogear/graphback/blob/a84dcddade1b55b2955841ac2764eefec261f4f4/packages/graphback-core/src/db/getPrimaryKey.ts#L35

machi1990: Agree. I copy pasted in a rush and our primary key detection will throw an error the copied code above.  

I meant to have take the below business case. 

```graphql
    """
    @model
    """
    type User {
      """
     @id
     """
      key: String!
      name: String!
      _id: GraphbackObjectID
      id: ID
    }
```

Here the primary key is `key`. This means that `_id` and `id` will not be included right?

craicoverflow: Got you. That is an uncounted use case - fixing it now.

craicoverflow: Fixed now in https://github.com/aerogear/graphback/pull/2171/commits/d4916a25e8cf40d055b383834f60fee323037c2e

machi1990 (approved):

machi1990:
Should we document this is schema breaking change?

craicoverflow:
This would mean releasing Graphback 2.0 - there was a discussion about this on Thursday and the conclusion was to not document it as a breaking change as when it comes to using Graphback from the client you will not be passing these IDs when you expect them to be created for you. And since there is no tooling which depends on the Graphback schema (except for Offix but this was having to do hacks until this issue is fixed).

machi1990:
Really minor, I suppose this file https://github.com/aerogear/graphback/blob/6a7a17f822814add4fdc4a2d3610f48d7400ec3f/packages/graphback-runtime-knex/tests/data/db.sqlite should have been gitignored. Rebase with the master maybe?

Issue #2184, opened 2020-10-15
Use an in-memory Knex database in unit tests


craicoverflow:
**Is your feature request related to a problem? Please describe.**

Some of the Knex related unit tests are creating a SQLite database file and cleaning it up after each test runs. This marginally impacts test completion time, and sometimes the files do not successfully get removed if tests are stopped early, which can make them susceptible to being committed to source control.

**Describe the solution you'd like**

Where possible we should use the in-memory database available when using [Knex.js](http://knexjs.org/).
This solution may not be appropriate for all test suites, they need to be assessed on an individual basis.

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

RinkiyaKeDad:
I'd like to work on this. Can you please tell me which tests are being referred here?

machi1990:
Hi @RinkiyaKeDad, doing a `.sqlite` search on the project will reveal which test files using the file for storage. There is still a couple of them. I'll take one example file for further clarification.

Le'ts take the file:

https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-knex/tests/service/DefaultCRUDService.test.ts

You can see that, Knex is configured to use sqlite database with a connection to a file system: 

https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-knex/tests/service/DefaultCRUDService.test.ts#L36

The idea here will be to use an in memory database like: https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-knex/tests/knexQueryMapper.test.ts#L12

To improve test speed and also make sure that we do not have to worry about the database file being left around. 

https://github.com/aerogear/graphback/blob/master/packages/graphback-runtime-knex/tests/service/DefaultCRUDService.test.ts#L15-L19

The patch needs to be applied to all occurrences where a connection to a file system is used to setup the database. Thanks.

craicoverflow:
I actually think now that this will not work at all - the problem is that `migrateDB` does not accept a database, just a config which it uses to create a connection. This creates two separate in-memory databases (one during migrations) and another in the tests but this one does not have the migration history.

PR #2192, opened 2020-10-19
chore: using in-memory knex database in unit tests

RinkiyaKeDad:
Fix for issue #2184 

I found two occurrences where SQLite database file was being created and have fix them both. Please let me know if some changes need to be made. Thanks.

## Types of changes

- [ ] Other (please specify)

## Checklist

### DIFF
--- a/packages/graphback-runtime-knex/tests/data/KnexDbDataProvider.test.ts
+++ b/packages/graphback-runtime-knex/tests/data/KnexDbDataProvider.test.ts
@@ -36,9 +36,7 @@ const setup = async (schemaStr: string, config: { seedData?: { [tableName: strin
 
   const dbConfig = {
     client: 'sqlite3',
-    connection: {
-      filename: dbPath,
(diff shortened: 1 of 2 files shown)


craicoverflow (changes requested):
I can see that tests are failing

RinkiyaKeDad:
Hey @craicoverflow! I was looking into those only and seems like I implemented it the wrong way. I've tried to do it again for ```DefaultCRUDService.test.ts ``` and have a doubt there which I've added as a comment. Can you please review that and tell me if I'm in the right direction. Once that it done I'll make similar changes to the other file also. Much thanks.

craicoverflow:
I actually think now that this will not work at all - the problem is that `migrateDB` does not accept a database, just a config which it uses to create a connection. This creates two separate in-memory databases (one during migrations) and another in the tests but this one does not have the migration history.

We will close this, but thanks for your PR!

Issue #2086, opened 2020-09-19
Add categories to templates - community, external etc.


wtrocki:
**Is your feature request related to a problem? Please describe.**
Please describe the problem you are trying to solve.

**Describe the solution you'd like**
Please describe the desired behavior.

**Describe alternatives you've considered**
Please describe alternative solutions or features you have considered.

machi1990:
/cc @craicoverflow, @machi1990
Automatically generated comment to notify maintainers

wtrocki:
More context here. We will be working with partners to try out their work and we need a way to include their templates into graphback for recognition and also association with technologies we cannot or have time to investigate.

craicoverflow:
Would the templates live in the repo, or will they be pulled from multiple sources, so that they do not all need to live inside the repo?

wtrocki:
Never in the repo. Entire concept will be to have them as community ones that might get outdated etc. after a while.

machi1990:
> Never in the repo. Entire concept will be to have them as community ones that might get outdated etc. after a while. This is one of the strategies that we will cover to promote graphback - it actually requires technical investment because I would not expect creator to add this capability (separate sort for community ones) - just add link to repo and that will be it.

We already have the ability to indicate template's url https://github.com/aerogear/graphback/blob/375dc5105455a4ee5c29c5db70634c7cdc693443/packages/create-graphback/src/index.ts#L28

If I understood correctly, the use case here is to be able to indicate that an template is a community provided or an external template if the url does not match graphback's templates repository?

craicoverflow:
IIRC that flag does not work.

wtrocki:
Ah.. idea here is to actually embedd them into some separate json file that people can edit and list them on init (but with header - those are community maintained).
So people can get something they feel familiar with.

machi1990:
> Ah.. idea here is to actually embedd them into some separate json file that people can edit and list them on init (but with header - those are community maintained).
> So people can get something they feel familiar with.

Even better.

RinkiyaKeDad:
Hi! Can I work on this? As far as I understand the goal is that people should be able to provide a link to their template (a JSON file) by using something like: ```graphback init --templateUrl=<link-goes-here>```? What would be a good starting point for solving this? Thanks.

wtrocki:
I was thinking we can actually create templates.json file in repo where community can add their templates.
Graphback-cli project have them hard coded for the moment so we can split between community and base templatez

PR #2215, opened 2020-11-17
feat(templates): adding external templates

RinkiyaKeDad:
Fix for #2086 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Other (please specify)

## Checklist

## Further comments
I've created a `community-templates.json` file in the root directory. Here people can specify their templates which they want to appear in the CLI. Then in the `starterTemplates.ts` file I import these templates as `externalTemplates` and then create the `externalTemplatesArray` which I merge with the already existing `allTemplates` array. 

I didn't know how to go about testing this so I haven't done that. Will write documentation for this once the feature is completed and tested. I also had to add 
```
{
  "compilerOptions": {
    "resolveJsonModule": true,
    "esModuleInterop": true  
  }
}
```
in the `tsconfig.json` in order to [import json files](https://hackernoon.com/import-json-into-typescript-8d465beded79).
Please let me know if you have any feedback or suggestions. Thanks.


community-templates.json
```diff
@@ -0,0 +1,19 @@
+[
+  {
+    "name": "a-sample-test-template",
+    "description": "Fetching current templates in the sample template",
+    "repos": [
+      {
+        "uri": "https://github.com/aerogear/graphback",
+        "branch": "templates-1.0.0",
```
wtrocki, line 8:
```suggestion
        "branch": "master",
```
```suggestion
        "branch": "master",
```

community-templates.json
```diff
@@ -0,0 +1,19 @@
+[
+  {
+    "name": "a-sample-test-template",
```
wtrocki, line 3: We need some community template example. We do not have any so I think that having JS file will be better:

- we can put comment with example (as documentation)
- We can leave that empty for the moment

packages/create-graphback/src/init/starterTemplates.ts
```diff
@@ -113,6 +114,9 @@ export const allTemplates: Template[] = [
   }
 ];
 
+const externalTemplatesArray: Template[] = externalTemplates;
```
wtrocki, line 117: I would also improve how we present those templates - Graphback and community templates when executing create-graphback cmd

RinkiyaKeDad: Thanks for reviewing the code @wtrocki. Do you have any suggestions about how I could go on implementing this? All templates must still reside in the `allTemplates` array for the code to function as expected, right? So how can I go about differentiating the templates in the CLI? A simple solution I can think of is appending `(community)` to the name of each template I import from the json file.

community-templates.json
RinkiyaKeDad: Yeah I'll leave this file empty for now. I only added this as a refernence. When I write the docs I'll mention the way and syntax of adding templates.

packages/create-graphback/src/init/starterTemplates.ts
wtrocki: https://github.com/aerogear/graphback/blob/master/packages/create-graphback/src/init/components.ts#L31-L53

Experiment with this methods. Now all will be printed there. 
If we can split them into two categories etc. that will be awesome.

RinkiyaKeDad:
Hey @wtrocki! I've tried to split the templates into categories in the CLI by adding an `external` property ( just like the `disabled` one already present ). I then choose the templates which get loaded into the [`displayedTemplates`](https://github.com/aerogear/graphback/blob/master/packages/create-graphback/src/init/components.ts#L31) variable based on whether this property is set to `true` or not. Please let me know if you have any other suggestions.

Thanks!

P.S. I've let a sample template still remain in `community-templates.json`. Will remove that once I write the docs.

community-templates.json
```diff
@@ -14,6 +14,7 @@
         "branch": "templates-1.0.0",
         "path": "/templates/ts-apollo-postgres-backend"
       }
-    ]
+    ],
+    "external": true
```
wtrocki, line 18: Idea is that this file will be for external ones so do we need flag

packages/create-graphback/src/init/components.ts
```diff
@@ -32,7 +32,7 @@ async function chooseTemplate(filter: string = ''): Promise<Template> {
     {
       type:'list',
       name:'typeOfTemplates',
-      message:'Choose between community or officail Graphback templates',
+      message:'Choose between community or official Graphback templates',
```
wtrocki, line 35: Would it be possible to print them separately? Too much ask?

wtrocki:
Really good work. I will verify that tommorow

community-templates.json
RinkiyaKeDad: We would need the flag somewhere. If not here then I could have it for the internal templates, something like `internal: true`?

packages/create-graphback/src/init/components.ts
RinkiyaKeDad: Separately as in no choosing and just all of them listed with some gap (or a line saying that 'following are the community templates') in between?

RinkiyaKeDad:
Thanks a lot! Will look into why the build is failing once we've decided on the way to show the templates.
Thanks for reviewing! :)

wtrocki:
```
Graphback supported templates:
- 
- 
_ 

Community templates
- 
_ 
_ 
``

File: community-templates.json (hunk 2 above)
Comment (wtrocki) [reply community-templates.json:3]: yes!

**Comment by wtrocki** [comment]
Like new display style. Would review PR today.

**Comment by RinkiyaKeDad** [comment]
Hey! My build keeps failing each time because of the way things are imported in some files which aren't even relevant to this PR. Am I doing something wrong? Should I go and fix all these imports? Please let me know. Thanks.

For example:
`github` and `request` were imported in [starterTemplates.ts](https://github.com/aerogear/graphback/blob/master/packages/create-graphback/src/init/starterTemplates.ts#L4-L5) but I had to change the way they were imported otherwise the build would keep failing. Similarly for `DataLoader` in the [CRUDServices.ts](https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/runtime/CRUDService.ts#L1) file.

@wtrocki @craicoverflow

**Comment by wtrocki** [comment]
I took a look into it but there seem to be some issues with the lining etc. 
I will check that this week and merge! Thank you so much!

**Comment by wtrocki** [comment]
@RinkiyaKeDad I'm worried that we cannot accept change due to numerous issues.

I think we should

- Keep the file close to the code and keep it as .ts file rather than json (Using json forces us to change compilation rules that we do not want to do.
- Add information into contributing guide how to add new community template

I have done some work on https://github.com/aerogear/graphback/pull/2226 so feel free to change and based on this. If you require any help or advice give us shout. 

You can ignore website builds. They will be gone once we rebase this and update dependencies.

**Comment by wtrocki** [comment]
@RinkiyaKeDad are you ok to continue working on this. We might actually need it to push our own templates at some point

**Comment by RinkiyaKeDad** [comment]
@wtrocki yeah definitely. I've written a draft documentation. What are the next steps?

File: COMMUNITY-TEMPLATES.md [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+      }
+    ],
+  }
+```
+
+Note that you can specify as many repositories as you want (with the desired branch and mount path). Files from the specified repositories would be picked up for your template.
+
+## Step 2: Adding your template object
+
+In the root directory of Graphback you'll find the file `community-templates.ts`. This exports an array called `externalTemplates`. It is in this array that you should add your template object for it to be picked up by the CLI. So the final file would look something like this:
```
wtrocki, line 35: We wanted typescript file in package. There is no way to use it outside package

COMMUNITY-TEMPLATES.md
```diff
@@ -0,0 +1,59 @@
...
+
+In case you want to use some other templates than the ones offered, Graphback also supports adding your templates. The process is fairly simple and is as follows:
+
+## Step 1: Creating an object for your template
+
+To add your template, you first need to have it in the form of an object which you'll later add to an array. This is best explained by an example:
+
+Let's say you have a template which you want to name `a-sample-template` and the files for it are located in the repository `https://github.com/sample-user/sample-project`. Then the object for your template would look like this:
+
+```
```
wtrocki, line 11: This should be ts file

RinkiyaKeDad: Do you mean the url should point to a particular ts file on the person's repository?

wtrocki:
We need templates in TS file like current templates.
JSON outside package will not work when publishing package.

RinkiyaKeDad:
Oh I was under the impression that we would be merging the PR you opened so I didn't change from json to a TS file in this branch. I've done that now. Please have a look and let me know the further steps. Thanks a lot!

wtrocki:
See my comment here: https://github.com/aerogear/graphback/pull/2215#issuecomment-739413263

RinkiyaKeDad:
Got it. So should I also bring over the other changes you made in your PR to this one?

wtrocki:
Like we talking about 5 minutes changes. You can simply start from scratch and just reapply docs and some of the changes from this PR. Closing this PR feel free to open new one.

PR #2226, opened 2020-11-30
feat(templates): ability to add external templates

wtrocki:
Fixing: https://github.com/aerogear/graphback/pull/2215


packages/create-graphback/src/init/starterTemplates.ts
```diff
@@ -1,16 +1,17 @@
 import { createWriteStream, mkdirSync, existsSync } from 'fs';
 import chalk from 'chalk';
 import ora from 'ora'
-import * as github from 'parse-github-url';
-import * as request from 'request';
+import gh from 'parse-github-url';
+import request from 'request';
 import * as tar from 'tar';
 import * as tmp from 'tmp';
+import externalTemplates from '../../community-templates.json';
```
craicoverflow, line 8: Getting `Cannot find module '../../community-templates.json'. Consider using '--resolveJsonModule' to import module with '.json' extensionts(2732)`

wtrocki: Actually having it as js file inside repository and contribution guide listing this file is much better than root. I have changed typescript config but it seems to be not applied.

packages/graphback-core/src/runtime/CRUDService.ts
```diff
@@ -1,4 +1,4 @@
-import * as DataLoader from "dataloader";
+import DataLoader from "dataloader";
```
craicoverflow, line 1: Even though the tsconfig is adjusted, I am getting:

```
src/SchemaCRUDPlugin.ts(724,39): error TS2351: This expression is not constructable.
  Type 'typeof DataLoader' has no construct signatures.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This is after running `yarn unlock && yarn clean && rm -rf node_modules && yarn && yarn build` to ensure a clean repo.

RinkiyaKeDad: Did we need to change the way we imported everything because of changes made to the `tsconfig.json` file?

wtrocki: Lets use ts file

Issue #2174, opened 2020-10-12
Subscriptions not working in GraphiQL Playground with Fastify


craicoverflow:
// TODO

* **Module**:
* **Version**:
* **Node.js / npm versions:**
* Example/snippet of the model that is causing issue

machi1990:
Automatically generated comment to notify maintainers
/cc @machi1990, @wtrocki

RinkiyaKeDad:
Hey @craicoverflow! I'd like to help solve this issue. 

> but it also might be

To be sure that it is not a Graphback issue I wanted to look at the resolvers which get generated. I was looking for them and as far as I could understand they get generated [here](https://github.com/aerogear/graphback/blob/master/templates/ts-fastify-mongodb-backend/src/index.ts#L35). But I still am not sure how to see the actually generated resolvers file. Can you please point me to it? Thanks.

craicoverflow:
Hi @RinkiyaKeDad!

The resolvers are created in the [SchemaCRUDPlugin](https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts)

RinkiyaKeDad:
Hi @craicoverflow! There were some Typescript types related errors in [SchemaCRUDPlugin](https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts). For example when adding the [create subscription resolver](https://github.com/aerogear/graphback/blob/master/packages/graphback-codegen-schema/src/SchemaCRUDPlugin.ts#L470) I was getting this error in the console:
```
Argument of type 'IUnionTypeResolver | IEnumTypeResolver | IInputObjectTypeResolver | IObjectTypeResolver<any, any, any> | IInterfaceTypeResolver<...>' is not assignable to parameter of type 'IObjectTypeResolver<any, any, any>'.
  Type 'IUnionTypeResolver' is not assignable to type 'IObjectTypeResolver<any, any, any>'.
    Type 'IUnionTypeResolver' is not assignable to type '{ [key: string]: IFieldResolver<any, any, any, any> | IFieldResolverOptions<any, any, any>; }'.
```
I am not entirely sure if this is responsible for subscriptions not working. Please let me know your views on this. Till then I'll try creating a Fastify server the same way as we did in the template and if that works I think it could confirm if the problem is with the way we created the template or not.

Thanks and happy holidays!

wtrocki:
[...]

PR #2248, opened 2021-01-01
fix(fastify-template): fix for subscriptions in fastify template

RinkiyaKeDad:
Fix for #2174 

## Types of changes

- [ ] Other (please specify)

## Checklist

## Further comments

As [discussed](https://github.com/aerogear/graphback/issues/2174#issuecomment-752543660) I've added resolvers in the plugin options of mercurius which should fix the issue.

## Verification

I tried verifying this by enabling the fastify template and then using the updated cli (by running `node dist/index.js` in the create-graphback folder) but I received this error due to which the template didn't load and I wasn't able to test it. Please let me know if there are some other steps I can take to test if the subscriptions are working or not. Thanks.
sendNew

### DIFF
--- a/templates/ts-fastify-mongodb-backend/src/index.ts
+++ b/templates/ts-fastify-mongodb-backend/src/index.ts
@@ -10,7 +10,7 @@ import { loadConfigSync } from 'graphql-config';
 import { makeExecutableSchema } from '@graphql-tools/schema';
 import { connectDB } from './db';
 import { noteResolvers } from './resolvers/noteResolvers';
-import { mergeResolvers } from '@graphql-tools/merge';
+import {mergeResolvers} from '@graphql-tools/merge';
 
 async function start() {
   const app = fastify();
@@ -21,9 +21,9 @@ async function start() {
   const config = loadConfigSync({
     extensions: [
       () => ({
-        name: graphbackExtension,
-      }),
-    ],
+        name: graphbackExtension
+      })
+    ]
   });
 
   const projectConfig = config.getDefault();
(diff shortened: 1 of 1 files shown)


templates/ts-fastify-mongodb-backend/src/index.ts
```diff
@@ -45,6 +45,7 @@ async function start() {
 
   app.register((mercurius as any), {
     schema,
+    resolvers,
```
machi1990, line 48: I think `resolvers` should include `noteResolvers` as in https://github.com/aerogear/graphback/blob/12d588718dacaecb5cc1d3ed8005ac4a949b82a4/templates/ts-fastify-mongodb-backend/src/index.ts#L40

machi1990 (changes requested):
Thanks for the PR. I've not verified it as I am away from laptop but it LGTM barring the small requested change inline. 

I'll let @craicoverflow @wtrocki do the verification.

wtrocki:
Let's start this year with this small PR. Looking at it now

wtrocki:
```
 ! [remote rejected]       pr/RinkiyaKeDad/2248-1 -> pr/RinkiyaKeDad/2248-1 (permission denied)
error: failed to push some refs to 'git@github.com:RinkiyaKeDad/graphback'
```

I cannot fix issues in PR. Can you add me to fork :) THX

wtrocki:
> I tried verifying this by enabling the fastify template and then using the updated cli (by running

How about running this from the code:
```
cd templates/ts-fastify-mongodb-backend
docker-compose up -d
yarn develop
```

PS: This doesn't seem to be working for me.

RinkiyaKeDad:
> I think `resolvers` should include `noteResolvers` as in

Thanks for pointing that out 😅 I've added them now @machi1990.




> Can you add me to fork :) THX

Added you @wtrocki :)







> PS: This doesn't seem to be working for me.




Same. It doesn't give an error in the terminal like last time, but shows that localhost refused to connect on the browser.



NewTest

RinkiyaKeDad:
> How about running this from the code:

Okay so this way of testing the template works if I don't specify the noteResolvers and simply leave the code like this:
```
  app.register((mercurius as any), {
    schema,
    resolvers,
    graphiql: true,
    context: contextCreator,
    subscription: true
  });
```
But even with this when I tested the subscriptions they were not working. Could it be because [`makeExecutableSchema`](https://github.com/aerogear/graphback/blob/master/templates/ts-fastify-mongodb-backend/src/index.ts#L39) is not creating the schema in a way acceptable by mercurius?

Will have to look for some other way to fix this then I guess 🤔

RinkiyaKeDad:
Would it be a good idea to switch from mercurius to [apollo-server-fastify](https://github.com/apollographql/apollo-server/tree/main/packages/apollo-server-fastify)?

wtrocki:
Sounds good?

RinkiyaKeDad:
Closing this since we're shifting to apollo-server-fastify. See #2256

PR #2281, opened 2021-05-06
fix: add license copyright

wtrocki:


craicoverflow:
This should be a `chore` IMO. This matters because the changelog is based off our commits.

wtrocki:
@craicoverflow testing CI/CD job with some giberish change to fix builds. Ignore that PR for the moment. Localy things work so this is something with github actions

PR #2294, opened 2021-06-22
fix(packages): Adds `GraphQLNonNull` to the `modelType` for ResultList.items

CoreyKovalik:
This makes our DX better. As a developer, we shouldn't have to assert that MyType is not null, because it never happens.

`items` is never `[MyType1, MyType2, null, null, MyType3]` or `[null, null, null, ...N]`

before:

```typescript
export type MyModelTypeResultList = {
  __typename?: 'MyModelTypeResultList';
  items: Array<Maybe<MyModelType>>;
  offset?: Maybe<Scalars['Int']>;
  limit?: Maybe<Scalars['Int']>;
  count?: Maybe<Scalars['Int']>;
};
```

after:

```typescript
export type MyModelTypeResultList = {
  __typename?: 'MyModelTypeResultList';
  items: Array<MyModelType>;
  offset?: Maybe<Scalars['Int']>;
  limit?: Maybe<Scalars['Int']>;
  count?: Maybe<Scalars['Int']>;
};
```

## Types of changes

- [ ] Other (please specify)

## Checklist

- [x] Lint and unit tests pass locally with my changes


CoreyKovalik:
ran `yarn test -- -u` to update snapshots

wtrocki (approved):

wtrocki:
Fully agree

craicoverflow:
I tend to agree also - this would mean deviating from the [GraphQL CRUD](https://graphqlcrud.org/) spec somewhat. Should we also propose it gets updated with this?

wtrocki:
I think that this was more ommision on our side.
If we look into rules, hasura or aws they all asume no nulls are returned

https://graphql-rules.com/rules/output-non-null

Sad part is we need to break graphback for this change and make tons of updates in graphqlcrud

CoreyKovalik:
> I think that this was more ommision on our side.
> If we look into rules, hasura or aws they all asume no nulls are returned
> 
> https://graphql-rules.com/rules/output-non-null
> 
> Sad part is we need to break graphback for this change and make tons of updates in graphqlcrud. doc etc.

While this is undoubtedly an improvement, you're infinitely more familiar with the codebase and maybe this PR opened a can of worms that is worth kicking down the road if it's going to cause a breaking change for the next release

